PR #92, opened 2019-01-10
USB driver

BowenLiu15:
Finished the UHCI, EHCI and the USB keyboard driver.


kevinaboos (changes requested):
I like the completeness of the code, but there are consistency issues and a desperate need for documentation and comments. Also, many of the functions can be moved into a struct implementation block, e.g., UsbDevice-related functions. 

Try to apply my comments across your whole codebase, not just the specific line. Many of them are repeated feedback that is applicable to a whole file or set of files. 

Finally, add a description to all of your Cargo.tomls and at the top of every lib.rs file (using `//!` top-level comments) so that we can quickly glance at each crate to know what it does.

kevinaboos:
The changeset specified in this PR has been moved to the `usb_driver` branch, since this is an old PR based on the old private version of this Theseus repo.

PR #179, opened 2019-07-24
Arm uefi

snoword:
Port Theseus to ARM
Initialize memories, exceptions and logs


applications/terminal/Cargo.toml
```diff
@@ -8,9 +8,15 @@ build = "../../build.rs"
 
 [dependencies]
 spin = "0.4.10"
+
```
kevinaboos: this is just some old leftover junk, the terminal shouldn't and doesn't need x86_64 or aarch64 stuff, you can remove these.

applications/terminal_print/Cargo.toml
```diff
@@ -8,8 +8,13 @@ build = "../../build.rs"
 
 [dependencies]
 spin = "0.4.10"
+
```
kevinaboos: same thing as above, none of these are needed.

cfg/aarch64-theseus.json
```diff
@@ -0,0 +1,25 @@
+{
+  "llvm-target": "aarch64-pc-windows-msvc",
```
kevinaboos: is this the right target triple, `pc-windows-msvc`? Doesn't it need to be `none`, and shouldn't we be using a gnu platform instead of microsoft?

kernel/fs_node/Cargo.toml
```diff
@@ -7,8 +7,13 @@ description = "defines the traits for File and Directory. These files and direct
 
 [dependencies]
 spin = "0.4.5"
+
```
kevinaboos: remove, `fs_node` doesn't need arch stuff.

kernel/memfs/Cargo.toml
```diff
@@ -7,8 +7,13 @@ description = "contains the MappedPages-backed implementation of the fs_node tra
 
 [dependencies]
 spin = "0.4.5"
+
```
kevinaboos: memfs doesn't need x86 either

kevinaboos (changes requested):
You've added a lot of new crates, of course, but some of them don't appear to be necessary. Mainly, things having to do with volatile types (`tock-registers`, `vcell`, `volatile-registers`, etc) can all be replaced by the `volatile` crate that we already use, which has types like `ReadOnly`, `ReadWrite`, and `WriteOnly`. Were you using those directly, or were they dependencies of other crates you're using? 

So, in general, you seem to have basically put aarch-specific stuff in like.. dozens of crates. That's really quite messy, and makes it even more difficult to add additional architectures in the future. What I was hoping for is that you'd separate the x86-specific stuff and the arm-specific stuff into separate crates, or something like that. Making _every_ crate need to be aware of the underlying architecture is surely a recipe for an unmaintainable mess.
For example, the interrupts and exceptions crates are just littered with cfg blocks, and I cannot accepts something like that into the mainline. 

Also, a note about the cfg blocks -- we don't support x86 (32-bit) and never intend to, so you don't need to insert that as an `any` condition everywhere. It's also not necessarily correct, a lot of those crates do _not_ support 32-bit x86, they only support x86_64. 

So far I have only reviewed up to the `memory` crate. I will post more comments later.

kevinaboos (changes requested):
Please undo all of these unnecessary formatting and whitespace changes. This makes it impossible to determine what actually changed to accommodate ARM and what was unnecessary.

In fact, I'm going to postpone the rest of my review until you address this. Thanks.

applications/terminal/Cargo.toml
snoword: removed

applications/terminal_print/Cargo.toml
snoword: removed

cfg/aarch64-theseus.json
snoword: This indicates that the project will be compiled as a UEFI application(UEFI applications can run on Windows). GNU will use a chainloader to load the UEFI binary file

kernel/fs_node/Cargo.toml
snoword: solved

kernel/memfs/Cargo.toml
snoword: deleted

applications/terminal/Cargo.toml
snoword: vcell, volatile-registers are introduces by the ARM64 crate `cortex-m`.
I have removed tock-registers.

PR #194, opened 2019-09-18
Support for Fat 32

rjcrjc:
Continuation of #178 from previous Fat 32 branch.


Makefile
```diff
@@ -591,7 +591,13 @@ bochs: $(iso)
 	bochs -f bochsrc.txt -q
 
 
-
+### Creates a fat32 filesystem that we can mount and test with:
+### Additional requirements: 
+fat32.img:
```
kevinaboos: i think this should all be in a separate script, not a makefile target. We can also make the QEMU targets depend on the fat32.img file to ensure that it always exists before starting QEMU

kevinaboos (changes requested):
How significant is the difference between lzma and other compression formats? LZMA support isn't included in most distributions by default, so if it's not significant we should choose a different format, like zip or bz2.

Before we merge, please cleanup unnecessary changes that were used for testing, e.g., those in `ls`, `captain`, and others. I think that scanning for filesystems can be done as part of device initialization after the storage devices are inited. That way, they are guaranteed to be done by (and in) the kernel.

The `PFS*` terms are old and should be changed; if they stand for "physical file system", then that's not really specific or worth anything in particular.

kernel/fat32/src/lib.rs
```diff
@@ -0,0 +1,2118 @@
...
+}
+
+impl ClusterChain {
+    pub fn new(cluster: u32, fs: FSRef, name: DiskName, parent: WeakDirRef, size: u32) -> ClusterChain {
+        ClusterChain {
+            filesystem: fs,
+            cluster,
+            name,
+            parent: parent.clone(),
+            parent_count: parent.upgrade().map(|_| 1).unwrap_or(0), // REVIEW this seems sketchy, but I think it's fair to do.
```
kevinaboos: in general it's a bad idea to have a strong reference to a parent node in a graph/tree data structure, so I think weak references are the way to go. 
however, is there a concurrency issue here? A reference count could change at any moment, so if you're storing it an using it later it could be an example of a TOCTTOU problem/race condition.

kevinaboos (changes requested):
For the functions that don't yet work or aren't implemented yet, we should insert a `return Err...` up at the top that explains that they're not yet implemented. You could also use `unimplemented!()` if you want.

I think there should be a bit more clarification on which structs are designed to be used by other crates outside of fat32, and which are designed for internal use only. There's quite a lot of structs, perhaps it should be split into a couple of modules (or better yet, separate crates). More structure would be helpful, there's a lot of categories (e.g., name/string handling, cluster handling, internal FAT32 structures that are packed to match on-disk formats, external FAT32 structures used for integrating with Theseus traits, etc).

kernel/fat32/src/fs_core.rs
```diff
@@ -0,0 +1,518 @@
...
+/// # Arguments
+/// * `sd`: the storage device that contains the FAT32 filesystem structure
+/// 
+/// # Return
+/// If the drive passed in contains a fat filesystem, this returns the filesystem structure needed to run operations on the disk 
+pub fn init_fs(sd: storage_device::StorageDeviceRef) -> Result<Filesystem, &'static str>  {
+    info!("Attempting to initialize a fat32 filesystem");
+    // Once a FAT32 filesystem is detected, this will create the Filesystem structure from the drive
+    if detect_fat(&sd) == true {
+        let header = match Header::new(sd.clone()){
```
kevinaboos: you can simplify these two `match` blocks using the `ok_or()` and `map_err()` combinator funtions

kevinaboos (changes requested):
The docs for block_cache and block_io seem to be mixed up a bit. Isn't block_cache just for caching whole blocks, while block_io is for byte granularity conversions?

kernel/block_cache/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+use storage_device::{StorageDevice};
+
+/// A cache to store read and written blocks from a storage device.
+pub struct BlockCache {
+    /// The cache of blocks (sectors) read from the storage device,
+    /// a map from sector number to data byte array.
+    cache: InternalCache,
+}
+
+impl BlockCache {
```
kevinaboos: Shouldn't the backing storage device be a part of the `BlockCache` struct itself, such that you don't need to pass it into every method?  (apologies if that was from my old design)
What you have here allows for unusual behavior:  what if you read/write/flush a block to a different storage device than what it originally came from? That would allow for a single `BlockCache` instance to have multiple blocks mixed from a variety of different storage devices, which I don't think is a good idea.

kevinaboos (changes requested):
I suppose we're going to break things up into separate crates in the next PR?

kevinaboos:
Closing because it's unfinished & out of date. Original code has been made accessible on private branch.

PR #211, opened 2019-10-18
Libc PR

Ramla-I:


kevinaboos (commented):
CString, CStr, and memchr are already implemented in separate crates on crates.io. Just use those rather than copying their code directly into our code base.  The same is likely true for other modules in your libc crate, such as all of the typedefs in `types.rs` and potentially all of the string functions too.

see my other comments on Slack

kevinaboos:
Closed in favor of #330

PR #282, opened 2020-05-01
IPC disconnection rendezvous

NamiLiy:


applications/test_channel/src/lib.rs
```diff
@@ -164,7 +164,10 @@ fn rendezvous_test_oneshot() -> Result<(), &'static str> {
 
     let t1 = spawn::new_task_builder(|_: ()| -> Result<(), &'static str> {
         warn!("rendezvous_test_oneshot(): Entered sender task!");
-        sender.send("hello")?;
+        sender.send("hello").map_err(|error| {
+            warn!("Sender task failed due to : {:?}", error);
+            return "Sender task failed";
```
kevinaboos: in all Rust expressions, the last value will be returned automatically. Here you just need 
```
"Sender task failed"
```
without the return or semicolon.

kevinaboos: apply this to all of your other `map_err` closures

kernel/rendezvous/src/lib.rs
```diff
@@ -133,13 +135,51 @@ pub fn new_channel<T: Send>() -> (Sender<T>, Receiver<T>) {
...
     (
         Sender   { channel: channel.clone() },
         Receiver { channel: channel }
     )
 }
 
+/// Indicates whether channel is Connected or Disconnected
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum ChannelStatus {
+    /// Channel is working. Initially channel is created with Connected status.
```
kevinaboos: the docs for this variant should be way more specific. The phrase "channel is working" means absolutely nothing and isn't informative. What really matters here is that at least one sender and one receiver endpoint exist

kernel/rendezvous/src/lib.rs
```diff
@@ -249,30 +296,40 @@ impl <T: Send> Sender<T> {
...
         // Now we need to handle other stuff (like notifying waiters) without holding the sender_slot lock.
+
+        
         match retval {
             Some(Ok(receiver_to_notify)) => {
                 drop(receiver_to_notify);
                 return Ok(());
             }
-            Some(Err(e)) => {
+            Some(Err(_e)) => {
```
kevinaboos, line 306: what is the actual error type here? 

I'm not really sure what the meaning of `InvalidInitialState` is, it's hard to take action to correct that. When can it happen, why, and what can a caller/user do about it?

kernel/rendezvous/src/lib.rs
```diff
@@ -249,30 +296,40 @@ impl <T: Send> Sender<T> {
...
         loop {
             {
+                
                 let exchange_state = sender_slot.0.lock();
+                
                 match &*exchange_state {
                     ExchangeState::WaitingForReceiver(blocked_sender, ..) => {
                         if blocked_sender.task() != curr_task {
-                            return Err("BUG: CURR TASK WAS DIFFERENT THAN BLOCKED SENDER");
+                            return Err(ChannelError::TaskSubsystemFailure);
```
kevinaboos: it's fine to return this, but you should still log the bug message with an `error!()` statement.

kevinaboos (changes requested):
The Drop routines seem very complicated; can you elaborate on the logic therein and describe what is going on (using comments in the source)? 

In general, it's a bad idea to do infinite loops in a drop handler. I don't see why it's necessary to do that when dropping the Sender/Receiver itself -- can't we just set a disconnection status like in the async channel? Then any future use of the channel will know it's been disconnected (just like async channel). 

It also seems like the droppability of the exchange slot itself isn't really being leveraged. For example, there are still lots of cases where we call `replace_sender/receiver_slot()`, which in my opinion shouldn't ever be required if SenderSlot/ReceiverSlot implement drop.

PR #313, opened 2020-09-04
IPC benchmark fix

Ramla-I:
- Added a check to make sure that a crate is fully unloaded before spawning a new task in IPC benchmark.


Ramla-I:
The spawn benchmark previously failed on qemu but not on hardware. One reason is that it doesn't remove the hello crate fast enough before the next thread is spawned. So I added a check to make sure that doesn't happen.
Also, for 10,000 iterations I start getting page fault and double fault errors in qemu, I'm still not sure why. So I lowered the number of iterations for this benchmark. I still get the same results on real HW even with a lower number of iterations.

kevinaboos (approved):

PR #320, opened 2020-10-30
Use the page allocator for all virtual addresses

kevinaboos:
- Redesigned page allocator and moved into separate crate. 
- Redesigned stack and stack allocator to use the page allocator instead of a dedicated range of virtual addresses. 
- Remove usage of `map_pages()`, other variants will be removed shortly.


kernel/multiple_heaps/src/lib.rs
```diff
@@ -111,20 +112,16 @@ pub fn switch_to_multiple_heaps() -> Result<(), &'static str> {
...
     let mut kernel_mmi = kernel_mmi_ref.lock();
 
     let mut frame_allocator = get_frame_allocator_ref()
         .ok_or("create_heap_mapping(): couldnt get FRAME_ALLOCATOR")?
         .lock();
 
-    let pages = PageRange::from_virt_addr(starting_address, size_in_bytes);
-    let heap_flags = HEAP_FLAGS;
-    let mp = kernel_mmi.page_table.map_pages(pages, heap_flags, frame_allocator.deref_mut())?;
+    let pages = page_allocator::allocate_pages_by_bytes_at(starting_address, size_in_bytes)
```
Ramla-I, line 122: The heap will fail here if the page at this address is not available. It's unlikely since the chunk which contains these addresses is after the kernel text sections. It can be made more generic where the heap is can request any page as long as it's 8k aligned.

kernel/page_allocator/src/lib.rs
```diff
@@ -0,0 +1,437 @@
...
+	}
+
+	// NOTE: we don't actually need to keep track of allocated chunks here, it's just for debugging.
+	if let Some(pr) = allocated_page_range {
+		let new_chunk = Chunk {
+			allocated: true,
+			pages: pr
+		};
+		let ret = new_chunk.as_allocated_pages();
+		locked_list.push_back(new_chunk).unwrap();
```
Ramla-I: A deadlock is possible at this point if the heap has requested a page from the page allocator, and the page allocator then requests memory from the heap for the linked list.

kernel/multiple_heaps/src/lib.rs
Ramla-I: The heap will fail here if the page at this address is not available. It's unlikely since the chunk which contains these addresses is after the kernel text sections. It can be made more generic where the heap can request any page as long as it's 8k aligned.

Ramla-I (commented):
We can bypass the 2 known issues for now by setting the variable in the multiple_heaps crate to a larger value so that no allocation is done for the slab allocators after initialization:
`const PAGES_PER_SIZE_CLASS: usize;`

kernel/multiple_heaps/src/lib.rs
kevinaboos: Yes, this is just to preserve compatibility with the existing/current heap and how it uses addresses. The heap should be refactored later to use arbitrary pages directly.

kernel/page_allocator/src/lib.rs
kevinaboos: agreed, will resolve with a different design later.

Ramla-I (approved):
Looks good. Some changes are still required to allow the heap to allocate pages from any address which we'll address in another PR.

PR #321, opened 2020-11-07
Change rls recommendation to rust-analyzer

Absolucy:
`rust-analyzer` is typically more reliable than rls, has more features, and [will eventually become Rust's official LSP implementation](https://github.com/rust-lang/rfcs/pull/2912)

Also, I added links to the plugins in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,20 +148,9 @@ There are similar commands for building the Theseus book:
 Our personal preference is to use Visual Studio Code (VS Code), which has excellent, official support from the Rust language team. Other options are available [here](https://areweideyet.com/), but we don't recommend them.
 
 For VS Code, recommended plugins are:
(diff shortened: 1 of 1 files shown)


kevinaboos:
Thanks for the PR, @aspenluxxxy!  I will definitely look into this and probably change to using it over RLS. First I need to update our current Rustc nightly version to something more recent, since we're a little (ha) behind on that. 

Once I do that, we'll experiment with newer plugins, such as this and also clippy.

kevinaboos:
I experimented with Rust Analyzer, and while I do like it, it's not perfect. As such, I've decided to mention it as an alternative but not to remove the mention of RLS. 

Recent changes to the README supersede the changes herein, thus I'm closing this as addressed by commit 168c14e.

Thanks again for the recommendation!

PR #330, opened 2020-11-13
Libc beginnings

Ramla-I:
-  I've shifted all the code from the other theseus repo to here.
-  The libc code was last tested by Arav and was working. Right now it won't compile because Theseus's interfaces have changed, like they way we allocate frames.


kevinaboos:
Closed in favor of other libc projects, which are further along now.

PR #331, opened 2020-11-14
Correct some book typos

octronics:
While reading the book I find that some words was missing or incorrect. 

After investigation I found the errors come from the e8cbd23 commit that convert the documentation embedded as crate
modules to mdbook markdown files. Four characters was removed on some wrapped lines.

Correct them and while at it remove the trailing whitespace.


kevinaboos:
@octronics thank you very much for going through all of this! I have no idea what happened or how that got merged in, looks like some kind of odd find & replace error. 

Some of that trailing whitespace is probably to ensure line breaks or paragraph breaks, but it's fine without them. The whole book needs to be rewritten anyway, but I'll merge this as is for now since it's definitely a major improvement. 

Much appreciated!

kevinaboos (approved):

PR #332, opened 2020-12-10
Ixgbe driver and language-level virtualization support

Ramla-I:
**Part 1:**
- _device_manager_  - added detection of the 82599 NIC

- _nic_initialization_ - removed register types and replaced with a register struct 

- _nic_queues_ - Added QueueRegister traits so that queues can store their own registers, and we can create these queue objects for all drivers. Added back in the send and receive functions for the queue since it is common code between the drivers.

- _intel_ethernet_ - added advanced transmit descriptor and removed register types since we now have a register trait.

- _e1000_ - moved registers to a separate file. Registers are divided into four sets so that we can separate out the Rx and Tx registers. Also reintroduced the queues so that we can combine the send and receive operations that are common in both drivers.

**Part 2:**
- _ixgbe_  
lib.rs - I've added comments in the places I think are most relevant to look at
regs.rs - same as the e1000, I've divided the registers up into multiple structs
queue_registers.rs - I've created a struct that stores the registers with a shared reference to their backing memory pages. I like to think of it as my own specific version of an OwnedRef
test_packets.rs - functions to create packets that can be used for testing

- _test_ixgbe_ - test application for the ixgbe device for basic functionality. It's not meant to be thorough but a place where we can start and extend testing as we make changes.

**Part 3:**

- _physcial_nic_ - a trait a NIC has to implement if it wants to support virtualization. It defines functions that return Rx and Tx queues to the original nic when dropped.

- _virtual_nic_ - a struct which contains a set of queues, and implements the NetworkInterfaceCard trait. Its drop handler returns the queues to the physical nic.

- _ixgbe_  
virtual_function.rs - functions to create a virtual NIC from the ixgbe device


kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1192 @@
...
+        let nic_regs3_mapped_page = allocate_memory(mem_base + GENERAL_REGISTERS_1_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES + GENERAL_REGISTERS_2_SIZE_BYTES + TX_REGISTERS_SIZE_BYTES + MAC_REGISTERS_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES, GENERAL_REGISTERS_3_SIZE_BYTES)?;
+
+        // Map the memory as the register struct and tie the lifetime of the struct with its backing mapped pages
+        let regs1 = BoxRefMut::new(Box::new(nic_regs1_mapped_page)).try_map_mut(|mp| mp.as_type_mut::<IntelIxgbeRegisters1>(0))?;
+        let regs2 = BoxRefMut::new(Box::new(nic_regs2_mapped_page)).try_map_mut(|mp| mp.as_type_mut::<IntelIxgbeRegisters2>(0))?;
+        let regs3 = BoxRefMut::new(Box::new(nic_regs3_mapped_page)).try_map_mut(|mp| mp.as_type_mut::<IntelIxgbeRegisters3>(0))?;
+        let mac_regs = BoxRefMut::new(Box::new(nic_mac_regs_mapped_page)).try_map_mut(|mp| mp.as_type_mut::<IntelIxgbeMacRegisters>(0))?;
+        
+        // Divide the pages of the Rx queue registers into multiple 64B regions
```
Ramla-I, line 461: Since the Rx and Tx register sets are 64B each, it's almost like we have to create a special heap for the NIC that can track these allocations from the backing MappedPages.

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1192 @@
...
+        assert!(mp.size_in_bytes() >= QUEUES_IN_MP * RX_QUEUE_REGISTERS_SIZE_BYTES);
+
+        let starting_address = mp.start_address();
+
+        // We share the backing mapped pages among all the queue registers
+        let shared_mp = Arc::new(mp);
+        let mut pointers_to_queues = Vec::with_capacity(QUEUES_IN_MP);
+
+        for i in 0..QUEUES_IN_MP {
+            // This is safe because we have checked that the number of queues we want to partition from these mapped pages fit into the allocated memory,
```
Ramla-I, line 486: We ensure that no memory overlaps. This is the core of the unsafe code and the point where the developer has to ensure correctness.

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1192 @@
...
+    }
+
+    /// Reads status and clears interrupt
+    fn clear_interrupt_status(&self) -> u32 {
+        self.regs1.eicr.read()
+    }
+
+    /// Removes `num_queues` Rx queues from this "physical" NIC device and gives up ownership of them.
+    /// This function is used when creating a virtual NIC that will own the returned queues.
+    fn remove_rx_queues(&mut self, num_queues: usize) -> Result<Vec<RxQueue<IxgbeRxQueueRegisters, AdvancedRxDescriptor>>, &'static str> {
```
Ramla-I: functions used when creating the virtualNIC

kernel/ixgbe/src/queue_registers.rs
```diff
@@ -0,0 +1,89 @@
...
+    boxed::Box
+};
+use core::ops::{Deref, DerefMut};
+use core::mem::ManuallyDrop;
+use nic_queues::{RxQueueRegisters, TxQueueRegisters};
+use memory::MappedPages;
+
+
+/// Struct that stores a pointer to registers for one ixgbe receive queue 
+/// as well as a shared reference to the backing MappedPages where these registers are located.
```
Ramla-I: We tie the lifetime of the register struct with the backing mapped pages manually, rather than through an owned ref. Since the members of this struct can't be accessed outside of this crate, it's not possible to separate the two.

kernel/device_manager/src/lib.rs
```diff
@@ -86,6 +86,17 @@ pub fn init(key_producer: Queue<Event>, mouse_producer: Queue<Event>) -> Result<
                 add_to_network_interfaces(e1000_interface);
                 continue;
             }
+            if dev.vendor_id == ixgbe::INTEL_VEND && dev.device_id == ixgbe::INTEL_82599 {
+                const LOCAL_IP: &'static str = "192.168.0.13/24"; 
+                const GATEWAY_IP: [u8; 4] = [192, 168, 0, 1];
```
kevinaboos: I know this will eventually change, but is there any reason you didn't use the existing DEFAULT_IP... ?

Cargo.lock
```diff
@@ -1044,6 +1045,42 @@ dependencies = [
...
+ "apic",
+ "bit_field 0.7.0",
+ "hashbrown",
+ "hpet",
+ "intel_ethernet",
+ "interrupts",
+ "irq_safety",
+ "kernel_config",
+ "lazy_static",
+ "log 0.3.9",
```
kevinaboos: please update the ixgbe Cargo.toml to use log 0.4.8 like the rest of the OS now

kernel/nic_queues/src/lib.rs
```diff
@@ -18,23 +20,55 @@ use alloc::{
...
+);
+
+/// The register trait that gives access to only those registers required for receiving a packet.
+/// The Rx queue control registers can only be accessed by the physical NIC.
+pub trait RxQueueRegisters {
+    fn update_rdbal(&mut self, value: u32);
+    fn update_rdbah(&mut self, value: u32);
+    fn update_rdlen(&mut self, value: u32);
+    fn update_rdh(&mut self, value: u32);
+    fn update_rdt(&mut self, value: u32);
```
kevinaboos: Nit:  is `update_foo()` the right naming convention? Do NIC manuals use that? Typically Rust uses things like `set_` or `write_`.

kernel/nic_queues/src/lib.rs
```diff
@@ -43,21 +77,106 @@ pub struct RxQueue<T: RxDescriptor> {
...
+    pub cpu_id: Option<u8>,
+    /// Pool where `ReceiveBuffer`s are stored.
+    pub rx_buffer_pool: &'static mpmc::Queue<ReceiveBuffer>,
+    /// The filter id for the physical NIC filter that is set for this queue
+    pub filter_num: Option<u8>
 }
 
+impl<S: RxQueueRegisters, T: RxDescriptor> RxQueue<S,T> {
+    /// Retrieves the ethernet frames from one queue
+    pub fn remove_frames_from_queue(&mut self) -> Result<(), &'static str> {
```
kevinaboos: I can guess here, but the behavior of this fn isn't 100% clear from the docs and name. I _assume_ it removes _all existing_ frames that are "ready" (ready means what?) from this queue, but I'm not sure. Docs/name improvements could help.

kernel/intel_ethernet/src/descriptors.rs
```diff
@@ -20,10 +20,16 @@ pub const TX_CMD_RS:                       u8 = 1 << 3;
 pub const TX_CMD_RPS:                      u8 = 1 << 4;     
 /// Tx Command: VLAN Packet Enable
 pub const TX_CMD_VLE:                      u8 = 1 << 6;     
+/// Tx Command: Descriptor Extension (Advanced format)
+pub const TX_CMD_DEXT:                     u8 = 1 << 5;
```
kevinaboos, line 22: any reason this is out of order?

kernel/intel_ethernet/src/descriptors.rs
```diff
@@ -20,10 +20,16 @@ pub const TX_CMD_RS:                       u8 = 1 << 3;
...
+/// Tx Command: Descriptor Extension (Advanced format)
+pub const TX_CMD_DEXT:                     u8 = 1 << 5;  
 /// Tx Command: Interrupt Delay Enable
 pub const TX_CMD_IDE:                      u8 = 1 << 7;     
 /// Tx Status: descriptor Done
 pub const TX_STATUS_DD:                    u8 = 1 << 0;
+/// Tx Descriptor Type: advanced
+pub const TX_DTYP_ADV:                     u8 = 0x3 << 4;
+/// Tx Descriptor paylen shift
+pub const TX_PAYLEN_SHIFT:                 u8 = 46 - 32; //(actual offset - offset of variable)
```
kevinaboos, line 35: i think this one warrants some more documentation. I can't tell if it's a definition copied from a TRM or just something Theseus-specific that we chose.

kernel/intel_ethernet/src/descriptors.rs
```diff
@@ -320,4 +326,50 @@ impl fmt::Debug for AdvancedRxDescriptor {
...
+/// Advanced Transmit Descriptor used in the Ixgbe driver.
+/// It has 2 modes: Read and Write Back, both of which use the whole 128 bits. 
+/// There is one transmit descriptor per transmit buffer that can be converted between these 2 modes.
+/// Read contains the addresses that the driver writes.
+/// Write Back contains information the hardware writes on receiving a packet.
+/// More information can be found in the 82599 datasheet.
+#[derive(FromBytes)]
+#[repr(C)]
+pub struct AdvancedTxDescriptor {
+    /// Starting physcal address of the receive buffer for the packet.
```
kevinaboos: typo physcal

kernel/intel_ethernet/src/descriptors.rs
```diff
@@ -320,4 +326,50 @@ impl fmt::Debug for AdvancedRxDescriptor {
...
+    pub packet_buffer_address:  Volatile<u64>,
+    /// Length of data buffer
+    pub data_len: Volatile<u16>,
+    /// dtyp = Descriptor Type
+    /// mac = option to apply LinkSec and time stamp
```
kevinaboos: Is this field two items combined into one `u8`? If so, it might be good to define which bits are used by whom, e.g.,
```
dtyp occupied bits [0:4), mac occupies bits [4:8)
```
You could also use [0:3] and [4:7] if you prefer inclusive ranges

kernel/intel_ethernet/src/descriptors.rs
```diff
@@ -320,4 +326,50 @@ impl fmt::Debug for AdvancedRxDescriptor {
...
+    pub dtyp_mac_rsv : Volatile<u8>,
+    /// Command bits
+    pub dcmd:  Volatile<u8>,
+    /// paylen = size in bytes of the data buffer in host memory (not including the fields that the hardware adds)
+    /// popts = options to offload checksum calculation
+    /// sta = status of the descriptor, if it's in use or not
+    pub paylen_popts_cc_idx_sta: Volatile<u32>,
```
kevinaboos: Same comment as above, it seems like this is a multi-field u32. Docs needed.

kernel/intel_ethernet/src/descriptors.rs
kevinaboos: Another way to do it is to use the bitfield type, but I dunno if that would interfere with how this field is actually used in the code. Take a look at it though.

kernel/e1000/src/regs.rs
```diff
@@ -1,3 +1,169 @@
...
+//! The registers are divided into multiple structs because we need to separate out the 
+//! receive and transmit queue registers and store them separately in a per-queue struct. 
+//! Though the e1000 device only has 1 pair of receive and transmit queues, we still structure
+//! the design this way to be able to use code shared by all network drivers.
+//! 
+//! The 4 structs which cover the registers of the entire memory-mapped region are:
+//! * `E1000Registers`
+//! * `E1000RxRegisters`
+//! * `E1000TxRegisters`
+//! * `E1000MacRegisters`
```
kevinaboos: Great docs! 

I'd recommend having a one-line summary at the top of every file, since that's what Rustdoc expects for its "short one-liner" brief of a module or crate.

kernel/e1000/src/regs.rs
```diff
@@ -1,3 +1,169 @@
...
+    _padding8d:                     [u8;  2048],
+    _padding8e:                     [u8;   512],
+    _padding8f:                     [u8;   256],
+    _padding8g:                     [u8;   236],
+    _padding8h:                     [u8;    12],
+    // End of all register structs should be at offset 0x20000 (128 KiB in total size).
+
+} // 28 4KiB pages
+
+const_assert_eq!(core::mem::size_of::<E1000MacRegisters>(), 28 * 4096);
```
kevinaboos, line 129: Now that you've split the e1000Registers into 4 separate structs, should we add another const_assert_eq here that checks that all 4 register structs add up to the expected size, which is  defined on Line 124?

kernel/e1000/src/regs.rs
```diff
@@ -1,3 +1,169 @@
...
+
+///struct to hold registers related to one receive queue
```
kevinaboos: Fix doc formatting, proper sentences, etc.

kernel/e1000/src/lib.rs
```diff
@@ -89,114 +83,51 @@ lazy_static! {
...
-    /// The receive descriptor head index, which points to the next available receive descriptor.
-    pub rdh:                        Volatile<Rdh>,          // 0x2810
-    _padding1:                      [u8; 4],                // 0x2814 - 0x2817
-    /// The receive descriptor tail index, which points to the last available receive descriptor.
-    pub rdt:                        Volatile<Rdt>,          // 0x2818
-}
 
+/// A struct which contains the receive queue registers and implements the `RxQueueRegisters` trait,
+/// which is required to store the registers in an `RxQueue` object.
+struct E1000RxQueueRegisters(BoxRefMut<MappedPages,E1000RxRegisters>);
```
kevinaboos: nit: spacing after comma

kernel/e1000/src/lib.rs
```diff
@@ -385,18 +265,33 @@ impl E1000Nic {
     /// # Arguments
     /// * `device`: reference to the nic device
     /// * `mem_base`: the physical address where the NIC's memory starts.
-    fn map_e1000_regs(device: &PciDevice, mem_base: PhysicalAddress) -> Result<BoxRefMut<MappedPages, E1000Registers>, &'static str> {
-        let nic_mapped_page = allocate_device_register_memory(device, mem_base)?;
-        let regs = BoxRefMut::new(Box::new(nic_mapped_page)).try_map_mut(|mp| mp.as_type_mut::<E1000Registers>(0))?;
-        Ok(regs)
+    fn map_e1000_regs(_device: &PciDevice, mem_base: PhysicalAddress) 
+        -> Result<(BoxRefMut<MappedPages, E1000Registers>, BoxRefMut<MappedPages, E1000RxRegisters>, BoxRefMut<MappedPages, E1000TxRegisters>, 
+            BoxRefMut<MappedPages, E1000MacRegisters>), &'static str> {
```
kevinaboos: hard to read, needs better formatting, something like:
```rust
fn map_e1000_regs(
    _device: &PciDevice,
    mem_base: PhysicalAddress
) -> Result<(
    BoxRefMut<MappedPages, E1000Registers>,
    BoxRefMut<MappedPages, E1000RxRegisters>,
    BoxRefMut<MappedPages, E1000TxRegisters>,
    BoxRefMut<MappedPages, E1000MacRegisters>,
), &'static str> {
    // fn body here
    ...
}
```

kevinaboos: or 
```rust
fn map_e1000_regs(
    _device: &PciDevice,
    mem_base: PhysicalAddress
) -> Result<
    (
        BoxRefMut<MappedPages, E1000Registers>,
        BoxRefMut<MappedPages, E1000RxRegisters>,
        BoxRefMut<MappedPages, E1000TxRegisters>,
        BoxRefMut<MappedPages, E1000MacRegisters>,
    ), 
    &'static str>
{
    // fn body here
    ...
}

File: kernel/e1000/src/lib.rs [hunk 19]
```diff
@@ -458,17 +352,17 @@ impl E1000Nic {
...
+        rx_regs.update_rdt((E1000_NUM_RX_DESC - 1) as u32); 
         // TODO: document these various e1000 flags and why we're setting them
         regs.rctl.write(regs::RCTL_EN| regs::RCTL_SBP | regs::RCTL_LBM_NONE | regs::RTCL_RDMTS_HALF | regs::RCTL_BAM | regs::RCTL_SECRC  | regs::RCTL_BSIZE_2048);
 
         Ok((rx_descs, rx_bufs_in_use))
     }           
     
     /// Initialize the array of tramsmit descriptors and return them.
-    fn tx_init(regs: &mut E1000Registers) -> Result<BoxRefMut<MappedPages, [LegacyTxDescriptor]>, &'static str> {
+    fn tx_init(regs: &mut E1000Registers, tx_regs: &mut E1000TxQueueRegisters) -> Result<BoxRefMut<MappedPages, [LegacyTxDescriptor]>, &'static str> {
```
kevinaboos: this also needs reformatting. Use this format for any "long" functions, lets  say beyond 100 chars wide:
```rust
fn tx_init(
    regs: &mut E1000Registers,
    tx_regs: &mut E1000TxQueueRegisters
) -> Result<BoxRefMut<MappedPages, [LegacyTxDescriptor]>, &'static str> {
    ....
}
```

kevinaboos: one day i'll set up rustfmt... lol

kevinaboos (changes requested):
Part 1 review complete. Everything looks pretty good, just some minor improvements! 
I'll wait until you address these (should be quick) before moving onto Parts 2/3.

kernel/device_manager/src/lib.rs
Ramla-I: it was just what I was using, no need to keep them

kernel/nic_queues/src/lib.rs
Ramla-I: more just how I think of writing to a register as "updating" it. I'll change them to set

kernel/e1000/src/regs.rs
Ramla-I: It would always pass for now, but it doesn't hurt to have more checks. Might be helpful if we ever change how we split up the registers again.

kernel/ixgbe/Cargo.toml
```diff
@@ -0,0 +1,93 @@
+[package]
+name = "ixgbe"
+version = "0.1.0"
+description = "Support for the 10 GbE devices and driver"
```
kevinaboos: does this NIC have other names? I see references so some number like 82599. Might be useful to put that number in here, and also Intel (if this NIC is intel-specific).

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
+//! An ixgbe driver for a 82599 10GbE Network Interface Card.
```
kevinaboos, line 1: oh yea, this is good. Add this stuff to the Cargo.toml description, since I use those for high-level overviews.

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
...
+//! 
+//! Currently we support basic send and receive, Receive Side Scaling (RSS), 5-tuple filters, and MSI interrupts. 
+//! We also support language-level virtualization of the NIC so that applications can directly access their assigned transmit and receive queues.
+//! When using virtualization, we disable RSS since we use 5-tuple filters to ensure packets are routed to the correct queues.
+//! We also disable interrupts, since we do not yet have support for allowing applications to register their own interrupt handlers.
+//! 
+//! Before using the driver, make sure to set the link type as either 1GB or 10GB by setting the `IXGBE_10GB_LINK` variable.
```
kevinaboos: hmm, this is pretty difficult to deal with. It requires a full recompile of this crate, meaning that it's not really a user-facing or app developer-facing option, it's only a kernel dev-facing option. 

Does this need to be a const value? Can we allow users (callers) of this NIC object to change it at their desire?

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
...
+/// If we increase the number to >1, then we also have to add the interrupt handlers in
+/// the NIC initialization function.
+/// Currently we have only tested for 16 interrupts.
+const NUM_MSI_VEC_ENABLED:                  u8      = 1; //IXGBE_NUM_RX_QUEUES_ENABLED;
+
+/***********************************************************/
+
+
+/// The single instance of the 82599 NIC.
+pub static IXGBE_NIC: Once<MutexIrqSafe<IxgbeNic>> = Once::new();
```
kevinaboos: this seems like an odd design choice. Shouldn't all of this (and the config options up top) be tied to a given IxgbeNic struct or something? What if we had multiple ixgbe NICs?

kevinaboos: I don't recall this being a design choice in `e1000`. Why assume there's only a single device instance

kevinaboos: On a related note, IIRC, there was some kind of way to get the list of network interfaces on the system, and then choose one to use. Not sure if that's the best way of doing this, but this singleton makes that impossible. 
The general goal is to register each device somewhere such that the system is aware of all available devices, at least for each device type (storage, networking, etc).

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
...
+        let mac_addr_hardware = Self::read_mac_address_from_nic(&mut mapped_registers_mac);
+
+        // initialize the buffer pool
+        init_rx_buf_pool(RX_BUFFER_POOL_SIZE, IXGBE_RX_BUFFER_SIZE_IN_BYTES, &RX_BUFFER_POOL)?;
+
+        // create the rx desc queues and their packet buffers
+        let (mut rx_descs, mut rx_buffers) = Self::rx_init(&mut mapped_registers1, &mut mapped_registers2, &mut rx_mapped_registers)?;
+        
+        // create the vec of rx queues
+        let mut rx_queues = Vec::new();
```
kevinaboos: minor optimization: could use Vec::with_capacity() here

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
...
+            rx_queues.push(rx_queue);
+            id += 1;
+        }
+
+
+        // create the tx descriptor queues
+        let mut tx_descs = Self::tx_init(&mut mapped_registers2, &mut mapped_registers_mac, &mut tx_mapped_registers)?;
+        
+        // create the vec of tx queues
+        let mut tx_queues = Vec::new();
```
kevinaboos: minor optimization: could use Vec::with_capacity() here

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
...
+        let nic_rx_regs1_mapped_page = allocate_memory(mem_base + GENERAL_REGISTERS_1_SIZE_BYTES, RX_REGISTERS_SIZE_BYTES)?;
+        let nic_regs2_mapped_page = allocate_memory(mem_base + GENERAL_REGISTERS_1_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES, GENERAL_REGISTERS_2_SIZE_BYTES)?;        
+        let nic_tx_regs_mapped_page = allocate_memory(mem_base + GENERAL_REGISTERS_1_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES + GENERAL_REGISTERS_2_SIZE_BYTES, TX_REGISTERS_SIZE_BYTES)?;
+        let nic_mac_regs_mapped_page = allocate_memory(mem_base + GENERAL_REGISTERS_1_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES + GENERAL_REGISTERS_2_SIZE_BYTES + TX_REGISTERS_SIZE_BYTES, MAC_REGISTERS_SIZE_BYTES)?;
+        let nic_rx_regs2_mapped_page = allocate_memory(mem_base + GENERAL_REGISTERS_1_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES + GENERAL_REGISTERS_2_SIZE_BYTES + TX_REGISTERS_SIZE_BYTES + MAC_REGISTERS_SIZE_BYTES, RX_REGISTERS_SIZE_BYTES)?;        
+        let nic_regs3_mapped_page = allocate_memory(mem_base + GENERAL_REGISTERS_1_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES + GENERAL_REGISTERS_2_SIZE_BYTES + TX_REGISTERS_SIZE_BYTES + MAC_REGISTERS_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES, GENERAL_REGISTERS_3_SIZE_BYTES)?;
```
kevinaboos: this is unreadable, lol. Just keep a running tally of the mem_base and do `mem_base += latest addition` between each `allocate_memory()` call.

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
...
+            //clear bits in the swsm register
+            let swsm = regs.swsm.read() & !(SWSM_SMBI) & !(SWSM_SWESMBI);
+            regs.swsm.write(swsm);
+
+            Ok(false)
+        }
+    }
+
+    /// Release the semaphore synchronizing between software and firmware.
+    fn release_semaphore(regs: &mut IntelIxgbeRegisters3) {
```
kevinaboos, line 661: shouldn't we use guard types (like the regular mutex types) to ensure that the semaphore is released? 
Without that, we're not guaranteed that the semaphore will be released on an early return or panic.

kevinaboos: I see that the usage of the semaphore is pretty limited right now, so we could save that for a future PR. If you choose to do that, file an issue to remind yourself.

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
...
+        Ok(())
+    }
+
+    /// Returns value of (links, links2) registers
+    pub fn link_status(&self) -> (u32, u32) {
+        (self.regs2.links.read(), self.regs2.links2.read())
+    }
+
+    /// Returns link speed in Mb/s
+    pub fn link_speed(&self) -> u32 {
```
kevinaboos: perfect use case for an enum, something like:
```rust
enum LinkSpeedMbps: u32 {
    LS_100: 0x1 << 28, 
    LS_1000: ...
}
```
that's pseudocode but you get the idea

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
...
+        let speed = self.regs2.links.read() & LINKS_SPEED_MASK;
```
kevinaboos, line 771: typo? `LINK_SPEED_MASK`

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
...
+        } else if speed == LS_1000 {
+            1000
+        } else if speed == LS_10000 {
+            10_000
+        } else {
+            0
+        }
+    }
+
+    /// Wait for link to be up for upto 10 seconds.
```
kevinaboos, line 775: why 10 seconds, is that an arbitrary value you chose or a value suggested by the ixgbe TRM?
If the former, then make it a parameter into this function. If the latter, document it. Former is probably better regardless.

kernel/ixgbe/src/lib.rs
```diff
@@ -0,0 +1,1235 @@
...
+    }
+
+    /// Reads status and clears interrupt
+    fn clear_interrupt_status(&self) -> u32 {
+        self.regs1.eicr.read()
+    }
+
+    /// Removes `num_queues` Rx queues from this "physical" NIC device and gives up ownership of them.
+    /// This function is used when creating a virtual NIC that will own the returned queues.
+    fn remove_rx_queues(
```
kevinaboos: "Remove" seems odd, in Rust terms it's used to indicate deletion/freeing. 
I think the rusty term here might be `take`? As in `Option::take()`

kernel/ixgbe/src/regs.rs
```diff
@@ -0,0 +1,671 @@
...
+
+    /// Link Status Register
+    pub links:                          Volatile<u32>,          // 0x42A4;
+
+    /// Auto-Negotiation Control 2 Register    
+    pub autoc2:                         Volatile<u32>,          // 0x42A8;
+    _padding14a:                        [u8; 98],               // 0x42AC - 0x4323
+    _padding14b:                        [u8; 22],
+
+    pub links2:                         Volatile<u32>,          // 0x4324
```
kevinaboos, line 199: missing docs. If intentional and it should be hidden, use `#[doc(hidden)]`

kernel/ixgbe/src/regs.rs
```diff
@@ -0,0 +1,671 @@
...
+    
+    /// Receive Address High
+    pub rah:                            Volatile<u32>,          // 0xA204;
+    _padding3a:                         [u8; 8192],             // 0xA208 - 0xCBFF
+    _padding3b:                         [u8; 2048],
+    _padding3c:                         [u8; 256],
+    _padding3d:                         [u8; 236],
+    _padding3e:                         [u8; 12],
+
+    pub txpbsize:                       [Volatile<u32>;8],      // 0xCC00
```
kevinaboos, line 288: missing docs too

kernel/ixgbe/src/regs.rs
```diff
@@ -0,0 +1,671 @@
...
+    pub srrctl:                         Volatile<u32>,          // 0x1014 //specify descriptor type
+
+    /// Receive Descriptor Tail
+    pub rdt:                            Volatile<u32>,          // 0x1018
+    _padding1:                          [u8;12],                // 0x101C - 0x1027
+
+    /// Receive Descriptor Control
+    pub rxdctl:                         Volatile<u32>,          // 0x1028
+    _padding2:                          [u8;20],                // 0x102C - 0x103F                                            
+} // 64B
```
kevinaboos, line 469: probably worth it to add const_assert_eq!() for the size of RegistersRx/Tx too

kernel/ixgbe/src/virtual_function.rs
```diff
@@ -0,0 +1,72 @@
...
+use nic_queues::{RxQueue, TxQueue};
+use alloc::vec::Vec;
+use network_interface_card::NetworkInterfaceCard;
+
+/// Create a virtual NIC from the ixgbe device.
+/// 
+/// # Arguments
+/// * `ip_addresses`: set of ip addresses that will be assigned to the allocated receive queues.
+/// The number of ip addresses is equal to the number of queue pairs that will be assigned to the vNIC.
+/// Packets with the destination ip addresses specified here will be routed to the vNIC's queues.
```
kevinaboos: nit:  put a couple of spaces of indentation to make it clear this is part of the above bullet point. I know it doesn't show up in rustdoc but it helps when viewing the code. 
```rust
/// * `ip_addresses`: set 
///    The number of ...
///    Packets with ...
```

kernel/ixgbe/src/virtual_function.rs
```diff
@@ -0,0 +1,72 @@
...
+/// Create a virtual NIC from the ixgbe device.
+/// 
+/// # Arguments
+/// * `ip_addresses`: set of ip addresses that will be assigned to the allocated receive queues.
+/// The number of ip addresses is equal to the number of queue pairs that will be assigned to the vNIC.
+/// Packets with the destination ip addresses specified here will be routed to the vNIC's queues. 
+/// * `default_rx_queue`: the queue that will be polled for packets when no other queue is specified.
+/// * `default_tx_queue`: the queue that packets will be sent on when no other queue is specified.
+pub fn create_virtual_nic(
+    ip_addresses: Vec<[u8;4]>,
```
kevinaboos: nit: space `[u8; 4]`

kernel/ixgbe/src/virtual_function.rs
```diff
@@ -0,0 +1,72 @@
...
+pub fn create_virtual_nic(
+    ip_addresses: Vec<[u8;4]>, 
+    default_rx_queue: usize, 
+    default_tx_queue: usize
+) -> Result<
+    VirtualNic<IxgbeRxQueueRegisters, AdvancedRxDescriptor, IxgbeTxQueueRegisters, AdvancedTxDescriptor>, 
+&'static str> {
+
+    let num_queues = ip_addresses.len();
+    if num_queues == 0 {return Err("need to request >0 number of queues");}
```
kevinaboos: nit: one liners are okay, but put spaces between curly braces

kernel/virtual_nic/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    pub fn id(&self) -> u8 {
+        self.id
+    }
+
+    /// Send a packet on the specified queue.
+    #[allow(dead_code)]
+    pub fn send_packet_on_queue(&mut self, qid: usize, transmit_buffer: TransmitBuffer) -> Result<(), &'static str> {
+        if qid >= self.tx_queues.len() {return Err("Invalid qid");}
```
kevinaboos: nit: curly brace spacing. Here and on L82, 90

kevinaboos (changes requested):
Your changes to my comments for Part 1 look good. Moving onto Parts 2 & 3.

Everything looks pretty good. Let's discuss alternatives to unsafety on Slack. 

My main gripe is this: 
In general the const values at the top of `ixgbe` are a mixed group of configurable settings and fixed values that describe the NIC hardware. It's hard to tell what's what. For the items that are configurable by the user or app developer, why are they consts? Can we make them normal variables tied to the NIC object instance?

kernel/ixgbe/src/lib.rs
Ramla-I: good point. I'll file the issue.

kernel/ixgbe/src/lib.rs
Ramla-I: no, the register is named links

kernel/ixgbe/src/lib.rs
Ramla-I: I've used an enum, but still an if-else statement. The match statement won't compile because the compiler won't interpret the enum as a u32.

kernel/ixgbe/src/lib.rs
Ramla-I: Even in the E1000 there's an assumption of one device, but I agree, it shouldn't be the final design. We can create a vec of IxgbeNics here, and register them all in the device manager crate.

kernel/ixgbe/src/lib.rs
Ramla-I: I can improve the init function to take certain parameters for each NIC (like which features are enabled), but resetting them on the fly is a bit more complicated. I would have to see the procedure from the reference manual because it's possible that some changes would require a software reset.
For now I'll just remove some of these constants and make them options at the time of initializing the NIC.

kernel/ixgbe/src/lib.rs
kevinaboos: Yeah you can't avoid a match or if, it's just better to use an enum than raw integers if not every integer value is valid.
I think the way to fix that compiler issue is `#[repr(u32)] `

kernel/ixgbe/src/lib.rs
kevinaboos: That's totally fine. For now, accept parameters for only the config options you know you can change dynamically, for others, specify them as consts but split them off into a separate section of the code that makes it clear they are to be modified by a developer.

Ramla-I:
I've made 2 major changes:

- Moved many of the constants to initialization options. The main changes for that are in the init function. The remaining constants I've specified to be either hardware constraints or developer constrains.

- Multiple Ixgbe NICs can be initialized. The way we identify them is the pci location.

kernel/ixgbe/src/lib.rs
```diff
@@ -232,9 +221,69 @@ impl NetworkInterfaceCard for IxgbeNic {
...
+    /// * `rx_buffer_size_kbytes`: The size of receive buffers. 
+    /// * `num_rx_descriptors`: The number of descriptors in each receive queue.
+    /// * `num_tx_descriptors`: The number of descriptors in each transmit queue.
+    pub fn init(
+        ixgbe_pci_dev: &PciDevice,
+        dev_id: PciLocation,
+        link_speed: LinkSpeedMbps,
+        virtualization_enabled: bool,
+        interrupts: Option<Vec<HandlerFunc>>,
+        rss_enabled: bool,
```
kevinaboos: "rss_enabled" sounds like the caller of this function already knows whether something is enabled, and is telling this function yes or no. 
If that's the case, then keep it this way. Otherwise, perhaps you meant "enable_rss" to indicate that the caller is choosing a setting of whether to enable RSS (?)

kernel/ixgbe/src/lib.rs
```diff
@@ -678,20 +726,25 @@ impl IxgbeNic {
...
+            },
+            LinkSpeedMbps::LS10000 => {
+                let val = regs2.autoc.read() & !(AUTOC_LMS_CLEAR);
+                regs2.autoc.write(val | AUTOC_LMS_10_GBE_S); // value should be 0xC09C_6004
+    
+                let val = regs2.autoc.read() & !(AUTOC_10G_PMA_PMD_CLEAR);
+                regs2.autoc.write(val | AUTOC_10G_PMA_PMD_XAUI); // value should be 0xC09C_6004
+    
+                // let val = regs2.autoc2.read() & !(AUTOC2_10G_PMA_PMD_S_CLEAR);
+                // regs2.autoc2.write(val | AUTOC2_10G_PMA_PMD_S_SFI); // value should be 0xA_0000
```
kevinaboos: is this part not necessary, is it not understood yet, or is it a future addition? Maybe a short comment line explaining why the lines below were commented out would help.

kernel/ixgbe/src/lib.rs
```diff
@@ -1198,10 +1248,49 @@ impl IxgbeNic {
...
+        } else if item == (0x2 << 28) {
+            Self::LS1000
+        } else if item == (0x3 << 28) {
+            Self::LS10000
+        } else {
+            Self::LSUnknown
+        }
+    }
+}
+
```
kevinaboos: Generally the `From` trait is the right idea, but unfortunately this documentation is not available to the caller of this because it's a trait implementation. So there's no way to actually view this documentation. Thus, the caller of this type would most likely expect that calling `LinkSpeedMbps::from(100u32)` would give an object `LS100`, when in reality it will give `LSUnknown`. 

As such, instead of using the `From` trait, I think you just ought to just define a direct conversion function that makes it abundantly clear what the u32 value represents, basically what you already wrote above.

For example:
```rust
impl LinkSpeedMbps {
    [pub] fn from_links_register_value(register_value_or_whatever_its_called: u32) -> LinkSpeedMbps {
        match value {
            1 << 28 => Self::LS100,
            2 << 28 => Self::LS1000,
            3 << 28 => Self::LS10000,
             _ => Self::Unknown,
        }
    }
```

kernel/ixgbe/src/lib.rs
```diff
@@ -1213,29 +1302,29 @@ pub enum FilterProtocol {
 }
 
 /// A helper function to poll the nic receive queues (only for testing purposes).
-fn rx_poll_mq(qid: usize) -> Result<ReceivedFrame, &'static str> {
-    let nic_ref = get_ixgbe_nic().ok_or("ixgbe nic not initialized")?;
+fn rx_poll_mq(qid: usize, nic_id: PciLocation) -> Result<ReceivedFrame, &'static str> {
+    let nic_ref = get_ixgbe_nic(nic_id).ok_or("ixgbe nic not initialized")?;
```
kevinaboos: should this error message be changed to reflect that the requested nic_id didn't exist (or wasn't inited)?

kernel/ixgbe/src/lib.rs
```diff
@@ -1213,29 +1302,29 @@ pub enum FilterProtocol {
...
     let frame = nic.rx_queues[qid as usize].return_frame().ok_or("no frame")?;
     Ok(frame)
 }
 
 /// A helper function to send a test packet on a nic transmit queue (only for testing purposes).
-fn tx_send_mq(qid: usize) -> Result<(), &'static str> {
+fn tx_send_mq(qid: usize, nic_id: PciLocation) -> Result<(), &'static str> {
     let packet = test_packets::create_dhcp_test_packet()?;
-    let nic_ref = get_ixgbe_nic().ok_or("ixgbe nic not initialized")?;
+    let nic_ref = get_ixgbe_nic(nic_id).ok_or("ixgbe nic not initialized")?;
```
kevinaboos: should this error message be changed to reflect that the requested nic_id didn't exist (or wasn't inited)?

kernel/ixgbe/src/test_packets.rs
```diff
@@ -1,11 +1,12 @@
...
-use super::{IXGBE_NIC, NetworkInterfaceCard, TransmitBuffer};
+use super::{get_ixgbe_nic, NetworkInterfaceCard, TransmitBuffer};
+use pci::PciLocation;
 
 /// Sends a dhcp request packet on the ixgbe NIC.
-pub fn dhcp_request_packet() -> Result<(), &'static str> {
+pub fn dhcp_request_packet(nic_id: PciLocation) -> Result<(), &'static str> {
     let transmit_buffer = create_dhcp_test_packet()?;
-    let ixgbe_nc = IXGBE_NIC.try().ok_or("ixgbe NIC hasn't been initialized yet")?;
+    let ixgbe_nc = get_ixgbe_nic(nic_id).ok_or("ixgbe NIC hasn't been initialized yet")?;
```
kevinaboos: same here

kernel/pci/src/lib.rs
```diff
@@ -186,7 +186,7 @@ fn scan_pci() -> Vec<PciBus> {
 
 /// The bus, slot, and function number of a given PCI device.
 /// This offers methods for reading and writing the PCI config space. 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, PartialEq)]
```
kevinaboos: good call, while you're at it you can derive `Eq` too, and `Hash` if we want. Those typically go together.

kevinaboos (changes requested):
looks great! Just a few tiny suggested changes.

kernel/ixgbe/src/lib.rs
Ramla-I: updated get_ixgbe_nic function to return a result that covers the two possibilities

kernel/ixgbe/src/lib.rs
Ramla-I: same as above

kernel/ixgbe/src/test_packets.rs
Ramla-I: same as above

kernel/ixgbe/src/lib.rs
```diff
@@ -256,7 +255,7 @@ impl IxgbeNic {
         if (link_speed == LinkSpeedMbps::LSUnknown) | (link_speed == LinkSpeedMbps::LS100) {
             return Err("Ixgbe driver can only be configured for 1 Gbps or 10 Gbps link speeds");
         }
-        if (virtualization_enabled & interrupts.is_some()) | (virtualization_enabled & rss_enabled) {
+        if (enable_virtualization & interrupts.is_some()) | (enable_virtualization & enable_rss) {
```
kevinaboos: I think this should be a logical boolean AND (`&&`) instead of a bitwise operator (`&`), right?

kevinaboos (changes requested):
just one tiny thing -- double check all of your bitwise operators vs logical conditional operators.

Ramla-I:
fixed the logical/ bitwise operators. I'm surprised Rust doesn't issue a warning for that. 
But, the bitwise operators work the same way as logical ones for boolean values except they don't short circuit.

kevinaboos (approved):

PR #337, opened 2021-01-06
Remove `rlibc` dependency.

Acciente717:
- Remove `rlibc` dependency from `nano_core` and `libtheseus`.
- Instruct `cargo` to build our own `compiler_builtin` crate with the `mem` feature turned on.

### DIFF
--- a/cfg/Config.mk
+++ b/cfg/Config.mk
@@ -21,6 +21,18 @@ CFG_DIR := $(ROOT_DIR)/cfg
 KERNEL_PREFIX ?= k\#
 APP_PREFIX    ?= a\#
 
+
+## Enable the following two unstable flags.
+CARGOFLAGS += -Z unstable-options
+
+## Instruct cargo that we want to build the `compiler_builtins` crate by ourselves,
+## so that we can enable the "mem" feature in our built version.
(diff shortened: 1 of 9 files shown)


cfg/Config.mk
```diff
@@ -21,6 +21,18 @@ CFG_DIR := $(ROOT_DIR)/cfg
...
+CARGOFLAGS += -Z unstable-options
+
+## Instruct cargo that we want to build the `compiler_builtins` crate by ourselves,
+## so that we can enable the "mem" feature in our built version.
+CARGOFLAGS += -Z build-std
+
+## Enable "mem" feature so that memory utility functions (e.g. memcpy) are not
+## mangled. Rust implicitly relies on these functions.
+CARGOFLAGS += -Z build-std-features=compiler-builtins-mem
+
```
kevinaboos, line 35: do we need these? I was under the impression that xargo handled these things for us, and that perhaps the feature option for "compiler-builtins-mem" needs to be passed to xargo instead?

kevinaboos (changes requested):
looks good, except i had one question about the cargoflags stuff.

kevinaboos:
Closed in favor of #338

PR #342, opened 2021-01-19
Redesign mouse pointer shape

pingiun:
While I was looking through the code to understand it when I found the cursor definition, just for fun I redesigned it.

It now looks like this:

image


kevinaboos:
Thanks for the contribution! Our whole graphics stack is kind of ad-hoc (a quick proof of concept) and needs to be rewritten entirely, but this is a nice improvement that i'd been hoping someone would tackle for a while. 

I'll try it out soon and then accept the PR assuming it all works well, since the code looks good to me.

kevinaboos (approved):
Looks good to me! Thanks very much for the contribution!

PR #344, opened 2021-01-21
Fix task_fs offset calculation

pingiun:
The correct slice range is now calculated, previously the end of the slice range did not account for the offset.


kevinaboos (approved):
Thanks again for the fix! Looks good and it works on my end.

PR #346, opened 2021-01-22
Upgrade Rust edition to 2018

pingiun:
I was reading up on editions and wanted to see if Theseus could build on the 2018 edition. After the automated cargo fixes, everything seems to run fine.

I think the `r#try` syntax is quite ugly, and we need to upgrade spin to get rid of it. I have done that in #347

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,7 +22,7 @@ dependencies = [
  "pit_clock",
  "rsdp",
  "rsdt",
- "spin",
+ "spin 0.4.10",
  "volatile",
 ]
 
@@ -62,7 +62,7 @@ dependencies = [
  "memory",
  "scheduler",
  "spawn",
- "spin",
+ "spin 0.4.10",
  "stack",
  "tlb_shootdown",
 ]
@@ -82,7 +82,7 @@ dependencies = [
  "owning_ref",
  "pit_clock",
  "raw-cpuid",
- "spin",
+ "spin 0.4.10",
  "static_assertions",
  "volatile",
  "x86_64",
@@ -101,7 +101,7 @@ dependencies = [
  "log",
  "print",
  "serial_port",
- "spin",
+ "spin 0.4.10",
  "stdio",
  "task",
  "window_manager",
@@ -126,7 +126,7 @@ dependencies = [
  "hpet",
  "log",
  "mpmc",
- "spin",
+ "spin 0.4.10",
  "task",
  "wait_queue",
 ]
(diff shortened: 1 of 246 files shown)


pingiun:
It seems some files have been left with an abundance of newlines, I think it would be nice to fix this by running `cargo fmt` on all the code (and checked in CI). I would be willing to do that, alternatively I could also just clean up the newlines by hand

kevinaboos:
Thanks! Really appreciate the work that went into this!

I've previously considered migrating to the 2018 edition; however, I couldn't really determine any significant benefits (aside from the convenience of not having to use `extern crate` statements in most cases, and (macro) import paths), so it didn't seem worthwhile. Perhaps the 2021/22 edition will be more worth it?
 
Admittedly, I'm not completely informed on differences between editions -- are there significant benefits to moving to this edition? It seems like a ton of changes for no actual benefit. I'd rather delay the changes until we actually _need_ 2018-specific functionality for a given crate. For example, we do have some async/await support crates coming down the line soon, and we obviously need 2018 editions for that.

Mostly, I'm wary of accepting a single huge PR that touches hundreds of files, simply because it creates a significant headache for other folks working on various subsystems who now have to deal with hundreds of merge conflicts. I think it'd be best for a few other large PRs to land (unfortunately private ones at the moment) before considering this huge change. 

I agree with the ugliness of `r#try` syntax too, so I would certainly accept #347 together with this.

pingiun:
Yeah to be quite honest i'm not fully aware of the advantages of 2018 edition either. The work on this was prompted by rust-analyzer complaining about `try` being a keyword. 

I could rework the spin update PR to drop the other changes and only do remove the try keyword usage by upgrading spin. Then we can leave this PR open for a later time.

kevinaboos:
Thanks, i do appreciate the effort you've undergone here and definitely do not want to be dismissive of it. This is a great start and we should definitely merge it in at some point. 

Also, I think that a mass change like this is a good opportunity to do some other minor cleanup tasks, such as re-organizing all Cargo.toml files and making them more consistent (e.g., the `rlib` crate type parts).

kevinaboos:
@pingiun thanks again for making all of these changes. I think I'd be willing to accept this, but i'd like to accept #347 first. We may also need/want to break this PR up into multiple smaller ones, since it's quite hard to review 250+ changed files at the same time.

bytesnail:
Directly upgrade Rust edition to 2021 may be a better idea. Rust edition 2021 will be release in 1.56 soon.

SmiteWindows:
1.56🆗

kevinaboos:
We've decided to close this issue based on the view that it's difficult to justify so many changes for no difference in functionality. Merging this in would unnecessarily complicate the version history and make future merges more difficult.

I think it's best to update each crate to a later edition on a case-by-case basis when needed, e.g., #524.

PR #347, opened 2021-01-22
Upgrade spin crate to 0.9.0

pingiun:
This builds on #346 and additionally upgrades `spin`, so the `r#try` syntax isn't needed anymore.

I choose to vendor @kevinaboos  forks of `owning_ref` and `stable_deref_trait`, because they are quite specialized and it might be nice to just have them in tree? I'm not sure about this, let me know if I should make PRs in those repos instead.

Also I have a question, could stable_deref_trait be replaced by the newly introduced Pin trait or is that a totally different thing? I'm not sure I understand the purpose correctly haha

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ dependencies = [
  "fadt",
  "hpet",
  "ioapic",
- "irq_safety",
+ "irq_safety 0.1.1 (git+https://github.com/theseus-os/irq_safety)",
  "kernel_config",
  "lazy_static",
  "log",
@@ -22,7 +22,7 @@ dependencies = [
  "pit_clock",
  "rsdp",
  "rsdt",
- "spin",
+ "spin 0.7.1",
  "volatile",
 ]
 
@@ -55,14 +55,14 @@ version = "0.1.0"
 dependencies = [
  "apic",
  "interrupts",
- "irq_safety",
+ "irq_safety 0.1.1 (git+https://github.com/theseus-os/irq_safety)",
  "kernel_config",
  "lazy_static",
  "log",
  "memory",
  "scheduler",
  "spawn",
- "spin",
+ "spin 0.7.1",
  "stack",
  "tlb_shootdown",
 ]
@@ -74,7 +74,7 @@ dependencies = [
  "atomic",
  "atomic_linked_list",
  "bit_field 0.7.0",
- "irq_safety",
+ "irq_safety 0.1.1 (git+https://github.com/theseus-os/irq_safety)",
  "kernel_config",
  "lazy_static",
  "log",
@@ -82,7 +82,7 @@ dependencies = [
  "owning_ref",
  "pit_clock",
  "raw-cpuid",
- "spin",
+ "spin 0.7.1",
  "static_assertions",
  "volatile",
  "x86_64",
@@ -101,7 +101,7 @@ dependencies = [
  "log",
  "print",
  "serial_port",
- "spin",
+ "spin 0.7.1",
  "stdio",
  "task",
  "window_manager",
@@ -126,7 +126,7 @@ dependencies = [
  "hpet",
  "log",
  "mpmc",
- "spin",
+ "spin 0.7.1",
  "task",
  "wait_queue",
 ]
@@ -139,7 +139,7 @@ dependencies = [
  "log",
  "pci",
  "port_io",
- "spin",
+ "spin 0.7.1",
  "storage_device",
 ]
 
@@ -279,7 +279,7 @@ dependencies = [
  "exceptions_full",
  "first_application",
  "interrupts",
- "irq_safety",
+ "irq_safety 0.1.1 (git+https://github.com/theseus-os/irq_safety)",
  "kernel_config",
  "keycodes_ascii",
  "log",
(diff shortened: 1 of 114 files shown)


pingiun:
As for reworking this to drop the 2018 edition changes, any comments on vendoring the two libs @kevinaboos ?

kevinaboos:
I like the idea of vendoring, except for the space usage of course. I once explored a related idea of going further and setting up a registry for all Theseus crates so we could avoid specifying each crate's relative path in the Cargo.toml files; if that ever happened, we could add these repos to that registry too. If we do adopt the idea of vendoring, I'd have to see how much space would be used, and I'd prefer to vendor _every_ crate for completeness.

The choice consistent with current project setup would be to shift the ownership of those repos to the Theseus org, which I had previously done for other repos but didn't get around to all of them. The goal of course is to ensure that we're not dependent upon repos that are out of our control.

kevinaboos:
I filed #349 to track this.

kevinaboos (changes requested):
Thanks again for these changes. Nearly everything looks good, and I'd like to accept this PR!

One request -- can you remove the local copies of  `owning_ref` and `stable_deref_trait` in the `libs` directory and the changes to the corresponding `Cargo.toml` files that depend upon those? I would still like to keep those separate, at least for now. That's an unrelated change that shouldn't be a part of this PR, even if we do decide to merge them into the main Theseus repo directly. 

Alternatively, with your permission, I can take these changes and make the suggested fixes myself by pushing them up to a separate branch on the `theseus-os/Theseus` repo. I will also make the requisite changes to the other external libraries that depend on `spin`.

kevinaboos:
@pingiun I had time to handle and close out #349 yesterday. Would you like to make the aforementioned fixes and merge this in, or would you prefer if i went ahead and did it?

pingiun:
I'll take some time to fix up this PR, you can expect it soon

kevinaboos:
> I'll take some time to fix up this PR, you can expect it soon

Thanks, and no rush at all.

pingiun:
The project is currently not building, I'm getting this error:

```
error[E0308]: mismatched types
  --> libs/cow_arc/src/lib.rs:61:28
   |
61 |         MutexGuardRef::new(self.arc.inner_arc.lock())
   |                            ^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `spin::mutex::MutexGuard`, found struct `MutexGuard`
   |
   = note: expected struct `spin::mutex::MutexGuard<'_, T>`
              found struct `MutexGuard<'_, T>`
   = note: perhaps two different versions of crate `spin` are being used?

error[E0308]: mismatched types
  --> libs/cow_arc/src/lib.rs:75:40
   |
75 |             Some(MutexGuardRefMut::new(self.arc.inner_arc.lock()))
   |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `spin::mutex::MutexGuard`, found struct `MutexGuard`
   |
   = note: expected struct `spin::mutex::MutexGuard<'_, T>`
              found struct `MutexGuard<'_, T>`
   = note: perhaps two different versions of crate `spin` are being used?

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0308`.
error: could not compile `cow_arc`
```

Not entirely sure why I wasn't getting this with the previous vendored version of the lib

pingiun:
I think it might be solved with my PRs in https://github.com/theseus-os/owning-ref-rs/pull/1 and https://github.com/theseus-os/stable_deref_trait/pull/1. However I'm not sure how to test it, so let me know if you see any problems

kevinaboos:
Thanks a lot! Yes, i'm looking into this, and I think I'm going to go ahead and upgrade all crates and submodule repos involved in Theseus to use the latest version of `spin`, which is `0.9.0` as of 10 days ago. Might as well move up to the latest version. 

These include (in dependency order):
* `stable_deref_trait`  (extending your PR)
* `owning_ref`  (extending your PR)
* `irq_safety`  (will create a PR myself)

Once I accept the PRs for those first two crates, I also intend to update them with the latest changes from their upstreams, since it's a good time to do so. 
After that, we can update this PR #347 to depend on those newly-changed crates, and then finally accept it.

Thanks again! I'll post an update here soon.

kevinaboos:
@pingiun I've made all of the aforementioned changes. 

We're upgrading to spin 0.9.0, which adds quite a few new features, so the dependency on spin should be specified as such: 
```toml
[dependencies.spin]
version = "0.9.0"
default-features = false
features = ["mutex", "spin_mutex", "rwlock", "once", "barrier"]
```

This is a large number of changes to make, so I'm happy to help. If you want to make all the fixes, that's fine, but I will probably need to do quite a bit of integration checks before merging everything in. Thus, it might be better to change the base branch of this PR to something other than `theseus_main`, but we can do that later. Just let me know. :+1:

kevinaboos:
Actually it appears that it's not necessary to set all of those features for the `spin` crate; I misread their new list of features. We can just keep it simple with `spin = "0.9.0"`. 

I have the working set of changes to upgrade to spin 0.9.0 ready to go on a separate branch. I plan to change the base branch of this PR to another branch, merge my changes in on top of that, and then merge that into `theseus_main`. That way your work won't go uncredited :smiley:

kevinaboos (approved):
Approved, see my prior comment about changing the base branch.

pingiun:
Nice!

PR #360, opened 2021-03-25
Porting Theseus onto ARM MCU

Acciente717:
Currently `memfs` works on ARM microcontroller. The design rationale is:
- Use a dummy frame allocator that always return success on request.
- Use a page allocator to manage memory allocation, i.e., conceptually treat the physical address space on MCU as a virtual address space. The mapping is always identical mapping.

Environment setup:
- xPack QEMU fork: https://xpack.github.io/qemu-arm/
- Add the directory containing the patched QEMU to PATH.
- Nightly rust (cargo) tool chain.
- `rustup target add thumbv7em-none-eabi`

Run:
- `make arm` to build the ARM version.
- `make run-arm` to run with QEMU.


kevinaboos:
thanks a lot! Let's discuss how to proceed with this on Slack so we can collaborate.

kevinaboos:
closed in favor of an identical PR #361  that uses a different branch as the upstream

PR #376, opened 2021-04-20
Avoid accessing PMU-specific MSRs on machines without PMU hardware

kevinaboos:
Move all PMU-specific logic into the `pmu_x86` crate, nothing should be in the `nmi_handler` or anywhere else.

### DIFF
--- a/kernel/pmu_x86/src/lib.rs
+++ b/kernel/pmu_x86/src/lib.rs
@@ -811,7 +811,7 @@ pub fn find_function_names_from_samples(sample_results: &SampleResults) -> Resul
 /// * Returns `Ok(true)` if a PMU sample occurred and was handled. 
 /// * Returns `Ok(false)` if PMU isn't supported, or if PMU wasn't yet initialized, 
 ///   or if there was not a pending sampling interrupt. 
-/// * Returns an `Err` if PMU is supported and initialized and a sample was pendind, 
+/// * Returns an `Err` if PMU is supported and initialized and a sample was pending, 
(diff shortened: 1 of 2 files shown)


kernel/pmu_x86/src/lib.rs
```diff
@@ -793,22 +793,51 @@ pub fn find_function_names_from_samples(sample_results: &SampleResults) -> Resul
...
+/// # Argument
+/// * `stack_frame`: the stack frame that was pushed onto the stack automatically 
+///    by the CPU and passed into the interrupt/exception handler. 
+///    This is used to determine during which instruction the sampling interrupt occurred.
+///
+/// # Return
+/// * Returns `Ok(true)` if a PMU sample occurred and was handled. 
+/// * Returns `Ok(false)` if PMU isn't supported, or if PMU wasn't yet initialized, 
+///   or if there was not a pending sampling interrupt. 
+/// * Returns an `Err` if PMU is supported and initialized and a sample was pendind,
```
Ramla-I: was pending

Ramla-I (dismissed):
Just a minor typo. The conditions at the top of the handle_sample function seem correct.

kernel/pmu_x86/src/lib.rs
kevinaboos: thanks for catching that!

kevinaboos:
sorry I didn't dismiss your review, that was an automatic dismissal due to Githuh branch protection settings. I've disabled that setting. 

All is good now, thanks for the review!

PR #389, opened 2021-05-24
Update macOS build setup script to support newer M1 macs

kevinaboos:
thanks to @shsym for figuring this out and confirming these fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,11 +89,11 @@ If you're on WSL, also do the following steps:
     rm -rf $HOME/theseus_tools_src $HOME/theseus_tools_opt /opt/local/bin/make
     ```
 
-  * If you're building Theseus on an M1-based Mac, you may need to use x86 emulation:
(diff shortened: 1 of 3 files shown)


shsym:
To run the script with Rosetta2, one may need to run:
`arch -x86_64 zsh` (or any other shell)
`export PATH=/usr/local/Homebrew/bin:$PATH` (if both arm and x86 versions of brew binaries are installed, like me)
`sh scripts/mac_os_build_setup.sh`

kevinaboos:
> To run the script with Rosetta2, one may need to run:
> `arch -x86_64 zsh` (or any other shell)
> `export PATH=/usr/local/Homebrew/bin:$PATH` (if both arm and x86 versions of brew binaries are installed, like me)
> `sh scripts/mac_os_build_setup.sh`

Thanks, I'll add that to the "Building on Mac OS" section of the README.

Issue #382, opened 2021-05-14
Use min const generics to re-simplify padding on MMIO struct


kevinaboos:
The newer versions of the `zerocopy` crate [now support this](https://fuchsia.googlesource.com/fuchsia/+/df1162fe6146a1e95ba51b7649119e11ee8b10dc), so we can get rid of this stupid design that breaks padding bytes into multiple arrays with lengths of a supported size. 

For example, the `E1000Registers` struct currently looks like this:
```rust
#[derive(FromBytes)]
#[repr(C)]
pub struct E1000Registers {
    pub ctrl:                       Volatile<u32>,          // 0x0
    _padding0:                      [u8; 4],                // 0x4 - 0x7
    pub status:                     ReadOnly<u32>,          // 0x8
    _padding1a:                     [u8; 128],              // 0xC - 0xBF,  180 bytes
    _padding1b:                     [u8;  52],              
    
    /// Interrupt control registers
    pub icr:                        ReadOnly<u32>,          // 0xC0   
    _padding2:                      [u8; 12],               // 0xC4 - 0xCF
    pub ims:                        Volatile<u32>,          // 0xD0
    _padding3:                      [u8; 44],               // 0xD4 - 0xFF 

    /// Receive control register
    pub rctl:                       Volatile<u32>,          // 0x100
    _padding4a:                     [u8; 512],              // 0x104 - 0x3FF,  764 bytes
    _padding4b:                     [u8; 236],
    _padding4c:                     [u8;  16],             
    
    /// Transmit control register
    pub tctl:                       Volatile<u32>,          // 0x400
    _padding5a:                     [u8; 4096],             // 0x404 - 0x1FFF
    _padding5b:                     [u8; 2048],
    _padding5c:                     [u8; 512],
    _padding5d:                     [u8; 256],
    _padding5e:                     [u8; 128],
    _padding5f:                     [u8; 64],
    _padding5g:                     [u8; 60],
    
} // 2 4KiB pages
```

All of the `_padding5*` fields can be combined together again, into just one `_padding5` of the exact size it needs to be: `[u8; 7164]`.

Apply this to all other MMIO register structs, all of which derive the `zerocopy::FromBytes` trait.

kevinaboos:
Commit c6c8cf1 shows more examples of structs that can be changed back.

PR #390, opened 2021-05-25
Simplify struct padding for MMIO registers using min const generics

kevinaboos:
Addresses issue #382

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2923,7 +2923,7 @@ checksum = "6a02e4885ed3bc0f2de90ea6dd45ebcbb66dacffe03547fadbb0eeae2770887d"
 [[package]]
 name = "volatile"
 version = "0.2.7"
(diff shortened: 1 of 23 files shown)


kevinaboos:
@Ramla-I requesting your review since this touches some e1000 and ixgbe register struct definitions, but doesn't change them. Just want to have another set of eyes on the changes to catch any accidental mistakes.

Ramla-I (approved):

bytesnail:
Would it be better to separate the upgrade of crate 'zerocopy' to another PR?

kevinaboos:
> Would it be better to separate the upgrade of crate 'zerocopy' to another PR?

The stabilization of min const generics is what led to the ugprade of zerocopy, which is what led to us being able to re-simplify the struct padding using byte arrays. So that chain of dependencies dictates that we must upgrade zerocopy along with these changes; plus, they are semantically related too.

Issue #379, opened 2021-05-13
Auto-generate Theseus book on each commit


kevinaboos:
How to do so: https://github.com/rust-lang/mdBook/blob/master/guide/src/continuous-integration.md

apogeeoak:
Hello. I would like to implement this, but I have some questions.

Would it be possible to separate the book into a dedicated repository? A decoupled book makes contributing to the book easier. Orientation is easier as there is less material to orient to. Building is easier and more idiomatic. Testing and the other functionality of `mdBook` are easier to access.

For the actual integration, should it be implemented using Travis, similar to the main project and the linked instructions, or using GitHub Actions as mentioned in the wiki?

kevinaboos:
Hi @apogeeoak, thanks for volunteering to help out! Sorry for the delay, I'm on vacation.

For now I'd like to keep the book in the same repository, as it's easier to link against source code or other scripts/artifacts more easily. We may move it to a separate repo in the future.

I think using github actions to build the documentation (both the book and rustdoc) is a good approach. The Travis yaml file that is in the repo is actually not being used, it's an old artifact that can be removed. I don't have much experience with CI tools myself, so I'm open to advice & recommendations here. Let me know how you'd like to proceed and I'll help out where I can. Thanks!

kevinaboos:
Closed in ea7447a34d3a20f21d352213a36122f56844e9bb

PR #393, opened 2021-06-21
Add GitHub Actions workflow to build and deploy documentation.

apogeeoak:
Added a workflow to build the book documentation on every push or pull request.

The documentation is deployed to the `github_pages` branch on a push to the `theseus_main` branch.

This pull request is a draft because the source documentation build is commented out. Building the source documentation fails when attempting to compile `packed_simd_2`.

Errors:

- error[E0557] feature has been removed
    const_fn
- error: attributes starting with `rustc` are reserved for use by the `rustc` compiler
    #[rustc_args_required_const(2)]

If the workflow is used as is (with the source documentation build commented out) and merged into `theseus_main` the GitHub Pages website will not include the source documentation. This can be resolved by either fixing the build errors, or as a temporary workaround manually including the old source documentation.

Partial fix of #379


kevinaboos:
@apogeeoak thanks so much, this is fantastic! I will look into that packed_simd_2 crate and fix the error there; that's an old dependency that we're not using any more. Ping me again if you don't hear an update about this by the end of the week.

kevinaboos:
Is it necessary to build the source code's rustdoc output into `./docs/` instead of `./build/doc/`? I'm not super familiar with github actions, but that seems like it may be unnecessary. It'd be nice if we could keep all build output in `./build/`

apogeeoak:
It is not necessary for GitHub Actions but GitHub Pages needs to collect all the published source files under one directory. I used the `./docs/` directory because that is how GitHub Pages is currently setup. The `make doc` rule can be changed back to output into `./build/doc/` instead if desired. The files in `./build/doc/` would then need to be copied over to `./docs/doc/` on deploy.

kevinaboos:
> It is not necessary for GitHub Actions but GitHub Pages needs to collect all the published source files under one directory. I used the `./docs/` directory because that is how GitHub Pages is currently setup. The `make doc` rule can be changed back to output into `./build/doc/` instead if desired. The files in `./build/doc/` would then need to be copied over to `./docs/doc/` on deploy.

Okay, thanks for the info. Given that, I think you made the right call, and we can stick with all built documentation being outputted into `./docs/`. :+1:

kevinaboos:
@apogeeoak I've just merged in a fix for the `packed_simd_2` crate, among other things. Kindly rebase this PR and see if things work as expected. 
Note that there are still a bunch of warnings that pop up during the `cargo doc` build, but we can fix those in a separate PR. 
Thanks!

apogeeoak:
I pushed a rebased version of the pull request. I also added a build cache to help occasionally speed up the workflow.

If desired I could also incorporate spell checking into the workflow. The spell checking can be limited to only the book source files and set to only warn on misspellings.

Closes #379

kevinaboos:
Fantastic, thanks so much! This is great. I'll test it on my own machine today or tomorrow and then approve & merge it in. 

I think spell checking is a great idea, but I'm curious how it'd work with a bunch of technical terms that aren't standard English. I'm not familiar with those tools, but if there's a way to add our own custom dictionary then I'd definitely be in favor of it.

apogeeoak:
GitHub Actions need to be run through GitHub but the workflow can be tested on a personal fork of the project in order to keep the main project clean.

I forgot to mention this but in order to publish the docs the GitHub Pages settings might need to be changed. Right now Pages is set to publish the `docs/` directory of the `github_pages` branch. The automatic deployment only deploys the `docs/` directory and removes the nesting so the content is placed on the root directory. Unless GitHub automatically switches directories when the previous directory no longer exists the Pages settings will need to change from the `docs/` directory to the root directory. The branch is kept the same.

For spell checking I use [cspell](https://github.com/streetsidesoftware/cspell), a spell checker for source code. `cspell` is open source and they provide an [extension](https://github.com/streetsidesoftware/vscode-spell-checker) for Visual Studio Code as well as a [GitHub Action](https://github.com/streetsidesoftware/cspell-action). `cspell` includes code specific dictionaries as well configuration files to add valid words.

kevinaboos:
> GitHub Actions need to be run through GitHub but the workflow can be tested on a personal fork of the project in order to keep the main project clean.

True, to be clear I meant just testing the local doc build on my PC.

> I forgot to mention this but in order to publish the docs the GitHub Pages settings might need to be changed. Right now Pages is set to publish the `docs/` directory of the `github_pages` branch. The automatic deployment only deploys the `docs/` directory and removes the nesting so the content is placed on the root directory. Unless GitHub automatically switches directories when the previous directory no longer exists the Pages settings will need to change from the `docs/` directory to the root directory. The branch is kept the same.

That's totally fine, the existing GitHub Pages settings for this repo are old junk and can be completely removed/overwritten. Happy to do it however you see fit; it doesn't have to be in the `docs/` directory, it could be anything else. 

> For spell checking I use [cspell](https://github.com/streetsidesoftware/cspell), a spell checker for source code. `cspell` is open source and they provide an [extension](https://github.com/streetsidesoftware/vscode-spell-checker) for Visual Studio Code as well as a [GitHub Action](https://github.com/streetsidesoftware/cspell-action). `cspell` includes code specific dictionaries as well configuration files to add valid words.

Nice, I'll look into it. I think we can add that in as part of a future PR.

kevinaboos:
@apogeeoak I've made some minor changes to fix the doc- and book-related Makefile targets when building things locally, but everything else looks great. Thanks again for the hard work.

kevinaboos:
@apogeeoak If everything looks good to you, let me know and I can merge it in. I can also adjust any necessary GitHub Pages settings and GitHub actions settings to ensure your new workflow gets used properly. In addition to the doc deployment, it'd be great to include doc/book building as a check on all PRs submitted against `theseus_main`.

Makefile
```diff
@@ -496,23 +495,25 @@ ifneq ($(shell mdbook --version > /dev/null 2>&1 && echo $$?), 0)
 	@echo -e "    cargo +stable install mdbook-linkcheck --force"
 	@exit 1
 endif
-	@mdbook build $(BOOK_SRC)
-	@rm -rf $(BOOK_OUT)
-	@mkdir -p $(BOOK_OUT)
-	@cp -rf $(BOOK_SRC)/book/html/* $(BOOK_OUT)
+	@mdbook build $(BOOK_SRC) -d $(BOOK_OUT)
```
apogeeoak, line 498: Building the book directly to the output directory makes the build command much shorter but adds unneeded nesting to the published output. Since the entire `docs/` directory is published to GitHub Pages the final output for the book would be https://theseus-os.github.io/Theseus/book/html/index.html. The `html/` nesting is avoided by the copy. Another drawback to building directly is that the `linkcheck/` directory is also published to GitHub Pages.

docs/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Generated documentation
+book/
+doc/
```
apogeeoak: Since the entire `docs/` directory is included in the git commit that deploys the documentation this ignore file would cause the `book/` and `doc/` directories to be ignored. Only the `index.html` and `README.md` files would be included.

docs/index.html
```diff
@@ -6,4 +6,4 @@ <h1>Theseus Documentation</h1>
 <p>Welcome to the online hosted docs for Theseus! This is currently a work in progress. </p>
 <p>The top-level Theseus <a href="https://github.com/theseus-os/Theseus/blob/theseus_main/README.md">README</a> is the most up-to-date resource and a great place to get started. </p>
 <p>Source-level <a href="./doc/___Theseus_Crates___/index.html">documentation</a> is also available.</p>
-<p>A higher-level overview of Theseus is available as <a href="./book/index.html">book-style documentation</a>.</p>
+<p>A higher-level overview of Theseus is available as <a href="./book/html/index.html">book-style documentation</a>.</p>
```
apogeeoak: Revert back unless nesting is desired.

Makefile
```diff
@@ -496,23 +495,25 @@ ifneq ($(shell mdbook --version > /dev/null 2>&1 && echo $$?), 0)
...
-	wslview "$(shell realpath --relative-to="$(ROOT_DIR)" "$(BOOK_OUT)")" &
+	wslview "$(shell realpath --relative-to="$(ROOT_DIR)" "$(BOOK_OUT_FILE)")" &
 else
-	@xdg-open $(BOOK_OUT) > /dev/null 2>&1 || open $(BOOK_OUT) &
+	@xdg-open $(BOOK_OUT_FILE) > /dev/null 2>&1 || open $(BOOK_OUT_FILE) &
 endif
 
 
+### Removes all built documentation
+clean_doc:
```
apogeeoak: I would change this to `clean-doc` for consistency.

apogeeoak (commented):
Only the GitHub Pages settings might need to be changed.

The workflow is currently setup to run on every push or pull request. The documentation is only deployed when pushing to `theseus_main` (which includes merging pull requests).

docs/.gitignore
kevinaboos: Oh, interesting! I didn't realize the gitignore was used by GitHub pages to exclude certain content from being published. Thanks for pointing that out!

Makefile
kevinaboos: Good points. I was only considering the local build process when making these changes, not the published Pages actions. Oops.

Makefile
kevinaboos: Agreed.

docs/.gitignore
apogeeoak: GitHub Pages just published whatever is in the branch at the directory specified. But the deployment action uses git which will not commit ignored file. There is probably a workaround but I think the best solution is just to move these directories to the `.gitignore` file on the root directory.

kevinaboos:
Based on your comments above (and thinking more long term), I feel it's best to have two workflows that are somewhat separate:
1. Building and viewing the book and source docs locally
2. Publishing some content on github pages

Right now I'm conflating the two, especially by assuming that (2) above includes only the docs from (1). I imagine in the future we may want to publish other content to GitHub pages beyond just generated docs.

Thus, my new proposal is to create a `github_pages` or `pages` folder in the root, instead of `docs`, and have the documentation copied there from the regular `build/doc/` and `build/book/` folders by the CI action. 

I think this leverages the best of both of our ideas: your idea to clean up what's published and avoid unnecessary content and nesting, and my idea of minimizing local operations and simplifying the Makefile. I also feel that the cleanest option is to have all build artifacts outputted into a single `build` folder, including docs, but I suppose that's just an opinion.

I'll make these changes and then confirm with you that the `docs.yaml` workflow is correct.

docs/.gitignore
kevinaboos: Yes, agreed. I'm making that change today. Thanks so much for clarifying!

kevinaboos:
@apogeeoak I've made the changes described above, thanks for your insightful corrections! Take a look at the `docs.yaml` and let me know if everything looks good.

docs/index.html
kevinaboos: Done, thanks!

apogeeoak:
The published content includes everything in the `docs/` directory, not just the book and source documentation. But I think `github_pages/` is a better name for this directory if other content gets added.

Collecting build artifacts into a single `build/` directory is a good idea. It makes cleaning and other maintenance easier.

The only change I would make would be to separate moving the build artifacts into a dedicated step that is only run on a push to main. Neither directory (`github_pages/book/` or `github_pages/doc/`) should exist before the Action is run but this step removes the directories and moves the build artifacts into `github_pages/`. Move is used instead of copy since it is more efficient and the `build/` directory is destroyed after the Action completes anyway. I pushed a commit with this change.

But other than that everything look good.

kevinaboos:
> The only change I would make would be to separate moving the build artifacts into a dedicated step that is only run on a push to main. Neither directory (`github_pages/book/` or `github_pages/doc/`) should exist before the Action is run but this step removes the directories and moves the build artifacts into `github_pages/`. Move is used instead of copy since it is more efficient and the `build/` directory is destroyed after the Action completes anyway. I pushed a commit with this change.

Great point, thanks for that change! That'll be great when using the actions as a doc-building check for new PRs. 

All looks good to me, I will merge shortly. Thanks again for everything!

kevinaboos (approved):

PR #396, opened 2021-06-25
Logging over UART

jacob-earle:
Implements functionality for uart communication and replaces semihosting logger with uart logger.


kernel/stm32f4_discovery/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+if #[cfg(target_vendor = "stm32f407")] {
+    use stm32f4::stm32f407;
+    use core::cell::RefCell;
+    use cortex_m::interrupt::{self, Mutex};
+    use lazy_static::lazy_static;
+
+    lazy_static!{
+        pub static ref STM_PERIPHERALS : Mutex<RefCell<stm32f407::Peripherals>> = {
+            let p = stm32f407::Peripherals::take().unwrap();
+            Mutex::new(RefCell::new(p))
```
Acciente717: Adding some comments would clarify why here inside a `Mutex` requires another `RefCell`.

kernel/uart/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+                        usart.dr.write(|w| w.dr().bits(u16::from(*byte)));
+                    }
+                });
+                Ok(())
+            }
+        }
+
+        impl SerialPort {
+            pub fn get_uart() ->  SerialPort {
+                uart_init();
```
Acciente717: A `get` method should not do the initialization, otherwise we will end up with initializing it again and again each time we get the UART interface.

Acciente717 (changes requested):
Great job overall! There are a few small points that I'd suggest to change before merging.

1. Create a new crate called `kernel/peripherals` and then move the current`kernel/stm32f4_discovery/` into it as a module, i.e. `kernel/peripherals/stm32f4_discovery`. Then add some cfg gates in `kernel/peripherals/lib.rs` to export functions from the `stm32f4_discovery` module.
2. Move the peripheral initialization from `kernel/uart` to `kernel/peripherals`.

kevinaboos:
> Great job overall! There are a few small points that I'd suggest to change before merging.
> 
> 1. Create a new crate called `kernel/peripherals` and then move the current`kernel/stm32f4_discovery/` into it as a module, i.e. `kernel/peripherals/stm32f4_discovery`. Then add some cfg gates in `kernel/peripherals/lib.rs` to export functions from the `stm32f4_discovery` module.
> 2. Move the peripheral initialization from `kernel/uart` to `kernel/peripherals`.

This is a good idea, but keep it as a separate crate inside kernel/peripherals. In Theseus, we prefer to have many small crates instead of using rust's source-level module system, for the sake of easy evolution & composability.

You can see an example of how I do this in `kernel/acpi`: there's one top level crate called `acpi` and then a bunch of smaller crates within that one.

PR #398, opened 2021-06-28
Add spell check to Actions workflow.

apogeeoak:
The spell check action is configured to only check files in the book directory. This can be changed to include all files or other directories.

The spell check action is set to strict mode except when pushing to `theseus_main`. Strict mode fails on a misspelling. Setting strict mode to mostly on allows pull requests to be marked as failing if there are misspellings but allows `theseus_main` builds to succeed and deploy even with misspellings. Alternatively, strict mode can be set to always on or always off.

The `tools/spellcheck/cspell.json` file is used for configuration. Words can be added to the `cspell` dictionary by adding them to the configuration file.

By default only files modified or added in the commit are checked. This can be changed to check all files.

This commit should be accompanied by another commit that fixes any misspellings or adds any valid unrecognized words to the configuration file. This would prevent the spell check action from failing when an unrelated misspelling is present in a modified file. I can add this other commit (which may affect many files) if this spell check action is desired.


kevinaboos:
@apogeeoak thanks very much for this PR! Yes, I think this is a great idea -- let's enable spell check for _all_ files in the book, not just newly-added files. 

Unfortunately I don't have the bandwidth this week to go through all the book files and spell check them, but hopefully they're not too bad. If you'd like to do so, you could either include spelling fixes as part of this PR, or open up a new PR for spelling fixes only and merge that one in before this one.

apogeeoak:
I can work on fixing the spelling this week. I will submit a different PR with the spelling fixes.

apogeeoak:
Spelling errors were fixed in #405. Another commit was added to this pull request to add additional words to the spellcheck dictionary. The documentation build action on this commit is failing since #405 is not merged in yet.

kevinaboos (changes requested):
With the addition of #405, everything here looks good to me, except one thing. The `tools/` directory is intended for programs used in building, running, or testing Theseus. Since the `cspell.json` file only applies to the book source, I think it belongs in `book/cspell.json` instead of in `tools/`.

apogeeoak:
Moved configuration file to book directory. Added spellcheck section to book readme.

kevinaboos (approved):

kevinaboos:
@apogeeoak thanks again, everything looks great!

PR #400, opened 2021-06-28
Simplify view-book makefile target.

apogeeoak:
Utilize the functionality of mdbook to open the Theseus book. The make dependency on book can be removed as the book is rebuild before opening. The check for a mdbook install would also be removed in this case. A somewhat useful error is still printed if mdbook cannot be found.

The `build` command could also be replaced with a `watch` command which rebuilds the book whenever the source changes.

```shell
@(&> /dev/null mdbook watch --open $(BOOK_SRC) -d $(BOOK_OUT) &)
```

The method used to suppress the output may not be portable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -502,11 +502,7 @@ endif
 ### Opens the Theseus book.
 view-book: book
 	@echo -e "Opening the Theseus book in your browser..."
-ifneq ($(IS_WSL), )
-	wslview "$(shell realpath --relative-to="$(ROOT_DIR)" "$(BOOK_OUT_FILE)")" &
-else
-	@xdg-open $(BOOK_OUT_FILE) > /dev/null 2>&1 || open $(BOOK_OUT_FILE) &
-endif
+	@mdbook build --open $(BOOK_SRC) -d $(BOOK_OUT)
 
 
 ### Removes all built documentation
--- a/Makefile
+++ b/Makefile
@@ -591,6 +591,7 @@ help:
 	@echo -e "\t Builds the Theseus book using the mdbook Markdown tool."
 	@echo -e "   view-book:"
 	@echo -e "\t Builds the Theseus book and then opens it in your default browser."
+	@echo -e "\t If the book doesn't open in your browser, install the latest version of mdbook."
 	@echo -e "   clean-doc:"
 	@echo -e "\t Remove all generated documentation files."
 	@echo ""
--- a/book/README.md
+++ b/book/README.md
@@ -8,7 +8,7 @@ The book is written in Markdown and uses [mdBook](https://rust-lang-nursery.gith
 
 ## Building the book
 
-First, install `mdbook`:
+First, install `mdbook`, version `0.4.13` or higher:
 ```sh
 cargo +stable install mdbook
 ```


kevinaboos:
This is a great idea, I'd love to avoid dealing with platform-specific things like `wslview` or `xdg-open`. 

Unfortunately, `mdbook build --open` doesn't work for me on WSL or WSL2, which is important because several other folks working on Theseus do use WSL. Perhaps there's some environment variable to set when using `mdbook build --open` in order to allow it to work properly on WSL, but I'm not sure what the solution is. 

It appears to use the `open` crate; that functionality was added to `mdBook` in this commit: https://github.com/rust-lang/mdBook/commit/3ffd24df631fba46fefe750719732624e3b8c4d2

If we can get it to work on WSL (and ideally macOS too), then I'm 100% in favor of merging this in. Also, you gave me the idea to check whether rustdoc supports this, and they do too using `cargo doc --open`.  Unfortunately, like `mdbook build --open`, the cargo doc one also doesn't work on WSL, at least for me.

kevinaboos:
Looking at the source code of the `open` crate, the version used by mdBook is 1.1, which doesn't support WSL. The newer version of `open` 1.7 _does_ support WSL using the `wslview` executable, so once mdBook is updated to use that, then we can merge this in.

apogeeoak:
I was unaware that WSL was not supported. I will try to submit a request to mdBook or rustdoc to update.

kevinaboos:
Looks like the necessary PR was merged into mdBook last week: https://github.com/rust-lang/mdBook/pull/1643

Pending testing on WSL, I think we can unblock this. :+1:

apogeeoak:
If mdBook is installed or updated through `cargo install mdbook` then `mdbook build --open` works on WSL. The latest binary release is not yet updated with the new functionality.

`cargo doc --open` now works under the latest stable compiler version.

kevinaboos:
> If mdBook is installed or updated through `cargo install mdbook` then `mdbook build --open` works on WSL. The latest binary release is not yet updated with the new functionality.
> 
> `cargo doc --open` now works under the latest stable compiler version.

Great to hear! I think I can unblock this now thanks to your confirmation. Also thanks a lot for pushing this through the mdBook repo too!

kevinaboos:
Will test soon on my WSL setup and then merge if everything works.

kevinaboos:
This does work on WSL if we install mdbook from their git repo, but not if we just install it from crates.io, which makes sense because mdbook hasn't been published to crates.io since August.

I'm going to re-mark this as blocked until mdbook publishes an update to crates.io.

apogeeoak:
If you did not want to wait for a new `mdbook` release, the `open` crate provides a small binary with the open functionality. The `open` crate is on a quick release cycle and it could also be used for opening the source documentation. However, this would add another dependency.

kevinaboos:
That's true, however, using the `open` crate isn't as much of an improvement (if any) over the current direct call to `wslview` as using a built-in `--open` flag in `mdBook` or `cargo doc`. No rush, we can wait on this until things stabilize.

kevinaboos:
Oh, funnily enough it looks like mdBook was just updated: https://github.com/rust-lang/mdBook/commit/3e91f9cd5d5e3104e61cbc82f2a18ab9d63eecab

I'll unblock this and test again soon.

kevinaboos:
Tested and working with the latest mdbook version. Thanks again for the contribution, @apogeeoak!

kevinaboos (approved):

apogeeoak:
You are welcome.

PR #403, opened 2021-07-01
Cortex-m4 Logging

jacob-earle:
Adding logging functionality over the UART on the STM32F4-Discovery board (excludes changes necessary to run Theseus on ARM boards)


Acciente717 (commented):
Great work! All looks good to me.

kernel/boards/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+//! This crate does not implement any functionality itself, but instead contains a subcrate that implements the platform-specific functionality for each ARM microcontroller we support.
```
kevinaboos: Needs line breaks

kernel/boards/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "boards"
+version = "0.1.0"
+authors = ["Jacob Earle <jacob.earle@yale.edu>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
kevinaboos: Remove

kernel/boards/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "boards"
+version = "0.1.0"
+authors = ["Jacob Earle <jacob.earle@yale.edu>"]
```
kevinaboos: Add brief `description`

kernel/boards/stm32f4_discovery/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "stm32f4_discovery"
+version = "0.1.0"
+authors = ["Jacob Earle <jacob.earle@yale.edu>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
kevinaboos: Remove.

Add `description`

kernel/boards/stm32f4_discovery/src/lib.rs
```diff
@@ -0,0 +1,31 @@
+//! Implements platform specific functionalities for the STM32F4_Discovery board, exposing its peripherals and providing ways of interacting with them.
```
kevinaboos: Line breaks

kernel/boards/stm32f4_discovery/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+            let p = stm32f407::Peripherals::take().unwrap();
+            MutexIrqSafe::new(p)
+        };
+    }
+
+    pub mod uart;
+
+    /// Initializes device peripherals for use.
+    pub fn init_peripherals () {
+    }
```
kevinaboos, line 40: Use the `todo!` or `unimplemented! ` macro here, whichever is appropriate

kernel/boards/stm32f4_discovery/src/uart/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+
+impl SerialPort {
+    pub fn new(uart_address: *const usart1::RegisterBlock) ->  SerialPort {
+        SerialPort {
+            uart_address,
+        }
+    }
+}
+
+// /// Macro used to print a formatted string over a serial port on the UART
```
kevinaboos: Is this old? If so, remove.

kernel/logger/Cargo.toml
```diff
@@ -12,7 +12,7 @@ log = "0.4.8"
 path = "../serial_port"
 
 [dependencies.irq_safety]
-git = "https://github.com/theseus-os/irq_safety"
+path = "../../libs/irq_safety"
```
kevinaboos: This is outdated I think

kernel/serial_port/serial_port_x86/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+build = "../../../build.rs"
+
+[dependencies]
+spin = "0.9.0"
+bare-io = { version = "0.2.1", features = [ "alloc" ] }
+
+[dependencies.port_io]
+path = "../../../libs/port_io"
+
+[dependencies.irq_safety]
```
kevinaboos: This doesn't exist, use the git dependency

kevinaboos (changes requested):
Thanks @jacob-earle and @Acciente717. Looks good in general! 

Let's discuss the crate organization on Slack; for example, I think the `uart` module should be its own crate.

Also, kindly remove the addition of the `irq_safety` module from `libs/` and submit a separate PR against that repo so we can add arm support to it. There shouldn't really be any code duplication needed there except for the part about held interrupts.

Makefile
```diff
@@ -446,6 +446,33 @@ preserve_old_modules:
...
+		-o target/thumbv7em-stm32f407-theseus-eabi/theseus.bin \
+		target/thumbv7em-stm32f407-theseus-eabi/release/libnano_core.a
+
+### Run the build for Cortex-m4 with QEMU.
+run-arm: arm
+	@qemu-system-gnuarmeclipse \
+		--board STM32F4-Discovery \
+		-nographic \
+		-semihosting-config enable=on,target=native \
+		-kernel target/thumbv7em-none-eabi/theseus.bin
```
kevinaboos: for these targets, kindly use the existing Makefile variables in order to standardize targets across the whole build system.  For example, look at the x86_64 invocation of cargo:
```
	RUST_TARGET_PATH="$(CFG_DIR)" RUSTFLAGS="$(RUSTFLAGS)" cargo build $(CARGOFLAGS) $(BUILD_STD_CARGOFLAGS) $(RUST_FEATURES) --all --target $(TARGET)
```
Here you'll just set the proper variables for `TARGET` and also change `--all` to `-p nano_core`

kernel/boards/stm32f4_discovery/src/uart/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+static USART1_SERIAL_PORT: Once<MutexIrqSafe<SerialPort>> = Once::new();
+static USART2_SERIAL_PORT: Once<MutexIrqSafe<SerialPort>> = Once::new();
+static USART3_SERIAL_PORT: Once<MutexIrqSafe<SerialPort>> = Once::new();
+static USART6_SERIAL_PORT: Once<MutexIrqSafe<SerialPort>> = Once::new();
+
+
+/// Initialize UART for use.
+pub fn uart_init(uart_address: *const usart1::RegisterBlock) {
```
kevinaboos: a safe function cannot accept an arbitrary pointer and then dereference it. This either has to be marked as unsafe (which is undesirable IMO) or accept a different argument such as a reference to or an owned instance of the `RegisterBlock` type

kernel/boards/stm32f4_discovery/src/uart/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    };
+    sp.call_once(|| {
+        uart_init(address_pointer);
+        MutexIrqSafe::new(SerialPort::new(address_pointer))
+    })
+}
+
+/// The `SerialPort` struct implements the Write trait that is necessary for use with uprint! macro.
+pub struct SerialPort {
+    uart_address: *const usart1::RegisterBlock,
```
kevinaboos: Do we need to store raw pointers everywhere? Can we not obtain an instance of `usart1::RegisterBlock` or at least a reference to it? (Sorry if dumb, i'm not familiar with the svd2rust APIs)

kernel/boards/stm32f4_discovery/src/uart/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+} 
+
+unsafe impl Send for SerialPort {}
```
kevinaboos: this is almost always a bad idea, unless you're implementing a mutual exclusion type yourself.

kernel/serial_port/serial_port_arm/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "serial_port_arm"
+version = "0.1.0"
+authors = ["Jacob Earle <jacob.earle@yale.edu>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
kevinaboos: remove

kernel/serial_port/serial_port_arm/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+#[macro_use]
+extern crate cfg_if;
+
+cfg_if ! {
+    if #[cfg(target_vendor = "stm32f407")] {
+        extern crate stm32f4_discovery;
+        pub use stm32f4_discovery::uart::{get_serial_port, SerialPort, SerialPortAddress};
+    } 
+
+    // Dummy implementation for when no physical device is present, in which case semihosting will be used
```
kevinaboos: Please add a block of crate-level documentation explaining what semihosting is and why the complex target_vendor-specific configuration is required here.

kernel/serial_port/serial_port_x86/src/lib.rs
```diff
@@ -0,0 +1,305 @@
+//! Support for basic serial port access, including initialization, transmit, and receive.
```
kevinaboos: In this file, there is now a strange mix of 2-space tabs and actual tab characters. In keeping with the rest of the project, please use 4-space tab indentation, as it was before.

kernel/serial_port/src/lib.rs
```diff
@@ -1,305 +1,20 @@
...
-//! 3. `0x08` again. 
-//!
-//! This isn't necessarily a bad idea, as it "clears out" whatever character was there before,
-//! presumably to prevent rendering/display issues for a deleted character. 
-//! But, this isn't required, and I personally believe it should be handled by a higher layer,
-//! such as a shell or TTY program. 
-//! We don't do anything like that here, in case a user of this crate wants to send binary data
-//! across the serial port, rather than "smartly-interpreted" ASCII characters.
-//! 
+//! This crate provides an abstraction on top of the separate serial port implementations for x86 and ARM, located in `./serial_port_x86` and `./serial_port_arm`.
```
kevinaboos: Excellent job with this crate! This is exactly the kind of abstraction I was envisioning. 

We could strengthen it further with a serial port trait to guarantee that all arch-specific implementations provide the same API, but that can be done in the next PR.

kevinaboos (changes requested):
Re-reviewing since some changes have been made.

- the `TriState` concept is unnecessarily duplicated, let's move it up a level into a common entity
- there seems to actually be 3 types of support for serial ports now: x86, arm (stm device), and arm semihosting. The crate/module structure should reflect that instead of stuffing the semihosting code into the top-level `lib.rs` file in serial_port_basic_arm
- Some Serial port-related code is actually in `kernel/boards/stm.../src/uart/mod.rs`, which is quite confusing. Not sure why that code is there instead of in a serial port crate. Is there a true distinguishable difference between UART/USART on arm and a serial port on arm?
- Is there only a single serial port / UART device on all ARM devices? If so, do you need the `SerialPortAddress` enum concept?
- There's an API mismatch between the x86 serial port basic and the arm ones. You've covered the implementation of `fmt::Write`, but that's pretty much it. It might be worthwhile to introduce a top-level `SerialPortTrait` that defines the minimal set of functions that each arch-specific Serial Port instance should implement, in order to ensure API completeness and matching.

kevinaboos:
Closing due to inactivity and the fact that the Cortex-M4 port will likely not be upstreamed.

PR #404, opened 2021-07-06
mlx5 

Ramla-I:
Two crates are added in this PR:
- mlx5 crate which contains the starting initialization steps of the driver
- mellanox_ethernet crate which defines structs used by the driver


book/src/running/virtual_machine/pci_passthrough.md
```diff
@@ -52,6 +52,13 @@ e.g. `echo 15b3 1019 > /sys/bus/pci/drivers/vfio-pci/new_id`
 
 Now, QEMU can be launched with direct access to the device.
 
+### Return device to the Host OS
+To reset the device, you can either reboot the system or return the device to the host OS using the following commands (replacing $slot_info with the value previously retrieved):
```
kevinaboos: for consistency, surround "$slot_info" with backticks:
```
`$slot_info`
```

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+    SignatureErr        = 0x1,
+    TokenErr            = 0x2,
+    BadBlockNumber      = 0x3,
+    BadOutputPointer    = 0x4,
+    BadInputPointer     = 0x5,
+    InternalErr         = 0x6,
+    InputLenErr         = 0x7,
+    OutputLenErr        = 0x8,
+    ReservedNotZero     = 0x9,
+    BadCommandType      = 0x10, //Should this be 10 or 16??
```
kevinaboos: potentially confusing comment

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+    BadInputLen         = 0x50,
+    BadOutputLen        = 0x51,
+    BadResourceState    = 0x10,
+    BadPkt              = 0x30,
+    BadSize             = 0x40,
+    Unknown
+}
+
+/// Possible values of the opcode modifer when the opcode is ManagePages
+pub enum ManagePagesOpmod {
```
kevinaboos: is `Opmod` a standard abbreviation used by mellanox? If not, i think something a bit longer would be far more clear, e.g., `OpcodeMod` or just the full `OpcodeModifier`

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+    /// Physically-contiguous command queue entries
+    entries: BoxRefMut<MappedPages, [CommandQueueEntry]>,
+    /// Per-entry boolean flags to keep track of which entries are in use
+    available_entries: Vec<bool>,
+    /// A random number that needs to be different for every command, and the same for all mailboxes that are part of a command.
+    token: u8,
+    /// Pages used for input mailboxes 
+    mailbox_buffers_input: Vec<Vec<(MappedPages, PhysicalAddress)>>, 
+    /// Pages used for output mailboxes 
+    mailbox_buffers_output: Vec<Vec<(MappedPages, PhysicalAddress)>>
```
kevinaboos: out of curiosity, are the number of command queue entries known statically, or must they be determined at runtime?

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+    mailbox_buffers_input: Vec<Vec<(MappedPages, PhysicalAddress)>>, 
+    /// Pages used for output mailboxes 
+    mailbox_buffers_output: Vec<Vec<(MappedPages, PhysicalAddress)>> 
+}
+
+impl CommandQueue {
+
+    /// Create a command queue object.
+    ///
+    /// # Arguments
```
kevinaboos: Moving forward, let's use `##` instead of `#` for section headers in doc comments. It's a lot easier to read and seems to be what the rust stdlib does.

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+    /// * `entries`: physically contiguous memory that is mapped as a slice of command queue entries.
+    /// * `num_cmdq_entries`: number of entries in the queue.
+    pub fn create(entries: BoxRefMut<MappedPages, [CommandQueueEntry]>, num_cmdq_entries: usize) -> Result<CommandQueue, &'static str> {
+        
+        // initially, all command entries are available
+        let mut available_entries = Vec::with_capacity(num_cmdq_entries);
+        for _ in 0..num_cmdq_entries { available_entries.push(true); }
```
kevinaboos: probably more performant to use the `vec![true; num_cmdq_entries]` macro

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+    }
+
+    /// Fill in the fields of a command queue entry.
+    /// At the end of the function, the command is ready to be posted using the doorbell in the initialization segment. 
+    /// Returns an error if no entry is available to use.
+    ///
+    /// # Arguments
+    /// * `opcode`: opcode for command that the driver wants to execute
+    /// * `opmod`: opcode modifer, only applicable for certain commands
+    /// * `allocated_pages`: physical address of pages that need to be passed to the NIC. Only used in the MANAGE_PAGES command.
```
kevinaboos: doc comment here should link to the `ManagePages` variant of the `CommandOpcode` enum

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+
+        // Assign a random number to token, another OS (Snabb) starts off with 0xAA
+        let token = 0xAA;
+
+        Ok(CommandQueue{ entries, available_entries, token, mailbox_buffers_input, mailbox_buffers_output })
+    }
+
+    /// Find an command queue entry that is not in use
+    fn find_free_command_entry(&self) -> Option<usize> {
+        self.available_entries.iter().position(|&x| x == true)
```
kevinaboos: A more rusty (and faster) way to do this is to return a reference to the entry itself rather than its index, assuming it's possible to change the code structure to use that directly. 

The `Iterator::find` function may be better for that too.

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+
+        // set the ownership bit to HW-owned.
+        // This bit will be cleared when the command is complete.
+        cmdq_entry.change_ownership_to_hw();
+
+        match opcode {
+            CommandOpcode::EnableHca => {
+                cmdq_entry.set_input_length_in_bytes(12);
+                cmdq_entry.set_output_length_in_bytes(8);
+                cmdq_entry.set_input_inline_data(opcode, opmod, None, None);
```
kevinaboos: For each one of these `cmdq_entry.set...` functions, there's probably a better way to represent it than just a bunch of constants, it's a bit hard to understand the logic. For example, a table or enum of length values per command opcode, perhaps a function implemented on the `CommandOpcode` enum.

kevinaboos: something something "rule of representation", as Lin would say :)

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+
+        Ok(entry_num)
+    }
+
+    /// Sets the mailbox pointer in a command entry with the physical address of the first mailbox.
+    fn set_mailbox_pointer_in_cmd_entry(&mut self, cmdq_entry: &mut CommandQueueEntry, entry_num: usize, mailbox_type: MailboxType) {
+        let mailbox_ptr = if mailbox_type == MailboxType::Input {
+            self.mailbox_buffers_input[entry_num][0].1.value()
+        } else {
+            self.mailbox_buffers_output[entry_num][0].1.value()
```
kevinaboos: in coordination with my other comments about avoiding indexing and avoiding passing around `entry_num`, this one has a double index and should definitely be restructured. 

Here's, I think it'd be better to make this an associated function (rather than a method that takes `&mut self`) and pass a mutable reference to the `mailbox_buffers_*` and then modify that parameter directly. 

you may wish to break it up into two functions, one for `MailboxType::Input` and one for `Output`

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+        };
+        cmdq_entry.input_mailbox_pointer_h.write(U32::new((mailbox_ptr >> 32) as u32));
+        cmdq_entry.input_mailbox_pointer_l.write(U32::new((mailbox_ptr & 0xFFFF_FFFF) as u32));
+    }
+
+    /// Initialize output mailboxes for the QUERY_ISSI command.
```
kevinaboos: anywhere the comments mention a command, I think it's best to link to the Rust enum/struct version of that rather than plaintext (which i assume is styled according to the mellanox manual). 

Consider that most people reading these docs won't have the mellanox manual in front of them

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+        while self.entries[entry_num].owned_by_hw() {}
+        self.available_entries[entry_num] = true;
+        (self.entries[entry_num].get_delivery_status(), self.entries[entry_num].get_return_status())
+    }
+
+    /// Get the current ISSI version and the supported ISSI versions, which is the output of the QUERY_ISSI command.  
+    pub fn get_query_issi_command_output(&self, entry_num: usize) -> Result<(u16, u8), &'static str> {
+        self.check_command_output_validity(entry_num, CommandOpcode::QueryIssi)?;
+
+        // TODO: This offset is not correct! need to check!
```
kevinaboos: please address before merging

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+        write!(f, "CQE:command_input_inline_data_0: {:#X} \n", self.command_input_inline_data_0.read().get())?;
+        write!(f, "CQE:command_input_inline_data_1: {:#X} \n", self.command_input_inline_data_1.read().get())?;
+        write!(f, "CQE:command_output_status: {:#X} \n", self.command_output_status.read().get())?;
+        write!(f, "CQE:command_output_syndrome: {:#X} \n", self.command_output_syndrome.read().get())?;
+        write!(f, "CQE:command_output_inline_data_0: {:#X} \n", self.command_output_inline_data_0.read().get())?;
+        write!(f, "CQE:command_output_inline_data_1: {:#X} \n", self.command_output_inline_data_1.read().get())?;
+        write!(f, "CQE:output_mailbox_pointer_h: {:#X} \n", self.output_mailbox_pointer_h.read().get())?;
+        write!(f, "CQE:output_mailbox_pointer_l: {:#X} \n", self.output_mailbox_pointer_l.read().get())?;
+        write!(f, "CQE:output_length: {} \n", self.output_length.read().get())?;
+        write!(f, "CQE:token_signature_status_own: {:#X} \n", self.token_signature_status_own.read().get())
```
kevinaboos: This technically works, but isn't flexible -- it doesn't support Rust's various fmt modifiers like `{:?}` vs `{:#?}` vs `{:#X?}`, etc. 

The proper way to do it is to use the fmt debug helper functions, like `debug_struct()` and the `field()` functions. Search the code base for `.debug_struct` to see examples of this. 

links: https://doc.rust-lang.org/core/fmt/struct.Formatter.html#method.debug_struct

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+            0x101   => {CommandOpcode::QueryAdapter}, 
+            0x102   => {CommandOpcode::InitHca}, 
+            0x103   => {CommandOpcode::TeardownHca}, 
+            0x104   => {CommandOpcode::EnableHca}, 
+            0x105   => {CommandOpcode::DisableHca}, 
+            0x107   => {CommandOpcode::QueryPages}, 
+            0x108   => {CommandOpcode::ManagePages}, 
+            0x10A   => {CommandOpcode::QueryIssi}, 
+            0x10B   => {CommandOpcode::SetIssi}, 
+            _       => {CommandOpcode::Unknown}
```
kevinaboos: best to implement this as a conversion trait:

```rust
impl From<integer_type> for CommandOpcode { ... }
```

kevinaboos: Then you can simply call `.from` or `.into` here

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+        }
+    }
+
+    /// Returns the first 16 bytes of output data that are written inline in the command.
+    /// The valid values for each field are different for every command, and can be taken from Chapter 23 of the PRM.
+    fn get_output_inline_data(&self) -> (u8, u32, u32, u32) {
```
kevinaboos: could be confusing to deal with 4-element tuples of similar values.... perhaps place these into a struct? Depends on usage, I suppose.

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+        } else if status == 7 {
+            CommandDeliveryStatus::InputLenErr
+        } else if status == 8 {
+            CommandDeliveryStatus::OutputLenErr
+        } else if status == 9 {
+            CommandDeliveryStatus::ReservedNotZero
+        } else if status == 10 {
+            CommandDeliveryStatus::BadCommandType
+        } else {
+            CommandDeliveryStatus::Unknown
```
kevinaboos: similar to my above comment, use a conversion trait (e.g., `From`) that incorporates a match statement instead of a series of if...else statements.

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+        } else if status == 0x51 {
+            CommandReturnStatus::BadOutputLen
+        } else if status == 0x10 {
+            CommandReturnStatus::BadResourceState
+        } else if status == 0x30 {
+            CommandReturnStatus::BadPkt
+        } else if status == 0x40 {
+            CommandReturnStatus::BadSize
+        } else {
+            CommandReturnStatus::Unknown
```
kevinaboos: same comment here about conversion traits and match statements

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+}
+
+
+#[derive(FromBytes)]
+#[repr(C)]
+/// Layout of mailbox used to pass extra input and output command data that doesn't fit into the command entry.
+struct CommandInterfaceMailbox {
+    /// Data in the mailbox
+    mailbox_data:           Volatile<[u8; 512]>,
+    _padding:               ReadOnly<[u8; 48]>,
```
kevinaboos: does this `_padding` field actually need to be read from? If not, just specify it as a plain array like we usually do: `_padding: [u8; 48]`

kernel/mellanox_ethernet/src/command_queue.rs
```diff
@@ -0,0 +1,740 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        for i in 0..(512/4) {
+            let data = self.mailbox_data.read();
+            write!(f, "mailbox data: {:#X} {:#X} {:#X} {:#X} \n", data[i*4], data[i*4+1], data[i*4+2], data[i*4 +3])?;
+        }
+        
+        write!(f, "next pointer h: {:#X} \n", self.next_pointer_h.read().get())?;
+        write!(f, "next pointer l: {:#X} \n", self.next_pointer_l.read().get())?;
+        write!(f, "block number: {:#X} \n", self.block_number.read().get())?;
+        write!(f, "token ctrl signature: {:#X} \n", self.token_ctrl_signature.read().get())
```
kevinaboos: see my other comment about `.debug_struct()`

kernel/mellanox_ethernet/src/lib.rs
```diff
@@ -0,0 +1,138 @@
+//! This crate defines the layout of memory objects that make up the software interface between the Mellanox hardware and the driver,
+//! as well as functions to access different fields of these objects.
+//!
+//! The Mellanox ethernet card is referred to as both the NIC (Network Interface Card) and the HCA (Host Channel Adapter).
+//!
+//! All information is taken from the Mellanox Adapters Programmer’s Reference Manual (PRM) [Rev 0.54], unless otherwise specified.
```
kevinaboos: can you provide a link to the PRM? Or is it under NDA

kernel/mellanox_ethernet/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+//! as well as functions to access different fields of these objects.
+//!
+//! The Mellanox ethernet card is referred to as both the NIC (Network Interface Card) and the HCA (Host Channel Adapter).
+//!
+//! All information is taken from the Mellanox Adapters Programmer’s Reference Manual (PRM) [Rev 0.54], unless otherwise specified. 
+
+ #![no_std]
+ #![feature(slice_pattern)]
+ #![feature(core_intrinsics)]
+ #![allow(dead_code)] //  to suppress warnings for unused functions/methods
```
kevinaboos: Generally this is a bad idea because unused code is typically a design/impl error; if truly needed, put it as an attribute on a specific item instead of a crate-wide attribute.

kernel/mellanox_ethernet/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    internal_timer_l:           ReadOnly<U32<BigEndian>>,
+    _padding4:                  [u8; 8],
+    /// Advanced debug information
+    health_counter:             ReadOnly<U32<BigEndian>>,
+    _padding5:                  [u8; 44],
+    real_time:                  ReadOnly<U64<BigEndian>>,
+    _padding6:                  [u8; 12228],
+}
+
+// const_assert_eq!(core::mem::size_of::<InitializationSegment>(), 16400);
```
kevinaboos: uncomment (?)

kernel/mellanox_ethernet/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+impl fmt::Debug for InitializationSegment {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Initialization Segment \n")?;
+        write!(f, "Firmware: {}.{}.{}, command interface: {} \n", self.fw_rev_major.read().get(), self.fw_rev_minor.read().get(), self.fw_rev_subminor.read().get(), self.cmd_interface_rev.read().get())?;
+        write!(f, "Command queue address: {:#X} {:#X} \n", self.cmdq_phy_addr_high.read().get(), self.cmdq_phy_addr_low.read().get())?;
+        write!(f, "Command doorbell vector: {:#X} \n", self.command_doorbell_vector.read().get())?;
+        write!(f, "Initializing state: {:#X} \n", self.initializing_state.read().get())
```
kevinaboos: use `.debug_struct()`

kernel/mellanox_ethernet/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+/// The possible values of the initialization state of the device as taken from the intialization segment.
+pub enum InitializingState {
+    NotAllowed = 0,
+    WaitingPermetion = 1, // Is this a typo in the PRM?
```
kevinaboos: perhaps check linux or another system implementation? "Permetion" definitely sounds like a typo, lol

kernel/mlx5/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+authors = ["Ramla Ijaz <ijazramla@gmail.com>"]
+name = "mlx5"
+description = "Support for the Mellanox connectx-5 NIC and driver"
```
kevinaboos: nit: is this crate not the driver itself? Your comment makes it sound like this is a support crate and the driver might be elsewhere

kernel/mlx5/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Command Queue
+    command_queue: CommandQueue,
+    /// Boot pages passed to the NIC. Once transferred, they should not be accessed by the driver.
+    boot_pages: Vec<MappedPages>,
+    /// Init pages passed to the NIC. Once transferred, they should not be accessed by the driver.
+    init_pages: Vec<MappedPages>,
+}
+
+
+/// Functions that setup the NIC struct.
```
kevinaboos, line 69: i'm not sure if `///` triple slash doc comments atop an implementation block actually appear anywhere. Does this throw a warning when building the documentation?

kevinaboos (changes requested):
I think it's rustfmt convention to place the `#[derive()]` attribute items between the doc comments and the struct/enum/function definition, like so:
```rust
/// Command opcode written by SW in opcode field of the input data in the command entry.
#[derive(PartialEq, Debug)]
pub enum CommandOpcode { ... }
```
There are a few places where this should be changed; it's easier to read, especially if doc comments are very very long. 

Also, in general, it's best to avoid indexing operations (e.g., `my_list[i]`) because it requires an expensive bound check and panic branch to be inserted. That's why Rust-style iterators are preferred over C-style loops. 

On a related note, it seems that nearly every functions basically works by indexing into a list of something using `entry_num`, which is expensive. It'd be better (more rusty, safe, etc) to pass around mutable references to entries or other lists directly rather than having each function perform many indexing operations itself. This may require refactoring some of the functions to _not_ take `&mut self` as a parameter and just take references to the specific data fields they need; this is actually preferred as it follows the principle of least privilege for each associated function. 

Finally, I think it's best to just go ahead and start with strongly-typed errors now rather than to add them later. This does require a bit of effort but is much better than returning `&'static str` everywhere. Check out my examples in the `block_io` crate for how to do this, and also how to implement conversions to `&'static str` such that other functions can easily call these using the `try` (`?`) operator.

kernel/mellanox_ethernet/src/command_queue.rs
Ramla-I: It's the standard. Opcode Modifier is used in only a couple of place in the text, everywhere else it's opmod

kernel/mellanox_ethernet/src/command_queue.rs
Ramla-I: they are determined at runtime

kernel/mellanox_ethernet/src/command_queue.rs
kevinaboos: Ok. How about `OpMod` then?

kernel/mellanox_ethernet/src/lib.rs
Ramla-I: I can provide a link to an older version, the one I have I got directly from Nvidia.

kernel/mellanox_ethernet/src/command_queue.rs
kevinaboos: Ok cool. I was wondering if we could use const generics here, but we cannot.

kernel/mellanox_ethernet/src/lib.rs
kevinaboos: that's good. Explain that link in the comments and then also maybe provide a link on how to obtain the one under NDA

kernel/mlx5/src/lib.rs
Ramla-I: No, and the comment appears at the top of the implementation part of the docs

kernel/mellanox_ethernet/src/command_queue.rs
Ramla-I: The bits that are actually going to be used from each field are different for each command. It's a private function and is only used within other functions of the command queue that then extract the relevant bit fields.
I don't personally think it's too confusing, but it wouldn't take much effort to change this to a struct.

kernel/mlx5/src/lib.rs
kevinaboos: oh very cool! I didn't know about that, perhaps it's a new addition to rustdoc.

kernel/mellanox_ethernet/src/command_queue.rs
kevinaboos: Your choice. If the valid values change per command, I'd recommend the use of an enum to represent the possible sets of valid values, such that you can more strongly enforce correctness using typing. The raw values could be converted into that enum and then returned from either this function or a surrounding one.

book/src/running/virtual_machine/pci_passthrough.md
```diff
@@ -52,6 +52,13 @@ e.g. `echo 15b3 1019 > /sys/bus/pci/drivers/vfio-pci/new_id`
 
 Now, QEMU can be launched with direct access to the device.
 
+### Return device to the Host OS
+To reset the device, you can either reboot the system or return the device to the host OS using the following commands (replacing `$slot_info` with the value previously retrieved):  
+```sh
+echo 1 >/sys/bus/pci/devices/$slot_info/remove    
+echo 1 >/sys/bus/pci/rescan
```
kevinaboos: to address this, you can either add "rescan" as a known word in our dictionary (or at the end of this file), or you can disabled checking for this code block. 

I assume there's a way to disable spell check for all code blocks, that setting should be enabled by default. Perhaps @apogeeoak knows how we could do that?

apogeeoak: I created a pull request #415 to disable spell check for code blocks.

kevinaboos: thanks so much!

Ramla-I:
I've completed all the changes except for the refactoring of the CommandQueue methods to avoid indexing into a  list in every function. That portion of the code will be revised in the upcoming PRs for the mlx5 driver.

kernel/mellanox_ethernet/src/command_queue.rs
kevinaboos: Postponed for a future PR.

kevinaboos (approved):

PR #405, opened 2021-07-06
Fix spelling errors in book source files.

apogeeoak:
Most additional valid words are added to the `cspell.json` configuration file from #398. When a word is only applicable to a single file, html comments are used to ignore the word or temporarily disable spell checking. These comments are not displayed but are still present in the html source.

`intro.md` was changed to `index.md` to align better with the makefile output and remove the unnecessary duplication of the content into `intro.html` and `index.html`.

While checking the spelling I noticed some [markdown lint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) errors. Most of the errors are blank line or other whitespace errors. There is also some trailing whitespace in some files. Trailing whitespace can be problematic in markdown files since ending a line with two or more spaces creates a line break which is difficult to see in the editor.

I can create another pull request which fixes these lint errors if desired.


kevinaboos:
LGTM! Thanks very much for fixing these typos and mistakes. 

Regarding the markdown lint warnings, I actually used to use that extension but disabled it because I felt its formatting recommendations were way too strict, especially regarding whitespace. If there are any really egregious errors exposed by the markdown lint, then we can fix those, but I generally feel that it's not worth the effort.

kevinaboos (approved):

PR #415, opened 2021-07-23
Disable spell check for markdown code blocks.

apogeeoak:
Everything between fenced code blocks is ignored by `cspell`. This change makes many of the words added to the `ignoreWords` array unnecessary.


kevinaboos (approved):
Thanks so much!

Issue #350, opened 2021-01-27
Generalize and merge `VirtualAddress` and `PhysicalAddress`


kevinaboos:
Good for code deduplication and establishing a consistent interface.

There are three sets of structures we should merge:

- [x] `VirtualAddress` and `PhysicalAddress`
- [x] `Page` and `Frame`
- [x] `PageRange` and `FrameRange`

kevinaboos:
Closed in b09ad9b and 6854306

PR #417, opened 2021-07-27
Generalize VirtualAddress and PhysicalAddress into a single macro implementation

kevinaboos:
No functional changes. 

Also, return an `Option` to indicate that `VirtualAddress::new()` or `PhysicalAddress::new()` failed.


kevinaboos:
Addresses #350 , but we should also do the same for `Page` and `Frame`, and `PageRange` and `FrameRange`.

PR #424, opened 2021-08-12
Adjust safety of deallocate() functions

### LINKED ISSUE #414: Safety of slabmalloc's deallocate()
Hello, I was testing a static analysis tool that validates unsafe usage with Theseus and would like to share the finding from it.

https://github.com/theseus-os/Theseus/blob/2bc18d736ac0a910cf747e45bde8b1f91c700fb1/kernel/slabmalloc/src/sc.rs#L311-L327

https://github.com/theseus-os/Theseus/blob/2bc18d736ac0a910cf747e45bde8b1f91c700fb1/kernel/slabmalloc/src/zone.rs#L188-L194

`slabmalloc` and `slabmalloc_unsafe` libraries expose `SCAllocator::deallocate` and `ZoneAllocator::deallocate` as public safe APIs. However, they transmute the argument pointer to a slab page without check. Thus, they have a safety invariant on the caller side that affects memory safety and should probably be defined as unsafe functions.

Their impact is minimal, since Theseus is the only user of the libraries, and as far as I can tell all use in Theseus (multiple_heaps) seem correct. Nevertheless, I think defining them as unsafe might better express the API safety and conform to Rust's safety rules.

I wonder if the maintainers are willing to accept a PR that updates the signature of those functions. The update can be limited to `slabmalloc` and `slabmalloc_unsafe`, or it can additionally include `slabmalloc_safe` for symmetricity.

Thanks!

qwaz:
This PR adjusts the declared safety of `deallocate()` functions. It also made `SCAllocator` private to its corresponding crate since the allocation/deallocation functionality is exposed through `ZoneAllocator` APIs. Fixes #414.

### DIFF
--- a/kernel/slabmalloc/src/lib.rs
+++ b/kernel/slabmalloc/src/lib.rs
@@ -32,7 +32,6 @@ mod sc;
 mod zone;
 
 pub use pages::*;
-pub use sc::*;
 pub use zone::*;
 
 use core::alloc::Layout;
--- a/kernel/slabmalloc/src/sc.rs
+++ b/kernel/slabmalloc/src/sc.rs
@@ -88,11 +88,6 @@ impl<'a, P: AllocablePage> SCAllocator<'a, P> {
         new_sc_allocator!(size)
     }
 
-    /// Returns the maximum supported object size of this allocator.
-    pub fn size(&self) -> usize {
-        self.size
-    }
-
     /// Add page to partial list.
     fn insert_partial_slab(&mut self, new_head: &'a mut P) {
         self.slabs.insert_front(new_head);
--- a/kernel/slabmalloc/src/sc.rs
+++ b/kernel/slabmalloc/src/sc.rs
@@ -308,7 +308,11 @@ impl<'a, P: AllocablePage> SCAllocator<'a, P> {
     /// May return an error in case an invalid `layout` is provided.
     /// The function may also move internal slab pages between lists partial -> empty
     /// or full -> partial lists.
-    pub fn deallocate(&mut self, ptr: NonNull<u8>, layout: Layout) -> Result<(), &'static str> {
+    ///
+    /// # Safety
+    /// The caller must ensure that `ptr` argument is returned from [`Self::allocate()`]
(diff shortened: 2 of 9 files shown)


kevinaboos:
hi @Qwaz , thanks so much for the pull request!  I've requested @Ramla-I to take a look at this since she was the original implementer of these various slabmalloc alternative crates, which are primarily used for benchmarking. 

Note that each of these slabmalloc variants is toggled on or off by means of cfg settings; have you ensured that all 3 variants build and run correctly after your changes? The cfg combinations are shown here in the `multiple_heaps` crate:
```rust
#[cfg(all(not(unsafe_heap), not(safe_heap)))]
extern crate slabmalloc;

#[cfg(unsafe_heap)]
extern crate slabmalloc_unsafe;

#[cfg(safe_heap)]
extern crate slabmalloc_safe;
```

You can enable them like so:
```sh
make run THESEUS_CONFIG="safe_heap"
```

kevinaboos:
@Ramla-I  One alternative would be to remove these crates entirely since they're not used for anything beyond paper benchmarks. We could preserve them in the `old_crates` folder.

qwaz:
Thanks for the feedback @kevinaboos. Yes, I've tested all 3 variants when creating a PR by using `RUSTFLAGS`, and ran them again with `make build THESEUS_CONFIG="[safe|unsafe]_heap"` after seeing your comment.

kevinaboos:
Great, thanks! I appreciate the extra testing. 

@Ramla-I is traveling through the weekend but will review these changes next week once she returns.

Ramla-I:
@Qwaz everything looks good to me!

Ramla-I (approved):
Everything looks good.

Ramla-I:
@kevinaboos, I agree that we should probably just remove these crates and preserve them. We can do it after this PR is merged in so that these fixes are applied first.

kevinaboos:
Sounds good, I think moving them to another directory will help clean things up.

kevinaboos:
Thanks again @Qwaz

PR #427, opened 2021-08-13
Switch Docker container to Ubuntu 21.04

### LINKED ISSUE #426: `target` and  `build` directory leak into the docker container
Hi. I tried to use docker method, but got an error where `demangle_readelf_file` requires `GLIBC_2.33` from `libc.so.6`. The problem is that Ubuntu 18.04 only supports `GLIBC_2.27`.

zetanumbers:
FIxes #426

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,7 @@
 ### that is sufficient to build Theseus and run it using QEMU.
 
 
-FROM ubuntu:18.04
+FROM ubuntu:21.04
 LABEL author=kevinaboos@gmail.com
 
 


kevinaboos:
Hi @ZetaNumbers, thanks for the PR! 

If this fixes your issue, then this change is okay with me. However, I think it'd be best to go with a LTS version of the distro instead of a short release; does 20.04 work too? If so, let's go with 20.04 instead.

zetanumbers:
Sadly no. I tried it at first but it wasn't enough with GLIBC_2.31. This may be the rust issue but I didn't check that.

kevinaboos:
Ok thanks for checking. The docker image may be out of date in other aspects too, I don't think any of us actually use it for development. Let me look into it and see if there's another problem beyond just the Ubuntu version, since 18.04 definitely works on my other machines.

kevinaboos:
@ZetaNumbers I just tried building Theseus within Docker on a completely fresh machine, and everything worked properly. Thus, the docker image is still working even with Ubuntu 18.04 as the base image. 

Are there any other differences you can think of? Perhaps you can do a full docker image build, run, and then a Theseus build and upload the entire log here?

zetanumbers:
https://github.com/theseus-os/Theseus/issues/426#issuecomment-898678819

PR #439, opened 2021-08-24
Preliminary Intel VT-d IOMMU support

NicholasLindsay:
# Overview

Adds preliminary driver support for an Intel IOMMU implementing Intel VT-d (Intel Virtualization for Directed I/O). Currently all we do is detect the IOMMU, and if detected, print out some useful information. This sets the groundwork for the rest of the IOMMU implementation.

This preliminary driver performs the following:
* Initialize the kernel IOMMU object if it is detected in the ACPI DMAR table (see `acpi::init()` and `iommu::early_init()`)
   * Map the IOMMU's memory-mapped registers into virtual memory
* Output the interpreted contents of the following IOMMU registers to the log:
  * `Version Register`
  * `Capability Register`
  * `Extended Capability Register`
  * `Global Status Register`
* Further initialization will be performed by calling `iommu::init()` from  `device_manager::init()` - although currently `iommu::init()` doesn't do anything useful.

Additionally, the driver provides a mechanism for sending commands to the IOMMU `Global Command Register` (and waiting until the command is complete).

# Testing

To test IOMMU support, pass `IOMMU=yes` when invoking the Makefile QEMU target (`orun` etc). Details about the IOMMU will be printed to the log.

Please test with and without `IOMMU=yes`.

# Outstanding issues:
* Printing to the log the interpreted contents of the `Capability Register` and `Extended Capability Register` produces multiple lines, of which only the first line has the log prefix (e.g. `[I] ...`)
* Only a single IOMMU is supported. Support for multiple IOMMUs can be added in the future.
* Further initialization in `device_manager::init()` currently doesn't do anything useful but is there because I anticipate it will be needed in the future.

### DIFF
--- a/kernel/iommu/src/regs.rs
+++ b/kernel/iommu/src/regs.rs
@@ -1,12 +1,11 @@
-//! This file contains structs that simplify programming the IOMMU.
+//! Structures needed for interacting with the IOMMU.
 
 use zerocopy::FromBytes;
 use volatile::{ReadOnly, WriteOnly};
 use bitflags::bitflags;
 use core::fmt;
 
-/// Struct which allows direct access to memory mapped registers when
-/// overlayed over corresponding page.
+/// The layout of the IOMMU's MMIO registers.
 #[derive(FromBytes)]
 #[repr(C)]
 pub struct IntelIommuRegisters {
@@ -26,7 +25,8 @@ pub struct IntelIommuRegisters {
     _unimplemented:         [u8; 4096-0x20],   // 0x20-0xFFF
 }
 // TODO: Hardware may use more than 4kB, which means the registers may occupy
-// more than one contiguous page.
+//       more than one contiguous page.
+//       Currently we assume the IOMMU registers occupy only a single page.
 const_assert_eq!(core::mem::size_of::<IntelIommuRegisters>(), 4096);
 
 /// Helper struct for decoding and printing capability register
@@ -180,12 +180,8 @@ pub enum GlobalCommand {
 bitflags! {
     /// Global status register flags.
     ///
-    /// The lowest 22 bits are `RsvdZ`. This is Intel parleance meaning that
-    /// they may be used in the future, but for now, all writes to these
-    /// bits must have the value 0. Since these bits may have values in the
-    /// future, the `from_bits_truncate()` method should be used to construct
-    /// this object from the underlying bit representation. Note of course
-    /// that this conversion is not invertible.
+    /// The least significant bits `[22:0]` are `RsvdZ`,
+    /// meaning that they are reserved for future usage and must be set to 0.
     pub struct GlobalStatus: u32 {
         /// Compatibility Format Interrupt Status
         const CFIS  = 1 << 23;
--- a/kernel/iommu/src/regs.rs
+++ b/kernel/iommu/src/regs.rs
@@ -3,6 +3,7 @@
 use zerocopy::FromBytes;
 use volatile::{ReadOnly, WriteOnly};
 use bitflags::bitflags;
+use core::fmt;
 
 /// Struct which allows direct access to memory mapped registers when
 /// overlayed over corresponding page.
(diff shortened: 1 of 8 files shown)


NicholasLindsay:
This PR is very much a work-in-progress and there will be multiple iterations until IOMMU support is fully ready. I am publishing this PR early in the hope that I can get some early feedback on coding style (this is my first PR for Theseus with actual code).

kernel/iommu/src/regs.rs
```diff
@@ -0,0 +1,212 @@
...
+        writeln!(f, "PSI:     {}", self.psi())?;
+        writeln!(f, "NFR:     {}", self.nfr())?;
+        writeln!(f, "MAMV:    {}", self.mamv())?;
+        writeln!(f, "DWD:     {}", self.dwd())?;
+        writeln!(f, "DRD:     {}", self.drd())?;
+        writeln!(f, "FL1GP:   {}", self.fl1gp())?;
+        writeln!(f, "PI:      {}", self.pi())?;
+        writeln!(f, "FL5LP:   {}", self.fl5lp())?;
+        writeln!(f, "ESIRTPS: {}", self.esirtps())?;
+        writeln!(f, "ESRTPS:  {}", self.esrtps())?;
```
kevinaboos: for this, check out [`debug_struct`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct)

kevinaboos (commented):
The code style so far looks quite good! I don't see any glaring problems. I left a few small comments, nothing major. I'll let you guide the progress of this PR; feel free to reach out to me on Slack as well. 

There's probably a better way to represent the `Capability` and `ExtendedCapability` structs, but what you've done thus far isn't necessarily wrong or bad. I'd need to see more about how those types are used in order to suggest something better, but you could always look at `bitfield` or `bitflags` for inspiration. I will say that if you do stick with this approach, it's probably clearer and more efficient to test a specific bit or a range of bits rather than right-shift the values by that bit index and then test against `0x1`, but I'm not sure if it actually compiles down to something much different. 

My primary question is whether it's necessary to split IOMMU initialization into `early_init()` and regular `init()`. That's only really needed when we have to break cyclic dependencies, e.g., two subsystems that partially depend on each other. If not, then i'd just stick with one regular `init()` function for simplicity's sake.

NicholasLindsay:
Is it okay to force-push changes after a local rebase?

kevinaboos:
> Is it okay to force-push changes after a local rebase?

Sure, that's no problem on your own private local branch.

NicholasLindsay:
> > Is it okay to force-push changes after a local rebase?
> 
> Sure, that's no problem on your own private local branch.

For more safety better to use `--force-with-lease`, even on your own private local branch.

NicholasLindsay:
And is it okay to use `git commit --amend` to update PRs in response to comments? Or do you prefer separate commits (perhaps with a squash at the end)?

kevinaboos:
> And is it okay to use `git commit --amend` to update PRs in response to comments? Or do you prefer separate commits (perhaps with a squash at the end)?

Oh yeah, don't worry about commit history. Pushing new individual commits is much easier for me to track. I'll squash everything later when it's time to merge anyway. 

It sounds like you intend to make many more commits; what else are you planning for this PR?

NicholasLindsay:
I agree with your advice; there is no need for separate `early_init()` and`init()` functions - at least not currently. I have merged these two functions together into a single `init()` function which is called from `acpi::init()`.

NicholasLindsay:
I haven't decided exactly how I want to structure the PRs yet. My initial thought was that the push into the mainline could happen once the IOMMU is up and running and is providing isolation between devices and the CPU.

NicholasLindsay:
> I haven't decided exactly how I want to structure the PRs yet. My initial thought was that the push into the mainline could happen once the IOMMU is up and running and is providing isolation between devices and the CPU.

There are likely to be many commits between now and then due to the general complexity of IOMMUs. If you prefer, I can provide smaller, more incremental PRs.

NicholasLindsay:
> The code style so far looks quite good! I don't see any glaring problems. I left a few small comments, nothing major. I'll let you guide the progress of this PR; feel free to reach out to me on Slack as well.
> 
> There's probably a better way to represent the `Capability` and `ExtendedCapability` structs, but what you've done thus far isn't necessarily wrong or bad. I'd need to see more about how those types are used in order to suggest something better, but you could always look at `bitfield` or `bitflags` for inspiration. I will say that if you do stick with this approach, it's probably clearer and more efficient to test a specific bit or a range of bits rather than right-shift the values by that bit index and then test against `0x1`, but I'm not sure if it actually compiles down to something much different.
> 
> My primary question is whether it's necessary to split IOMMU initialization into `early_init()` and regular `init()`. That's only really needed when we have to break cyclic dependencies, e.g., two subsystems that partially depend on each other. If not, then i'd just stick with one regular `init()` function for simplicity's sake.

Perhaps something like this would be clearer for checking bit values: `(self.0) & (1 << 63) != 0`? It's slightly more convoluted than the C version since in Rust there is no implicit conversion from `u64` to `bool`.

NicholasLindsay:
> The code style so far looks quite good! I don't see any glaring problems. I left a few small comments, nothing major. I'll let you guide the progress of this PR; feel free to reach out to me on Slack as well.
> 
> There's probably a better way to represent the `Capability` and `ExtendedCapability` structs, but what you've done thus far isn't necessarily wrong or bad. I'd need to see more about how those types are used in order to suggest something better, but you could always look at `bitfield` or `bitflags` for inspiration. I will say that if you do stick with this approach, it's probably clearer and more efficient to test a specific bit or a range of bits rather than right-shift the values by that bit index and then test against `0x1`, but I'm not sure if it actually compiles down to something much different.
> 
> My primary question is whether it's necessary to split IOMMU initialization into `early_init()` and regular `init()`. That's only really needed when we have to break cyclic dependencies, e.g., two subsystems that partially depend on each other. If not, then i'd just stick with one regular `init()` function for simplicity's sake.

`bitflags` won't work since there are some fields than span multiple bits. And I had a look at `bitfield` but I thought it looked rather cumbersome. Additionally there are some IOMMU register fields which when read require further arithmetic (e.g. `fn nfr(&self)     -> u64  { ((self.0 >> 40) & 0xff) + 1 }`), which cannot be done cleanly with `bitfield`.

I have used the `bitflags!` macro for `GlobalStatus` since each IOMMU register field consists of a single bit. And I haven't used `bitflags!` at all for `GlobalCommand` since it hasn't been necessary, although it could be added in the future if needed.

NicholasLindsay:
> > The code style so far looks quite good! I don't see any glaring problems. I left a few small comments, nothing major. I'll let you guide the progress of this PR; feel free to reach out to me on Slack as well.
> > There's probably a better way to represent the `Capability` and `ExtendedCapability` structs, but what you've done thus far isn't necessarily wrong or bad. I'd need to see more about how those types are used in order to suggest something better, but you could always look at `bitfield` or `bitflags` for inspiration. I will say that if you do stick with this approach, it's probably clearer and more efficient to test a specific bit or a range of bits rather than right-shift the values by that bit index and then test against `0x1`, but I'm not sure if it actually compiles down to something much different.
> > My primary question is whether it's necessary to split IOMMU initialization into `early_init()` and regular `init()`. That's only really needed when we have to break cyclic dependencies, e.g., two subsystems that partially depend on each other. If not, then i'd just stick with one regular `init()` function for simplicity's sake.
> 
> Perhaps something like this would be clearer for checking bit values: `(self.0) & (1 << 63) != 0`? It's slightly more convoluted than the C version since in Rust there is no implicit conversion from `u64` to `bool`.

Stylistically I actually prefer the original version, e.g. `(self.0 >> 63) & 0x1 == 0x1`. As you said, with a good compiler, there shouldn't be much of a difference, since they are equivalent.

NicholasLindsay:
I have done the following:
* Replace `(self.0) >> 63) & 0x1 == 0x1` style bit checks with `(self.0) & (1 << 63) != 0` which should enable more compile time optimization.
* Remove `fmt::Display` implementations for `iommu::regs::Capability` and `iommu::regs::ExtendedCapability` and add `fmt::Debug` implementations, making use of `debug_struct`.

If there are no outstanding issues then I declare this PR ready-to-go, from my point of view.

kevinaboos:
Note to myself: this looks good, pending some minor changes that I plan to make.

kevinaboos (approved):

PR #441, opened 2021-08-26
Fix: Overflow in TSC (Timestamp counter)

Narasimha1997:
Fixed the panic due to overflow of 64-bit unsigned integer by extending the size to 128 bits. The changes are reflected on `kernel/tsc/src/lib.rs`, also the `kernel/libterm` crate depends on the value from timestamp counter for some internal calculation, so the necessary changes are also made in this crate (updated the size of blinking frequency from 64 to 128 bits).

### DIFF
--- a/kernel/tsc/src/lib.rs
+++ b/kernel/tsc/src/lib.rs
@@ -11,13 +11,14 @@ pub struct TscTicks(u128);
 
 impl TscTicks {
     /// Converts ticks to nanoseconds. 
+    ///
     /// Returns `None` if the TSC tick frequency is unavailable 
-    /// or if overflow occured.
+    /// or if overflow occured during the conversion.
     pub fn to_ns(&self) -> Option<u128> {
         get_tsc_frequency()
             .ok()
             .and_then(|freq| self.0.checked_mul(1000000000)
-                    .map(|checked_tsc| checked_tsc / freq)
+                .map(|checked_tsc| checked_tsc / freq)
             )
     }
 
--- a/kernel/tsc/src/lib.rs
+++ b/kernel/tsc/src/lib.rs
@@ -7,15 +7,18 @@ use core::sync::atomic::{AtomicUsize, Ordering};
 
 
 #[derive(Debug)]
-pub struct TscTicks(u64);
+pub struct TscTicks(u128);
 
 impl TscTicks {
     /// Converts ticks to nanoseconds. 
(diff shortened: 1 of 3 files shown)


kevinaboos:
Thanks for the PR, @Narasimha1997! This does seem to address the issue partially, but it doesn't really fix the core problem of overflow possibly occurring. I realize overflow is highly unlikely with a 128-bit integer, but we might as well address that possibility now. 

For that, we'd need to use some kind of overflow-aware mathematical operation, e.g., wrapping, saturating, or a checked multiplication/addition operation in all functions that operate on TSC values. A checked operation would probably be my recommendation, unless you had compelling reasons to use wrapping ones instead. Let me know what you think.

Narasimha1997:
@kevinaboos 
Yes, we can use `checked_mul` in `to_ns` function of `TscTicks` and keep `u128` as it is. Anyways the `to_ns` is gonna return an option, so it returns `None` in case of overflow. This the only place I see where overflow check is required right now.

Narasimha1997:
@kevinaboos 
I have added `checked_mul` in the latest commit as discussed.

kernel/tsc/src/lib.rs
```diff
@@ -7,15 +7,19 @@ use core::sync::atomic::{AtomicUsize, Ordering};
...
-    pub fn to_ns(&self) -> Option<u64> {
-         get_tsc_frequency().ok().map(|freq| {
-            (self.0 * 1000000000) / freq 
-         })
+    pub fn to_ns(&self) -> Option<u128> {
+        if let Some(freq) = get_tsc_frequency().ok() {
+            return self.0.checked_mul(1000000000).map(|checked_tsc| {
+                checked_tsc / freq
+            });
+        }
```
kevinaboos: Kindly update the docs as well to reflect the new conditions under which `None` may be returned. Also, let's stick with the combinator format, something like this:
```
get_tsc_frequency()
    .ok()
    .and_then(|freq| self.0.checked_mul(1000000000)
        .map(|checked_tsc| checked_tsc / freq)
    )
```

kevinaboos (changes requested):
Looks good; I left one comment about code style. Ready to merge after that. :+1:

Narasimha1997:
Made all the changes and rebased it to a single commit.

kevinaboos:
LGTM, I tested it and can confirm it works in debug build mode. Thanks again!

kevinaboos (approved):

PR #460, opened 2021-11-09
Support TLS sections in dynamically-loaded crate object files

kevinaboos:
Completes TLS support, which was introduced in 3e6c50a


kevinaboos:
See the https://github.com/rust-lang/rust/issues/29594 tracking issue for updated information about the `#[thread_local]` attribute.

PR #470, opened 2021-12-04
Remove the dependency from `heap` -> `memory_initialization`

kevinaboos:
* Adds a new `bootloader_modules` crate to break the unnecessary dependency chain from `mod_mgmt` -> `memory_initialization`.

This allows, for example, to use a bunch of memory-related or memory-dependent crates from most foreign crates. For one, the `heap` crate can now depend on `task` as @NamiLiy requested.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -254,6 +254,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "bootloader_modules"
+version = "0.1.0"
+dependencies = [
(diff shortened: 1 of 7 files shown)


NamiLiy (approved):
Can now recall task::get_my_current_task() fro the heap.
Thus satisfies my requirement.

kevinaboos:
Glad to hear it works. Will merge shortly.

PR #472, opened 2021-12-14
Support for running wasm-wasi binaries

vikrammullick:
This PR includes `applications/wasi_interpreter`, a wasi interpreter that interfaces between **Theseus** and **wasmi** and currently supports the following wasi system calls:

| System Call | Notes |
|-------------|-------|
| ProcExit|| 
|FdClose||
|FdWrite||
|FdSeek||
|FdRead||
|FdFdstatGet||
|EnvironSizesGet||
|EnvironGet||
|FdPrestatGet||
|FdPrestatDirName||
|PathOpen||
|FdFdstatSetFlags||
|ArgsSizesGet||
|ArgsGet||
|ClockTimeGet|TODO: return UNIX timestamp (currently returns 0)|

Additionally, this PR includes the following applications:
`applications/wgrep`: grep-2.0 compiled to wasi-wasm and run with wasi-interpreter [source](https://ftp.wayne.edu/gnu/grep/)
`applications/wzork`: zork compiled to wasi-wasm and run with wasi-interpreter [source](https://github.com/devshane/zork), [about](https://en.wikipedia.org/wiki/Zork#Zork_and_Dungeon)
`applications/wpee`: pee (command to write to files) compiled to wasi-wasm and run with wasi-interpreter [source](https://wapm.io/package/liftm/pee)
`applications/wasmi_simple_test`: Test application to verify that `wasmi` successfully runs in `no_std` environment
`applications/gen_test_files`: Helper application to generate test text files (opening crawls from the Star Wars Saga)


kevinaboos:
Thanks again for submitting this; I've left some comments on Slack.

kernel/wasi_interpreter/src/wasi_definitions.rs
```diff
@@ -0,0 +1,221 @@
...
+//! This module contains the following:
+//! * Macros for easily defining wasmi function signatures.
+//! * SystemCall enum type consisting of supported system calls.
+//! * Mapping from system call string to SystemCall type.
+//! * Mapping between system call number and SystemCall type.
+//! * Mapping from SystemCall type to wasmi signature.
+//! * Definitions of WASI rights for full file and directory permissions.
+//!
+//! Signature macro from tomaka/redshirt:
+//! https://github.com/tomaka/redshirt/blob/main/kernel/core/src/primitives.rs
```
kevinaboos: Good citation here, but I think this should be a permalink to a direct commit rather than the blob/main/ link. I think that github links like this will die if the main branch is updated to change/remove the `primitives.rs` file

kernel/wasi_interpreter/src/wasi_definitions.rs
```diff
@@ -0,0 +1,221 @@
...
+    FdPrestatGet,
+    FdPrestatDirName,
+    PathOpen,
+    FdFdstatSetFlags,
+    ArgsSizesGet,
+    ArgsGet,
+    ClockTimeGet,
+}
+
+impl SystemCall {
```
kevinaboos: These various conversion functions are good, but you can integrate them nicely with Rust's built-in [conversion traits](https://doc.rust-lang.org/std/convert/index.html) for easy interop. For example:
* For `from_fn_name()`, use [FromStr](httpsblo://doc.rust-lang.org/std/str/trait.FromStr.html)
* For `from_usize()`, use [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) for `usize` 
* For `to_usize()`, use [Into](https://doc.rust-lang.org/std/convert/trait.Into.html)
* For `to_usize()`, use [Into](https://doc.rust-lang.org/std/convert/trait.Into.html)
* Optionally, for `signature()`, you could use `Into` as well, but that's not required since it's not a typical conversion.

Each of these would go in a separate `impl` block.

applications/wasm/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "wasm"
+version = "0.1.0"
+authors = ["Vikram Mullick <vikram1.mullick@gmail.com>"]
+edition = "2018"
```
kevinaboos, line 5: add short description, e.g., the first line of `applications/wasm/src/lib.rs`

applications/wasm/src/lib.rs
```diff
@@ -0,0 +1,150 @@
+//! Application for running WASI-compliant WebAssembly binaries from Theseus command line.
+//!
+//! USAGE:
+//!    wasm </path/to/wasm/binary.wasm> [args ...]
```
kevinaboos: perhaps this should clarify options to the `wasm` application vs arguments that are passed directly to the WASM binary?

Not sure if this is accurate, but something like this is closer to how most posix apps' man pages look
```
    wasm [OPTION]...  WASM_BINARY_PATH  [ARG]...
```

Also change the `USAGE` string to reflect this.

applications/wasm/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    0
+}
+
+fn print_usage(opts: Options) {
+    println!("{}", opts.usage(USAGE));
+}
+
+const USAGE: &'static str = "USAGE:
+    wasm </path/to/wasm/binary.wasm> [args ...]
```
kevinaboos: see above comment

kernel/wasi_interpreter/src/wasi_definitions.rs
kevinaboos: in this case, it's `https://github.com/tomaka/redshirt/blob/4df506f68821353a7fd67bb94c4223df6b683e1b/kernel/core/src/primitives.rs`

kevinaboos (changes requested):
Looks excellent! just left a few small comments.

kevinaboos (approved):

PR #490, opened 2022-03-09
Mellanox driver PR 2

Ramla-I:
In this PR we've extended the Mellanox driver to support send and receive queues. Currently only transmission is working correctly.

The crate that has been added is:
- **test_mlx5**: an application crate used to test the driver

The crates that have been changed are:
- **device_manger**: initializes the Mellanox NIC
- **mlx5_ethernet**: extended the command queue to support more commands and added additional modules for send queues, receive queues, completion queues, flow tables, etc.
- **mlx5**: added additional command executions to create the send and receive queues, and supporting objects.


kevinaboos:
hey @Ramla-I, congrats on getting this PR out the door, and thanks for the contribution!

I'd like to tackle two things first before getting to the meat of this PR:
1. Move the changes to `MappedPages` (the `split()` fn) to a separate PR. My only real comments for that function are to follow the return type of `AllocatedPages::split()`, in which the error type is only `MappedPages` (the unmodified `self`).
2. Explain the need for the `logger` changes. EDIT: I see the only difference being that the mlx-specific one doesn't print location info. If that's truly needed, let's separate that change into another separate PR, since it's not really mlx-specific. 

    > ... bytes are printed out in a standard format used by other drivers.
   I'm not aware of other drivers that use different log formats... ?

That way, only mellanox stuff will be present in this PR.

We can chat on slack if you prefer.

kernel/mlx5/src/lib.rs
```diff
@@ -116,95 +188,453 @@ impl ConnectX5Nic {
...
+        
         // execute MANAGE_PAGES command to transfer boot pages to device
-        let cmdq_entry = cmdq.create_command(
-            CommandOpcode::ManagePages, 
-            Some(ManagePagesOpMod::AllocationSuccess as u16), 
-            Some(boot_pa)
+        let completed_cmd = cmdq.create_and_execute_command(
+            CommandBuilder::new(CommandOpcode::ManagePages)
+            .opmod(ManagePagesOpMod::AllocationSuccess as u16)
+            .allocated_pages(boot_pa),
```
kevinaboos: indent more

kernel/mlx5/src/lib.rs
```diff
@@ -55,23 +91,56 @@ pub fn get_mlx5_nic() -> Option<&'static MutexIrqSafe<ConnectX5Nic>> {
...
     /// Initializes the new ConnectX-5 network interface card that is connected as the given PciDevice.
     /// (steps taken from the PRM, Section 7.2: HCA Driver Start-up)
-    pub fn init(mlx5_pci_dev: &PciDevice) -> Result<&'static MutexIrqSafe<ConnectX5Nic>, &'static str> {
+    /// 
+    /// # Arguments
+    /// * `mlx5_pci_dev`: Contains the pci device information for this NIC.
+    /// * `num_tx_descs`: The number of descriptors in each transmit queue.
+    /// * `num_rx_descs`: The number of descriptors in each receive queue.
+    /// * `mtu`: Maximum Transmission Unit in bytes
+    pub fn init(mlx5_pci_dev: &PciDevice, num_tx_descs: usize, num_rx_descs: usize, mtu: u16) -> Result<&'static MutexIrqSafe<ConnectX5Nic> , &'static str> {
```
kevinaboos: use multiple lines like so:
```rust
pub fn init(
    mlx5_pci_dev: &PciDevice,
    num_tx_descs: usize,
    num_rx_descs: usize,
    mtu: u16
) -> Result<&'static MutexIrqSafe<ConnectX5Nic> , &'static str> {
```

kernel/mlx5/src/lib.rs
```diff
@@ -116,95 +188,453 @@ impl ConnectX5Nic {
...
+        // execute CREATE_EQ for page request event
+        // 1. Allocate pages for EQ
+        let num_eq_entries = 64;
+        let (eq_mp, eq_pa) = create_contiguous_mapping(num_eq_entries * core::mem::size_of::<EventQueueEntry>(), NIC_MAPPING_FLAGS)?;
+    
+        let completed_cmd = cmdq.create_and_execute_command(
+            CommandBuilder::new(CommandOpcode::CreateEq)
+            .allocated_pages(vec!(eq_pa))
+            .uar(uar)
+            .queue_size(num_eq_entries as u32),
```
kevinaboos: indent more

kernel/mlx5/src/lib.rs
```diff
@@ -116,95 +188,453 @@ impl ConnectX5Nic {
...
+        )?;
+        let (eqn, status) = cmdq.get_eq_number(completed_cmd)?;
+        // 2. Initialize the EQ
+        let event_queue = EventQueue::init(eq_mp, num_eq_entries, eqn)?;
+        trace!("CreateEq: {:?}, eqn: {:?}", status, eqn); 
+
+        #[cfg(mlx_logger)]
+        {
+            event_queue.dump()
+        }
```
kevinaboos: should this be something like a verbose mlx log cfg? (since we've removed the other logger stuff, this will never be enabled)

kernel/mlx5/src/lib.rs
```diff
@@ -116,95 +188,453 @@ impl ConnectX5Nic {
...
+        let (db_page, db_pa) = create_contiguous_mapping(core::mem::size_of::<CompletionQueueDoorbellRecord>(), NIC_MAPPING_FLAGS)?;
+        
+        let completed_cmd = cmdq.create_and_execute_command(
+            CommandBuilder::new(CommandOpcode::CreateCq) 
+            .allocated_pages(vec!(cq_pa))
+            .uar(uar)
+            .queue_size(NUM_CQ_ENTRIES_SEND as u32)
+            .eqn(eqn)
+            .db_page(db_pa)
+            .collapsed_cq(),
```
kevinaboos: indent

kernel/mlx5/src/lib.rs
```diff
@@ -116,95 +188,453 @@ impl ConnectX5Nic {
...
+        // 2. Allocate page for doorbell
+        let (db_page, db_pa) = create_contiguous_mapping(core::mem::size_of::<CompletionQueueDoorbellRecord>(), NIC_MAPPING_FLAGS)?;
+        
+        let completed_cmd = cmdq.create_and_execute_command(
+            CommandBuilder::new(CommandOpcode::CreateCq) 
+            .allocated_pages(vec!(cq_pa))
+            .uar(uar)
+            .queue_size(cq_entries_r as u32)
+            .eqn(eqn)
+            .db_page(db_pa),
```
kevinaboos: indent

kernel/mlx5/src/lib.rs
```diff
@@ -116,95 +188,453 @@ impl ConnectX5Nic {
...
+        init_rx_buf_pool(num_rx_descs, mtu, &RX_BUFFER_POOL)?;
+
+        // Create the RQ
+        let completed_cmd = cmdq.create_and_execute_command(
+            CommandBuilder::new(CommandOpcode::CreateRq) 
+            .allocated_pages(vec!(q_pa)) 
+            .queue_size(num_rx_descs as u32)
+            .db_page(db_pa)
+            .cqn(cqn_r) 
+            .pd(pd),
```
kevinaboos: indent

kernel/mlx5/src/lib.rs
```diff
@@ -216,4 +646,42 @@ impl ConnectX5Nic {
         let mp = allocate_memory(mem_base, core::mem::size_of::<InitializationSegment>())?;
         BoxRefMut::new(Box::new(mp)).try_map_mut(|mp| mp.as_type_mut::<InitializationSegment>(0))
     }
+
+    /// Allocates `num_pages` [`MappedPages`] each of the standard kernel page size [`PAGE_SIZE`].
+    /// Returns a vector of the [`MappedPages`] and a vector of the [`PhysicalAddress`] of the pages.
+    fn allocate_pages_for_nic(num_pages: usize) -> Result<(Vec<MappedPages>, Vec<PhysicalAddress>), &'static str> {
```
kevinaboos, line 656: any particular reason you use two separate vectors here? seems like semantically, a better fit would be to have one `Vec<(MappedPages, PhysicalAddress)>`, forcing them to always be together

kernel/mlx5/src/lib.rs
```diff
@@ -216,4 +646,42 @@ impl ConnectX5Nic {
...
+    /// Adds a packet to be sent to the transmit queue.
+    pub fn send_fastpath(&mut self, buffer_addr: PhysicalAddress, buffer: &[u8]) {
+        self.send_queue.send(buffer_addr, buffer);
+    }
 }
+
+/// Returns true if `num` is a power of 2.
+fn check_pow_2(num: usize) -> bool {
+    (num & (num - 1) == 0) && (num != 0)
+}
```
kevinaboos: use this: https://doc.rust-lang.org/std/primitive.u32.html#method.is_power_of_two

There are also related functions like `*_next_power_of_two()`

kernel/mlx_ethernet/src/command_queue.rs
```diff
@@ -210,29 +287,294 @@ pub enum QueryPagesOpMod {
...
+}
+
+/// The layout of output data for the command [`CommandOpcode::QueryNicVportContext`].
+/// This command is mainly used to retrieve the mac address of the NIC when working with the physical function.
+/// (PRM Section 8.6: NIC_Vport Context)
+#[derive(FromBytes)]
+#[repr(C)]
+struct NicVportContext {
+    /// This field is not necessarily padding, just unused fields that are not relevant.
+    _padding0: [u8; 36],
```
kevinaboos: then rename to `_unused`. same for two lines below

kernel/mlx_ethernet/src/command_queue.rs
```diff
@@ -247,328 +589,881 @@ impl CommandQueue {
...
-
-        // set the type of transport and token
-        cmdq_entry.set_type_of_transport();
-        cmdq_entry.set_token(self.token);
+        let num_pages = parameters.allocated_pages.as_ref().and_then(|pages| Some(pages.len())); 
 
-        // set the ownership bit to HW-owned.
-        // This bit will be cleared when the command is complete.
-        cmdq_entry.change_ownership_to_hw();
+        #[cfg(mlx_logger)]
```
kevinaboos: yeah, seems like `mlx_verbose_log` or something would be a good fit here

kernel/mlx_ethernet/src/command_queue.rs
```diff
@@ -771,3 +1744,209 @@ impl fmt::Debug for CommandInterfaceMailbox {
             .finish()
     }
 }
+
+/// Layout of output of command [`CommandOpcode::QueryHcaCap`]
+#[derive(FromBytes)]
+#[repr(C)]
+struct HCACapabilitiesLayout {
+    vhca_resource_manager:          Volatile<U32<BigEndian>>,
```
kevinaboos: are all of these volatiles necessary? seems odd to have so many volatile fields, and then a separate mirror struct without them. I'm puzzled here.

kernel/mlx_ethernet/src/uar.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub(crate) fn write_wqe_to_doorbell(&mut self, current_doorbell: &CurrentUARDoorbell, wqe_value: [U32<BigEndian>; 64]) {
+        match current_doorbell {
+            CurrentUARDoorbell::Even => self.db_blueflame_buffer0_even.write(wqe_value),
+            CurrentUARDoorbell::Odd => self.db_blueflame_buffer0_odd.write(wqe_value),
+        }
+    }
+}
+
+impl Default for UserAccessRegion {
+    /// We have to define our own default function since only array sizes up to 32 are supported
```
kevinaboos, line 61: i don't think that's true with the advent of const generics, e.g., `#![feature(min_const_generics)]` should do it (?)

kernel/mlx_ethernet/src/work_queue.rs
```diff
@@ -0,0 +1,405 @@
...
+#[derive(FromBytes, Default)]
+#[repr(C)]
+pub(crate) struct EthSegment {
+    _padding0:              u32,
+    /// Maximum Segment Size
+    mss:                    Volatile<U32<BigEndian>>,
+    flow_table_metadata:    Volatile<U32<BigEndian>>,
+    /// A multi-part field:
+    /// * `inline_header_size`: length of inlined packet headers in bytes, occupies bits [25:16]
+    /// * `inline_headers`: beginning of the inlined packet headers, occupies bits [15:0]
```
kevinaboos, line 349: Should probably break this apart into an actual struct, otherwise it's quite hard to access

kevinaboos (changes requested):
Looks generally pretty good.

`Vec`s are used a lot, where I think boxed slices would be better. Not only do they use less memory, but they also prevent unexpected growth. Generally if an array doesn't need to grow dynamically, I'd recommend using a boxed slice. You can always use a vector to construct it initially, and then use the `Vec::into_boxed_slice()` to convert it after creation/population.

There are also a lot of `write_*_to_mailbox()` functions. I wonder if we could introduce a trait like `MailboxWriter` or `WritableToMailbox` that reduces the number of individual functions we need. With such traits, you could implement the same `write_to_mailbox` trait using generics for multiple different items that could be written to the mailbox. 

Make sure all files have a trailing newline too.

kernel/mlx_ethernet/src/uar.rs
Ramla-I: they haven't supported default for all sizes of arrays yet.
see [here](https://doc.rust-lang.org/src/core/array/mod.rs.html#373)

kevinaboos: I see, interesting. I thought that had been solved, thanks for the correction!

kernel/mlx_ethernet/src/command_queue.rs
Ramla-I: Basically, `HCACapabilitiesLayout` gives the order of how different fields are arranged in memory when we are returned device capability information from the command `QueryHcaCap`. 
`HCACapabillites` is the actual values of these capabilities, which have to be extracted from subfields of each member of the `HCACapabilitiesLayout` struct. (see the get_capabilities() function).

I think the confusion is perhaps why I set each member always to a U32. It's because of the big endianness of the fields. If I use types smaller than u32, like u16 or u8, then I have to rearrange their order in the struct manually, and I believe that will lead to more confusion and chances of error.

But your pointing this out reminded me that all these fields only need to be ReadOnly. So thanks for that.

kernel/mlx_ethernet/src/work_queue.rs
Ramla-I: I prefer keeping fields at the u32 granularity. See my response to the comment above.

kernel/mlx5/src/lib.rs
Ramla-I: If I intended to store the physical addresses, yes that would be better. But I pass the physical addresses to the command interface where they are consumed and use the MappedPages to map to different objects. I could return them together at this function interface but at a future point I would separate them.

kernel/mlx_ethernet/src/command_queue.rs
```diff
@@ -330,12 +428,12 @@ pub enum HcaPortType {
 #[derive(FromBytes)]
 #[repr(C)]
 struct NicVportContext {
-    /// This field is not necessarily padding, just unused fields that are not relevant.
-    _padding0: [u8; 36],
+    /// Unused fields that are not relevant.
+    unused0: [u8; 36],
```
kevinaboos: still a good idea to prepend these with `_` to indicate they aren't used

kernel/mlx_ethernet/src/completion_queue.rs
```diff
@@ -13,8 +16,11 @@ use alloc::boxed::Box;
 use memory::{PhysicalAddress, MappedPages};
 use owning_ref::BoxRefMut;
 use num_enum::TryFromPrimitive;
+
+#[allow(unused_imports)]
 use crate::{
     *,
+    command_queue::CommandOpcode,
     work_queue::WQEOpcode
 };
```
kevinaboos: this import statement is suspect. Either use `crate::*` or specify the exact items. The glob is only preferred if there are dozens of things you need to import

kernel/mlx_ethernet/src/event_queue.rs
```diff
@@ -9,7 +9,8 @@ use byteorder::BigEndian;
 use alloc::boxed::Box;
 use memory::MappedPages;
 use owning_ref::BoxRefMut;
-use crate::*;
+#[allow(unused_imports)]
+use crate::{*, command_queue::CommandOpcode};
```
kevinaboos: same comment

kevinaboos (changes requested):
Just a couple small nitpicks, up to your discretion. Otherwise looks good.

kevinaboos (approved):

PR #491, opened 2022-03-11
MappedPages split function

Ramla-I:
Adds a function to split a MappedPages object into two. It follows the same logic as the function to split the contained AllocatedPages.


kevinaboos (approved):
LGTM

PR #492, opened 2022-03-13
Realtime scheduler

jacob-earle:
Adding support for the rate monotonic scheduling algorithm in the crates runqueue_realtime and scheduler_realtime and adding support for a basic task delay API that can be used to implement periodic tasks for a realtime system.


Makefile
```diff
@@ -790,7 +790,13 @@ bochs : export override THESEUS_CONFIG += apic_timer_fixed
 bochs: $(iso) 
 	bochs -f bochsrc.txt -q
 
+### Compile Theseus for realtime scheduling algorithm
+build-realtime: export override THESEUS_CONFIG += realtime_scheduler
+build-realtime: build
 
+### Configure Theseus to run realtime scheduling algorithm
+run-realtime: export override THESEUS_CONFIG += realtime_scheduler
+run-realtime: run
```
kevinaboos: good for testing, but let's remove these makefile changes

applications/test_realtime/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+        let periodic_tb1 = spawn::new_task_builder(task_delay_tester, 1).block();
+        let periodic_task_1 = periodic_tb1.spawn().unwrap();        // setting the periods of the tasks
+
+        scheduler::set_periodicity(&periodic_task_1, 1000).unwrap();
+
+        // starting the tasks
+        periodic_task_1.unblock();
+    }
+    else {
+        info!("Realtime scheduler not enabled!");
```
kevinaboos: Good stuff. To make it even more apparent, use `println` here and return an error.

applications/test_realtime/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+    }
+    0
+}    
+
+fn task_delay_tester(arg: usize) {
```
kevinaboos: are there any more complex tests that we should add? this one is good but it's extremely simple with only a single test. let me know if you have ideas and i can code them later, too.

kernel/interrupts/src/lib.rs
```diff
@@ -383,6 +384,11 @@ extern "x86-interrupt" fn lapic_timer_handler(_stack_frame: &mut ExceptionStackF
     
     // we must acknowledge the interrupt first before handling it because we switch tasks here, which doesn't return
     eoi(None); // None, because 0x22 IRQ cannot possibly be a PIC interrupt
+
+    // Callback to the sleep API to unblock tasks whose waiting time is over
+    // and alert to update the number of ticks elapsed
+    sleep::increment_tick_count();
+    sleep::unblock_delayed_tasks();
```
kevinaboos: I think this ought to go before the `eoi`, otherwise it might be weird with re-entrancy issues (shouldn't be a problem, but logically it makes sense to handle one interrupt's side effects before allowing another one to happen).

kernel/runqueue/src/lib.rs
```diff
@@ -10,15 +10,17 @@ extern crate irq_safety;
 extern crate atomic_linked_list;
 extern crate task;
 #[cfg(priority_scheduler)] extern crate runqueue_priority;
-#[cfg(not(priority_scheduler))] extern crate runqueue_round_robin;
+#[cfg(realtime_scheduler)] extern crate runqueue_realtime;
```
kevinaboos: now that we have more than 2 options, it's better to use the `cfg_if` crate. I use it elsewhere, but the main idea is that you can use regular if conditional syntax to specify complex if-else conditional statements.

kernel/runqueue_realtime/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+authors = ["Jacob Earle <jacob.earle@yale.edu>"]
```
kevinaboos: for your author credentials, do you have a permanent email contact? perhaps whatever you use for github? (that is, if the yale address isn't permanent)

kernel/interrupts/src/lib.rs
kevinaboos: Another issue of note (which we can resolve later), this interrupt fires only X milliseconds, in which X is the length of the timeslice used for context switches.

Obviously the choice of lapic timer rate now affects more than that, since we're using it to track ticks for sleep purposes. My concern is that this isn't super accurate, especially for something "important" like a pseudo-realtime scheduler. In particular, you wouldn't be able to sleep for a time period close to or significantly less than X.

In the future, we could also use a dedicated timer source for this, e.g., HPET or a different APIC-based timer. 

Also, while the "counting ticks" approach works, a more efficient method would be to set a timer interrupt to occur whenever it's time for the next sleeping task to be unblocked. This could optionally be done per core, too.

kernel/runqueue_realtime/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+
+/// A list of references to `Task`s (`RealtimeTaskRef`s). 
+/// This is used to store the `Task`s (and associated scheduler related data) 
+/// that are runnable on a given core.
+/// A queue is used for the round robin scheduler.
+/// `Runqueue` implements `Deref` and `DerefMut` traits, which dereferences to `VecDeque`.
+#[derive(Debug)]
+pub struct RunQueue {
+    core: u8,
+    queue: VecDeque<RealtimeTaskRef>,
```
kevinaboos, line 111: shouldn't this be a queue type, or something more like a sorted structure (e.g., binary heap)? That way, you wouldn't have to manually determine where to insert it. Non-periodic tasks could be sorted using a period value of 0 or infinity, depending on how you wanted them sorted.

again, we can leave this for a future PR too. Just something to note.

kernel/runqueue_realtime/src/lib.rs
```diff
@@ -0,0 +1,298 @@
+//! This crate contains the `RunQueue` structure, for a realtime scheduler using rate monotonic scheduling.
+//! `RunQueue` structure is essentially a list of `Task`s
+//! that is used for scheduling purposes.
+//!
```
kevinaboos, line 2: Here there should be a top-level description of the structure/format of the realtime RunQueue and how it actually works. You've added it as internal doc comments that kind of cover parts of it for each function, which is good, but it's hard to get a quick feel for how it works and how tasks are organized within the runqueue.

kernel/scheduler/src/lib.rs
```diff
@@ -7,15 +7,17 @@ extern crate apic;
 extern crate task;
 extern crate runqueue;
 #[cfg(priority_scheduler)] extern crate scheduler_priority;
-#[cfg(not(priority_scheduler))] extern crate scheduler_round_robin;
+#[cfg(realtime_scheduler)] extern crate scheduler_realtime;
+#[cfg(all(not(priority_scheduler), not(realtime_scheduler)))] extern crate scheduler_round_robin;
```
kevinaboos: same `cfg_if` tip here

kevinaboos (changes requested):

applications/test_realtime/src/lib.rs
jacob-earle: Done, but on the point of returning an error, how is that done with a return type of isize, I would assume just return -1?

applications/test_realtime/src/lib.rs
jacob-earle: Agreed, I am not too familiar with testing capabilities in Rust, but I think a good one to implement could be:

Let the program take in arguments n and p1, ..., pm. For each pi, we spawn a task with period pi, then instead of the logging statement, we put a statement that measures the time elapsed between successive calls and let each of the spawned tasks run for n periods. We can use these measurements to get statistics on how much the time between successive calls of the timing statement deviates from the expected time, i.e. the period pi.

Of course, to achieve this, we would need to have the capability to measure times pretty accurately.

kernel/interrupts/src/lib.rs
jacob-earle: Thanks for explaining this, fixed! Also, I agree that a more efficient time-keeping scheme would be a good topic for future work.

kernel/runqueue/src/lib.rs
jacob-earle: Fixed

kernel/runqueue_realtime/Cargo.toml
jacob-earle: I replaced that with my personal email

kernel/scheduler/src/lib.rs
jacob-earle: Fixed

kernel/runqueue_realtime/src/lib.rs
jacob-earle: I was thinking so myself, I was wondering whether you would know a good datatype for this use case, where we need:
1. Efficient insertion but also
2. The ability to iterate over the list. When we are selecting the next task, we want to select the task with the highest priority but with the condition that the task is currently Runnable. I don't think the BinaryHeap in alloc::collections would work as the into_iter() method does not iterate over the elements in sorted order, it only allows us to peek the min/max element.

I was thinking about using the priority queue implementation [here](https://docs.rs/priority-queue/latest/priority_queue/), but I tried using it as the priority queue implementation of the `sleep` crate but for some reason it pulls in `std` even when I set default-features = false in Cargo.toml?

If I can find the proper data type, I think I could implement this pretty quickly tonight

applications/test_realtime/src/lib.rs
kevinaboos: yep, any negative value like `-1` is fine for now

applications/test_realtime/src/lib.rs
kevinaboos: ok, if you *want* to add such a test, that's doable using the `hpet` clock source. it's very accurate. 
If you don't want to do this, no worries. Maybe just add a comment at the top of `test_realtime/src/lib.rs` describing what other kinds of tests could/should be done later.

kernel/runqueue_realtime/src/lib.rs
kevinaboos: Okay, I didn't realize you needed queue-like functionality.

That `priority-queue` crate does work for me, so you could certainly use it. The thing is, because we're in a `no_std` environment, you need to provide your own implementation of a hasher (which is normally provided by the std lib). Thus, you can't use the default constructors like `PriorityQueue::new()` or `with_capacity()`, you have to use the more specific constructors like this:

```rust
let mut q: PriorityQueue<&'static str, usize, ahash::RandomState> = priority_queue::PriorityQueue::with_default_hasher();
let mut q2: PriorityQueue<usize, usize, _> = PriorityQueue::with_hasher(ahash::RandomState::default());
```

where the `ahash` crate is providing the hasher implementation. So in the Cargo.toml file:
```toml
[dependencies]
priority-queue = "1.2.1"
ahash = { version = "0.7.6", default-features = false }
```

kernel/sleep/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+use alloc::collections::binary_heap::BinaryHeap;
+use irq_safety::MutexIrqSafe;
+use task::get_my_current_task;
+
+/// This struct will be used as an entry that contains the task id and the associated wakeup time for an entry in DELAYED_TASKLIST
+/// Entries will be sorted in increasing order of resume_time
+#[derive(Copy, Clone, Eq, PartialEq)]
+struct SleepingTaskNode {
+    resume_time: usize,
+    taskid: usize,
```
kevinaboos: Use `TaskRef` directly here instead of task id. The id lookup is slow.

kevinaboos (changes requested):
I did a lot of cleanup on code and docs. Can you please review my changes to ensure they're all correct?

Also see my comment about using TaskRef directly instead of task id in the `sleep` crate.

kevinaboos (approved):

PR #498, opened 2022-03-22
Update to Rust 1.61, nightly from 2022-03-21

kevinaboos:
* Remove `llvm_asm!()` and convert everything to the new `asm!()` syntax

* The `__truncdfsf2` family of functions is now in the compiler_builtins crate, so we can remove our own version.

* Remove lots of outdated or stabilized features.

* `Debug` and `Clone` derivations now warn about unused struct fields, so we allow those or prepend a `_` before the field name as necessary.

* The target spec `.json` files now need to be cleaned up to avoid warnings, with no fake comment lines.

* Other changed crates include `irq_safety`, `spin`, `x86_64`


Ramla-I (approved):
looks good to me

Acciente717 (approved):
Look good to me.

PR #499, opened 2022-03-24
Update our fork of `x86_64` to latest upstream version

kevinaboos:


phil-opp:
Hi Kevin, I just saw that you're using a fork of `x86_64`. Perhaps we can update the upstream crate to support your use case? The [diff](https://github.com/rust-osdev/x86_64/compare/master...theseus-os:master) shows only a new `LockedIdt` type, which might be possible to implement outside of the crate, and a public `handler_addr` function, which we could easily add.

kevinaboos:
> 

Hi Phil, yeah you're so quick that you beat me to it! haha.

I'd been using a few parts of your `x86_64` crate from like... 2017 or 2018, so that code was ancient. I figured I'd update it, and in doing so, I realized that I should rewrite some of our own code to simplify it. That actually changed the number of things I needed to modify in the current version of the `x86_64` crate, so now I think we only really need to publicize the `handler_addr` function as you mentioned.

I was planning to submit a quick PR, but I'll definitely do it now that you said something.

As a side note, I also plan to add strongly-typed defs (bitflags or something) for a variety of other MSRs we use in Theseus. But that isn't particularly high priority unless others need support for them too.

phil-opp:
Great, thanks a lot for submitting that PR!

> As a side note, I also plan to add strongly-typed defs (bitflags or something) for a variety of other MSRs we use in Theseus. But that isn't particularly high priority unless others need support for them too.

We're always happy to add more register abstractions, it just takes us a bit to review those sometimes. So please open PRs for anything that could be useful to others as well.

PR #519, opened 2022-05-12
Export more `libm` functions as `#[no_mangle]`.

kevinaboos:
Adds the following function wrappers:
* `ceil`
* `floor`
* `ceilf`
* `floorf`
* `trunc`
* `truncf`


kevinaboos:
Necessary because of https://github.com/rust-lang/libm/issues/258

PR #523, opened 2022-06-06
Ixgbe issues

Ramla-I:
Changes made to remove bugs in the ixgbe driver, and extend the ixgbe test application.

The following changes were made:
- **Makefile:** added a flag to prevent QEMU from trying to boot via pxe. 
- **test_ixgbe:** extended test application to allow testing the nic directly in addition to testing via the virtual nic interface.
- **device_manager**: change packet buffer to size to 2 KiB since the max ethernet packet size is 1500B.
- **intel_ethernet**: added `Debug` trait for `AdvancedTxDescriptor`.
- **ixgbe**: enabled bus mastering, disabled interrupts properly, removed steps to setup phy and link speed since that's done automatically, removed a source of deadlock when creating a virtual function.


kernel/intel_ethernet/src/descriptors.rs
```diff
@@ -379,7 +379,14 @@ impl TxDescriptor for AdvancedTxDescriptor {
...
+            // error!("tx desc status: {:#X}", self.paylen_popts_cc_idx_sta.read());
         } 
     }
 }
+
+impl fmt::Debug for AdvancedTxDescriptor {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{{Packet buffer address: {:#X},\n data len: {:#X},\n dtype_mac_rsv: {:#X},\n dcmd: {:#X},\n paylen_popts_cc_idx_sta: {:#X},\n }}",
+            self.packet_buffer_address.read(), self.data_len.read(), self.dtyp_mac_rsv.read(), self.dcmd.read(), self.paylen_popts_cc_idx_sta.read())
+    }
```
kevinaboos: for this it's better to use [`debug_struct()`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct) along with [`finish_non_exhaustive()`](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html#method.finish_non_exhaustive) since there are more fields in the struct that are hidden from the Debug impl

Makefile
```diff
@@ -757,7 +757,7 @@ endif
 ## Enable passthrough of a PCI device in QEMU by passing its slot information to VFIO.
 ## Slot information is its bus, device, and function number assigned by the host OS, e.g., 'vfio=59:00.0'.
 ifdef vfio
-	QEMU_FLAGS += -device vfio-pci,host=$(vfio)
+	QEMU_FLAGS += -device vfio-pci,host=$(vfio),romfile=
```
kevinaboos: what does this do, and is it supposed to be a blank arg?

applications/test_ixgbe/src/lib.rs
```diff
@@ -14,19 +23,47 @@
...
+use getopts::{Matches, Options};
+
+
+const DEST_MAC_ADDR: [u8; 6] = [0xa0, 0x36, 0x9f, 0x1d, 0x94, 0x4c];
+
+pub fn main(args: Vec<String>) -> isize {
+
+    let mut opts = Options::new();
+    opts.optflag("h", "help", "print this help menu");
+    opts.optflag("v", "virtual", "Test the ixgbe nic using the virtual nic interface.
```
kevinaboos: this is a bit unorthodox; most users expect `-v` to enable verbose printing/logging. I know it's only a test app, but it's worth considering. You could also simply remove the short opt for "-v" and leave only "--virtual". Up to you ultimately.

kevinaboos (changes requested):
LGTM, left a few small comments. I don't have any real way to test this locally but it's still fine to merge.

Makefile
Ramla-I: So initially when I used pci passthrough for the ixgbe device, QEMU was always trying to boot via pxe first and that added 5-10 seconds in the boot time.
This was just a workaround I found [here](https://askubuntu.com/questions/190929/how-do-i-disable-unwanted-ipxe-boot-attempt-in-libvirt-qemu-kvm) that prevented that.

I've removed it now because now it's booting from the cd rather than through pxe first, even without this flag.
If it happens again, I can submit another PR. There may be a better workaround such as specifying the boot order.

applications/test_ixgbe/src/lib.rs
Ramla-I: removed -v option, and only left --virtual.

Makefile
kevinaboos: I don't mind leaving it in or adding it back later, so long as we document it. Mostly, I was wondering whether it should *always* be used, or just with the vfio arg.

kevinaboos (approved):
looks good, thanks!

Makefile
Ramla-I: I think just with the vfio arg, because it's specifying something about that particular device.

PR #525, opened 2022-06-09
Remove doubled flag in book

wieczorek1990:
Line 19 and 25 contain doubled `-o` compiler flag.


kevinaboos (approved):
Hi @wieczorek1990, thanks for catching this! Your changes look good to me.

PR #527, opened 2022-06-10
Replace `unwrap_or` with `unwrap_or_else` and vice versa depending on cost of computing argument

### LINKED ISSUE #35: Use `unwrap_or_else` instead of `unwrap_or`, and `ok_or_else` instead of `ok_or`
The `unwrap_or(default)` and `ok_or(default)` functions are used ubiquitously, but they _eagerly_ evaluate the `default` value, even when the actual value is `Some` or `Ok`. That's bad, and really wasteful. 

Instead, we should convert them to `unwrap_or_else(|| closure)` and `ok_or_else(|| closure)`, because the closures that provide the default value are _not_ eagerly evaluated; they are only _lazily_ evaluated if the original value was `None` or `Err`.

xa888s:
To fix #35 

My editor auto formatted (with rustfmt) the files on save but the (non-whitespace/formatting) changes are quite small. I have finished `unwrap_or*` methods and now I'll work on the `ok_or*` methods


kevinaboos (changes requested):
Thanks Brandon, I appreciate the PR! 

Can you undo the rustfmt changes? Otherwise it's difficult to review and isolate exactly what changed.

I'm happy to handle rustfmt configuration in a separate PR, as it's an important part of linting along with clippy.

xa888s:
> I'm happy to handle rustfmt configuration in a separate PR, as it's an important part of linting along with clippy.

This is a good idea, I think I'll set this aside until we can merge a formatting PR.

kevinaboos:
Hi @xa888s, it appears that there are still quite a few unrelated formatting changes in this branch. Once you remove those, I am happy to review and accept this PR.

Also, kindly merge the latest changes from `theseus_main` into this branch, which remove the top-level `build.rs` script.

kevinaboos:
Closing due to inactivity and unrelated changes in favor of #641.

PR #530, opened 2022-06-11
Formatting

### LINKED ISSUE #529: Run `rustfmt` / `cargo fmt` on source code
Many files have inconsistent formatting which can make it harder to understand the code. Running rustfmt on all files would ensure that the formatting is consistent.

xa888s:
Fixes #529  

This is an initial draft for adding automatic formatting. All I have done so far is add a `rustfmt.toml` and fix some spots manually so that `cargo fmt` would run successfully. It has been suggested to add `cargo clippy` to CI https://github.com/theseus-os/Theseus/issues/526#issuecomment-1152703909, and I think running an automated formatting check on potential PRs would be a good first step towards this.

The changes are huge so it is difficult to review (although all changes are only whitespace/formatting excluding the `rustfmt.toml`)


xa888s:
@kevinaboos Should this PR be split into PRs for each subsystem to make it more manageable? In it's current form it is infeasible to review manually, but I think having a PR for each subsystem may bloat up the PR page with a ton of (highly related) PRs. Maybe I can find a way to group related subsystems so that there is a balance between changes per PR and the amount of PRs?

kevinaboos:
I agree that formatting needs to be standardized, but I'm generally not a huge fan of how `rustfmt` does things by default. I'd need to either be educated on or read up on the various options for `rustfmt`.

One thing that I take issue with is how `rustfmt` likes to break things across way too many lines such that you have to scroll 20 lines just to read one statement; I'm sure there are others too. Perhaps there is a setting to enable rustfmt to spit out warnings or suggestions rather than actually modifying the file?

Notably, I do value both soft and hard line width checks, e.g., >100 columns allowed but warned against, >110 columns prohibited. So this would be a good time to enforce that too.

kevinaboos:
Some other things of note:
I prefer this
```rust
#[cfg(...)] {
```
to this
```rust
#[cfg(...)]
{
```

Whitespace removal is _far_ too aggressive. Code that is aligned to operators or keywords using whitespace shouldn't be changed; there's no real need to remove a trailing space from a comment line, etc.

That being said, I do notice several improvements that `rustfmt` has made on various code sections in your other PR #527; much of that is all good with me.

kevinaboos:
> @kevinaboos Should this PR be split into PRs for each subsystem to make it more manageable? In it's current form it is infeasible to review manually, but I think having a PR for each subsystem may bloat up the PR page with a ton of (highly related) PRs. Maybe I can find a way to group related subsystems so that there is a balance between changes per PR and the amount of PRs?

Just realized I completely disregarded this question, sorry! I don't quite have as clear of an idea for how best to divide up this PR like I do for the clippy PR, but I think subsystem by subsystem is okay. You could start with all the `applications/` crates, then non-submodule `libs`, then maybe `tools`, and then save `kernel` crates for last? The kernel crates we could do alphabetically or by some other arbitrary category. Considering that this is just a visual, non-functional change, I don't think it matters. I'm happy to review them in whatever groupings you deem suitable, just use your best judgment so that it's easier to review. Thanks!

xa888s:
For a quick "fix" for now, I think it would be a good idea to add a `rustfmt.toml` file to the root with the content ([link](https://rust-lang.github.io/rustfmt/?version=v1.5.0&search=#ignore)):
```
ignore = ["/"]
```
This would ensure no files are accidentally formatted and makes the decision to avoid use of rustfmt explicit. It also helps by allowing those who have format to run automatically on save in their editor not have to disable and re-enable it depending on if they are working on this project. This also provides a path for adding formatting to specific subsystems at a time in the future (by adding a `rustfmt.toml` to each folder to override the project-global formatting rules).

kevinaboos:
> For a quick "fix" for now, I think it would be a good idea to add a `rustfmt.toml` file to the root with the content ([link](https://rust-lang.github.io/rustfmt/?version=v1.5.0&search=#ignore)):
> 
> ```
> ignore = ["/"]
> ```
> 
> This would ensure no files are accidentally formatted and makes the decision to avoid use of rustfmt explicit. It also helps by allowing those who have format to run automatically on save in their editor not have to disable and re-enable it depending on if they are working on this project. This also provides a path for adding formatting to specific subsystems at a time in the future (by adding a `rustfmt.toml` to each folder to override the project-global formatting rules).

That's a great idea! I like the explicit nature of that solution; I'm happy to accept a PR that adds that rustfmt file.

After that PR, it's probably a good idea to close this PR and open a separate tracking issue (as with my idea for a clippy tracking issue) that tracks the progress of which files/folders have been formatted and which formatting options are enabled.

EDIT: I'm blind, forgot about #529. Let's use that for the rustfmt tracking issue.

PR #532, opened 2022-06-14
Disable parallel make

jkugelman:
I have `MAKEFLAGS=-j8` in my environment which enables parallel builds
by default. Parallel building should be explicitly disabled until
target dependencies are all fixed.


kevinaboos (approved):
Thanks @jkugelman! Appreciate you discovering this issue.

PR #533, opened 2022-06-15
Fix mistake in `crate_metadata` documentation

tsoutsman:
A and B are confused when describing the dependencies.


kevinaboos (approved):
Hi @tsoutsman, thanks for catching this mistake and thanks for the PR! Much appreciated.

PR #534, opened 2022-06-15
Make intention to drop values explicit by calling drop()

jkugelman:
* Calling `drop()` makes it clear when values are being dropped and removes the need for explanatory comments and extra scope blocks.

### DIFF
--- a/kernel/task/src/lib.rs
+++ b/kernel/task/src/lib.rs
@@ -813,10 +813,12 @@ impl Task {
 
         // Now, as a final action, we drop any data that the original previous task 
         // prepared for droppage before the context switch occurred.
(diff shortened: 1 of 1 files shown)


kevinaboos (approved):
Thanks, looks good to me. I added back in one lexical scope block because I have previously experienced issues with a call to `drop(lock_guard)` not actually doing anything until the next lexical scope block ends. Hopefully new Rust versions would have fixed that behavior, but it's _very_ difficult to trigger those conditions when testing (especially in a subsystem as sensitive as task switching or unwinding).

In any case, I find that it's nice to have lexical blocks separate logically-distinct sections of the code, too; it can help clarify which pieces of code a given comment applies to.

PR #536, opened 2022-06-16
Fix `grub-mkrescue` command on Fedora

tsoutsman:
`grub-mkrescue` is called `grub2-mkrescue` on Fedora. I think this is a sound way of implementing it, but I am illiterate in Make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,8 +37,14 @@ else
 	UNMOUNT = umount
 	USB_DRIVES = $(shell lsblk -O | grep -i usb | awk '{print $$2}' | grep --color=never '[^0-9]$$')
 endif
-GRUB_MKRESCUE = $(GRUB_CROSS)grub-mkrescue
 
+ifneq (,$(shell command -v $(GRUB_CROSS)grub-mkrescue))
+	GRUB_MKRESCUE = $(GRUB_CROSS)grub-mkrescue
+else ifneq (,$(shell command -v $(GRUB_CROSS)grub2-mkrescue))
+	GRUB_MKRESCUE = $(GRUB_CROSS)grub2-mkrescue
+else
+	$(error grub-mkrescue not found)
+endif
 
 
 ###################################################################################################
--- a/Makefile
+++ b/Makefile
@@ -38,12 +38,13 @@ else
 	USB_DRIVES = $(shell lsblk -O | grep -i usb | awk '{print $$2}' | grep --color=never '[^0-9]$$')
 endif
 
+## Look for `grub-mkrescue` (Debian-like distros) or `grub2-mkrescue` (Fedora)
 ifneq (,$(shell command -v $(GRUB_CROSS)grub-mkrescue))
 	GRUB_MKRESCUE = $(GRUB_CROSS)grub-mkrescue
 else ifneq (,$(shell command -v $(GRUB_CROSS)grub2-mkrescue))
 	GRUB_MKRESCUE = $(GRUB_CROSS)grub2-mkrescue
 else
-	$(error grub-mkrescue not found)
+	$(error Error: could not find 'grub-mkrescue' or 'grub2-mkrescue', please install 'grub' or 'grub2')
 endif
 
 
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ git submodule update --init --recursive
 
 Currently, we support building Theseus on the following platforms:
  * Linux, 64-bit Debian-based distributions like Ubuntu, tested on Ubuntu 16.04, 18.04, 20.04. 
-   - Arch Linux has also been reported to work correctly. 
+   - Arch Linux and Fedora have also been reported to work correctly. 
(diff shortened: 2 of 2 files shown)


kevinaboos:
Hi @tsoutsman, thanks for the PR! We haven't yet experimented with builds on Fedora so there may be some rough edges there, but I'm always happy to support more build environments.

Yes, the way you implemented it seems fine to me, though I wonder if `uname -n` always outputs `fedora` on systems where the `grub-pc-bin` package installs `grub2-mkrescue` instead of `grub-mkrescue`. 

It seems like this might be a package manager-specific issue rather than a distro-specific issue. For example, OpenSUSE might also use `grub2-mkrescue` but probably wouldn't return `fedora` from a call to `uname -n`. That being said, I'm not sure how to check the current package manager in the build environment.

Also, I assume there are other differences in the Fedora set-up and build procedure. If you have a list of differences in said instructions, I'd love to add them to the README too.

kevinaboos:
@tsoutsman I like your idea of just using `grub2-mkrescue` if `grub-mkrescue` doesn't exist, rather than trying to ascertain which package manager or distro we're running on. 

The test I was thinking of is simply `command -v grub-mkrescue`, a POSIX-required program (often a shell built-in) that is more portable than `which` but works the same way. On my Ubuntu system, I get the following output, as expected:
```sh
~$ command -v grub-mkrescue
/usr/bin/grub-mkrescue
```

```sh
~$ command -v gru2b-mkrescue
(no output)
```

So we can use that to determine whether or not `grub-mkrescue` exists.

tsoutsman:
It should be right now, but `GRUB_CROSS` _might_ cause issues. Also, I assume your second example should say `grub2-mkrescue`.

I added the ever so slightly different package names in `dnf` to the README.

kevinaboos:
> It should be right now, but `GRUB_CROSS` _might_ cause issues. 

Cross-toolchain stuff shouldn't cause any issues. 

> Also, I assume your second example should say `grub2-mkrescue`.

Whoops, yeah. Fixed, thanks.

kevinaboos:
Does the  `dnf` registry have a legacy grub package? Perhaps we could just use that instead of `grub2`, which would simplify the Makefile.  I'm not sure which grub package on `dnf` is equivalent to what we use from `apt` on Ubuntu.

kevinaboos (approved):
Thanks for the fixes! Much appreciated.

tsoutsman:
I don't think it does:
```
> dnf search grub                                                                 15:24:24
Last metadata expiration check: 2 days, 1:05:32 ago on Thu 16 Jun 2022 14:18:57.
=============================== Name & Summary Matched: grub ================================
grub-customizer.x86_64 : Graphical GRUB2 settings manager
grub2-breeze-theme.x86_64 : Breeze theme for GRUB
grub2-common.noarch : grub2 common layout
grub2-efi-aa64-modules.noarch : Modules used to build custom grub.efi images
grub2-efi-arm-modules.noarch : Modules used to build custom grub.efi images
grub2-efi-ia32.x86_64 : GRUB for EFI systems.
grub2-efi-ia32-modules.noarch : Modules used to build custom grub.efi images
grub2-efi-x64.x86_64 : GRUB for EFI systems.
grub2-efi-x64-modules.noarch : Modules used to build custom grub.efi images
grub2-emu.x86_64 : GRUB user-space emulation.
grub2-emu-modules.x86_64 : GRUB user-space emulation modules.
grub2-pc-modules.noarch : Modules used to build custom grub images
grub2-ppc64le-modules.noarch : Modules used to build custom grub images
grub2-tools.x86_64 : Support tools for GRUB.
grub2-tools-efi.x86_64 : Support tools for GRUB.
grub2-tools-extra.x86_64 : Support tools for GRUB.
grub2-tools-minimal.x86_64 : Support tools for GRUB.
ostree-grub2.x86_64 : GRUB2 integration for OSTree
terminus-fonts-grub2.noarch : Clean fixed width font (grub2 version)
==================================== Name Matched: grub =====================================
boom-boot-grub2.noarch : A set of libraries and tools for managing boot loader entries
grub2-efi-ia32-cdboot.x86_64 : Files used to boot removeable media with EFI
grub2-efi-x64-cdboot.x86_64 : Files used to boot removeable media with EFI
grub2-pc.x86_64 : Bootloader with support for Linux, Multiboot, and more
grubby.x86_64 : Command line tool for updating bootloader configs
grubby-deprecated.x86_64 : Legacy command line tool for updating bootloader configs
=================================== Summary Matched: grub ===================================
ipxe-bootimgs.noarch : Network boot loader images in bootable USB, CD, floppy and GRUB
                     : formats
```
And according to [this](https://forums.fedoraforum.org/showthread.php?297497-How-to-install-grub-legacy-in-Fedora-20) post from 2014, the package was removed.

kevinaboos:
Great, thanks for double checking! We'll stick with your solution from this PR.

PR #538, opened 2022-06-17
Fix clippy `deny`-by-default errors

jkugelman:
Clippy generates a lot of diagnostics. Fixing all the warnings would be a big undertaking, but it's quite doable to fix its handful of `deny` errors.

`cargo clippy -- -A warnings`, which runs clippy with warnings suppressed, is now error-free.

### DIFF
--- a/kernel/pmu_x86/src/lib.rs
+++ b/kernel/pmu_x86/src/lib.rs
@@ -690,7 +690,14 @@ pub fn start_samples(event_type: EventType, event_per_sample: u32, task_id: Opti
 
     sampled_events.sample_count = sample_count;
 
-    if event_per_sample > core::u32::MAX || event_per_sample <= core::u32::MIN {
+    if event_per_sample == 0 {
+        return Err("Number of events per sample invalid: must be nonzero");
+    }
+    // This check can never trigger since `event_per_sample` is a `u32`
+    // and is therefore by definition in the range `u32::MIN..=u32::MAX`.
+    // We'll check anyways, just in case `event_per_sample`'s type is changed.
+    #[allow(clippy::absurd_extreme_comparisons)]
+    if event_per_sample > core::u32::MAX || event_per_sample < core::u32::MIN {
         return Err("Number of events per sample invalid: must be within unsigned 32 bit");
     }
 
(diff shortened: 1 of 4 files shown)


kevinaboos (approved):
Great, thanks so much! Everything looks good to me.

kevinaboos:
@jkugelman at your convenience, can you check out the clippy tracking issue #526 and let us know which categories of lints are addressed by this PR? We can then add a basic CI pass for this, which enables only the default `deny` warnings from clippy that have already been addressed, in order to ensure that future changes don't introduce new code that violates said warnings.

kernel/pmu_x86/src/lib.rs
```diff
@@ -690,7 +690,14 @@ pub fn start_samples(event_type: EventType, event_per_sample: u32, task_id: Opti
 
     sampled_events.sample_count = sample_count;
 
-    if event_per_sample > core::u32::MAX || event_per_sample <= core::u32::MIN {
+    if event_per_sample == 0 {
```
kevinaboos, line 693: @Ramla-I since you have more familiarity with this code, can you describe some of the constrains on this `event_per_sample` type? It's currently `u32` but I'm wondering if it might be more accurate to use an `i32` based on some docs from the ia64 manual. 

Please advise, thanks!

Ramla-I: Events_per_sample is the number of times the required event should occur before triggering an interrupt. I don't think making it an i32 makes sense, because you wouldn't want an event to occur a negative number of times. Though I agree the old check is excessive since the value is restrained by the type system anyway.

Each PMU counter value lies in the range [0, MAX_COUNT) where the MAX_COUNT is the maximum value that can be stored in the register depending on the bit width. We program an interrupt to be triggered when the value of the counter is equal to MAX_COUNT, and we set the starting value of the counter to be (MAX_COUNT - event_per_sample).

Is there somewhere in the manual you saw something different?

(There is actually an issue here because we are making the assumption that the bit width of the registers is 32, and actually that value should be taken from the CPUID register, but we can discuss that separately.)

kevinaboos: Thanks, yeah I agree that we should open a separate issue for this. I'll do that and cite the section of the manual that talks about this register.

kevinaboos: Resolving this after opening #539. This code will be rewritten anyway.

PR #541, opened 2022-06-19
Limine bootloader support

NathanRoyer:
Now that upstream limine works out of the box with theseus, I propose to merge my fork with the main repository.

This PR adds the following:
- the `use_limine` make flag which makes the switch at build time
- when this flag is set, all modules are compressed into one `modules.cpio.lz4` file
- `mod_mgmt` was modified to extract this file at boot time (if present)
- extraction required more memory, so the kernel heap was increased from 20MB to 50MB

Notes:
- the `use_limine` flag was documented in the `help` make target and in the readme
- packing the modules into one file was required because Limine becomes very slow if you pass it a lot of modules.
- the `modules.cpio.lz4` file is not compatible with the `lz4` linux program, because our file is prepended with the original size of `modules.cpio` (to reduce allocations at boot time).
- `mod_mgmt` only tries to extract a file if its name matches `modules.cpio.lz4` exactly.

I'm open to any change as you see fit.


Makefile
```diff
@@ -71,8 +80,8 @@ check-rustc:
 BUILD_DIR := $(ROOT_DIR)/build
 NANO_CORE_BUILD_DIR := $(BUILD_DIR)/nano_core
 iso := $(BUILD_DIR)/theseus-$(ARCH).iso
-GRUB_ISOFILES := $(BUILD_DIR)/grub-isofiles
-OBJECT_FILES_BUILD_DIR := $(GRUB_ISOFILES)/modules
+BOOTLOADER_ISOFILES := $(BUILD_DIR)/bootloader-isofiles
```
kevinaboos: let's just call this `isofiles` without any prefix:
```
ISOFILES := $(BUILD_DIR)/isofiles
```

Makefile
```diff
@@ -18,10 +18,28 @@ all: iso
 debug ?= none
 net ?= none
 
+## Should we use the limine bootloader?
+use_limine ?= NO
```
kevinaboos: I think it's more typical to set up a standard argument like the ones above, and give it a default value. For example:
```make
boot ?= grub
```
And then below, check for invalid values of `boot`, just like we do for `debug` or `net`.

kevinaboos: Correspondingly, we should rename the `boot` target to `usb` since that's what it actually is for.

Makefile
```diff
@@ -18,10 +18,28 @@ all: iso
...
+ifeq ($(use_limine),NO)
+	bootloader = grub
+
+	## Look for `grub-mkrescue` (Debian-like distros) or `grub2-mkrescue` (Fedora)
+	ifneq (,$(shell command -v $(GRUB_CROSS)grub-mkrescue))
+		GRUB_MKRESCUE = $(GRUB_CROSS)grub-mkrescue
+	else ifneq (,$(shell command -v $(GRUB_CROSS)grub2-mkrescue))
+		GRUB_MKRESCUE = $(GRUB_CROSS)grub2-mkrescue
+	else
+		GRUB_MKRESCUE = not-found
```
kevinaboos: I think we should restore the existing way this is done, since it simplifies the actual `grub` target.

Makefile
```diff
@@ -304,13 +317,39 @@ endif
...
 grub:
-	@cargo run --release --manifest-path $(ROOT_DIR)/tools/grub_cfg_generation/Cargo.toml -- $(GRUB_ISOFILES)/modules/ -o $(GRUB_ISOFILES)/boot/grub/grub.cfg
-	@$(GRUB_MKRESCUE) -o $(iso) $(GRUB_ISOFILES)  2> /dev/null
-
-
-### This target copies all extra files into the `GRUB_ISOFILES` directory,
+	@if [ "$(GRUB_MKRESCUE)" = "not-found" ]; then \
+		echo "could not find grub-mkrescue!"; \
+		error "grub-mkrescue not found"; \
+	fi
```
kevinaboos: let's avoid using bash-specific features within `Make` as much as possible, as it hurts portability. See my other comment about this.

kernel/mod_mgmt/Cargo.toml
```diff
@@ -13,6 +13,8 @@ qp-trie = "0.7.3"
 cstr_core = "0.2.3"
 rangemap = "0.1.13"
 const_format = "0.2.2"
+lz4_flex = { version = "0.9.3", default-features = false }
+cpio_reader = "0.1.0"
```
kevinaboos: These should be optional dependencies based on a cargo feature, perhaps called `limine`, that can be defined in `mod_mgmt/Cargo.toml` and optionally also in the top-level `theseus_features` crate. Then, the code should also be gated with a similar `#[cfg(feature = "limine")]` block (or `cfg_if!()` macro).

kevinaboos: Also, this feature should be set in the Makefile based on the value of the `boot=` argument. I think you'd want the typical `export override CARGOFLAGS+="--features limine"` or something like that.

kevinaboos (changes requested):
Hi Nathan, thanks for the PR and sorry for the delay in reviewing, I've come down with covid.

Things generally look pretty good. I have a few small requests:
* Do we have to have `limine` as a git submodule? It's absolutely huge and takes a lot of space/time to download. 
   * Instead of forcing everyone to download it, perhaps we can add it to the set-up instructions and have folks who want to build for `limine` install it on their system elsewhere? That's a common approach with third-party tools that aren't really code dependencies, which is what git submodules are intended for.
* The `limine.cfg` file should be moved into the `cfg` directory; the root dir is already getting too crowded.

I left a few other inline comments.

kernel/mod_mgmt/Cargo.toml
NathanRoyer: I named the feature `mod_unpack` as I think it could be used with any bootloader in the future.

NathanRoyer:
@kevinaboos I agree with all your suggestions and have patched my fork.

Can you test targets `tlibc`, `ctest` and `grub` on your side to check that everything still works as before for these targets?

Thanks

kevinaboos:
> @kevinaboos I agree with all your suggestions and have patched my fork.

Great, everything looks pretty good so far. 

Unfortunately we've merged in some recent changes to the Makefile in #565 and #562 so you'll need to merge those changes in first. I think most everything should be compatible and conflict-free with this PR though.

NathanRoyer:
Sorry for the force pushes, I just found out that my first commits had a different author somehow (my linux username rather than my github account). To solve this I replaced all my previous commits with a new unique commit including all limine-related changes.

kernel/mod_mgmt/src/lib.rs
```diff
@@ -161,36 +171,75 @@ fn parse_bootloader_modules_into_files(
...
+                    {
+                        let slice = mp.as_slice_mut(0, size)?;
+                        slice.copy_from_slice(bytes);
+                    }
+                    process_module(name, size, mp)?;
+                }
+                continue;
+            }
+            #[cfg(not(feature = "mod_unpack"))]
+            debug!("modmgmt: WARNING! `modules.cpio.lz4` encountered but the `mod_unpack` feature is disabled!");
```
kevinaboos: this should be a hard error, right? you can both log an `error!()` message and return an `Err`.

kernel/mod_mgmt/Cargo.toml
```diff
@@ -13,6 +13,11 @@ qp-trie = "0.8.0"
 cstr_core = "0.2.3"
 rangemap = "0.1.13"
 const_format = "0.2.2"
+lz4_flex = { version = "0.9.3", default-features = false, optional = true }
+cpio_reader = { version = "0.1.0", optional = true }
+
+[features]
+mod_unpack = ["lz4_flex", "cpio_reader"]
```
kevinaboos: document what this feature is used for.

Also, is there a clearer name for the feature, such as "unarchive_boot_modules" or "extract_..."?  I'm open to something else, but to be nitpicky, `mod_unpack` is a bit vague to my ear.

Makefile
```diff
@@ -893,8 +922,8 @@ endif  ## end of checking for WSL
 
 
 ### Creates a bootable USB drive that can be inserted into a real PC based on the compiled .iso. 
-boot : export override THESEUS_CONFIG += mirror_log_to_vga
-boot: check-usb $(iso)
+usb : export override THESEUS_CONFIG += mirror_log_to_vga
+usb: check-usb $(iso)
```
kevinaboos, line 934: There might be a conflict with having a target called `usb` _and_ an env var called `usb`. Even if it works, it's a bit confusing. 

We can change the environment variable to `drive` or something instead of `usb` (and document it as such).

Makefile
```diff
@@ -38,13 +39,19 @@ else
...
+ifeq ($(bootloader),grub)
+	## Look for `grub-mkrescue` (Debian-like distros) or `grub2-mkrescue` (Fedora)
+	ifneq (,$(shell command -v $(GRUB_CROSS)grub-mkrescue))
+		GRUB_MKRESCUE = $(GRUB_CROSS)grub-mkrescue
+	else ifneq (,$(shell command -v $(GRUB_CROSS)grub2-mkrescue))
+		GRUB_MKRESCUE = $(GRUB_CROSS)grub2-mkrescue
+	else
+$(error Error: could not find 'grub-mkrescue' or 'grub2-mkrescue', please install 'grub' or 'grub2')
+	endif
+else ifeq ($(bootloader),limine-iso)
```
kevinaboos: nitpick: can we use just `limine` instead? it's shorter and easier

Makefile
```diff
@@ -337,13 +343,36 @@ endif
...
+limine-iso:
+	@test -d limine/ || echo "Missing ./limine/ directory! Did you follow the Readme?"
+	@cd $(OBJECT_FILES_BUILD_DIR)/ && ls | cpio --no-absolute-filenames -o > $(ISOFILES)/modules.cpio
+	@cargo run -r --manifest-path $(ROOT_DIR)/tools/limine_compress_modules/Cargo.toml -- -i $(ISOFILES)/modules.cpio -o $(ISOFILES)/modules.cpio.lz4
+	@rm -r $(OBJECT_FILES_BUILD_DIR)
+	@rm $(ISOFILES)/modules.cpio
+	@cp cfg/limine.cfg $(ISOFILES)/
+	@cp limine/limine-cd.bin $(ISOFILES)/
+	@cp limine/limine-cd-efi.bin $(ISOFILES)/
+	@cp limine/limine.sys $(ISOFILES)/
```
kevinaboos: we should use a top-level variable to define this, i.e., `LIMINE_DIR`

Makefile
```diff
@@ -337,13 +343,36 @@ endif
...
+	tree $(ISOFILES)/
```
kevinaboos: is this a debug statement? if so, we can remove or comment it out.

Makefile
```diff
@@ -337,13 +343,36 @@ endif
...
+	@cargo run --release --manifest-path $(ROOT_DIR)/tools/grub_cfg_generation/Cargo.toml -- $(ISOFILES)/modules/ -o $(ISOFILES)/boot/grub/grub.cfg
+	@$(GRUB_MKRESCUE) -o $(iso) $(ISOFILES)  2> /dev/null
+
+### This target uses limine to build a bootable ISO.
+### This target should be invoked when all of contents of `ISOFILES` are ready to be packaged into an ISO.
+limine-iso:
+	@test -d limine/ || echo "Missing ./limine/ directory! Did you follow the Readme?"
+	@cd $(OBJECT_FILES_BUILD_DIR)/ && ls | cpio --no-absolute-filenames -o > $(ISOFILES)/modules.cpio
+	@cargo run -r --manifest-path $(ROOT_DIR)/tools/limine_compress_modules/Cargo.toml -- -i $(ISOFILES)/modules.cpio -o $(ISOFILES)/modules.cpio.lz4
+	@rm -r $(OBJECT_FILES_BUILD_DIR)
```
kevinaboos: isn't this already handled by a previous invocation of the `clean-old-build` target?

I only ask because sometimes when I'm debugging or manually hacking the build process, I need to be able to add arbitrary files to the `OBJECT_FILES_BUILD_DIR`  and then re-create the ISO by manually calling `make grub`. This seems to prevent that by removing that directory unconditionally.

Makefile
kevinaboos: you can combine all of these into one:
```make
@cp cfg/limine.cfg limine/limine-cd.bin limine/limine-cd-efi.bin limine/limine.sys $(ISOFILES)/
```

Makefile
```diff
@@ -337,13 +343,36 @@ endif
...
+	@cp limine/limine-cd-efi.bin $(ISOFILES)/
+	@cp limine/limine.sys $(ISOFILES)/
+	tree $(ISOFILES)/
+	@rm -f $(iso)
+	@xorriso -as mkisofs \
+		-b limine-cd.bin -no-emul-boot -boot-load-size 4 \
+		-boot-info-table --efi-boot limine-cd-efi.bin \
+		-efi-boot-part --efi-boot-image --protective-msdos-label \
+		$(ISOFILES)/ -o $(iso)
+	@$(MAKE) -C limine
```
kevinaboos: using `LIMINE_DIR` would help clarify this too, I mistakenly thought `limine` was a new target when reading this.

README.md
```diff
@@ -153,6 +153,15 @@ make full   ## or `make all`
 Run `make help` to see other make targets and the various command-line options.
 
 
+## Using the Limine bootloader instead of GRUB
+To use Limine instead of GRUB, clone pre-built limine and pass `bootloader=limine-iso` to make:
+```
+git clone https://github.com/limine-bootloader/limine.git --branch=v3.6-binary --depth=1
```
kevinaboos: I don't think this is futureproof, as the branch `v3.6-binary` could be removed at any time. Let's use a specific commit instead.

Also, please add the language tag to the top of this code block:
```
```sh
```

kevinaboos (changes requested):
I left a few minor comments, once those are addressed we can merge this in. Thanks!

Makefile
NathanRoyer: If I leave the directory, it ends up in the iso, so the iso gets twice as big though :thinking:

NathanRoyer: It's OK now, it gets removed later from the ISO

NathanRoyer: But now qemu wont boot the iso... I'll just leave it doubly in the iso

README.md
NathanRoyer: done

Makefile
NathanRoyer: done

Makefile
NathanRoyer: done

Makefile
NathanRoyer: done

Makefile
NathanRoyer: done

Makefile
NathanRoyer: done

Makefile
NathanRoyer: done

kernel/mod_mgmt/Cargo.toml
NathanRoyer: done, renamed it to "extract_boot_modules"

kernel/mod_mgmt/src/lib.rs
NathanRoyer: yep, done

Makefile
kevinaboos: Hmm we definitely don't want something duplicated unnecessarily in the .iso. 

Is this behavior specific to `limine`? If so, why would it be duplicated?

Oh, wait, is it just a simple issue of all the modules existing in separate files and then also together in the compressed archive?

PR #542, opened 2022-06-21
Shift `nano-core` symbol parsing to compile time

### LINKED ISSUE #431: Use (de)serialization to shift nano_core symbol parsing to build time
As mentioned in #425 `readelf` should be replaced with a proper "handmade" tool. One aspect of the proposed solution is serialization/deserialization, which can be addressed by the `serde` crate. Other than that, vendoring `serde` might help in the future with similar problems.

As you may know, `serde` does not support any serialization format. There are different paths to consider:
 - Vendor a textual format serializer/deserializer like `serde_json`;
 - Vendor a binary format serializer/deserializer like `postcard`;
 - Implement and have our own.

Every crate mentioned supports `no_std`.

tsoutsman:
Closes #431.

I'm not sure about the naming of the `offset` field on `SerializedSection`. Also, I didn't correctly rebase onto the upstream repository, and so the GitHub diff viewer is playing up. E.g. it's showing that I made changes to the README, but I didn't.


kevinaboos:
Thanks for the PR! Sorry for the delay, I came down with a bad case of covid last week. 

I merged in the latest changes from the `theseus_main` branch and now the github diff viewer seems correct.

kevinaboos (changes requested):
Everything looks pretty good. A few requested changes (including those I mentioned on Discord):

* Kindly restore the existing text-based `parse_nano_core()` option that uses the .sym file extension
* Use a different file extension for this, maybe something like `nano_core.serde`, and then use that extension to determine which of the 3 options should be used in the original `parse_nano_core`
* Minimize/undo unnecessary changes to all other files that don't need to be modified (e.g., comments that change "parse" to "deserialize")
* Clarify when we are deserializing or serializing structures; `load()` is a bit ambiguous
* Tool name should be something like `serialize_nano_core`
* Add a bullet point to `tools/README` about `serialize_nano_core`
* Add item-level, crate-level, and module-level documentation to all of the new files you added (and a `description` field in your new tool's `Cargo.toml`)


In a future PR, we can look at redesigning the `LoadedSection` type to make it easier or faster to (de)serialize, perhaps by using some more arena-like data structures or indices, rather than direct strong/weak references.

tsoutsman:
>     * Clarify when we are deserializing or serializing structures; `load()` is a bit ambiguous
`load()` doesn't actually deserialize the `SerializedCrate`, the call to `bincode::serde::decode_from_slice()` in `parse_nano_core` does.

tsoutsman:
@kevinaboos should be all done now, sans two points:
* Renaming `load()`: As I mentioned above, `load()` doesn't deserialize the `SerializedCrate`, so I didn't rename it. I'm happy to oblige if you still think I should rename it.
* On Discord, you mentioned comparing the sections in `nano_core` after parsing/deserialization with the output of `readelf`. I'm not quite sure how to do that because, to compare the output of our parsing with the output of `readelf`, we need to parse `readelf` to have something to compare to :). Also, during deserialization (during boot), we don't have access to the output of `readelf`, and if we were to add it as a boot module and reparse it, all the performance gains would obviously be lost. I think I may be misunderstanding your request.

kernel/mod_mgmt/src/serde.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct SerializedSection {
+    /// The full name of the section.
+    pub name: String,
+    /// The type of the scetion.
+    pub ty: SectionType,
+    /// Whether or not the section is global.
+    pub global: bool,
+    /// The starting [`VirtualAddress`] of the range covered by this section.
+    pub virtual_address: usize,
+    /// This field is identical to `self.virtual_address` unless `self.ty == SectionType::TlsData`.
```
kevinaboos: I think this also applies for `TlsBss`. So, any kind of TLS section.

tools/README.md
```diff
@@ -5,6 +5,7 @@ This directory contains tools used in Theseus's build process or for testing pur
 ## Build-related tools
 * `copy_latest_crate_objects`: a Rust program that selects the latest version of a compiled crate object file and copies it to the OS image for creating a GRUB image. 
 * `demangle_readelf_file`: a Rust program that demangles the output of `readelf`.
+* `serialize_nano_core`: A Rust program that creates a serialized representation of the symbols in the `nano_core` object file from the output of `demangle_readelf_file`.
```
kevinaboos: The `nano_core` actually is not an object file, it's the statically-linked kernel binary.

tools/serialize_nano_core/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "serialize_nano_core"
+version = "0.1.0"
+edition = "2021"
+authors = ["Klim Tsoutsman <klim@tsoutsman.com>"]
+description = "Creates a serialized representation of the symbols in the `nano_core` object file"
```
kevinaboos: same as other comment, it's not an object file

tools/serialize_nano_core/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[dependencies]
+crate_metadata = { path = "../../kernel/crate_metadata" }
+mod_mgmt = { path = "../../kernel/mod_mgmt" }
+memory = { path = "../../kernel/memory" }
+hashbrown = "0.11"
+log = "0.4"
```
kevinaboos: I don't think `log` is a necessary dependency here. Because this is a CLI tool, we should use `println!()` instead of `error!()`, and simply comment out `trace!()` statements.

tools/serialize_nano_core/src/parse.rs
```diff
@@ -0,0 +1,403 @@
+use crate_metadata::{SectionType, Shndx};
+use hashbrown::HashMap;
+use log::{error, trace};
```
kevinaboos: see my other comment about removing `log`.

kernel/mod_mgmt/src/serde.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/// A serialized representation of a section.
+///
+/// See [`LoadedSection`] for more detail on the fields of this struct.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SerializedSection {
+    /// The full name of the section.
+    pub name: String,
+    /// The type of the scetion.
```
kevinaboos: small typo here, `scetion`

kernel/mod_mgmt/src/parse_nano_core.rs
```diff
@@ -3,156 +3,170 @@
...
-/// We consider both `GLOBAL` and `WEAK` symbols to be global public symbols; this is necessary because symbols that are 
+/// We consider both `GLOBAL` and `WEAK` symbols to be global public symbols; this is necessary because symbols that are
 /// compiler builtins, such as memset, memcpy, etc, are symbols with weak linkage in newer versions of Rust (2021 and later).
-/// 
+///
 /// # Return
 /// If successful, this returns a tuple of the following:
 /// * `nano_core_crate_ref`: A reference to the newly-created nano_core crate.
-/// * `init_symbols`: a map of symbol name to its constant value, which contains assembler and linker constances.
+/// * `init_symbols`: a map of symbol name to its constant value, which contains assembler and linker constants.
```
kevinaboos, line 36: nice, thanks for catching this

kernel/mod_mgmt/src/lib.rs
```diff
@@ -45,6 +45,7 @@ pub use crate_metadata::*;
 
 pub mod parse_nano_core;
 pub mod replace_nano_core_crates;
+pub mod serde;
```
kevinaboos, line 48: do we want this to be public? I don't think it should be accessible by other crates.

kernel/mod_mgmt/src/serde.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// A set containing the `.data` and `.bss` sections of the crate.
+    pub data_sections: BTreeSet<Shndx>,
+    /// A map of symbol names to their constant values, which contain assembler and linker
+    /// constants.
+    pub init_symbols: BTreeMap<String, usize>,
+}
+
+impl SerializedCrate {
+    /// Load the crate and its sections.
+    pub fn load(
```
kevinaboos: I don't think this needs to be fully public, it could be `pub(crate) fn` but that's not important if we make the `serde` module non-public.
```rust
Convert this into a `LoadedCrate` and its sections into `LoadedSection`s.
pub(crate) fn into_loaded_crate(...
```

tools/serialize_nano_core/src/main.rs
```diff
@@ -0,0 +1,30 @@
+//! Tool that creates a serialized representation of the symbols in the `nano_core` object file.
+
+mod parse;
+
+use mod_mgmt::serde::SerializedCrate;
+use std::io::Write;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let arg = &std::env::args().collect::<Vec<String>>()[1];
```
kevinaboos: nit:  if you just want the first arg, you can do `args().nth(1)` instead of collecting them all into a vector

tools/serialize_nano_core/src/main.rs
```diff
@@ -0,0 +1,30 @@
...
+    let str = std::fs::read_to_string(arg)?;
```
kevinaboos: i'm surprised this actually compiles, since `str` is a reserved keyword. Let's change it to avoid potential future compiler errors.

kevinaboos (changes requested):
Changes look good in general. Can you restore the `parse_nano_core_binary()` function that you removed, and the code that invokes it? I'd still like for that to be an option available in the future.

Let's rename the `load()` function to be more explicit, i.e., `into_loaded_section()`. To my ear, `load()` implies you'll actually be loading something into memory or into the system somehow. Instead, using the term `into` fits well with Rust's typical lingo surrounding conversion, which is what we're doing -- converting a `SerializedSection` into a `LoadedSection`. Same goes for the `SerializedCrate::load()` function.

kevinaboos:
> On Discord, you mentioned comparing the sections in nano_core after parsing/deserialization with the output of readelf. I'm not quite sure how to do that because, to compare the output of our parsing with the output of readelf, we need to parse readelf to have something to compare to :). > Also, during deserialization (during boot), we don't have access to the output of readelf, and if we were to add it as a boot module and reparse it, all the performance gains would obviously be lost. I think I may be misunderstanding your request.

Ah sorry, to clarify, I didn't mean doing this at runtime, I meant just double-checking our work before we merge this in. We can simply log/dump out the details of all `LoadedSection`s in the `nano_core` `LoadedCrate` after we initially parse them, and then compare that with the output of `readelf` on linux. This is nothing more than adding some logging code to make sure that your (de)serialization routines create the correct sections that match those from the actual ELF file. You can commit the code for logging the section details to this branch, then share the log output and `readelf` output with me either here or on Discord. Then we can remove it once we're sure it's correct.

kernel/mod_mgmt/src/serde.rs
tsoutsman: I don't think it does. On the main branch the offset for `TlsBss` sections is [set to the virtual address (offset)](https://github.com/theseus-os/Theseus/blob/44d8df954560416d1d40490857b7503759a5dc46/kernel/mod_mgmt/src/parse_nano_core.rs#L731), whereas for `TlsData` [adds the virtual address (offset) to the address of the section](https://github.com/theseus-os/Theseus/blob/44d8df954560416d1d40490857b7503759a5dc46/kernel/mod_mgmt/src/parse_nano_core.rs#L711). Although `TlsBss` has the same comment as the `TlsData` code and so it *may* be a bug.

kernel/mod_mgmt/src/lib.rs
tsoutsman: `serialize_nano_core` relies on the types from `serde`. I'll still make the loading functions `pub(crate)`.

tsoutsman:
I've sent the `readelf` output through Discord output because GitHub was not having 1.3 MB of text in a comment.

kernel/mod_mgmt/src/lib.rs
kevinaboos: ah, right. Forgot about that.

kernel/mod_mgmt/src/serde.rs
kevinaboos: Yeah, on the main branch it's an unused placeholder value for `TlsBss` sections, since they don't actually exist anywhere in memory and are sort of placeholders themselves.

tsoutsman:
So I've made a diff between the symbols after parsing and after deserialization:
```bash
make run 2> serialize_output
# DO: as soon as QEMU opens, ctrl+c
# DO: manually remove all the lines around the printed-out types in serialize_output
sort -o serialize_output serialize_output_diff

make run > deserialize_output
# DO: when theseus boots (i.e. command line displayed), ctrl+c
# DO: manually remove all the lines around the printed-out types in deserialize_output
# the following command strips terminal escape codes and extra spaces
cut -c 45- < deserialize_output | rev | cut -c 5- | rev > deserialize_output_diff
sort -o deserialize_output_diff deserialize_output_diff

diff serialize_output_diff deserialize_output_diff
```

```diff
798,800d797
< ap_start_protected_mode|v0x100018|true
< ap_start_realmode_end|v0x100376|true
< ap_start_realmode|v0x100102|true
4809a4807
> .eh_frame|v0xFFFFFFFF803EB000|false
4831d4828
< _error|v0x1004EF|true
5078a5076
> .gcc_except_table|v0xFFFFFFFF80466000|false
5559,5561d5556
< initial_bsp_stack_bottom|v0xFFFFFFFF80801000|true
< initial_bsp_stack_guard_page|v0xFFFFFFFF80800000|true
< initial_bsp_stack_top|v0xFFFFFFFF80811000|true
6008,6009d6002
< long_mode_start_ap|v0x1000C0|true
< long_mode_start|v0x100513|true
7707d7699
< set_up_SSE|v0x1004C3|true
8376d8367
< start|v0x100377|true
```
The `ap_start`, `long_mode_start`, `set_up_SSE`, and `start` symbols are contained in `init_symbols` as they are in the `.init` section. The `initial_bsp_stack` symbols are contained in `init_symbols` as they are in the `.stack` section.

The `.eh_frame` and `.gcc_except_table`do not have `.symtab` entries and are added to `serialized_crate.sections` separately.

The output seems to indicate no bugs in the implementation of serialization or deserialization. I'm not sure how I'm supposed to test the parsing of `readelf`'s output.

Also, it would be nice to add this as a test when the kernel's testing infrastructure is built out.

Prior to merging, the print statements in `tools/serialize_nano_core`, and `kernel/mod_mgmt` need to be removed.

kernel/mod_mgmt/src/serde.rs
tsoutsman: In the most recent commit, I changed the documentation and changed the logic in `parse.rs` to reflect this. Can you check if I've resolved the issue correctly?

tsoutsman: In the most recent commit, I modified this comment and `parse.rs` to reflect this. Can you make sure the change to `parse.rs` in the most recent commit is what you meant?

kevinaboos:
@tsoutsman I merged in my port of `wasmtime`, which is massive, in #556. This relies on `bincode` that I ported as a submodule (long before the `bincode` issue you & I mentioned on Discord was fixed), so there are some conflicts there. 

You can ignore that conflict for now; once everything else in this PR is finalized, I will go ahead and modify the rest of Theseus to use the current version of `bincode` on git, the same one you use in this PR. Then there shouldn't be any conflicts.

kernel/mod_mgmt/src/serde.rs
kevinaboos: yeah I think that's fine.

kevinaboos:
#558 was merged in, so you can now use `version = "2.0.0-rc.1` for bincode deps.

tsoutsman:
Take two:
```sh
make run
# DO: when QEMU opens, ctrl+c
awk '{print $2 " " substr($0, index($0, $8))}' < readelf_output > readelf_output_diff 
sort -o readelf_output_diff readelf_output_diff

make run > deserialize_output
# DO: when theseus boots (i.e. command line displayed), ctrl+c
# DO: manually remove all the lines around the printed-out types in deserialize_output
# the following command strips terminal escape codes and extra spaces
cut -c 45- < deserialize_output | rev | cut -c 5- | rev > deserialize_output_diff
sort -o deserialize_output_diff deserialize_output_diff

diff readelf_output_diff deserialize_output_diff
```

```diff
1,2d0
<  
<  
5,13d2
< 0000000000100018 ap_start_protected_mode
< 00000000001000c0 long_mode_start_ap
< 0000000000100102 ap_start_realmode
< 0000000000100376 ap_start_realmode_end
< 0000000000100377 start
< 00000000001004c3 set_up_SSE
< 00000000001004ef _error
< 0000000000100513 long_mode_start
< ffffffff80000000 KERNEL_OFFSET
10277a10267,10268
> ffffffff803ea000 .eh_frame
> ffffffff80465000 .gcc_except_table
10389,10393d10379
< ffffffff80800000 initial_bsp_stack_guard_page
< ffffffff80801000 initial_bsp_stack_bottom
< ffffffff80811000 initial_bsp_stack_top
< table Symbol table '.symtab' contains 12098 entries:
< Value Name
```
The first two and last two differences are `readelf` output artifacts. The other symbols were explained in the comment above.

Also, it would be nice to add this as a test when the kernel's testing infrastructure is built out.

Prior to merging, the print statement in `kernel/mod_mgmt` needs to be removed, and the command to dump `readelf`'s output in the Makefile.

tsoutsman:
I restored `parse_nano_core_binary` but booting with it didn't work because it didn't add any global symbols that weren't either functions or objects to `init_symbols`. I changed this behaviour resulting in the following additions to `init_symbols`:
```
name: eputs
section: .text

name: start_high_ap
section: .text

name: initial_bsp_stack_bottom
section: .stack

name: ap_start_protected_mode
section: .init

name: start
section: .init

name: KEXIT
section: .text

name: KERNEL_OFFSET
section: 65521

name: long_mode_start
section: .init

name: _error
section: .init

name: initial_bsp_stack_top
section: .stack

name: set_up_SSE
section: .init

name: ap_start_realmode
section: .init

name: initial_bsp_stack_guard_page
section: .stack

name: ap_start_realmode_end
section: .init

name: long_mode_start_ap
section: .init

name: puts
section: .text

name: start_high
section: .text
```
All these symbols are present in `init_symbols` when parsing using the other methods.

The only change to `parse_nano_core_binary` was the condition for adding a section:
 ```diff
< typ == xmas_elf::symbol_table::Type::Func || typ == xmas_elf::symbol_table::Type::Object
> (typ == xmas_elf::symbol_table::Type::Func || typ == xmas_elf::symbol_table::Type::Object) || global
```

I also added a comment in the Makefile with the command to include the `nano_core` binary as a boot module.

**Comment by kevinaboos** [comment]
Thanks for fixing the binary parsing! We don't use it but it's always good to bring it up to date and back into a working state.

Looks like the verification step is all good, I will double-check it later today on my machine before approving this PR.

File: Makefile [hunk 12]
```diff
@@ -277,18 +277,25 @@ $(nano_core_binary): cargo $(nano_core_static_lib) $(assembly_object_files) $(li
...
+		<($(CROSS)readelf -s -W $(nano_core_binary) | sed '/OBJECT  LOCAL .* str\./d;/NOTYPE  LOCAL  /d;/FILE    LOCAL  /d;/SECTION LOCAL  /d;') \
+ 		>  $(ROOT_DIR)/readelf_output
 ## run "readelf" on the nano_core binary, remove irrelevant LOCAL symbols from the ELF file, demangle it, serialize it, and then output to a serde file
 	@cargo run --release --manifest-path $(ROOT_DIR)/tools/serialize_nano_core/Cargo.toml \
 		<(cargo run --release --manifest-path $(ROOT_DIR)/tools/demangle_readelf_file/Cargo.toml \
 		<($(CROSS)readelf -S -s -W $(nano_core_binary) \
 		| sed '/OBJECT  LOCAL .* str\./d;/NOTYPE  LOCAL  /d;/FILE    LOCAL  /d;/SECTION LOCAL  /d;')) \
 		> $(OBJECT_FILES_BUILD_DIR)/$(KERNEL_PREFIX)nano_core.serde
-## Original command: this doesn't parse the object file at compile time, instead placing the modified output of "readelf" into a boot module so it can then
+## `.sys`: this doesn't parse the object file at compile time, instead including the modified output of "readelf" as a boot module so it can then
```
kevinaboos: typo, should be 
```
`.sym`:
```

kernel/mod_mgmt/src/parse_nano_core.rs
```diff
@@ -3,89 +3,165 @@
...
+    (StrongCrateRef, BTreeMap<String, usize>, usize),
+    (&'static str, [Arc<Mutex<MappedPages>>; 3]),
+> {
+    let text_pages = Arc::new(Mutex::new(text_pages));
     let rodata_pages = Arc::new(Mutex::new(rodata_pages));
-    let data_pages   = Arc::new(Mutex::new(data_pages));
+    let data_pages = Arc::new(Mutex::new(data_pages));
+
+    /// Just like Rust's `try!()` macro, but packages up the given error message in a tuple
+    /// with an array of the above 3 MappedPages objected.
```
kevinaboos: oops, this is probably my typo:
`objected` --> `objects`

kernel/mod_mgmt/src/serde.rs
```diff
@@ -0,0 +1,200 @@
...
+        info!(
+            "Finished parsing nano_core crate, {} new symbols.",
+            num_new_syms
+        );
+        
+        let loaded_crate_ref = loaded_crate.lock_as_ref();
+        for (_, section) in loaded_crate_ref.sections.iter() {
+            trace!("{:016x} {}", section.address_range.start.value(), section.name);
+        }
+        drop(loaded_crate_ref);
```
kevinaboos: you can leave this in but just comment it out, since we might use it in the future to double check things.

kernel/mod_mgmt/src/serde.rs
```diff
@@ -0,0 +1,200 @@
...
+
+        trace!(
+            "SerializedCrate::load(): adding symbols to namespace {:?}...",
+            namespace.name
+        );
+        let num_new_syms = namespace.add_symbols(sections.values(), verbose_log);
+        trace!("SerializedCrate::load(): finished adding symbols.");
+
+        let mut loaded_crate_mut = loaded_crate.lock_as_mut().ok_or(
+            "BUG: SerializedCrate::load(): couldn't get exclusive mutable access to loaded_crate",
```
kevinaboos: function is no longer called `load`, the log statements need updating.

kernel/mod_mgmt/src/serde.rs
```diff
@@ -0,0 +1,200 @@
...
+            SectionType::Text => Arc::clone(text_pages),
+            SectionType::Rodata
+            | SectionType::TlsData
+            | SectionType::TlsBss
+            | SectionType::GccExceptTable
+            | SectionType::EhFrame => Arc::clone(rodata_pages),
+            SectionType::Data | SectionType::Bss => Arc::clone(data_pages),
+        };
+        let virtual_address = VirtualAddress::new(self.virtual_address)
+            .ok_or("SerializedSection::load(): invalid virtual address")?;
```
kevinaboos: function no longer called `load`

kernel/mod_mgmt/src/serde.rs
```diff
@@ -0,0 +1,200 @@
...
+            // since they are read-only initializer sections that would hold all zeroes.
+            // Thus, we just use a max-value mapped pages offset as a canary value here,
+            // as that value should never be used anyway.
+            mapped_pages_offset: match self.ty {
+                SectionType::TlsBss => usize::MAX,
+                _ => mapped_pages
+                    .lock()
+                    .offset_of_address(
+                        VirtualAddress::new(self.offset)
+                            .ok_or("SerializedSection::load(): invalid offset")?,
```
kevinaboos: function no longer called `load`

kevinaboos (changes requested):
I confirmed the verification results and everything looks good. I left a few small comments, once those are addressed I will approve and merge.

PR #543, opened 2022-06-21
Streamline task code with a little bit better naming and panic ergonomics

jkugelman:
# Make fallible and infallible versions of `get_my_current_task` and `get_my_current_task_id`

* Briefly during startup there may not be a task, but once the first task is bootstrapped there's always a task. Application code and OS code that is only run after bootstrapping shouldn't have to check for a `None` current task.

* I've made `get_my_current_task` infallible so it returns a `&'static TaskRef` directly, and added a fallible `try_get_my_current_task` method that returns a `Result` for low-level code that could run during early startup. The same goes for `get_my_current_task_id`.

* I updated all the usage sites of these functions -- and there are a lot. This allowed a lot of `match` blocks, unwraps, and other similar error handlers to be removed from a lot of code, particularly all of the applications. For low-level code that calls the `try_` versions I mainly just had to insert `try_` and then tweak them a little since the fallible functions return `Result` instead of `Option`.

* I made liberal use of the soon-to-be-stabilized `let`-`else` feature in the updated code.

* One knock-on effect is that the shell's `AppErr::NamespaceErr` variant is no longer needed.
 
# Drop `get_my` from function names

* Rename `task::get_my_current_task` to `task::current_task`.
* Rename `task::get_my_current_task_id` to `task::current_task_id`.
* Rename `apic::get_my_apic` to `apic::current_apic`.
* Rename `apic::get_my_apic_id` to `apic::current_apic_id`.

# Inline and remove `Task::get_namespace()`

* `Task::namespace` is already a public field. There's no need for a getter function. Getters aren't idiomatic in Rust.

### DIFF
--- a/applications/bm/src/lib.rs
+++ b/applications/bm/src/lib.rs
@@ -233,7 +233,7 @@ fn do_null() -> Result<(), &'static str> {
 }
 
 /// Internal function that actually calculates the time for null syscall.
-/// Measures this by calling `get_my_current_task_id` of the current task. 
+/// Measures this by calling `current_task_id` of the current task. 
 fn do_null_inner(overhead_ct: u64, th: usize, nr: usize) -> Result<u64, &'static str> {
 	let start_hpet: u64;
 	let end_hpet: u64;
@@ -245,7 +245,7 @@ fn do_null_inner(overhead_ct: u64, th: usize, nr: usize) -> Result<u64, &'static
 
 	start_hpet = hpet.get_counter();
 	for _ in 0..tmp_iterations {
-		mypid = task::get_my_current_task_id();
+		mypid = task::current_task_id();
 	}
 	end_hpet = hpet.get_counter();
 
@@ -321,8 +321,8 @@ fn do_spawn_inner(overhead_ct: u64, th: usize, nr: usize, _child_core: u8) -> Re
 	let hpet = get_hpet().ok_or("Could not retrieve hpet counter")?;
 
 	// Get path to application "hello" that we're going to spawn
-	let namespace = task::get_my_current_task().get_namespace().clone();
-	let namespace_dir = task::get_my_current_task().get_namespace().dir().clone();
+	let namespace = task::current_task().get_namespace().clone();
+	let namespace_dir = task::current_task().get_namespace().dir().clone();
 	let app_path = namespace_dir.get_file_starting_with("hello-")
 		.map(|f| Path::new(f.lock().get_absolute_path()))
 		.ok_or("Could not find the application 'hello'")?;
@@ -1543,12 +1543,12 @@ fn do_fs_delete_inner(fsize_b: usize, overhead_ct: u64) -> Result<(), &'static s
 
 /// Helper function to get the name of current task
 fn get_prog_name() -> String {
-    task::get_my_current_task().name.clone()
+    task::current_task().name.clone()
 }
 
 /// Helper function to get the PID of current task
 fn getpid() -> usize {
-    task::get_my_current_task().id
+    task::current_task().id
 }
 
 
@@ -1567,7 +1567,7 @@ fn hpet_2_time(msg_header: &str, hpet: u64) -> u64 {
 
 /// Helper function to get current working directory
 fn get_cwd() -> Option<DirRef> {
-	Arc::clone(&task::get_my_current_task().get_env().lock().working_dir)
+	Arc::clone(&task::current_task().get_env().lock().working_dir)
 }
 
 /// Helper function to make a temporary file to be used to measure read open latencies
--- a/applications/bm/src/lib.rs
+++ b/applications/bm/src/lib.rs
@@ -245,7 +245,7 @@ fn do_null_inner(overhead_ct: u64, th: usize, nr: usize) -> Result<u64, &'static
 
 	start_hpet = hpet.get_counter();
 	for _ in 0..tmp_iterations {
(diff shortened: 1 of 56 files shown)


jkugelman:
The first commit requires more scrutiny than the second and third. There were many places I had to choose whether to call `try_get_my_current_thread()` or `get_my_current_thread()`. Any time I chose the latter introduces a possible panic site if that code is called before the first OS task is created.

jkugelman:
I split the PR into three commits so you can pick and choose if you don't like all the changes.

kevinaboos (changes requested):
Thanks for the PR, and sorry for the delay in my review due to covid. I had a lengthy discussion typed up here but chrome crashed and i lost it. The high points are as follows:

* Kernel code should never assume the existence of a current task or the initialization of the task subsystem; merely because it works that way now doesn't mean it always will.
* Code that may execute within unwinding, interrupt, or task mgmt contexts should definitely not assume tasking exists, and it should also never panic.
* Returning an `Option`/`Result` is always better than panicking.
* Divergent error handling (panic/exception) may look nicer at the source code level but causes generated code to be much larger and more complex. That's worse than extra conditional branches for non-divergent error handling, not to mention that panics also cause extra runtime overhead via similar conditionals.
* Panicking requires unwinding to exist and to work properly; task management shouldn't rely on that if at all possible.

I'm fully on board with the function name changes. 

I do think applications should assume the existence of a current task context, but that should only be offered through the `libtheseus` crate, which is intended to be the sole crate that applications use to access Theseus-provided functionality. We just haven't gotten there yet, but I would accept a different PR that adds an infallible `current_task()` to `libtheseus`.

The existing `current_task()` should still return an `Option<TaskRef>`. 

Removal of the `Task::namespace()` function is fine, but we may need to add it back in if we refactor the `Task`/`TaskRef` struct co-design (as that's why it was originally there).

kevinaboos:
When reading the way you changed `set[_my]_kill_handler` and related functions, I realized that they shouldn't work like they currently do, and #551 was the reason I had originally added the `set_my_kill_handler()` function.... but I didn't quite do it properly initially.

See #551 for the proper fix, which you can incorporate to this PR.

jkugelman:
Today there is kernel code that panics if there's no current task. It seems to me that only higher-level kernel code does this. Lower-level code tends to return `Err` instead of panicking. That observation is what led to my first bullet:

> * Briefly during startup there may not be a task, but once the first task is bootstrapped there's always a task. Application code and OS code that is only run after bootstrapping shouldn't have to check for a None current task.

Given the existence of both panicking and non-panicking kernel code, are you sure you don't want two functions? My intention was primarily to commonize all the `expect` calls that are all over the place, put them in one place inside `current_task`. I tried not to add any panics to non-panicking code (and I can give my PR a once-over before you approve it to make sure that's the case).

**Alternate idea:** If you don't like having separate `current_task` and `try_current_task` methods in the kernel, would you like me to tackle this a different way? I could remove all the panics from kernel code. Get rid of any calls to `expect`s and `unwrap`s and propagate the errors instead.

I suspect that will have a ripple effect. There are probably functions that get the current task that don't return `Result`, so this might cause substantial error-propagation-rippleage.

jkugelman:
> The existing `current_task()` should still return an `Option<TaskRef>`.

How about `Result<TaskRef, &'static str>`, to make propagation simpler?

kevinaboos:
> Today there is kernel code that panics if there's no current task. It seems to me that only higher-level kernel code does this. Lower-level code tends to return `Err` instead of panicking. 

That should only be the case if there is no way to return an error from the current context. If you find cases where `unwrap`/`expect` is used but returning an error is possible, please let me know.

> Given the existence of both panicking and non-panicking kernel code, are you sure you don't want two functions? My intention was primarily to commonize all the `expect` calls that are all over the place, put them in one place inside `current_task`. I tried not to add any panics to non-panicking code (and I can give my PR a once-over before you approve it to make sure that's the case).

Yes, I don't really see the benefit of two functions, especially since I want to avoid hiding/masking panicking operations whenever possible. Another problem with "commonizing all the `expect` calls" is that it can be hard to determine where it came from, so location-specific error messages are preferable.

> **Alternate idea:** If you don't like having separate `current_task` and `try_current_task` methods in the kernel, would you like me to tackle this a different way? I could remove all the panics from kernel code. Get rid of any calls to `expect`s and `unwrap`s and propagate the errors instead.

Yes, this would be preferable. I'd thought we already do that, but it's possible we missed some places.

> How about Result<TaskRef, &'static str>, to make propagation simpler?

Generally, `Result` is best for when a function may have some kind of side effect, while `Option` is best for pure `get()`-like functions. The Rust docs also recommend this, so I think it's best to avoid surprising users/developers. I understand that makes things a bit tedious propagation-wise, but it should be easy enough; plus, allows the calling context to define its own error message. In contrast, a single system-wide generic error message there actually makes it harder to discover where the problem originated.

Also, there are cases where we want to propagate `None` using the `try`/`?` operator too, it's not always a `Result`.

kevinaboos:
In any case, I'm happy to accept this PR pending the recommended changes I suggested [above](https://github.com/theseus-os/Theseus/pull/543#pullrequestreview-1025567922)

jkugelman:
Let me close this so it's not just hanging out getting stale. I haven't been able to work on it for a few weeks. I'll reopen or re-create this PR when I get a chance to work on it again.

PR #544, opened 2022-06-23
Merge core logic of allocators

tsoutsman:
This is an implementation of the first dot point under "Rust Oriented" projects in the [wiki](https://github.com/theseus-os/Theseus/wiki). It uses generics on `Address<T>`, `Chunk<T>` and `ChunkRange<T>` to differentiate between physical and virtual memory (e.g. `Address<Physical>` is equivalent to `PhysicalAddress`, `Chunk<Physical>` is equivalent to `Frame`).

I'm not too sure about the naming of `Chunk` and `ChunkRange`. The Theseus book explicitly uses frames to refer to physical memory, and pages to refer to virtual memory, but online, the line is blurred (usually physical = "Page Frame" and virtual = "Virtual Page"). @jkugelman suggested renaming `Chunk` to `Page`, and I think that does make sense, but it would require changing the terminology across the entire project. The term `Chunk` also conflicts with the `Chunk` struct in the allocators.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1684,13 +1684,8 @@ dependencies = [
 name = "memory_structs"
 version = "0.1.0"
 dependencies = [
- "bit_field 0.7.0",
- "derive_more",
  "entryflags_x86_64",
  "kernel_config",
- "multiboot2",
- "paste",
- "xmas-elf",
  "zerocopy",
 ]
 
@@ -2183,12 +2178,6 @@ name = "parity-wasm"
 version = "0.42.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "be5e13c266502aadf83426d87d81a0f5d1ef45b8027f5a471c360abfe4bfae92"
-
-[[package]]
-name = "paste"
-version = "1.0.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "acbf547ad0c65e31259204bd90935776d1c693cec2f4ff7abb7a1bbbd40dfe58"
 
 [[package]]
 name = "path"
--- a/kernel/memory_structs/Cargo.toml
+++ b/kernel/memory_structs/Cargo.toml
@@ -7,6 +7,11 @@ edition = "2021"
 
 [dependencies]
 zerocopy = "0.5.0"
+intrusive-collections = "0.9.0"
+spin = "0.9.0"
+
+[dependencies.log]
+version = "0.4.8"
 
 [dependencies.kernel_config]
 path = "../kernel_config"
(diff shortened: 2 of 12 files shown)


jkugelman:
My thought on terminology is that "page" and "frame" are used right now to give short names to physical and virtual pages. "Page" and "frame" roll off the tongue a lot better than "virtual page" and "physical page". And then the documentation often wants to refer to "physical and virtual pages", but that's super awkward, so "chunks" it is.

In the spirit of this PR unifying the code, it would be great to also unify the terminology throughout the project, eh? Just say "page" everywhere and drop the terms "frame" and "chunk". What do you all think?

TL;DR - "Chunk" is a silly word. 😛

tsoutsman:
I just implemented `AllocatedChunks` (`AllocatedFrames` /  `AllocatedPages`) and `ChunkRangeWrapper` (a temporary rename of the old `Chunk` in the allocators), and let's just say it's less than ideal. The physical and virtual `AllocatedChunks` are identical, except in their `Drop` implementations. But `Drop` implementations can't be specialised to particular generics (i.e. different `Drop` implementation for `AllocatedChunks<Virtual>` vs `AllocatedChunks<Physical>` are not allowed), leading to `AllocatedChunks` being a wrapper around a non-generic inner type: what we are trying to avoid in the first place. `kernel/memory_structs/src/allocator/allocated.rs` is a mix of macros, associated types, trait bounds and is just generally hard to follow.

`ChunkRangeWrapper` contains different data for the physical and virtual implementations, again requiring an associated type in `MemoryType`. That's the general pattern I'm following. All structs whose data differs in the physical vs virtual instances have an extra `inner` field that has the type of an associated type of `MemoryType` (that's a mouthful). This associated type is bound by some trait that generalises over the internal data of the two implementations.

I think it will be better from here on because other than `Drop`, all traits and functions (such as `allocator::init`) can be specialised to differ for virtual and physical implementations. But, I'm also hesitant to push on without approval because I'm making significant changes to the memory module which make the code more Rusty (whatever that means) but have no tangible performance benefits and increase the code complexity by a lot.

kevinaboos:
Sorry for the delay in responding to this. I do appreciate the effort undertaken in this PR, but unfortunately we are in the midst of a formal verification effort to prove the correctness of memory management related to `MappedPages`, `AllocatedFrames`/`AllocatedPages`, and more. This will result in potential design changes to the `page_allocator` and `frame_allocator` in order to make them more amenable to verification.

The bullet point for this project idea in the wiki was intended to cover smaller-scope changes than those made here -- basically just deduplicate the code across `page_allocator` and `frame_allocator`. The major changes to `VirtualAddress` and `PhysicalAddress` seem to unnecessarily complicate the usage of address types, imo. If we can simplify this PR and limit the changes to just deduplicating the allocators, then I'd be open to accepting it.

cc'ing @Ramla-I who is leading the verification effort.

PR #545, opened 2022-06-23
Async sleep

jkugelman:
Tasks:
- [x] Convert `sleep` crate to be asynchronous.
- [ ] Update all code that uses `sleep` to be async.
- [ ] Integrate an async executor into Theseus.
- [ ] Allow application `main()` functions to be `async`.

### DIFF
--- a/kernel/sleep/src/lib.rs
+++ b/kernel/sleep/src/lib.rs
@@ -18,7 +18,7 @@ extern crate scheduler;
 use core::sync::atomic::{Ordering, AtomicUsize};
 use alloc::collections::binary_heap::BinaryHeap;
 use irq_safety::MutexIrqSafe;
-use task::{get_my_current_task, TaskRef};
+use task::{current_task, TaskRef};
 
 /// Contains the `TaskRef` and the associated wakeup time for an entry in DELAYED_TASKLIST.
 #[derive(Clone, Eq, PartialEq)]
@@ -112,7 +112,7 @@ pub fn sleep(duration: usize) {
     let current_tick_count = TICK_COUNT.load(Ordering::SeqCst);
     let resume_time = current_tick_count + duration;
 
-    let current_task = get_my_current_task().clone();
+    let current_task = current_task().clone();
     // Add the current task to the delayed tasklist and then block it.
     add_to_delayed_tasklist(SleepingTaskNode{taskref: current_task.clone(), resume_time});
     current_task.block();
--- a/kernel/sleep/src/lib.rs
+++ b/kernel/sleep/src/lib.rs
@@ -9,128 +9,219 @@
 //! TODO: use regular time-keeping abstractions like Duration and Instant.
 
 #![no_std]
+#![feature(let_else)]
+
 extern crate task;
 extern crate irq_safety;
 extern crate alloc;
 #[macro_use] extern crate lazy_static;
 extern crate scheduler;
 
-use core::sync::atomic::{Ordering, AtomicUsize};
 use alloc::collections::binary_heap::BinaryHeap;
+use alloc::sync::Arc;
+use core::cmp;
+use core::future::Future;
+use core::pin::Pin;
+use core::sync::atomic::{Ordering, AtomicUsize};
+use core::task::{Context, Poll, Waker};
+
 use irq_safety::MutexIrqSafe;
-use task::{current_task, TaskRef};
 
-/// Contains the `TaskRef` and the associated wakeup time for an entry in DELAYED_TASKLIST.
-#[derive(Clone, Eq, PartialEq)]
-struct SleepingTaskNode {
-    resume_time: usize,
-    taskref: TaskRef,
+/// A duration given in ticks of the [APIC timer].
+/// 
+/// [APIC timer]: interrupts::lapic_timer_handler
+pub type Ticks = usize;
+
+pub type AtomicTicks = AtomicUsize;
+
+/// A sleeper is the `Future` returned by the asynchronous sleep functions.
+#[derive(Debug, Clone)]
+struct Sleeper {
+    /// The absolute time in ticks at which the sleeper will wake up.
+    wake_time: Ticks,
+    /// Mutable inner state.
+    inner: Arc<MutexIrqSafe<SleeperInner>>,
 }
 
-// The priority queue depends on `Ord`.
-// Explicitly implement the trait so the queue becomes a min-heap
-// instead of a max-heap.
-impl Ord for SleepingTaskNode {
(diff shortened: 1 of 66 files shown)


kernel/sleep/src/lib.rs
```diff
@@ -9,128 +9,219 @@
...
+    let sleeper = Sleeper::new(wake_time);
 
-    if resume_time > current_tick_count {
-        sleep(resume_time - current_tick_count);
+    let mut sleepers = SLEEPERS.lock();
+    sleepers.push(sleeper.clone());
+
+    let next_wake_time = NEXT_WAKE_TIME.load(Ordering::SeqCst);
+    if sleeper.wake_time < next_wake_time {
+        NEXT_WAKE_TIME.store(sleeper.wake_time, Ordering::SeqCst);
```
tsoutsman, line 200:
```suggestion
    let _ = NEXT_WAKE_TIME.fetch_update(Ordering::SeqCst, Ordering::SeqCst, |next_wake_time| {
        if sleeper.wake_time < next_wake_time {
            Some(sleeper.wake_time)
        } else {
            None
        }
    });
```
I'm pretty sure the original code has a race condition. `next_wake_time` could be modified between the load and store, either by another core or by the current task being preempted.

kernel/sleep/src/lib.rs
```diff
@@ -9,128 +9,219 @@
...
-#[derive(Clone, Eq, PartialEq)]
-struct SleepingTaskNode {
-    resume_time: usize,
-    taskref: TaskRef,
+/// A duration given in ticks of the [APIC timer].
+/// 
+/// [APIC timer]: interrupts::lapic_timer_handler
+pub type Ticks = usize;
+
+pub type AtomicTicks = AtomicUsize;
```
tsoutsman, line 35:
```suggestion
pub type Ticks = u64;

pub type AtomicTicks = AtomicU64
```
I know ticks were stored as a `usize` before this pull request, but might as well comment on it now. `usize`s are generally used for computation related to memory, so ticks should be a `u64` (or `u32`?). It doesn't make sense for the type of `TICK_COUNT` to change based on the target architecture.

Also, more specific to this PR, is it really necessary to use type aliases here?

kernel/sleep/src/lib.rs
```diff
@@ -9,128 +9,219 @@
...
-    sleep_until(new_resume_time);
+/// Returns a timer that can be used to sleep for predictable durations correcting for drift.
+/// For example, if you wanted to sleep for precise 100ms intervals
+/// but there could be anywhere from 0-5ms of latency between calls to `sleep`,
+/// this would compensate for the variable delay and sleep up to 5ms less as needed.
+pub fn regular_timer() -> RegularTimer {
+    RegularTimer {
+        last_wake_time: tick_count(),
+    }
+}
```
tsoutsman, line 214: This should probably be an associated `new` method, rather than a separate function.

kernel/sleep/src/lib.rs
```diff
@@ -9,128 +9,219 @@
...
+    last_wake_time: Ticks,
+}
+
+impl RegularTimer {
+    /// Blocks asynchronously for `duration` ticks, correcting for any delay since the last
+    /// `sleep` call.
+    pub async fn sleep(&mut self, duration: Ticks) {
+        sleep_until(self.last_wake_time + duration).await;
+        self.last_wake_time += duration;
+    }
```
tsoutsman, line 226: What happens if a user performs some computation in a loop after calling `timer.sleep(duration)`?

For example, in the code:
```rust
let timer = regular_timer();
loop {
    timer.sleep(100);
    sleep(50);
}
```
If I understand correctly, the program will only sleep 100 ticks per iteration.

tsoutsman (changes requested):
I've left a review with a couple of thoughts.

I think some of the memory orderings could be changed to `Acquire`/`Release` but I don't dare touch orderings with a ten-foot pole.

kernel/sleep/src/lib.rs
jkugelman: Interrupts are disabled so there shouldn't be any preemption. I'll take a closer look at your suggestion when I get back from vacation. If it's equivalent and doesn't rely on preemption being disabled, all the better.

kernel/sleep/src/lib.rs
jkugelman: Agree, it deserves a comment. `usize` and `AtomicUsize` are carryovers from the old code. I had actually made them `u64` but then reverted when I looked into atomics more and saw that `AtomicU64` isn't supported on all platforms. `AtomicUsize` is the most portable atomic int. I presume that's why the original code used it.

jkugelman: Do you not like the type aliases? I was actually considering going one step further and making them newtypes, and/or using `Duration`s from a time crate. I don't like how the values lack units.

kernel/sleep/src/lib.rs
jkugelman: Yep, that's the intention.

kernel/sleep/src/lib.rs
tsoutsman: I don't think interrupts are disabled in this function. Even if they are, another core could mutate `NEXT_WAKE_TIME` between the two operations.

Enjoy your vacation :)

kernel/sleep/src/lib.rs
tsoutsman: I guess it's a matter of style. The issue with using concrete units (i.e. `Duration`) is that the LAPIC timer's frequency is dependent on the CPU's frequency, so the tick interval varies from machine to machine.

kevinaboos:
Apologies for the delay in responding. This looks interesting but it's a bit tough to review, as it seems to include many unrelated changes (perhaps those overlapping with #543?). 

My first recommendation would be to create a separate crate that provides asynchronous sleeping, as the current `sleep` crate is (1) poorly thought-out and (2) somewhat closely tied to the rate monotonic scheduler implementation that is partially complete.

I think we also may want to close this in favor of waiting on @tsoutsman's current work on a more generalized timekeeping abstraction for Theseus (?).

jkugelman:
> Apologies for the delay in responding. This looks interesting but it's a bit tough to review, as it seems to include many unrelated changes (perhaps those overlapping with https://github.com/theseus-os/Theseus/pull/543?).

Yes, it is based off of #543.

> I think we also may want to close this in favor of waiting on @tsoutsman's current work on a more generalized timekeeping abstraction for Theseus (?).

Sounds good. I'll repost this after I rework it.

PR #549, opened 2022-06-30
Use shared string references for section names to save heap memory

kevinaboos:
Previously, each namespace's symbol map had a duplicate copy of every section's full name, which is quite wasteful.

Now we deduplicate them using the `StrRef` type, effectively an `Arc<str>`, which reduces heap usage by about 20% (or more in `cfg(loadable)` mode).

Adds a frontend interface in the `deps` application to query which section a virtual address corresponds to, e.g., similar to `addr2line` symbolication.

Also fixes small bug around the parsing of `.gcc_exept_table` sections.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -562,9 +562,12 @@ dependencies = [
  "fs_node",
  "goblin",
  "hashbrown",
+ "lazy_static",
  "log",
  "memory",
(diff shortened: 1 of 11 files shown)


kevinaboos:
These changes need to be accounted for in #542, but that'll be pretty easy.

PR #553, opened 2022-07-05
Use a build script to clone the 'wasmtime' git repo on demand

kevinaboos:
In the future we should move this into a separate crate instead of having it reside in the `theseus_features` crate. Then `theseus_features` should optionally depend on and thus build that other crate if a "wasmtime"-like feature is enabled.

This is a bit of a hack to get around around two issues:
* cargo not being able to check out a git/registry dependency to a specific location in the project workspace (e.g., `ports/wasmtime`)
* git submodules not being "optional"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,9 @@ target/
 # Build directory
 build/
 
+# Dynamically cloned git repo for wasmtime
+ports/wasmtime/
+
 # Generated documentation, used for GitHub Pages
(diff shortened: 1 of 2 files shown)


kevinaboos:
Closed in favor of #556. `wasmtime` isn't actually that large (~40 MB), so we can simply add it as a regular git submodule; there's no need to bother with this clever yet confusing hack.

PR #559, opened 2022-07-08
Pre-calculate section offsets when serializing `nano_core`

### LINKED ISSUE #546: Improve nano_core (de)serialization to use pre-calculated section offsets
After #542 lands, another optimization we could do in the `serialize_nano_core` tool is to pre-calculate the `mapped_pages_offset` value for each `LoadedSection`. 
This would simplify and accelerate the procedure in `SerializedSection::load()` when calculating offsets.

The rules to do so are simple:
* `text_pages` start at `.init + KERNEL_OFFSET`
* `rodata_pages` start at `.rodata`
* `data_pages` start at `.data`
* Pre-use `usize::MAX` for all TLS BSS sections too

The "ordering" rules above are defined by the linker script, but also when remapping the kernel sections during memory/paging initialization: https://github.com/theseus-os/Theseus/blob/09ccbfc644da93e98ef11c9532ef926f1ee7b4f4/kernel/memory_x86_64/src/lib.rs#L158-L181

tsoutsman:
`serialize_nano_core` now parses the section header addresses and uses them
to pre-calculate the offsets of `.symtab` entries.

Closes #546.

### DIFF
--- a/kernel/mod_mgmt/src/parse_nano_core.rs
+++ b/kernel/mod_mgmt/src/parse_nano_core.rs
@@ -193,10 +193,10 @@ fn parse_nano_core_symbol_file_or_binary(
     new_crate_mut.global_sections = parsed_crate_items.global_sections;
     new_crate_mut.data_sections   = parsed_crate_items.data_sections;
     
-    // Dump loaded sections for verification. See pull request #559 for more details:
-    for (_, section) in new_crate_mut.sections.iter() {
-        trace!("{:016x} {} {}", section.address_range.start.value(), section.name, section.mapped_pages_offset);
-    }
+    // // Dump loaded sections for verification. See pull request #559 for more details:
+    // for (_, section) in new_crate_mut.sections.iter() {
+    //     trace!("{:016x} {} {}", section.address_range.start.value(), section.name, section.mapped_pages_offset);
+    // }
 
     drop(new_crate_mut);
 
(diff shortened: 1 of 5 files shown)


tsoutsman:
I'm yet to test it, but I think dumping out the sections in `SerializedCrate::into_loaded_crate()` before and after the change should be pretty straightforward. I don't think there's an easy way to compare it to the "source of truth" (i.e. `readelf`) because, as far as I know, it doesn't have an option to display offsets.

kevinaboos:
> I don't think there's an easy way to compare it to the "source of truth" (i.e. `readelf`) because, as far as I know, it doesn't have an option to display offsets.

That's correct, since offsets are a Theseus-specific concept related to `MappedPages`.

tsoutsman:
@kevinaboos this should be good now.

```sh
make run > debug_new
# DO: when theseus boots (i.e. command line displayed), ctrl+c
# DO: manually remove all the lines around the printed-out types in debug_new
# the following command strips terminal escape codes and extra spaces
cut -c 45- < debug_new | rev | cut -c 5- | rev > debug_new_diff
sort -o debug_new_diff debug_new_diff

# DO: Switch to `.sym` nano_core parsing in the Makefile

make run > debug_old
# DO: when theseus boots (i.e. command line displayed), ctrl+c
# DO: manually remove all the lines around the printed-out types in debug_old
# the following command strips terminal escape codes and extra spaces
cut -c 55- < debug_old | rev | cut -c 5- | rev > debug_old_diff
sort -o debug_old_diff debug_old_diff

diff debug_old_diff debug_new_diff # should be empty
```
This shows that the change keeps all the offsets the same.

Prior to merging, the debug statements in `kernel/mod_mgmt/src/serde.rs` and `kernel/mod_mgmt/src/parse_nano_core.rs` need to be removed.

kevinaboos (approved):
looks good after a few small changes from my end. Thanks!

PR #561, opened 2022-07-13
Fix most clippy::perf items in root workspace

nramos0:
Addresses #526 

The below items suggested by Clippy from clippy::perf were changed in the root workspace:

redundant-clone
or-fun-call
to-string-in-format-args
cmp-owned
single-char-pattern
vec-init-then-push
expect-fun-call

I noticed that calling Clippy from the root workspace will not reach some other crates in the codebase (like `./kernel/wasi_interpreter`, this doesn't seem to be excluded in the root Cargo.toml so I thought it should be included like most of the other kernel crates) but I wasn't able to immediately understand how some were not being included in the workspace compilation, so I didn't go on to call Clippy in other individual workspaces and just made the root workspace changes for now.

### DIFF
--- a/applications/less/src/lib.rs
+++ b/applications/less/src/lib.rs
@@ -68,14 +68,14 @@ fn get_content_string(file_path: String) -> Result<String, String> {
                         Ok(num) => num,
                         Err(e) => {
                             return Err(format!("Failed to read {:?}, error {:?}",
-                                               file_locked.get_name(), e).to_string())
+                                               file_locked.get_name(), e))
                         }
                     };
                     let read_string = match str::from_utf8(&string_slice_as_bytes) {
                         Ok(string_slice) => string_slice,
                         Err(utf8_err) => {
                             return Err(format!("File {:?} was not a printable UTF-8 text file: {}",
-                                               file_locked.get_name(), utf8_err).to_string())
+                                               file_locked.get_name(), utf8_err))
                         }
                     };
                     Ok(read_string.to_string())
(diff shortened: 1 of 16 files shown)


kevinaboos:
Thanks @nramos0 for the PR!

To build all crates in the workspace, you need to explicitly pass `--workspace` to a cargo command. This minimal command works for us:
```sh
cargo clippy -Z build-std=core,alloc -Z build-std-features=compiler-builtins-mem --target cfg/x86_64-theseus.json --workspace
```

Thanks for bringing that up, I will add it to the tracking issue.

kevinaboos:
There's also a known issue with clippy's CLI related to enabling/disabling sets of warnings, I filed it here https://github.com/rust-lang/rust-clippy/issues/9012

kevinaboos (changes requested):
Everything looks good to me!

I'll wait to merge until you're able to include the rest of the workspace crates, hopefully the command i posted works for you. That way we can mark those lints as fully complete in the tracking issue.

nramos0:
> Thanks @nramos0 for the PR!
> 
> To build all crates in the workspace, you need to explicitly pass `--workspace` to a cargo command. This minimal command works for us:
> 
> ```shell
> cargo clippy -Z build-std=core,alloc -Z build-std-features=compiler-builtins-mem --target cfg/x86_64-theseus.json --workspace
> ```
> 
> Thanks for bringing that up, I will add it to the tracking issue.

Good to know, thanks for clearing that up. Running the command
```
cargo clippy -Z build-std=core,alloc -Z build-std-features=compiler-builtins-mem --target cfg/x86_64-theseus.json --workspace --all-features -- -A clippy::all -W clippy::redundant-clone -W clippy::or-fun-call -W clippy::to-string-in-format-args -W clippy::cmp-owned -W clippy::vec-init-then-push -W clippy::single-char-pattern -W clippy::expect-fun-call
``` 
now yielded additional changes that I included in a new commit!

kevinaboos (approved):
Thanks, tested and working!

PR #563, opened 2022-07-14
Fix lint subset of clippy::complexity

nramos0:
Addresses #526.

Fixes all Clippy suggestions yielded from the command
```
cargo clippy -Z build-std=core,alloc -Z build-std-features=compiler-builtins-mem --target cfg/x86_64-theseus.json --workspace --all-features -- -A clippy::all -W clippy::while-let-loop -W clippy::bool-comparison -W clippy::manual-unwrap-or -W clippy::useless-asref -W clippy::filter_next -W clippy::nonminimal-bool -W clippy::bind-instead-of-map
```

including the lints from clippy::complexity below:
while-let-loop
bool-comparison
manual-unwrap-or
useless-asref
filter_next
nonminimal-bool
bind-instead-of-map

### DIFF
--- a/applications/shell/src/lib.rs
+++ b/applications/shell/src/lib.rs
@@ -889,10 +889,7 @@ impl Shell {
             if slash_ending {
                 ""
             } else {
-                match nodes.pop() {
-                    Some(node) => node,
-                    None => ""
-                }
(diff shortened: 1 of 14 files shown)


kevinaboos (approved):
Thanks, tested and working!

PR #564, opened 2022-07-14
small fixes in book

krystian-wojtas:


kevinaboos (changes requested):
Hi @krystian-wojtas, thanks so much for these fixes!

One small request -- can you specify the language for t]e new code blocks you added to `book/src/running/pxe.md`? I think `sh` is what we want there, e.g.,
```
```sh
sudo ...
```

Everything else looks good to me.

krystian-wojtas:
Hi Kevin,

Of course, changes applied. Thanks to take a look and review these few small fixes.

I'm really inspired by concepts you have and described in book and this new operating system Theseus :)

kevinaboos (approved):
Thanks for the fix and for the kind words :)

PR #570, opened 2022-07-20
Add environment variables

tsoutsman:
This is needed for an eventual `std` port.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -891,6 +891,7 @@ name = "environment"
 version = "0.1.0"
 dependencies = [
  "fs_node",
+ "hashbrown",
  "root",
 ]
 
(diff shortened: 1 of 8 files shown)


kevinaboos (approved):
Thanks Klim, looks good to me! I made a few small changes, feel free to submit another PR or lmk if any of those go against what you wanted.

Also, a [recent change to Theseus](https://github.com/theseus-os/Theseus/commit/f0d212ce811e8bf85d8083b757adf0cf691d3cfc) causes it to build only a subset of all crates by default. When making a new PR, we need to ensure all crates build. I've [recently added a make target](https://github.com/theseus-os/Theseus/commit/0e61ea815cbf2dc4857b8e484a5f39587e6bc25a) for that: `make full`. For example, in this PR, the `ports/theseus_std` crate wasn't included in the standard `make` build target, so some changes were missed there. We should add a CI build step to double check this.

PR #574, opened 2022-07-22
Add Github Action to run Clippy on marked lints in #526

nramos0:
Addresses #526.
Adds an action that runs on when PRs are opened, reopened, edited, or synchronized to run Clippy for all lints checked off in the tracking issue #526. It uses the Github CLI to fetch the issue body and then does some parsing to turn it into arguments for a clippy command.


nramos0:
It appears this PR itself is making use of the action and there are currently some places where the lint fails, I'll try to fix them

kevinaboos (changes requested):
@nramos0 great work with this PR, and thanks so much for tackling it!

The solution to parse issue 526 is quite clever, and normally I'd be in favor of that. However, that issue's text contents can be modified by the original author and any privileged maintainers, which could break your new clippy action.

Thus, I think it's probably best to hard-code the specific lints that we currently support, and then simply modify that list whenever more clippy lints are addressed. That would ensure no changes to 526 could break this action.

nramos0:
> @nramos0 great work with this PR, and thanks so much for tackling it!
> 
> The solution to parse issue 526 is quite clever, and normally I'd be in favor of that. However, that issue's text contents can be modified by the original author and any privileged maintainers, which could break your new clippy action.
> 
> Thus, I think it's probably best to hard-code the specific lints that we currently support, and then simply modify that list whenever more clippy lints are addressed. That would ensure no changes to 526 could break this action.

Alright, so I added a file at the root of the project called `lints.txt` that contains the clippy lints to check, and then the action just reads from there instead. Does that work better? The action searches for lines that don't contain `#` (to allow for "commented out" lints that can be easily added back in by removing the #, as well as titles for sections of lints) and feeds those into clippy

kevinaboos (approved):
Thanks @nramos0, the new changes look great. Much appreciated!

PR #575, opened 2022-07-23
Restructure `libtheseus` and rework sync primitives

tsoutsman:
EDIT: See latest comment


tsoutsman:
Converting to draft as per issues sent in Discord.

tsoutsman:
This PR restructures `libtheseus` to match its intended purpose as an interface between the kernel and `std`. It also reworks Theseus`s synchronisation primitives.

Sync primitives are now separated into two crates: `kernel/raw_sync` and `kernel/sync`. The primitives in `raw_sync` have a similar interface to the one that we have to implement in `std::sys` - e.g. the mutex does not have guards or store data. `sync` on the other hand, is essentially a port of `std::sync` sans a few points:
- `sync` only has mutexes, reader-writer locks, and condition variables
- `sync` mutexes and reader-writer locks do not track poisoning
- `raw_sync` (and `sync`) primitives have non-const initialisers. This is because internally they use a `VecDeque` which doesn't have a const initialiser. I believe the sync primitives implementations in `std::sys` require a const initialiser so we'll have to figure something out.
- `raw_sync` condition variables don't implement `wait_timeout`, and by extension `sync` condition variables don't implement `wait_timeout` or `wait_timeout_while`. This relies on a sleep implementation and `Instant` so we'll need to wait for #569. Although this is needed for the `std` port, it can be added in a later PR.

Something to note is that the implementations in `raw_sync` are untested, although they are based on papers/lectures which I've included links to. The source files aren't very long ~100 lines of code, but I've already fixed a couple of bugs after I thought I was done. Race conditions are very hard to spot and even harder to test.

This PR also makes changes to `ports/backtrace` and `ports/wasmtime` but I'm not sure how to include them in the PR. As you can see, right now the submodules point to invalid references.

I found a race condition in the way `mutex_sleep` was interacting with `wait_queue`. Although it's now fixed (as the mutex implementation doesn't rely on `wait_queue` anymore), I have a feeling other users of `wait_queue` may have a similar issue. However, the scope of this PR is already big enough and so I'll probably address this in a separate PR (if these issues exist).

Also `make libtheseus` will need to be changed, but I'm not great with Make so I haven't done that. Currently `libtheseus` builds (i.e. `cargo build --manifest-path libtheseus/Cargo.toml`) fail with the error:
```
error[E0277]: the trait bound `Volatile<u32>: FromBytes` is not satisfied
   --> /home/klim/Projects/Theseus/kernel/apic/src/lib.rs:199:10
    |
199 | #[derive(FromBytes)]
    |          ^^^^^^^^^ the trait `FromBytes` is not implemented for `Volatile<u32>`
    |
    = help: the following other types implement trait `FromBytes`:
              ()
              ApicRegisters
              I128<O>
              I16<O>
              I32<O>
              I64<O>
              PhantomData<T>
              PhysicalAddress
            and 24 others
note: required by a bound in `<ApicRegisters as FromBytes>::only_derive_is_allowed_to_implement_this_trait::ImplementsFromBytes`
   --> /home/klim/Projects/Theseus/kernel/apic/src/lib.rs:199:10
    |
199 | #[derive(FromBytes)]
    |          ^^^^^^^^^ required by this bound in `<ApicRegisters as FromBytes>::only_derive_is_allowed_to_implement_this_trait::ImplementsFromBytes`
    = note: this error originates in the derive macro `FromBytes` (in Nightly builds, run with -Z macro-backtrace for more info)
```
I think this has something to do with the lock file and incompatible versions.

kevinaboos:
This is an awesome set of changes! Can we separate out the various components into several smaller PRs so that it's easier to review? I'd vote one PR for each major sync primitive, and at the very least, definitely a separate PR for the libtheseus changes.

Issue #578, opened 2022-07-29
Use preemption-safe locks instead of interrupt-safe locks wherever possible


tsoutsman:
> Theseus uses the `MutexIrqSafe` type to ensure that interrupts are disabled when accessing the inner data. This is quite useful for sensitive contexts where deadlock could otherwise occur if another execution context is allowed to take over while a regular spinlock `Mutex` might be held. For example, usage of `MutexIrqSafe` is required whenever sharing data between a regular execution context and an interrupt handler context. 
> 
> However, we've used it far too ubiquitously in Theseus, primarily due to our initial goal to reduce unnecessary OS states; this did make sense in the pursuit of that goal, because whether or not preemption is currently enabled is another bit of per-core state that must be atomically accessible by all tasks on a given CPU core. 
> 
> In many cases it simply suffices to disable preemption (task switching) so we don't context switch to another task while a sensitive lock is held. Such contexts are most commonly found in the task management code itself, i.e., in the `task` and `spawn` crates' functions. This would improve general I/O performance and interactivity of the system by allowing interrupts to be enabled during more regular execution contexts that don't actually have anything to do with interrupt handling.

kevinaboos:
This is now in progress since #595 is merged. I will first work on using preemption guards in scheduling and task management, since that's the clearest avenue for efficiency wins.

kevinaboos:
As I mused earlier, another good idea may be to introduce a more fine-grained interrupt-safe lock crate that disables only certain interrupts by masking those IRQs.

For example, if a task is accessing state that's shared with only the e1000 NIC driver's interrupt handler, then the lock surrounding that shared state can simply mask the e1000 IRQ instead of disabling _all_ interrupts.

However, we'd have to create a lint or some form of code analysis to check this, otherwise we'd be walking head-first into a wall of potential deadlocks.

kevinaboos:
#600 paves the way for properly handling (enabling/disabling) preemption and interrupts during task switching.

kevinaboos:
#603 implements this for `schedule()` and `task_switch()`

kevinaboos:
#616 implements this for task lifecycle functions that cleanup task states after a task has exited.

kevinaboos:
I think the only remaining obvious area for this is runqueue management, which I can tackle soon.

kevinaboos:
With #629 merged in, I think this issue can be closed. 

Looking through the code base, the only other areas where interrupt-safe locks could be removed are:
[...]

PR #579, opened 2022-07-29
Implement preemption disabling

tsoutsman:
This commit is the first step towards closing #578.

The implementation uses a counter over a boolean to keep track of
recursive preemption disabling.

The test application isn't great, but I can't think of a more ergonomic
or thorough way of testing the change.

As a next step (and after #576 lands), we can create a new mutex similar
to `spin::Mutex` and `irq_safety::MutexIrqSafe` that disables preemption
when locked.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3357,6 +3357,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "test_preemption_disable"
+version = "0.1.0"
+dependencies = [
+ "scheduler",
+ "sleep",
+ "terminal_print",
+]
+
+[[package]]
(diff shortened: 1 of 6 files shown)


tsoutsman:
`enable_preemption` and `disable_preemption` unwrap the output of `get_my_apic`. I'm not sure if that's ok, but I didn't want them to return a result.

kevinaboos:
Thanks for the PR. There are a few issues here regarding safety and correctness. I think I'm going to take a stab at fixing them and then we can discuss the changes on discord.

kevinaboos:
Closed in favor of #595

PR #580, opened 2022-07-30
Mark `LoadedSection::as_func()` as unsafe

tsoutsman:
`LoadedSection::as_func()` is very unsafe. I didn't mark `swap_crates` as unsafe because it's used in a number of different contexts where the user provides a symbol to a function, but ideally, it would be marked unsafe as well.

### DIFF
--- a/kernel/crate_metadata/src/lib.rs
+++ b/kernel/crate_metadata/src/lib.rs
@@ -40,6 +40,7 @@
 //! [`B.sections_dependent_on_me`]: LoadedSectionInner::sections_dependent_on_me
 //! 
 
+#![deny(unsafe_op_in_unsafe_fn)]
 #![no_std]
 
 extern crate alloc;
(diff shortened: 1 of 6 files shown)


kevinaboos (changes requested):
Ironic, I had this exact thought a few weeks ago, completely randomly. 👍  Totally agree. 

It's safer than it used to be when exposed as a `MappedPages` method, but it still can't be fully safe until we have more RTTI, which we can perhaps obtain from newer mangling schemas.

This PR is missing quite a few places where as_func() is invoked. Would you mind also changing them? e.g., `#[cfg(loadable)]` blocks, some simd_personality stuff, etc. If you grep for `as_func(` you'll see them all.

tsoutsman:
Sorry, I thought it was good to go if `make full` built.

Yea, I have a couple of ideas to fix the _in god we trust_ the signature matches approach.

kevinaboos (approved):
Thanks, looks good now!

PR #581, opened 2022-07-30
Add `Environment::chdir()`

tsoutsman:
I'm not sure about the error handling. Currently, the usual Theseus
result type is `Result<T, &'static str>'`. However, this limits what
the function caller can do with an error. Pattern matching on strings
is ... less than ideal. So instead, `chdir()` returns an [`Error`] enum,
as is common in Rust. However, both variants are file system errors
rather than environment errors and so ideally, we would define this type
in `fs_node`. But, `fs_node`'s API doesn't expose functions that would
use these error variants, so ...

For reference, `std::env::set_current_dir()` returns an
`std::io::Result`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -412,6 +412,7 @@ checksum = "79c2681d6594606957bbb8631c4b90a7fcaaa72cdb714743a437b156d6a7eedd"
 name = "cd"
 version = "0.1.0"
 dependencies = [
+ "environment",
  "fs_node",
  "getopts",
  "log",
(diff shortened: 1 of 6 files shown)


kevinaboos (approved):
Looks good! 

The error handling thing is fine for now. I fully intend to go through all Theseus crates and use proper error types/enums, but I'm waiting for the `Error` trait to be moved from `std` into `core`, which Jane Lusby is working on. I've discussed this at length with her and I think it's not worth the massive effort of migrating from error strings to error types until that trait is available in `no_std` environments.

PR #582, opened 2022-07-31
Add `random` crate

tsoutsman:
The crate mimics the `RngCore` interface. It was named `random` to not
conflict with `rand`'s symbols.

I've also replaced locally defined RNGs in the `ixgbe` and
`ota_update_client` crates, that used the ... HPET timer count as
a seed.

A further development could be to expose an `add_entropy` function.
I'm not sure it's a major concren because a CSPRNG seeded with 32 bytes
of entropy should be plenty secure. However, if RDSEED or RDRAND don't
exist, it is absolutely necessary. Adding entropy would require locking
the CSPRNG, generating a random seed, combining it with the entropy, and
reseeding the CSPRNG. This would somehow need to be arbitrated to ensure
the the CSPRNG isn't being constantly reseeded, impacting performance.

Our QEMU configuration currently doesn't support RDSEED, so the backup
RDRAND is used.

### DIFF
--- a/kernel/rand/src/lib.rs
+++ b/kernel/rand/src/lib.rs
@@ -1,0 +1,137 @@
+#![no_std]
+
+use rand_chacha::{
+    rand_core::{RngCore, SeedableRng},
+    ChaCha20Rng,
+};
+use spin::{mutex::Mutex, once::Once};
+
+/// The global random number generator.
+///
+/// This PRNG is cryptographically-secure. It can be directly accessed using
+/// [`next_u32`], [`next_u64`], and [`fill_bytes`] for one-off randomness.
+/// However, it should mostly be used to seed local PRNGs using [`init_prng`].
+///
+/// Using a single global CSPRNG allows us to feed it with entropy from device
+/// drivers and such.
+static CSPRNG: Once<Mutex<ChaCha20Rng>> = Once::new();
+
+/// Initialises the global CSPRNG.
+///
+/// Only the first call will initialise the CSPRNG. Subsequent calls will
+/// return the already initialised CSPRNG.
+pub fn init_once() -> &'static Mutex<ChaCha20Rng> {
+    CSPRNG.call_once(|| {
+        let seed = rdseed_seed()
+            .or_else(|_| rdrand_seed())
+            .unwrap_or_else(|_| tsc_seed());
+        Mutex::new(ChaCha20Rng::from_seed(seed))
+    })
+}
+
+/// Tries to generate a 32 byte seed using the RDSEED x86 instruction.
+fn rdseed_seed() -> Result<[u8; 32], ()> {
+    match rdrand::RdSeed::new() {
+        Ok(mut generator) => {
+            let mut seed = [0; 32];
+            match generator.try_fill_bytes(&mut seed) {
+                Ok(_) => {
+                    log::info!("using RDSEED for CSPRNG seed");
+                    Ok(seed)
+                }
+                Err(_) => {
+                    log::warn!("failed to generate seed from RDSEED");
+                    Err(())
+                }
+            }
+        }
+        Err(_) => {
+            log::warn!("failed to initialise RDSEED");
+            Err(())
+        }
+    }
+}
+
+/// Tries to generate a 32 byte seed using the RDRAND x86 instruction.
+fn rdrand_seed() -> Result<[u8; 32], ()> {
+    match rdrand::RdRand::new() {
+        Ok(mut generator) => {
+            let mut seed = [0; 32];
+            match generator.try_fill_bytes(&mut seed) {
+                Ok(_) => {
+                    log::info!("using RDRAND for CSPRNG seed");
+                    Ok(seed)
+                }
+                Err(_) => {
+                    log::warn!("failed to generate seed from RDRAND");
+                    Err(())
+                }
+            }
+        }
+        Err(_) => {
+            log::warn!("failed to initialise RDRAND");
+            Err(())
+        }
+    }
+}
+
+/// Generates a 32 byte seed using the TSC.
+///
+/// The TSC isn't a high quality source of randomness and so it is only used as
+/// a last resort if both RDSEED and RDRAND fail.
+fn tsc_seed() -> [u8; 32] {
+    let mut seed = [0; 32];
+
+    for i in 0..seed.len() {
+        // The last byte is the _most_ random.
+        seed[i] = u128::from(tsc::tsc_ticks()).to_be_bytes()[15];
+    }
+
+    // The TSC isn't a high quality source of randomness.
+    log::error!("using TSC for CSPRNG seed - this is not ok");
+    seed
+}
+
+/// Returns a random [`u32`].
+///
+/// Consider using [`init_rng`] if calling this function in a loop, or if you
+/// don't require cryptographically-secure random numbers.
+pub fn next_u32() -> u32 {
+    let mut csprng = init_once().lock();
+    csprng.next_u32()
+}
+
+/// Returns a random [`u64`].
+///
+/// Consider using [`init_rng`] if calling this function in a loop, or if you
+/// don't require cryptographically-secure random numbers.
+pub fn next_u64() -> u64 {
+    let mut csprng = init_once().lock();
+    csprng.next_u64()
+}
+
+/// Fills `dest` with random data.
+///
+/// Consider using [`init_rng`] if calling this function in a loop, or if you
+/// don't require cryptographically-secure random numbers.
+pub fn fill_bytes(dest: &mut [u8]) {
+    let mut csprng = init_once().lock();
+    csprng.fill_bytes(dest);
+}
+
+/// Initialises a `T` RNG.
+///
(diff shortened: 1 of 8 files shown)


NathanRoyer:
This looks great!
Have you thought about implementing the RngCore trait?
Would that be a good thing?
https://docs.rs/rand_core/latest/rand_core/trait.RngCore.html

tsoutsman:
We only expose one source of randomness, so implementing the trait seems unnecessary. The functions would have to be converted to methods on a ZST, and the caller would have to import the `RngCore` trait. It also means we expose `rand` in the public API, which is notoriously unstable.

kernel/ota_update_client/src/lib.rs
```diff
@@ -324,11 +319,11 @@ fn download_files<S: AsRef<str>>(
     }
 
     let startup_time = hpet_ticks!();
-    let mut rng = SmallRng::seed_from_u64(startup_time);
     let rng_upper_bound = u16::max_value() - STARTING_FREE_PORT;
 
     // the below items may be overwritten on each loop iteration, if the socket was closed and we need to create a new one
-    let mut local_port = STARTING_FREE_PORT + (rng.next_u32() as u16 % rng_upper_bound);
+    // TODO: This doesn't provide a uniform distribution of port numbers.
```
kevinaboos: this TODO is interesting, can you elaborate? Are you saying that `random::next_u32()` doesn't generate truly random numbers?

kernel/random/src/lib.rs
```diff
@@ -0,0 +1,88 @@
+//! A crate for generating random numbers.
+
+#![no_std]
+
+use rand_chacha::{
```
kevinaboos, line 25: do we need to use `rand_chacha`? I guess the real question is whether we need to provide a cryptographically-secure RNG as the default randomness provider in Theseus.... I don't think that we do.

Previously we were using `rand_core` via `rand` (with no features enabled); I think that's the way to go, and IIRC, is what `rand` recommends doing. We could also use other crates like `fastrand`, but considering that `wasmtime` uses `rand` directly, we might as well just stick with that.

We can also expose a crypto-secure RNG from this crate, or a separate crate, but I feel like that shouldn't be the only/default choice.

kernel/random/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+// TODO: Ideally we'd use core::cell::LazyCell (or core::lazy::Lazy, or whatever
+// they choose to call it), but for some reason it doesn't yet allow mutable
+// access despite being !Sync.
+lazy_static::lazy_static! {
+    static ref CSPRNG: Mutex<ChaCha20Rng> = {
+        let seed = rdseed_seed()
+            .or_else(|_| rdrand_seed())
+            .unwrap_or_else(|_| {
+                log::error!("Could not generate hardware seed - using a predetermined seed. THIS IS NOT OK.");
+                [1; 32]
```
kevinaboos: this is okay, but it'd probably be better to explicitly initialize our RNG with a backup seed value obtained from something at least partially random, perhaps a current timer value (PIT/RTC/TSC/HPET) or something else.

For this, you can use the `spin::Once` type or OnceCell or something similar that allows you to pass in values to the init routine, instead of `lazy_static`. Might as well do that since you're initializing it from the `captain` anyway.

kevinaboos (changes requested):
Looks generally pretty good to me, I think it'll be an excellent addition to Theseus with a few suggested changes.

FWIW, I do think it would be beneficial to have tighter integration with the `rand` crate, perhaps by implementing one or more of its traits. We can do that in a future PR, but it seems like a quick & easy addition that would help to standardize `random`;s public interface.

kernel/ota_update_client/src/lib.rs
tsoutsman: Take a particularly egregious example:
```rust
/// Generates a random number between 0-20.
fn gen_random() -> u8 {
    // generate
}

let x = gen_random();
let y = x % 15;
```
`y` is as likely to be in the range 0-4 as it is to be in the range 5-14.

Looking over the code again, removing the cast to `u16` would fix this as the difference in probability would be incredibly small.

kernel/random/src/lib.rs
tsoutsman: Yea that's probably best. We'd need to ensure that `rand` is only initialised after the timers then.

kernel/random/src/lib.rs
tsoutsman: My logic behind this was the user can seed whatever RNG they'd like with this global CSPRNG. Although I can see the benefit of having a global non-cryptographically secure PRNG so that crates don't have to depend on their own implementation (e.g. `fastrand`).

kernel/ota_update_client/src/lib.rs
kevinaboos: ok yeah, so to fix it we just want to truncate the `rng.next_u32()` value after we modulo it with the upper bound, not before.

kevinaboos: we'll do that in the future PR that changes other crates to use separate PRNGs based on our new singleton CSPRNG

kevinaboos:
So, to summarize the key points of our discussion:
* Document why we have a single CSPRNG, and why we're using a CS one instead of non-CS
* Describe how other crates would want to use this (use the CSPRNG to seed a non-CS PRNG)
* Undo the changes to other crates that were using various random sources (we can address than in a new PR)
* Add explicit initialization, perhaps from the captain, but another place could work too (device_manager?)

As part of a future PR we could:
* Implement `rand::RngCore` for our CSPRNG such that other crates could use it to seed another instance of anything that implements `rand::SeedableRng`
* Change crates across Theseus to use this new `random` crate, perhaps adding convenience interfaces to it that create new non-CS PRNG instances on demand for each caller

tsoutsman:
I've changed the PR to simply add a new unused `rand` interface addressing the points of our discussion. I've added explicit initialization to `captain` but I'm not sure if we need it because the CSPRNG would just be initialised the first time we access it.

Running `tsc_seed` on my computer yields the following seed:
```rust
[
    1,
    110,
    179,
    224,
    12,
    55,
    96,
    138,
    175,
    208,
    241,
    18,
    51,
    84,
    117,
    148,
    179,
    212,
    243,
    18,
    49,
    80,
    111,
    142,
    175,
    206,
    237,
    29,
    60,
    93,
    126,
    157,
]
```
The bytes are very clearly increasing in a loop. I tried using just the last bit (calling `tsc_ticks()` 256 times) but this resulted in (I think) an even worse seed:
```rust
[
    177,
    169,
    181,
    85,
    170,
    170,
    85,
    90,
    170,
    85,
    74,
    170,
    85,
    170,
    165,
    85,
    170,
    85,
    74,
    170,
    85,
    42,
    169,
    85,
    42,
    173,
    85,
    170,
    213,
    82,
    170,
    85,
]
```

kernel/rand/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+    rand_core::{RngCore, SeedableRng},
+    ChaCha20Rng,
+};
+use spin::{mutex::Mutex, once::Once};
+
+/// The global random number generator.
+///
+/// This PRNG is cryptographically-secure. It can be directly accessed using
+/// [`next_u32`], [`next_u64`], and [`fill_bytes`] for one-off randomness.
+/// However, it should mostly be used to seed local PRNGs using [`init_prng`].
```
kevinaboos: i don't see a `init_prng` function, did you mean `init_rng`?

kernel/rand/src/lib.rs
```diff
@@ -0,0 +1,137 @@
+#![no_std]
```
kevinaboos: Kindly add some crate-level docs that describe how this crate offers a CS PRNG, why we do that,  why it's a singleton, and what it should be used for / how to use it.
It'd be nice to add a small blurb about how lazy init works, and how we try to initialize it with three different seed generators, etc.
Also mention that it's based on `rand_chacha`'s `ChaCha20Rng` implementation (or really is just a thin wrapper around it).

The docs on the `static CSPRNG` object are good, but nobody can see them in any generated documentation, hence the request for crate-level docs.

kevinaboos (changes requested):
So now that you've made an infallible init function, we can simply use `lazy_static` as before. I didn't know you were going to go that route, hence why i suggested `Once` for an explicit init function that could fail. No need to call it from the captain either.

Also, kindly rename it back to `random` since `rand` is already a widely-known crate in the Rust ecosystem.

tsoutsman:
> infallible init

That's because `tsc` is incorrectly infallible; it doesn't check whether the counter exists. But that's a separate issue.

kevinaboos:
> That's because `tsc` is incorrectly infallible; it doesn't check whether the counter exists. But that's a separate issue.

Hmm, i see. While that technically might be a concern for ancient CPUs, I doubt it's of _practical_ concern to us since we don't even support anything prior to 64-bit x86. AFAICT TSC has existed since Pentium was first introduced, so I think we'd be okay.

kevinaboos (approved):
Excellent job with the docs, looks all good to me now.

PR #586, opened 2022-08-05
Separate cargo flags from cargo features

kevinaboos:
This commit introduces a new `Makefile` variable called `FEATURES`,
which is passed directly into Theseus's `cargo build` invocation.
It has a default value of `--workspace`, just as `CARGOFLAGS` used to.

This is necessary because `CARGOFLAGS` is reused for correctly
configuring out-of-tree builds against a prebuilt instance of Theseus,
so we cannot overload it with cargo features.

For more info, see the crate-level documentation for `theseus_features`.

At this point, `CARGOFLAGS` is only really used to specify `--release`,
while `FEATURES` is used for controlling which features/crates
are included in a build of Theseus.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,7 @@ export override RUSTFLAGS += $(patsubst %,--cfg %, $(THESEUS_CONFIG))
 
 ### Convenience targets for building the entire Theseus workspace
 ### with all optional features enabled. 
-### See `theseus_cargo/Cargo.toml` for more details about what this includes.
+### See `theseus_features/src/lib.rs` for more details on what this includes.
 all: full
 full : export override FEATURES += --workspace --all-features
 full: iso
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ $(error Error: could not find 'grub-mkrescue' or 'grub2-mkrescue', please instal
 else ifeq ($(bootloader),limine)
 	## Check if the limine directory exists. 
 	ifneq (,$(wildcard $(LIMINE_DIR)/.))
-		export override CARGOFLAGS += --features extract_boot_modules
+		export override FEATURES += --features extract_boot_modules
 	else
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -166,9 +166,11 @@ endif
 export override RUSTFLAGS += $(patsubst %,--cfg %, $(THESEUS_CONFIG))
 
 
-### Convenience targets for building Theseus with all features enabled.
+### Convenience targets for building the entire Theseus workspace
+### with all optional features enabled. 
+### See `theseus_cargo/Cargo.toml` for more details about what this includes.
```
tsoutsman:
```suggestion
### See `theseus_features/src/lib.rs` for more details on what this includes.
```

Makefile
```diff
@@ -166,9 +166,11 @@ endif
...
 all: full
-full : export override CARGOFLAGS += --all-features
+full : export override FEATURES += --workspace --all-features
```
tsoutsman, line 173:
```suggestion
full: export override FEATURES += --all-features
```
`FEATURES` already contains `--workspace`. You can check this by inspecting the Cargo command printed out when running `make full`.

tsoutsman (changes requested):
Just some minor changes.

p.s. I'm probably not the greatest person to review changes to the Makefile.

Makefile
kevinaboos: yep, that's true, and i double-checked this. However, I left it in because if someone sets `FEATURES` on the make command line, then it won't contain `--workspace`, and i think it should it someone is running `make full`.

For ex, if you run `make full FEATURES="--feature hello"` then `--workspace` will not be passed into `cargo build`, so some crates will be excluded (the ones not used as dependencies by any of the optional features.

I have also confirmed that it's okay to pass in `--workspace` multiple times, so it's better to have it appear twice than to not have it appear at all.

PR #593, opened 2022-08-09
Depend on bincode via git repository in kernel/mod_mgmt

NathanRoyer:
This would allow us to depend on theseus via git in libstd


kevinaboos:
Should we merge this into `theseus_main` too? The only other crate that uses our `patch`-ed bincode is wasmtime.

Perhaps we should wait until bincode publishes the latest version to crates.io (?)

PR #594, opened 2022-08-09
[WIP] async components implementation

kavan-mevada:


kevinaboos:
HI @kavanmevada, can you kindly explain what this PR contains? It appears to consist of thousands of lines of code copied from another project (which we cannot accept due to unknown licensing restrictions), with no real integration into the rest of Theseus. Kindly add a bit more detail to the description box above. 

Even better would be submitting a series of smaller PRs that iteratively build on each other, such that things are smaller in scope and easier to understand.

PR #595, opened 2022-08-10
Add initial preemption manager crate

kevinaboos:
The crate is unused so far, but in the scheduler, we would use it like so:
```rust
pub fn schedule() -> bool {
    let preemption_guard = preemption::hold_preemption();
    // If preemption was not previously enabled (before we disabled it above),
    // then we shouldn't perform a task switch here.
    if !preemption_guard.preemption_was_enabled() {
        return false;
    }

    ...
```

We'd need to do a few other things to avoid disabling interrupts during `schedule()` and only disable preemption:
* remove the call to `hold_interrupts()` 
* move the preemption guard from the curr task's stack to the next task's stack, similar to how we use `drop_after_task_switch`
* potentially remove the unconditional call to `enable_interrupts()` in `spawn::task_wrapper_internal()`.

### DIFF
--- a/kernel/apic/Cargo.toml
+++ b/kernel/apic/Cargo.toml
@@ -3,6 +3,7 @@ authors = ["Kevin Boos <kevinaboos@gmail.com>"]
 name = "apic"
 description = "APIC (Advanced Programmable Interrupt Controller) support for Theseus (x86 only), including apic/xapic and x2apic"
 version = "0.1.0"
+edition = "2018"
 
 [dependencies]
 spin = "0.9.0"
--- a/kernel/preemption/src/lib.rs
+++ b/kernel/preemption/src/lib.rs
@@ -1,0 +1,115 @@
+//! Manages preemption on a per-CPU core basis.
+//! 
+//! Supports enabling and disabling preemption for the purpose of 
+//! safe task state management, e.g., through preemption-safe locks.
+
+#![no_std]
+
+extern crate alloc;
+
+use core::sync::atomic::{AtomicU8, Ordering};
+
+/// Theseus uses a `u8` to hold each CPU core's ID (apic ID),
+/// so the maximum number of cores it supports is `u8::MAX` (256).
+const MAX_CPU_CORES: usize = u8::MAX as usize;
+
+const ATOMIC_U8_ZERO: AtomicU8 = AtomicU8::new(0);
+
+/// The per-core preemption count, indexed by a CPU core's APIC ID.
+/// 
+/// If a CPU's count is `0`, preemption is enabled.
+/// If a CPU's count is greater than `0`, preemption is disabled.
+static PREEMPTION_COUNT: [AtomicU8; MAX_CPU_CORES] = [ATOMIC_U8_ZERO; MAX_CPU_CORES];
+
+
+/// Prevents preemption (preemptive task switching) from occurring
+/// until the returned guard object is dropped.
+pub fn hold_preemption() -> PreemptionGuard {
+    let apic_id = apic::get_my_apic_id();
+    let prev_val = PREEMPTION_COUNT[apic_id as usize].fetch_add(1, Ordering::Relaxed);
+    // If the previous counter value was 0, that indicates we are transitioning
+    // from preemption being enabled to disabled on this CPU.
+    let preemption_was_enabled = prev_val == 0;
+    // Create a guard here immediately after incrementing the counter,
+    // in order to guarantee that a failure below will drop it and decrement the counter.
+    let guard = PreemptionGuard {
+        apic_id,
+        preemption_was_enabled,
+    };
+
+    if preemption_was_enabled {
+        // When transitioning from preemption being enabled to disabled,
+        // we must disable the local APIC timer used for preemptive task switching.
+        apic::get_my_apic()
+            .expect("BUG: hold_preemption() couldn't get local APIC")
+            .write()
+            .enable_lvt_timer(false);
+    } else if prev_val == u8::MAX {
+        // Overflow occurred and the counter value wrapped around, which is a bug.
+        panic!("BUG: Overflow occurred in the preemption counter for CPU {}", apic_id);
+    }
+
+    guard
+}
+
+
+/// A guard type that ensures preemption is disabled as long as it is held.
+/// 
+/// Call [`hold_preemption()`] to obtain a `PreemptionGuard`.
+/// 
+/// Preemption *may* be re-enabled when this guard is dropped,
+/// but not necessarily so, because other tasks on this CPU 
+/// may also have acquired a `PreemptionGuard`.
+pub struct PreemptionGuard {
+    /// The APIC ID of the CPU on which preemption was held.
+    /// This is not necessary, it's just used for sanity checks or debugging.
+    apic_id: u8,
+    /// Whether preemption was enabled when this guard was created.
+    preemption_was_enabled: bool,
+}
+
+impl PreemptionGuard {
+    /// Returns `true` if preemption was originally enabled
+    /// when this guard was created and disabled them.
+    pub fn preemption_was_enabled(&self) -> bool {
+        self.preemption_was_enabled
+    }
+}
+
+impl Drop for PreemptionGuard {
+    fn drop(&mut self) {
+        let apic_id = apic::get_my_apic_id();
+        assert!(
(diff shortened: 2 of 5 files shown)


kevinaboos:
@tsoutsman requesting your review since I believe this fixes the potential tricky interleaving/safety issues in #579

kernel/preemption/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+impl Drop for PreemptionGuard {
+    fn drop(&mut self) {
+        let apic_id = apic::get_my_apic_id();
+        assert!(
+            self.apic_id == apic_id,
+            "PreemptionGuard::drop(): BUG: APIC IDs did not match! \
+            This indicates an unexpected task migration across CPUs."
+        );
```
tsoutsman, line 102: To enforce this `PreemptionGuard` should be marked `!Sync`.

kernel/preemption/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+/// so the maximum number of cores it supports is `u8::MAX` (256).
+const MAX_CPU_CORES: usize = u8::MAX as usize;
+
+const ATOMIC_U8_ZERO: AtomicU8 = AtomicU8::new(0);
+
+/// The per-core preemption count, indexed by a CPU core's APIC ID.
+/// 
+/// If a CPU's count is `0`, preemption is enabled.
+/// If a CPU's count is greater than `0`, preemption is disabled.
+static PREEMPTION_COUNT: [AtomicU8; MAX_CPU_CORES] = [ATOMIC_U8_ZERO; MAX_CPU_CORES];
```
tsoutsman, line 23:
```suggestion
static PREEMPTION_COUNT: [u8; MAX_CPU_CORES] = [u8; MAX_CPU_CORES];
```
`PREEMPTION_COUNT[apic_id]` is only accessed by the current thread so it's ok if it's not atomic.

Any reads or writes will have to be marked unsafe, but we won't need to check for overflows.

kernel/preemption/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+}
+
+
+/// Returns `true` if preemption is currently enabled on this CPU.
+/// 
+/// Note that this value can't be used as a lock indicator or property,
+/// as it is just a snapshot offers no guarantee that preemption will be enabled
+/// or disabled on the next attempt to enable/disable it
+/// or acquire a preemption-safe lock.
```
tsoutsman: It actually can be used as a lock indicator if it returns false. Only the current processor can reenable preemption, and since preemption is disabled, only the caller is running on the current processor.

kernel/preemption/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// The APIC ID of the CPU on which preemption was held.
+    /// This is not necessary, it's just used for sanity checks or debugging.
+    apic_id: u8,
+    /// Whether preemption was enabled when this guard was created.
+    preemption_was_enabled: bool,
+}
+
+impl PreemptionGuard {
+    /// Returns `true` if preemption was originally enabled
+    /// when this guard was created and disabled them.
```
tsoutsman:
```suggestion
    /// Returns `true` if preemption was enabled when this guard was created.
    ///
    /// I.e. this will return false if multiple guards exist on the same thread.
```

kernel/apic/Cargo.toml
```diff
@@ -3,6 +3,7 @@ authors = ["Kevin Boos <kevinaboos@gmail.com>"]
 name = "apic"
 description = "APIC (Advanced Programmable Interrupt Controller) support for Theseus (x86 only), including apic/xapic and x2apic"
 version = "0.1.0"
+edition = "2018"
```
tsoutsman, line 6: Is there a reason we don't go straight to the `2021` edition?

kernel/preemption/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+
+/// A guard type that ensures preemption is disabled as long as it is held.
+/// 
+/// Call [`hold_preemption()`] to obtain a `PreemptionGuard`.
+/// 
+/// Preemption *may* be re-enabled when this guard is dropped,
+/// but not necessarily so, because other tasks on this CPU 
+/// may also have acquired a `PreemptionGuard`.
```
tsoutsman: This is _technically_ true because a task can hold the guard and call `scheduler::schedule()`, but in practice, the counter is in case the same task recursively disables preemption.

kernel/preemption/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+pub struct PreemptionGuard {
+    /// The APIC ID of the CPU on which preemption was held.
+    /// This is not necessary, it's just used for sanity checks or debugging.
+    apic_id: u8,
```
tsoutsman: Should we only enable this field (and any modifications of this field) in debug builds?
```suggestion
    /// The APIC ID of the CPU on which preemption was held.
    ///
    /// This is not necessary; it's just used for sanity checks or debugging.
    #[cfg(debug_assertions)
    apic_id: u8,
```

tsoutsman (changes requested):
I've left a few comments.

We should also port `applications/test_preemption_disable` from #579.

kernel/preemption/src/lib.rs
tsoutsman: I now realise I didn't do that in #579.

kevinaboos: yea that's a good idea. Actually we can go even further and say that it shouldn't be marked `Send` either. Technically that isn't truly necessary since it's not protecting task-specific access to an underlying resource (but a PreemptSafeMutexGuard _should_ be `!Send`), but it's probably a good idea nonetheless.

kernel/preemption/src/lib.rs
kevinaboos: I don't think this is true. We could do that, but we'd have to enforce atomicity (e.g., by disabling interrupts) for entire operation of changing the counter, checking the counter, and then disabling/enabling the Local APIC timer. With atomic variables, it's impossible for a different task to observe a stale value of its current CPU's `PREEMPTION_COUNT`, even if a context switch occurs between reading the counter, checking it, and setting it.

kernel/preemption/src/lib.rs
kevinaboos: Yeah, i should change the docs. I actually started using it as such at the top of `schedule()` as an indicator, as seen in the description box of this PR.

kernel/apic/Cargo.toml
kevinaboos: only that i haven't looked into it and don't know what `2021` includes :grimacing:

kernel/preemption/src/lib.rs
kevinaboos: yes, that's a better example. I will mention the case of the same task repeatedly disabling preemption by obtaining multiple preemption guards along the same call stack path.

kernel/preemption/src/lib.rs
kevinaboos: i would, but we basically never build in debug mode because it runs prohibitively slowly. Considering that we're using it for sanity checks now, it's best to just change the comment to reflect that.

kernel/preemption/src/lib.rs
tsoutsman: Wait sorry I mixed up send and sync. It can be sync, because the drop handler is only run by the thread owning the guard.

kernel/preemption/src/lib.rs
kevinaboos: hmm, actually on second thought, this function alone cannot be used as an indicator of anything. I thought this comment was referring to `PreemptionGuard::preemption_was_enabled()`, which can be useful.

kevinaboos: Regardless, I've update the docs

kernel/preemption/src/lib.rs
kevinaboos: Right. done :+1:

kernel/preemption/src/lib.rs
kevinaboos: changed, thanks.

tsoutsman (approved):

PR #602, opened 2022-08-14
Add `local_rust` option to Makefile

tsoutsman:
I haven't worked with make much, so this may be a terrible way to do it. It definitely feels terrible.

This way of doing it unnecessarily forces us to rebuild `core` and `alloc`, but I'm not sure if it's possible to clean out just `std` using `x.py`. This is something we can look into in the future.

Running `x.py` from anywhere other than the rust repo directory doesn't work, so we `cd $(local_std)` rather than run `$(local_std)/x.py`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ debug ?= none
 net ?= none
 merge_sections ?= yes
 bootloader ?= grub
+local_rust ?= none
 
 ## test for Windows Subsystem for Linux (Linux on Windows)
 IS_WSL = $(shell grep -is 'microsoft' /proc/version)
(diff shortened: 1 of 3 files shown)


kevinaboos (changes requested):
Hmm, yea this is a bit of a strange approach. I think the more canonical way of doing this would be to define separate make targets that each handle these specific behaviors rather than using env vars to completely change which build commands are run. 

The benefit of adding new dedicated targets for this is that you can set/override variables for _just_ that target (check out how you we do that for various targets, e.g., `loadable:`). That also avoids messing up the rest of our CI checks.

You probably want to set these targets as PHONY to ensure they'll always run, too, since we let Rust/cargo itself handle build caching, not make.

tsoutsman:
The issue is I want to run the entire `build` target (e.g. merge sections, copy crate objects, strip files, etc.), but I'm not sure how to create a new target without duplicating all of `$(nano_core_binary)` and switching it to rely on my different `cargo` target.

The CI build fails because it uses `--all-features` which also enables the newly added `std` feature which relies on this special build process. I'm merging this into `std_dep` so I'm not sure how big of an issue that is.

kevinaboos:
I see. Can you not just define a new target that sets whatever makefile env vars you need, and then either depends on the existing `build`, etc targets, or invokes them directly?

PR #603, opened 2022-08-16
Disable preemption, not interrupts, during task switching

kevinaboos:


kernel/spawn/src/lib.rs
```diff
@@ -587,8 +588,10 @@ fn task_wrapper_internal<F, A, R>() -> Result<R, task::KillReason>
     // Since the `task_switch()` function was originally invoked from an interrupt handler,
     // interrupts were disabled but never had the chance to be re-enabled
     // because we did not return from the interrupt handler as usual.
-    // Therefore, we need to re-enabled interrupts here to ensure that things continue
-    // to run as normal, with interrupts enabled so we can properly preempt this task.
+    // Therefore, we need to re-enable interrupts and preemption here to ensure that
+    // things continue to run as normal, with interrupts and preemption enabled
+    // such that we can properly preempt this task.
+    drop(recovered_preemption_guard);
     enable_interrupts();
```
tsoutsman, line 640: Do we still need this? Do interrupt handlers automatically disable interrupts when called?

tsoutsman (approved):
This generally looks good to me.

However, in my initial PR I exposed a `disable` method on `PreemptionGuard`. What if, instead of storing the guard in the next task, we just disabled the existing guard (or used `std::mem::forget`?), and then constructed a new `PreemptionGuard` after the context switch?

The guard constructor would be marked unsafe.

kernel/spawn/src/lib.rs
kevinaboos: yes, and yes.

kevinaboos:
> What if, instead of storing the guard in the next task, we just disabled the existing guard (or used `std::mem::forget`?), and then constructed a new `PreemptionGuard` after the context switch?

Yeah, i considered doing that instead of using the guard type. However, I really like the guarantees that guard types provide: 
* if you forget to explicitly enable or disable preemption, the guard type will handle it automatically
* This includes exceptional execution paths, i.e., after a panic, such that our unwinder can drop it.

Semantically, it's clearer to have a specific guard object that is moved around until we need it, and it also prevents logic bugs in a variety of scenarios. Once we implement proper per-CPU locals, we won't need to move things into the Task struct like we do now.

> The guard constructor would be marked unsafe.

In general, any alternative to `unsafe` and `mem::forget` is preferable. Offering an `unsafe` constructor for a preemption guard or any other way to circumvent the semantics of the preemption guard will likely only introduce headaches later when they're used in weird circumstances -- that'd be very difficult to debug.

kevinaboos:
Note: this isn't quite ready, needs more testing.

kevinaboos:
Fully tested and working correctly with all timeslice values fuzzed.

PR #606, opened 2022-08-17
Avoid `lazy_static` for TLS init by using const `RangeMap::new()`

kevinaboos:
* Added our own fork of `rangemap`, waiting on PR acceptance.


kevinaboos:
The `patch` statements can be removed once https://github.com/jeffparsons/rangemap/pull/52 is accepted.

PR #615, opened 2022-08-20
Add `time` crate

tsoutsman:
This PR is a simplification of #569. It adds the timer abstraction without implementing or using it anywhere.

We would implement `ClockSource` for various ZSTs e.g. `Pit`, `Tsc`, `Hpet`.
Then in `captain` we'd register them as so:
```rust
time::register_early_sleeper::<pit::Pit>().expect("couldn't set PIT as early sleeper");

// Discover devices including the HPET.
device_manager::early_init(kernel_mmi_ref.lock().deref_mut())?;

// Now we try set the HPET as the early sleeper. We prefer it over the PIT as it's more accurate.
time::register_early_sleeper::<hpet::Hpet>();

time::register_clock_source::<rtc::Rtc>().expect("failed to initialise a realtime clock");

time::register_clock_source::<tsc::Tsc>()
    .or_else(|_| time::register_clock_source::<hpet::Hpet>())
    .or_else(|_| time::register_clock_source::<pit::Pit>())
    .expect("failed to initialise a monotonic clock");
```

You mentioned in our discussion that you would prefer clock sources are registered upon discovery, like device drivers. The issue with that, is certain clock sources like TSC aren't _discovered_. In the case of TSC, we have to query CPUID to determine whether or not it's supported.

I feel like choosing a clock source based on certain characteristics exposed by the trait rather than a fixed order will significantly complicate things. A number of different factors should be considered when choosing a clock including accuracy, access speed, whether or not it uses interrupts, and resolution.


kevinaboos:
> You mentioned in our discussion that you would prefer clock sources are registered upon discovery, like device drivers. The issue with that, is certain clock sources like TSC aren't discovered. In the case of TSC, we have to query CPUID to determine whether or not it's supported.

Yes, I do think that's the best way forward. In this case, by "discovery" i mean that the crate that implements support for each clocksource should be the one to determine whether that clock source is valid before configuring it and registering it with whatever central timekeeping crate or clock-provider crate. There's really no other way to do it, since each clocksource crate is effectively the only source for whether a clocksource exists and how to configure it. 

In your TSC example, the `tsc` crate should be the one issuing CPUID commands to determine whether the TSC exists and whether it's a stable clocksource, and to determine its period/frequency/resolution/whatever other clocksource characteristics. The alternative is that the `captain` or whatever other init crate has to encode that information within it (?), which is poor design for myriad reasons.

It is indeed true that those `init` routines (e.g., `tsc::init()`) will have to be explicitly invoked, but that's true of many devices. That doesn't stop them from being "discoverable", it just means that we have to explicitly invoke their "discovery" (`init`) routines in order to check whether they exist. IMO, that's the job of the `device_manager`'s `init`/`early_init` routines.

-----

> I feel like choosing a clock source based on certain characteristics exposed by the trait rather than a fixed order will significantly complicate things. A number of different factors should be considered when choosing a clock including accuracy, access speed, whether or not it uses interrupts, and resolution.

Absolutely, it's totally fine to have a fixed order of clocksources. That fixed order can either use specific clocksources by name, or simply keep track of a priority list of each clocksource category based on characteristics. In the general case, it's impossible to identify a fixed order by name since that's arch-specific and even platform-specific, but it's okay for now.

-----
The clocksource trait is missing quite a few components describing the characteristics of the clocksource. In my mind, that trait exists solely for two reasons: to abstract access to concrete hardware clocks, and to describe the frequency, resolution, etc of each clock.

I don't mind the idea of using an associated type to establish what kind of clocksource it is, but I do question why it's beneficial to do it that way. Typically an associated type is used as more than just a simple marker trait, e.g., in `Iterator` the `Item` associated type is used to set the return type of `next()`. We're not really using it as anything beyond a marker, so it'd be more ergonomic and easier to understand if we just returned the type of the clocksource via a standard trait method.

If we ever wanted to offer specific methods that differ between a monolithic clock or wall-clock clocksource, then the associated type-based design would be more appropriate.

In general, I feel like this design forces the usage of generic types for no benefit, both in the clocksource trait and in the registration function.

tsoutsman:
Aaaand GitHub removed everything that I typed up :smile:.

> In your TSC example, the `tsc` crate should be the one issuing CPUID commands to determine whether the TSC exists and whether it's a stable clocksource, and to determine its period/frequency/resolution/whatever other clocksource characteristics. The alternative is that the `captain` or whatever other init crate has to encode that information within it (?), which is poor design for myriad reasons.

I'm not sure if I understand. In the API proposed by this crate, the only information about clock sources encoded in `captain` (or any other init crate) is the order. For example the CPUID commands would be run in `tsc::Tsc`'s implementation of `ClockSource::exists`, found in the `tsc` crate.

> It is indeed true that those `init` routines (e.g., `tsc::init()`) will have to be explicitly invoked, but that's true of many devices. That doesn't stop them from being "discoverable", it just means that we have to explicitly invoke their "discovery" (`init`) routines in order to check whether they exist. IMO, that's the job of the `device_manager`'s `init`/`early_init` routines.

`ClockSource::exists` is the "discovery", not `ClockSource::init`. `init` is only run if `exists` succesfuly "discovers" the clock. This is already done in `register_hardware_clock`.

> The clocksource trait is missing quite a few components describing the characteristics of the clocksource. In my mind, that trait exists solely for two reasons: to abstract access to concrete hardware clocks, and to describe the frequency, resolution, etc of each clock.

How would we expose the frequency.? Would we define functions like so:

```rust
pub fn frequency(ty: ClockType) -> usize;
```

As soon as we include more than one function, we can no longer technically use atomics because there could be race conditions if someone calls a function while we're in the middle of registering a clock. In practice, I don't think this is of much concern because we don't ever modify the atomics after startup. I don't mind adding the functions, but I'm not quite sure of the use case. When would a user need the frequency of the currently running clock?

Also, what does etc. specifically include? Resolution is just the inverse of frequency, and I don't think there's a way of determining the accuracy of a hardware clock.

> I don't mind the idea of using an associated type to establish what kind of clocksource it is, but I do question why it's beneficial to do it that way. Typically an associated type is used as more than just a simple marker trait, e.g., in `Iterator` the `Item` associated type is used to set the return type of `next()`.

IIUC you're proposing something like this:
```rust
pub trait ClockSource {
	fn exists() -> bool;
	fn init() -> Result<(), &'static str>;
	fn now() -> Duration;
	fn clock_type() -> ClockType;
}

enum ClockType {
	Realtime,
	Monotonic,
}

impl ClockType {
	fn func_addr(&self) -> &'static AtomicUsize {
		match self {
			Self::Realtime => &REALTIME_CLOCK_FUNCTION,
			Self::Monotonic => &MONOTONIC_CLOCK_FUNCTION,
		}
	}
}

pub fn register_clock_source<T>() -> Result<(), RegisterError>
where
	T: ClockSource,
{
	if T::exists() {
		T::init()?;
		T::clock_type().func_addr().store(T::now as usize, Ordering::SeqCst);
		Ok(())
	} else {
		Err(RegisterError::NonExistent)
	}
}

pub fn now(ty: ClockType) {
	let addr = ty.func_addr().load(Ordering::SeqCst);
	if addr == 0 {
		panic!("time function not set");
	} else {
		let f: fn() -> Duration = unsafe { core::mem::transmute(addr) };
		f()
	}
}
```
(excuse the atrocious indenting)

I thought that these kinds of type shenanigans weren't uncommon in Rust. I guess the only thing that GATs provide are a compile time guarantee that the clock type for a clock source won't change and a tiny performance improvement but I don't think either of those are particularly important. I'm fine with changing it to use enums.

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+static EARLY_SLEEP_FUNCTION: AtomicUsize = AtomicUsize::new(0);
+static EARLY_SLEEPER_FREQUENCY: AtomicU64 = AtomicU64::new(0);
+
+static MONOTONIC_NOW_FUNCTION: AtomicUsize = AtomicUsize::new(0);
+static INSTANT_TO_DURATION_FUNCTION: AtomicUsize = AtomicUsize::new(0);
+static DURATION_TO_INSTANT_FUNCTION: AtomicUsize = AtomicUsize::new(0);
+static MONOTONIC_FREQUENCY: AtomicU64 = AtomicU64::new(0);
+
+static REALTIME_NOW_FUNCTION: AtomicUsize = AtomicUsize::new(0);
+static REALTIME_FREQUENCY: AtomicU64 = AtomicU64::new(0);
```
kevinaboos: We ought to use a bit more type safety here by using the `Atomic` wrapper type from crossbeam (I use it elsewhere) to store actual `fn` references. That also prevents the undefined behavior (likely a page fault) if we try to invoke code at address 0.

kevinaboos: The initial values can be dummy functions that just return 0 or some other canary value.

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+    T: EarlySleeper,
+{
+    let old_frequency = EARLY_SLEEPER_FREQUENCY.load(Ordering::SeqCst);
+    if frequency > old_frequency {
+        if T::INIT_REQUIRED {
+            // FIXME: The source may be double initialised: once here and once in
+            // register_clock_source. This doesn't currently cause issues but needs to be
+            // fixed. This will probably be fixed if we separate interrupts from clocks in a
+            // future PR.
+            T::init()?;
```
kevinaboos: hmm, isn't this backwards from what we discussed a couple weeks ago? By that I mean that a given clocksource would be initialized separately (e.g., by the device manager) and then as part of its initialization routine, it would register itself as a clocksource/early sleeper.

Personally I find this to be a bit confusing in terms of control flow.

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+/// A clock source.
+pub trait ClockSource {
+    /// The type of clock (either [`Monotonic`] or [`Realtime`]).
+    type ClockType: ClockType;
+
+    /// Whether the clock source exists on the system.
+    fn exists() -> bool;
+
+    /// Initialise the clock source.
+    fn init() -> Result<(), &'static str>;
```
kevinaboos: same question as above -- I thought we agreed to invert the control flow here to a more typical init routine:
* device manager invokes init routine for a given clock device
* clock device's init routine sets itself up, reading whatever config values from hardware it needs
* clock device's init routine registers itself as a clock, sets itself as a sleeper, etc

I don't think there's any reason for `exists()` to exist (heh), and `init()` also shouldn't be part of a `ClockSource` trait imo.

kevinaboos (changes requested):
The "new" design doesn't seem to follow the results of our latest discussion, I left some comments related to that.

kevinaboos:
this crate may also be of interest, though we probably don't need to use it directly https://lib.rs/crates/embedded-time

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+    /// The type of clock (either [`Monotonic`] or [`Realtime`]).
+    type ClockType: ClockType;
+
+    /// The current time according to the clock.
+    ///
+    /// For monotonic clocks this is usually the time since boot, and for
+    /// realtime clocks it's the time since 12:00am January 1st 1970 (i.e.
+    /// Unix time).
+    ///
+    /// This function must only be called after [`init`](ClockSource::init).
```
kevinaboos: outdated, remove

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+    /// # Note to Implementors
+    ///
+    /// The default implementation of this function uses [`ClockSource::now`] -
+    /// it can only be used if [`ClockSource::now`] doesn't rely on
+    /// interrupts.
+    fn sleep(duration: Duration) {
+        let start = Self::now();
+        while Self::now() < start + duration {}
+    }
+}
```
kevinaboos: in both the trait docs and docs for `sleep()` we should mention that this definition of "sleep" doesn't mean the task will be put to sleep, but rather that it just spins/blocks the current task until the duration is up.

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,283 @@
...
+///
+/// The current early sleeper will only be overwritten by `T` if `frequency` is
+/// larger than the frequency of the current early sleeper.
+///
+/// Returns whether the early sleeper was overwritten.
+///
+/// # Errors
+///
+/// Returns an error if [`INIT_REQUIRED`](EarlySleeper::INIT_REQUIRED) is `true`
+/// and [`ClockSource::init`] returned an error.
```
kevinaboos: outdated, remove

kevinaboos (changes requested):
Left a few small comments.

It's a bit funky and likely unexpected for a clocksource implementor to call `register_clock_source()` without actually passing in an instance of the clocksource `T`, but I think that will have to change if we want to save all registered clocksources in the future. For now it's fine since I'm sure other unforeseen changes will crop up once we start implementing and using the clocksource trait for actual clock devices.

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/// Register the clock source that can be used to sleep when interrupts are
+/// disabled.
+///
+/// The current early sleeper will only be overwritten by `T` if `frequency` is
+/// larger than the frequency of the current early sleeper.
+///
+/// Returns whether the early sleeper was overwritten.
+pub fn register_early_sleeper<T>(frequency: u64) -> bool
```
kevinaboos: what is the unit for `frequency` here? Let's create a newtype called `Frequency` that clearly denotes the units such that comparisons are always sane.

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+impl Instant {
+    const ZERO: Self = Self { counter: 0 };
+
+    /// Returns the amout of time elapsed from another instant to this one, or
+    /// zero duration if that instant is later than this one.
+    pub fn duration_since(&self, earlier: Self) -> Duration {
+        let instant = Instant {
+            counter: match self.counter.checked_sub(earlier.counter) {
+                Some(value) => value,
+                None => return Duration::ZERO,
```
kevinaboos, line 53: is this typical Rust-y behavior? Seems like a more canonical approach would be to return an Option (or panic, but i'd prefer to not do that)

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+}
+
+impl core::ops::Add<Duration> for Instant {
+    type Output = Self;
+
+    fn add(self, rhs: Duration) -> Self::Output {
+        let instant = duration_to_instant(rhs);
+        Self {
+            counter: self.counter + instant.counter,
```
kevinaboos: don't we need to check for overflow in these two arithmetic ops impls? Otherwise we could end up creating a completely invalid inner value

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+/// A clock source.
+pub trait ClockSource {
+    /// The type of clock (either [`Monotonic`] or [`Realtime`]).
+    type ClockType: ClockType;
+
+    /// The current time according to the clock.
+    ///
+    /// For monotonic clocks this is usually the time since boot, and for
+    /// realtime clocks it's the time since 12:00am January 1st 1970 (i.e.
+    /// Unix time).
```
kevinaboos: This is good, but it's be even better to be concrete and further clarify that
* monotonic clocks return the time since system boot expressed as an `Instant`, whereas
* wall time clocks return a `Duration` representing the elapsed time since the Unix epoch.

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// interrupts, and can be used prior to the scheduler being initiated.
+    ///
+    /// # Note to Implementors
+    ///
+    /// The default implementation of this function uses [`ClockSource::now`] -
+    /// it can only be used if [`ClockSource::now`] doesn't rely on
+    /// interrupts.
+    fn sleep(duration: Duration) {
+        let start = Self::now();
+        while Self::now() < start + duration {}
```
kevinaboos, line 244: overflow check (?)

kernel/time/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "time"
+version = "0.1.0"
+authors = ["Klim Tsoutsman <klim@tsoutsman.com>"]
+description = "Abstraction for hardware timers"
```
kevinaboos: see my top-level comment in which I recommend distinguishing between a clock(source) and a timer

kernel/time/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+edition = "2021"
+
+[dependencies]
+log = "0.4.8"
+static_assertions = "1.1.0"
+
+[dependencies.crossbeam]
+version = "0.8.2"
+default-features = false
```
kevinaboos: to minimize dependencies, we can just pull in `crossbeam-utils`:
```toml
crossbeam-utils = { version = "0.8.12", default-features = false }
```

kevinaboos (changes requested):
Just got a few overly pedantic comments (sorry):
* It's probably best to avoid the keyword "realtime" since it's so overloaded. I know it's not technically wrong, but we probably want to use "wall time" or "wall clock time" instead, which is still accurate and has no connotation of hard- or pseudo-realtime behavior.
* The term hardware "timer" is used in various places, but we really mean hardware clock or clocksource. In the future we will add a timer abstraction, but this isn't really that.

kernel/time/src/lib.rs
tsoutsman: Yes https://doc.rust-lang.org/std/time/struct.Instant.html#method.duration_since

kernel/time/src/lib.rs
tsoutsman: there is no way to get the time since boot from an `Instant`, so from the perspective of the API consumer, it just represents an arbitrary point in time.

kernel/time/src/lib.rs
tsoutsman: the overflow check is now done in the addition

kernel/time/src/lib.rs
kevinaboos: ah ok. We should then add the `checked` version of that function too, which is what I was mistakenly thinking of, and then have this function call that checked one.

https://doc.rust-lang.org/src/std/time.rs.html#330-332

kernel/time/src/lib.rs
kevinaboos: right, that is true, thanks for pointing that out.

kevinaboos (changes requested):
Just requested one more small addition, a `checked_duration_since` function.

I also wonder if it's better to represent frequency not in Hz but as the inverse, i.e., a "period" value expressed in nanoseconds or something. The main potential problems with Hz are related to floating point numbers, which we'd need to use in order to assure accurate calculations. Introducing floating point is sort of undesirable though, since it tends to be infectious (other code must then use it or deal with it), the arithmetic ops on it are much slower, it can be harder to compare (test for equality), etc.

I don't think we need to bother with changing it now in this PR, I'm just mentioning it as a forward-looking thought and to see if you have any comments on that concern.

tsoutsman:
There are ways of avoiding floating point operations by expressing the frequency in GHz (1 GHzonemeans 1 tick per nanosecond) when calculating subsecond intervals. It is probably simpler to use the period as that is a simple multiplication `period * counter`.

FWIW I _think_ Linux uses frequency internally and OS Dev Wiki suggests [calculating the frequency of the HPET](https://wiki.osdev.org/HPET), although neither of these is particularly important.

The HPET provide the period, whereas the TSC and PIT provide the frequency.

None of this is currently a problem, as we only use the frequency to compare clocks. However, if we know the frequency (or period), we don't need the `instant_to_duration` and `duration_to_instant` functions as they are `period * counter` and `duration / period`, respectively, regardless of the clock source.

kernel/time/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+///
+/// Monotonic clocks return an [`Instant`] whereas wall time clocks return a
+/// [`Duration`] signifying the time since 12:00am January 1st 1970 (i.e. Unix
+/// time).
+///
+/// # Panics
+///
+/// This function will panic if called prior to registering a clock using
+/// [`register_clock_source`]. [`register_clock_source`] must return [`Ok`] and
+/// the [`ClockType`] of the registered clock must be the same as `T`.
```
kevinaboos: i was just about to merge this when I saw this doc comment. Is this still accurate? I don't see any panics possible in this code path, so it's likely outdated.

kevinaboos (changes requested):

kevinaboos (approved):

PR #617, opened 2022-08-23
Add additional documentation regarding `rip` in context switches

tsoutsman:

### DIFF
--- a/kernel/context_switch_regular/src/lib.rs
+++ b/kernel/context_switch_regular/src/lib.rs
@@ -21,11 +21,18 @@ pub struct ContextRegular {
     r12: usize,
     rbp: usize,
     rbx: usize,
+    /// The instruction pointer.
+    ///
+    /// `rip` is implicitly pushed onto the stack when a function is called, and
+    /// popped off when returning. When a task's stack is set to an instance of
+    /// [`ContextRegular`], [`context_switch_regular`] will execute `ret` when
+    /// the stack pointer is pointing to the value of `rip`. Hence, the program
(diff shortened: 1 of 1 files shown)


kevinaboos:
Hmm... I wonder why this is necessary. A context switch wouldn't be able to push/pop the current value of the instruction pointer, that doesn't really make sense because you cannot load a value directly into the IP/PC register on most architectures. IP/PC values aren't generally part of a context switch's stack switching operation. As such, I'm not sure what this actually clarifies; if anything I find it a bit more confusing to add it there, as it makes it seems like we're deliberately skipping a typical component of context switching or doing something in a weird/unique manner.

tsoutsman:
Perhaps, I put the comment in the wrong place. I think it's confusing that `ContextRegular` contains `rip` but it never seems to be accessed. Specifically, it's unclear how `spawn::setup_context_trampoline` works. I think at least a doc comment on the `rip` field of `ContextRegular` saying that it's stored at the bottom of the stack or something would help.

kevinaboos:
Ohhh, yes, I see. That would make more sense, I didn't realize that was the confusing part.

kevinaboos:
Yes, I'd be happy to accept a PR that more clearly documents how the various `Context*` structs work and why `rip` is the final field. I do think those comments should be atop those `Context*` structs themselves. 

It's currently very briefly mentioned in `ContextRegular::new()` but it doesn't really go so far as to explain why and how it's there, just what it's used for and what it does.

kevinaboos (approved):
Looks good to me!

PR #636, opened 2022-09-12
Uncomment passing complexity lints

dcampbell24:
My editor automatically removed the trailing white space, please let me know if I should remove that commit. I just wanted to remove the unneeded import that was giving me a warning even though I had everything turned off.

I added all the complexity lints that are passing. I will add the others but this is a first chunk. I am a new contributor to the project so wanted to make sure my ducks are in a row before doing too much.

### DIFF
--- a/kernel/page_table_entry/src/lib.rs
+++ b/kernel/page_table_entry/src/lib.rs
@@ -1,13 +1,13 @@
 //! Defines the structure of Page Table Entries (PTEs) on x86_64.
 //!
 //! This crate is also useful for frame deallocation, because it can determine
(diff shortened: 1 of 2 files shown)


tsoutsman (approved):

tsoutsman:
Thanks!

The project [doesn't use `rustfmt`](https://github.com/theseus-os/Theseus/pull/530#issuecomment-1154547570) and so we do have trailing white space sprinkled around the place. This kind of drive-by formatting can be troublesome for larger PRs because it can be hard to separate formatting changes from functional changes, but in this case, it's completely fine.

PR #637, opened 2022-09-13
Add rustfmt.toml, ignore all for now

hecatia-elegua:
instead of just `/` I added all the subfolders of root which contain rust code within them, so we can remove these if we're ever done formatting them

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,12 @@
+ignore = [
+    "/applications",
+    "/compiler_plugins",
+    "/kernel",
+    "/libs",
+    "/libtheseus",
+    "/old_crates",
+    "/ports",
+    "/theseus_features",
+    "/tlibc",
+    "/tools"
+]


kevinaboos:
Hi @hecatia-elegua, thanks for the PR. 

I don't want to be rude, and we do appreciate all contributions, but what's the purpose of this if it doesn't have any real effect?

hecatia-elegua:
@kevinaboos this: https://github.com/theseus-os/Theseus/pull/530#issuecomment-1155969831
and also because I would like to go through a lot of the repo and fix/refactor/format/clippy while/after reading the code :)
So I would then update this file with more specific paths, or maybe upstream will have some better way to handle big codebases.

Oh and just to be clear: I know that you have used manual formatting in a lot of places, I will try to bend rustfmt to keep that intact

kevinaboos:
> @kevinaboos this: [#530 (comment)](https://github.com/theseus-os/Theseus/pull/530#issuecomment-1155969831) and also because I would like to go through a lot of the repo and fix/refactor/format/clippy while/after reading the code :) So I would then update this file with more specific paths, or maybe upstream will have some better way to handle big codebases.

I see. I agree that formatting is important, we just haven't had the time to devote to making rustfmt behave nicely. Your plan to follow up with more changes sounds good to me, we'd greatly appreciate that! I'll approve this PR.

> Oh and just to be clear: I know that you have used manual formatting in a lot of places, I will try to bend rustfmt to keep that intact

Great! Yeah, that's one of our biggest barriers to using rustfmt: myself and a few others just don't like the way it formats things. I appreciate that you'll try to customize it, and I'll do my best to work with you to tailor it to our tastes.

I can start off by reviewing the features of rustfmt and adding a comment to #529 with some of my initial goals, e.g., max num columns, etc.

kevinaboos (approved):

PR #639, opened 2022-09-18
Add `vfs` API and `memory_vfs` implementation

tsoutsman:
This PR reworks the file system API. It also includes a port of `memfs`. The new API should also be simple to integrate with `fatfs` and other physical file systems.

The API uses forest mounting (i.e. Windows flavour), which is explained in `vfs`'s crate-level documentation. I chose this mounting method because it's simpler to implement.

Key differences from `fs_node`:
- Nodes use interior mutability - Certain operations may not need mutable access, e.g. when interacting with physical file systems; renaming the file doesn't require mutable access to the struct as the changes are made on the storage device, not in memory. Interior mutability simplifies the API and means mutexes are only used where necessary. Also, data can be stored behind multiple mutexes allowing for more parallel access, e.g. the contents and name of a memory file can be changed simultaneously.
- `FileSystem` trait - The critical thing lacking from `fs_node` is a clear distinction between different file systems, which is what this trait and the mount points aim to solve.
- Arbitrary nodes can't be added to a directory - `insert_file` and `insert_directory` (and their entry counterparts) take a path (or name) and return a new empty node. This means consumers can't add nodes from one file system into another, e.g. memory file into a FAT directory. Copying should be implemented by creating a new node and then writing all the bytes from the source file into the destination file.

Some issues:
- Trailing slashes
- Invalid file/dir names e.g. `.`, `..`, `/`
- Format of paths for `get_node` - I chose the `file_system:/path/to/file` because it's simple and is similar to what Redox uses, but I don't have any excellent reasons to use it.
- Resolving `.` and `..` in paths relies on the `trait_upcasting` feature, which is incomplete. This _should_ be resolved soon rust-lang/rust#101718 (aaaaaaand it's closed :))

### DIFF
--- a/kernel/vfs/src/lib.rs
+++ b/kernel/vfs/src/lib.rs
@@ -1,0 +1,74 @@
+//! File system abstractions for Theseus.
+//!
+//! These are expected to eventually replace `fs_node`.
+//!
+//! Theseus uses forest mounting to keep track of different file systems. The
+//! differences between forest mounting and tree mounting are explained below:
+//!
+//! - Forest mounting (i.e. Windows flavour): There are multiple directory tree
+//!   structures, with each one being a contained file system. For example,
+//!   Windows' drives (e.g. `C:`). Technically Windows does support folder mount
+//!   points but we're ignoring that for the sake of simplicity.
+//! - Tree mounting (i.e. Unix flavour): There is a single directory tree
+//!   structure originiating in the root directory. Filesystems are mounted to
+//!   subdirectories of the root.
+//!
+//! Theseus file systems are similar to Windows drives, except they use an
+//! arbitrary string identifier rather than a letter.
+
+#![cfg_attr(not(test), no_std)]
+#![feature(trait_upcasting)]
+#![allow(incomplete_features)]
+
+extern crate alloc;
+
+use alloc::{string::String, sync::Arc, vec::Vec};
+use spin::RwLock;
+
+mod node;
+mod path;
+
+pub use path::*;
+pub use node::*;
+
+/// The currently mounted file systems.
+///
+/// We use a [`Vec`] rather than a [`HashMap`] because it's more performant when
+/// the number of entries is in the tens. It also avoids the indirection of
+/// `lazy_static`.
+static FILE_SYSTEMS: RwLock<Vec<(String, Arc<dyn FileSystem>)>> = RwLock::new(Vec::new());
+
+/// A file system.
+pub trait FileSystem: Send + Sync {
+    /// Returns the root directory.
+    fn root(&self) -> Arc<dyn Directory>;
+}
+
+/// Returns the file system with the specified `id`.
+pub fn file_system(id: &str) -> Option<Arc<dyn FileSystem>> {
+    FILE_SYSTEMS.read().iter().find(|s| s.0 == id).map(|(_, fs)| fs).cloned()
+}
+
+/// Mounts a file system.
+///
+/// # Errors
+///
+/// Returns an error if a file system with the specified `id` already exists.
+#[allow(clippy::result_unit_err)]
+pub fn mount_file_system(id: String, fs: Arc<dyn FileSystem>) -> Result<(), ()> {
+    let mut file_systems = FILE_SYSTEMS.write();
+    if file_systems.iter().any(|s| s.0 == id) {
+        return Err(());
+    };
+    file_systems.push((id, fs));
+    Ok(())
+}
+
+/// Unmounts the file system with the specified `id`.
+///
+/// Returns the file system if it exists.
+pub fn unmount_file_system(id: &str) -> Option<Arc<dyn FileSystem>> {
+    let mut file_systems = FILE_SYSTEMS.write();
+    let index = file_systems.iter().position(|s| s.0 == id)?;
+    Some(file_systems.remove(index).1)
+}
--- a/kernel/vfs/src/lib.rs
+++ b/kernel/vfs/src/lib.rs
@@ -70,5 +70,5 @@ pub fn mount_file_system(id: String, fs: Arc<dyn FileSystem>) -> Result<(), ()> 
 pub fn unmount_file_system(id: &str) -> Option<Arc<dyn FileSystem>> {
     let mut file_systems = FILE_SYSTEMS.write();
     let index = file_systems.iter().position(|s| s.0 == id)?;
-    Some(file_systems.remove(index).1)
+    Some(file_systems.swap_remove(index).1)
 }
--- a/kernel/vfs/src/node/directory.rs
+++ b/kernel/vfs/src/node/directory.rs
@@ -35,15 +35,15 @@ pub trait Directory: Node {
         }
     }
 
-    /// Inserts a file with the specified name, returning it.
+    /// Creates a file with the specified name.
     ///
     /// `name` must not contain slashes.
-    fn insert_file_entry(self: Arc<Self>, name: &str) -> Result<Arc<dyn File>, &'static str>;
+    fn create_file_entry(self: Arc<Self>, name: &str) -> Result<Arc<dyn File>, &'static str>;
 
-    /// Inserts a directory with the specified name, returning it.
+    /// Creates a directory with the specified name.
     ///
     /// `name` must not contain slashes.
-    fn insert_directory_entry(
+    fn create_directory_entry(
         self: Arc<Self>,
         name: &str,
     ) -> Result<Arc<dyn Directory>, &'static str>;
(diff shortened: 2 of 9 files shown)


kernel/vfs/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    Ok(())
+}
+
+/// Unmounts the file system with the specified `id`.
+///
+/// Returns the file system if it exists.
+pub fn unmount_file_system(id: &str) -> Option<Arc<dyn FileSystem>> {
+    let mut file_systems = FILE_SYSTEMS.write();
+    let index = file_systems.iter().position(|s| s.0 == id)?;
+    Some(file_systems.remove(index).1)
```
NathanRoyer: It would be better to use swap_remove here I think

kernel/vfs/src/node/directory.rs
```diff
@@ -0,0 +1,147 @@
+// TODO: We would rather these methods take &Arc<Self>, but that isn't currently
+// possible.
```
NathanRoyer, line 2: These could be associated associated functions which would take Arcs, what do you think?

kernel/vfs/src/node/directory.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// `name` must not contain slashes.
+    fn insert_file_entry(self: Arc<Self>, name: &str) -> Result<Arc<dyn File>, &'static str>;
+
+    /// Inserts a directory with the specified name, returning it.
+    ///
+    /// `name` must not contain slashes.
+    fn insert_directory_entry(
+        self: Arc<Self>,
+        name: &str,
+    ) -> Result<Arc<dyn Directory>, &'static str>;
```
NathanRoyer: Having a common `insert_entry` (method taking a NodeKind) could be good here too, no?

kernel/vfs/src/node/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+/// A specific file system node kind.
+pub enum NodeKind {
+    Directory(Arc<dyn Directory>),
+    File(Arc<dyn File>),
+}
+
+/// Gets the node at the `path`.
+///
+/// The path consists of a file system id, followed by a colon, followed by the
+/// path. For example, `tmp:/a/b` or `nvme:/foo/bar`.
```
NathanRoyer, line 40: The use of colon here could be debated, I'm not particularly against it though.

kernel/vfs/src/path.rs
```diff
@@ -0,0 +1,109 @@
+/// A file system path.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct Path<'a> {
+    inner: &'a str,
```
NathanRoyer, line 4: We could allow filesystems to use slashes in node names if we used an iterable here directly rather than a consolidated path, a vec for instance

NathanRoyer (commented):
Overall: great work imo :ok_hand:

kernel/vfs/src/node/directory.rs
tsoutsman: The API intentionally doesn't let you insert an existing `dyn Node` (or `dyn File`/`dyn Directory`/`NodeKind`) so that different file system nodes aren't mixed. On another note I think create is a better name than insert.

kernel/vfs/src/node/mod.rs
tsoutsman: As I said in the PR, I just copied Redox but this is definitely up for debate if anyone has any better ideas.

kernel/vfs/src/path.rs
tsoutsman: Allowing slashes in node names sounds like asking for trouble. Especially because the `std` fs API uses string paths and so we'd somehow have to escape slashes in node names.

kernel/vfs/src/node/directory.rs
tsoutsman: I'm not quite sure what you mean, could you give an example?

The reason it can't currently be done is that the trait needs to be [object safe](https://doc.rust-lang.org/reference/items/traits.html#object-safety).

NathanRoyer: Ah, indeed I forgot about that object safety trait stuff

kernel/vfs/src/node/directory.rs
NathanRoyer: > On another note I think create is a better name than insert.

I'd say "insert" refers to slot insertion (one may think it implies there's a growable array underneath) whereas "create" is more general/natural.

kernel/vfs/src/node/mod.rs
NathanRoyer: If we used the first path node name as fs name (`/root-fs/usr/bin/gcc` or `/usb-drive-2/photos/camping-trip-2016` for instance), that would ease filesystem enumeration but could also result in confusement and vulnerabilities (php's path traversal issues come to mind).

kernel/vfs/src/path.rs
kevinaboos: yeah we cannot allow slashes in fs node names, that's a fairly typical limitation.

kernel/vfs/src/node/mod.rs
kevinaboos: i'd probably prefer to have a more Unix-style set of mount points, but for simplicity's sake it doesn't have to be arbitrary. We could have all FSes mounted at `/` or `/mnt/`, for example.

kevinaboos (changes requested):
Good start so far. Agreed that we obviously do need a way to register and support multiple distinct filesystem instances. I also like the idea of forcing interior mutability from the primary VFS interface. There is one major issue I see that requires a redesign: we should not and cannot use `Arc` everywhere. 

Aside from being inefficient, copious usage of `Arc` only really works in the context of an in-memory FS; a real physical FS wouldn't be able to use `Arc`s to represent a relationship between nodes when stored on disk, e.g., a directory entry containing multiple file entries. For that, inodes are commonly used, but we don't have to explicitly include the concept of inodes in the VFS layer. Instead, we just need an abstract way to represent the concept of one entry that refers to one or more other entries.

So i'd recommend that we create an actual concrete outer type for `File` and `Directory` instead of using trait objects like `Arc<dyn File>` and `Arc<dyn Directory>`, and then the inner fields of that struct are dependent on what the real filesystem needs to represent the inter-node relationships. For a simple heap-based fs (like the current`heapfile`), the inner field of the top-level `Directory` could indeed by an `Arc` reference to a set of directory entries, which would make sense because it's an efficient way to represent those relationships when you know everything is in heap memory.

Other than that, I left a few small comments on existing discussions.

Since you're focused on other things right now, I'd be happy to take a crack at extending this PR  with a new attempt at the design I've described above. Feedback also welcome here or on Discord.

tsoutsman:
Hm, yea that's definitely an issue.

Even if we have concrete outer types, eventually we will have to use trait objects. Generics won't work because the type of filesystem returned by something like `open("/dev1/foo/bar")` is only decided at runtime.

I don't have any particularly good ideas at the moment so definitely feel free to extend the PR.

tsoutsman:
I've been thinking about this a bit, and I've come up with a potential solution.

We can't guarantee that the nodes returned from `get` are valid because another thread may immediately delete them, so we can't return a reference. We can work around this by tracking open files.

If we do something like this:
```rust
pub trait FileSystemImplementation {
    type File: File;
    type FileRef: FileRef<File = Self::File>;
    type Directory: Directory;

    fn root(&self) -> Self::Directory;
}

pub trait FileRef {
    type File: File;

    fn upgrade(self) -> Option<Arc<dyn File>>;
}

pub struct FileSystem {
    inner: Box<
        dyn FileSystemImplementation<
            File = dyn File,
            FileRef = dyn FileRef<File = dyn File>,
            Directory = dyn Directory,
        >,
    >,
    opened: HashMap<Path, OpenFile>,
}

pub struct OpenFile {
    inner: Arc<dyn File>,
}

pub trait File {
    // ...
}

pub trait Directory {
    // ...
}

pub struct Path {
    // ..
}
```

We can enforce that files aren't deleted if they are open. File system operations would be done on the `FileSystem` struct, which would perform the necessary modifications to the `opened` field. For example, deletion would first check the strong count of `OpenFile.inner`.

`Directory::get` would return a `FileSystem::FileRef` whereas `Directory::open` would return an `OpenFile`. For example, in a FAT file system, the `FileRef` would contain a reference to the filesystem, which it would then query in `upgrade`. A memory filesystem `FileRef` would be a weak reference to the file system node.

Although this implementation also extensively uses `Arc`s, they're used as a layer on top of the filesystem to prevent/reduce race conditions rather than a way to model the file system's structure.

Hypothetically, we can also just return a `FileSystem::File` and assume it won't be deleted. E.g. [`fatfs::File`](https://docs.rs/fatfs/latest/fatfs/struct.File.html) or in the case of `memfs` an `Arc<MemoryFile>`.

tsoutsman:
I've played around with this some more, and I don't think it will be possible till traits with GATs are object safe.

PR #640, opened 2022-09-23
Use crate metavariable in print crates

tsoutsman:
Avoids having to import the `print` macro when using `println` with newer Rust versions (without `#[macro_use]`).


kevinaboos (approved):
Looks good, thanks!

PR #648, opened 2022-10-06
Add `MappedPagesBorrowed[Mut]<T>`, a self-referential struct that borrows its contained `MappedPages` as `&T`/`&mut T`

kevinaboos:
* `BorrowedMappedPages<T>` stores an immutable reference to a struct of type `T` that has been overlaid onto a range of the underlying memory owned by a given `MappedPages` object.
   * `BorrowedMappepdPagesMut<T> is the same but stores a mutable reference to that type `T`

* This builds on `MappedPages::as_type[_mut]()` to borrow a `MappedPages` object as an immutable or mutable reference to a struct of a different type `&'m T`/`&'m mut T`, where `'m` is the lifetime of the `MappedPages` object. 
   * This is a self-referential struct that is sound only because `MappedPages` can never move in memory after creation -- its virtual address is fixed for the entire lifetime of the `MappedPages` object.

* Offers an alternative to the unsound `owning_ref` crate's `BoxRef[Mut]` type

* Doesn't require heap allocation (`Box`) to fix the owner's address in memory, because `MappedPages` is already fixed.


kernel/memory/src/paging/mapper.rs
```diff
@@ -867,3 +867,104 @@ pub fn mapped_pages_unmap(
...
+/// When dropped, the borrow ends and the contained `MappedPages` is dropped and unmapped.
+pub struct BorrowedMappedPages<T: FromBytes> {
+    ptr: NonNull<T>,
+    mp: MappedPages,
+}
+impl<T: FromBytes> Deref for BorrowedMappedPages<T> {
+    type Target = T;
+    fn deref(&self) -> &T {
+        // SAFETY:
+        // ✅ The pointer is properly aligned because we use `NonNull`.
```
tsoutsman: `NonNull` doesn't guarantee that the pointer is aligned. `NonNull::dangling()` does, but we don't use the dangling pointer.

The guarantee of the pointer being properly aligned should come from the fact that `as_type` returns a valid reference to `T` but, as I mention below, that may not be true.

kernel/memory/src/paging/mapper.rs
```diff
@@ -867,3 +867,104 @@ pub fn mapped_pages_unmap(
...
+        offset: usize,
+    ) -> Result<BorrowedMappedPages<T>, (MappedPages, &'static str)> {
+        let mut borrowed_mp = BorrowedMappedPages {
+            ptr: NonNull::<T>::dangling(),
+            mp,
+        };
+        borrowed_mp.ptr = match borrowed_mp.mp.as_type::<T>(offset) {
+            Ok(r) => r.into(),
+            Err(e_str) => return Err((borrowed_mp.mp, e_str)),
+        };
```
tsoutsman:
```suggestion
        let borrowed_mp = BorrowedMappedPages {
            ptr: match mp.as_type::<T>(offset) {
                Ok(r) => r.into(),
                Err(e_str) => return Err((mp, e_str)),
            },
            mp,
        };
```
The call to `NonNull::dangling()` seems unecessary.

kernel/memory/src/paging/mapper.rs
```diff
@@ -867,3 +867,104 @@ pub fn mapped_pages_unmap(
...
+    /// describing the error.
+    pub fn try_into_borrowed(
+        mp: MappedPages,
+        offset: usize,
+    ) -> Result<BorrowedMappedPages<T>, (MappedPages, &'static str)> {
+        let mut borrowed_mp = BorrowedMappedPages {
+            ptr: NonNull::<T>::dangling(),
+            mp,
+        };
+        borrowed_mp.ptr = match borrowed_mp.mp.as_type::<T>(offset) {
```
tsoutsman: This is an issue with `as_type`, but I can't comment on its source code as it's not part of the PR.

`as_type` doesn't check alignment which I believe is unsound. The [`zerocopy::FromBytes`](https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html) documentation states that:

> `FromBytes` types can safely be deserialized from an untrusted sequence of bytes because any byte sequence corresponds to a valid instance of the type.

But it doesn't mention anything about alignment. However, the existence of the [`zerocopy::Unalign`](https://docs.rs/zerocopy/latest/zerocopy/struct.Unalign.html) type suggests that if we aren't using `zerocopy::Unalign` we should be checking pointer alignment.

kernel/memory/src/paging/mapper.rs
```diff
@@ -867,3 +867,104 @@ pub fn mapped_pages_unmap(
...
+        offset: usize,
+    ) -> Result<BorrowedMutMappedPages<T>, (MappedPages, &'static str)> {
+        let mut borrowed_mp = BorrowedMutMappedPages {
+            ptr: NonNull::<T>::dangling(),
+            mp,
+        };
+        borrowed_mp.ptr = match borrowed_mp.mp.as_type_mut::<T>(offset) {
+            Ok(r) => r.into(),
+            Err(e_str) => return Err((borrowed_mp.mp, e_str)),
+        };
```
tsoutsman:
```suggestion
        let borrowed_mp = BorrowedMutMappedPages {
            ptr: match mp.as_type_mut::<T>(offset) {
                Ok(r) => r.into(),
                Err(e_str) => return Err((mp, e_str)),
            },
            mp,
        };
```
Same as above.

kernel/memory/src/paging/mapper.rs
```diff
@@ -867,3 +867,104 @@ pub fn mapped_pages_unmap(
...
+    /// describing the error.
+    pub fn try_into_borrowed_mut(
+        mp: MappedPages,
+        offset: usize,
+    ) -> Result<BorrowedMutMappedPages<T>, (MappedPages, &'static str)> {
+        let mut borrowed_mp = BorrowedMutMappedPages {
+            ptr: NonNull::<T>::dangling(),
+            mp,
+        };
+        borrowed_mp.ptr = match borrowed_mp.mp.as_type_mut::<T>(offset) {
```
tsoutsman: Same alignment issue as above.

tsoutsman (changes requested):

kernel/memory/src/paging/mapper.rs
NathanRoyer: > The call to NonNull::dangling() seems unecessary.

But the `borrowed_mp` variable doesn't exist yet in your snippet at line 956, does it?

kernel/memory/src/paging/mapper.rs
NathanRoyer: > The call to NonNull::dangling() seems unecessary.

`borrowed_mp` doesn't exist yet when you use it :thinking:

NathanRoyer (commented):
I'm not familiar with zerocopy but this looks good, thanks Klim for checking alignment :ok_hand:

kernel/memory/src/paging/mapper.rs
tsoutsman: Right, good catch. I've modified the snippet to reference `mp` rather than `borrowed_mp.mp` which should work.

kernel/memory/src/paging/mapper.rs
tsoutsman: Same as above.

kernel/memory/src/paging/mapper.rs
kevinaboos: I was previously moving `mp` into `borrowed_mp` in order to make it explicitly clear that the lifetime is tied to that of something owned by `borrowed_mp`, but yeah i agree it's not necessary.

kernel/memory/src/paging/mapper.rs
kevinaboos: ah yes, alignment. Thanks for bringing this up. So several years ago I actually had experimented with enforcing that the `offset` used into `MappedPages::as_*` was properly aligned according to the alignment requirements of the type `T`. However, it was unnecessarily restrictive and threw runtime errors for some of our code at the time, because there were indeed valid cases where we wanted to overlay a type at an offset that wasn't an even multiple of that type's alignment.

I've never had any issues with unaligned pointers returned by `MappedPages::as_*` on any machine, both real hardware and emulated, so I'm not really sure if that's an actual problem. Nevertheless, it's probably a good idea to avoid unaligned accesses just for perf reasons at the last, so I will dig into this again to see if I can force all `MappedPages::as_*` to check that alignment is correct, and revisit the cases where that caused a problem.

kernel/memory/src/paging/mapper.rs
kevinaboos: ack'ed

kevinaboos:
Blocked on #651

kernel/memory/src/paging/mapper.rs
kevinaboos: Fixed now

kernel/memory/src/paging/mapper.rs
kevinaboos: Fixed

kevinaboos:
alignment is now checked as of #651, so the safety comments here are now fully upheld correctly.

kernel/memory/src/paging/mapper.rs
```diff
@@ -774,11 +777,15 @@ impl MappedPages {
...
+    /// * `byte_offset`: the offset (in number of bytes) into the memory region
+    ///    at which the slice should start.
+    ///    This `byte_offset` must be properly aligned with respect to the alignment requirements
+    ///    of type `T`, otherwise an error will be returned.
+    /// * `length`: the length of the slice, i.e., the number of elements of type `T` in the slice. 
+    ///    Thus, the slice's address bounds will span from `byte_offset` to
+    ///    `byte_offset + (size_of::<T> * length)`.
     /// 
-    /// Returns a reference to the new slice that is formed from the underlying memory region,
+    /// Returns a reference to the new slice that a reference to the underlying memory region,
```
tsoutsman: missing "is"?

kernel/memory/src/paging/mapper.rs
```diff
@@ -774,11 +777,15 @@ impl MappedPages {
...
-    /// * `byte_offset`: the offset (in number of bytes) into the memory region at which the slice should start.
-    /// * `length`: the length of the slice, i.e., the number of `T` elements in the slice. 
-    ///   Thus, the slice will go from `offset` to `offset` + (sizeof(`T`) * `length`).
+    /// * `byte_offset`: the offset (in number of bytes) into the memory region
+    ///    at which the slice should start.
+    ///    This `byte_offset` must be properly aligned with respect to the alignment requirements
+    ///    of type `T`, otherwise an error will be returned.
+    /// * `length`: the length of the slice, i.e., the number of elements of type `T` in the slice. 
+    ///    Thus, the slice's address bounds will span from `byte_offset` to
+    ///    `byte_offset + (size_of::<T> * length)`.
```
tsoutsman:
```suggestion
    ///    `byte_offset + (size_of::<T>() * length)`.
```

tsoutsman (changes requested):
Just some tiny documentation nitpicks.

NathanRoyer (approved):
Looks good

kernel/memory/src/paging/mapper.rs
kevinaboos: thanks, fixed.

tsoutsman (approved):

PR #657, opened 2022-10-09
Make `Task::block` and `Task::unblock` fallible

tsoutsman:
Previously, exited tasks could be blocked or unblocked leading to a panic during context switching.

### DIFF
--- a/applications/shell/src/lib.rs
+++ b/applications/shell/src/lib.rs
@@ -1419,9 +1419,11 @@ impl Shell {
             if let Ok(job_num) = job_num.parse::<isize>() {
                 if let Some(job) = self.jobs.get_mut(&job_num) {
                     for task_ref in &job.tasks {
-                        let _ = task_ref.unblock();
-                        // TODO: Do we want to set the job status of an exited task to running?
-                        job.status = JobStatus::Running;
+                        if let Err(_) = task_ref.unblock() {
+                            job.status = JobStatus::Stopped;
+                        } else {
+                            job.status = JobStatus::Running;
+                        }
                     }
                     self.clear_cmdline(false)?;
                     self.redisplay_prompt();
@@ -1451,9 +1453,11 @@ impl Shell {
                 if let Some(job) = self.jobs.get_mut(&job_num) {
                     self.fg_job_num = Some(job_num);
                     for task_ref in &job.tasks {
-                        let _ = task_ref.unblock();
-                        // TODO: Do we want to set the job status of an exited task to running?
-                        job.status = JobStatus::Running;
+                        if let Err(_) = task_ref.unblock() {
+                            job.status = JobStatus::Stopped;
+                        } else {
+                            job.status = JobStatus::Running;
+                        }
                     }
                     return Ok(());
                 }
--- a/applications/shell/src/lib.rs
+++ b/applications/shell/src/lib.rs
@@ -106,11 +106,12 @@ pub fn main(_args: Vec<String>) -> isize {
         };
     }
 
+    // block this task, because it never needs to actually run again
+    task::get_my_current_task().unwrap().block().unwrap();
+    scheduler::schedule();
+
     loop {
-        // block this task, because it never needs to actually run again
-        if let Some(my_task) = task::get_my_current_task() {
-            my_task.block().unwrap();
-        }
+        warn!("BUG: blocked shell task was scheduled in unexpectedly");
     }
 
     // TODO: when `join` puts this task to sleep instead of spinning, we can re-enable it.
--- a/applications/shell/src/lib.rs
+++ b/applications/shell/src/lib.rs
@@ -28,7 +28,7 @@ extern crate libterm;
 #[macro_use] extern crate alloc;
 #[macro_use] extern crate log;
 
-use event_types::{Event};
+use event_types::Event;
 use keycodes_ascii::{Keycode, KeyAction, KeyEvent};
 use alloc::string::{String, ToString};
 use alloc::vec::Vec;
@@ -109,7 +109,7 @@ pub fn main(_args: Vec<String>) -> isize {
     loop {
         // block this task, because it never needs to actually run again
         if let Some(my_task) = task::get_my_current_task() {
-            my_task.block();
+            my_task.block().unwrap();
         }
     }
 
@@ -462,8 +462,9 @@ impl Shell {
                 app_io::lock_and_execute(&|_flags_guard, _streams_guard| {
                     // Stop all tasks in the job.
                     for task_ref in task_refs {
-                        if task_ref.has_exited() { continue; }
-                        task_ref.block();
+                        if task_ref.block().is_err() {
+                            continue;
+                        }
 
                         // Here we must wait for the running application to stop before releasing the lock,
                         // because the previous `block` method will NOT stop the application immediately.
@@ -770,7 +771,7 @@ impl Shell {
 
                 // All IO streams have been set up for the new tasks. Safe to unblock them now.
                 for task_ref in &new_job.tasks {
-                    task_ref.unblock();
+                    task_ref.unblock().unwrap();
                 }
 
                 // Allocate a job number for the new job. It will start from 1 and choose the smallest number
@@ -1418,9 +1419,8 @@ impl Shell {
             if let Ok(job_num) = job_num.parse::<isize>() {
                 if let Some(job) = self.jobs.get_mut(&job_num) {
                     for task_ref in &job.tasks {
-                        if !task_ref.has_exited() {
-                            task_ref.unblock();
-                        }
+                        let _ = task_ref.unblock();
+                        // TODO: Do we want to set the job status of an exited task to running?
                         job.status = JobStatus::Running;
                     }
                     self.clear_cmdline(false)?;
@@ -1451,9 +1451,8 @@ impl Shell {
                 if let Some(job) = self.jobs.get_mut(&job_num) {
                     self.fg_job_num = Some(job_num);
                     for task_ref in &job.tasks {
-                        if !task_ref.has_exited() {
-                            task_ref.unblock();
-                        }
+                        let _ = task_ref.unblock();
+                        // TODO: Do we want to set the job status of an exited task to running?
                         job.status = JobStatus::Running;
                     }
                     return Ok(());
--- a/applications/test_wait_queue/src/lib.rs
+++ b/applications/test_wait_queue/src/lib.rs
@@ -55,7 +55,7 @@ fn rmain() -> Result<(), &'static str> {
     let t3 = spawn::new_task_builder(
         |tref| {
             warn!("DeezNutz:  testing spurious wakeup on task {:?}", tref);
-            tref.unblock();
+            tref.unblock().unwrap();
         }, 
         t1.clone(),
         )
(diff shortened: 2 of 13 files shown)


applications/shell/src/lib.rs
```diff
@@ -1451,9 +1451,8 @@ impl Shell {
                 if let Some(job) = self.jobs.get_mut(&job_num) {
                     self.fg_job_num = Some(job_num);
                     for task_ref in &job.tasks {
-                        if !task_ref.has_exited() {
-                            task_ref.unblock();
-                        }
+                        let _ = task_ref.unblock();
+                        // TODO: Do we want to set the job status of an exited task to running?
```
kevinaboos: no, if the unblock operation fails because the task is exited here, there we should mark it as such and remove the job.

applications/test_wait_queue/src/lib.rs
```diff
@@ -55,7 +55,7 @@ fn rmain() -> Result<(), &'static str> {
     let t3 = spawn::new_task_builder(
         |tref| {
             warn!("DeezNutz:  testing spurious wakeup on task {:?}", tref);
```
kevinaboos, line 57: lol can't believe I left that in. Feel free to remove haha

kernel/serial_port/src/lib.rs
```diff
@@ -186,12 +186,20 @@ impl SerialPort {
                 match SerialPortAddress::try_from(base_port) {
                     Ok(SerialPortAddress::COM1 | SerialPortAddress::COM3) => {
                         INTERRUPT_ACTION_COM1_COM3.call_once(|| 
-                            Box::new(move || { deferred_task.unblock(); })
+                            Box::new(move || {
+                                deferred_task
+                                    .unblock()
+                                    .expect("BUG: com_1_com3_interrupt_handler: couldn't unblock deferred task");
+                            })
```
kevinaboos: i find this kind of formatting really tedious and hard to read. Personally would prefer something like:
```rust
Box::new(move || deferred_task.unblock()
    .expect(...)
)
```

kernel/serial_port/src/lib.rs
```diff
@@ -186,12 +186,20 @@ impl SerialPort {
...
                         );
                     }
                     Ok(SerialPortAddress::COM2 | SerialPortAddress::COM4) => {
                         INTERRUPT_ACTION_COM2_COM4.call_once(|| 
-                            Box::new(move || { deferred_task.unblock(); })
+                            Box::new(move || {
+                                deferred_task
+                                    .unblock()
+                                    .expect("BUG: com_2_com4_interrupt_handler: couldn't unblock deferred task");
+                            })
```
kevinaboos: same formatting comment as above

kernel/sleep/src/lib.rs
```diff
@@ -88,7 +88,7 @@ fn add_to_delayed_tasklist(new_node: SleepingTaskNode) {
 fn remove_next_task_from_delayed_tasklist() {
     let mut delayed_tasklist = DELAYED_TASKLIST.lock();
     if let Some(SleepingTaskNode { taskref, .. }) = delayed_tasklist.pop() {
-        taskref.unblock();
+        let _ = taskref.unblock();
```
kevinaboos: see top-level review comment about forwarding `Result`s instead of `unwrap`ping. This however is probably the least good option since it just silently masks failures.

kernel/task/src/lib.rs
```diff
@@ -678,13 +678,59 @@ impl Task {
     }
 
     /// Blocks this `Task` by setting its runstate to [`RunState::Blocked`].
-    pub fn block(&self) {
-        self.runstate.store(RunState::Blocked);
+    ///
+    /// Returns the previous runstate. Trying to block an exited or reaped task
+    /// will return an error.
+    pub fn block(&self) -> Result<RunState, ()> {
```
kevinaboos: 1. Should probably return the current runstate on error for extra info
2. Clarify the docs to state that the function only succeeds if the current runstate is `Runnable`
3. Clarify the docs to state that it returns the previous runstate on success and the current runstate on error

kernel/task/src/lib.rs
```diff
@@ -678,13 +678,59 @@ impl Task {
...
+            Ok(RunState::Runnable)
+        } else if self
+            .runstate
+            .compare_exchange(RunState::Blocked, RunState::Blocked)
+            .is_ok()
+        {
+            Ok(RunState::Blocked)
+        } else {
+            Err(())
+        }
```
kevinaboos: IMO these should be re-ordered to move the common success case to the top, where the current runstate is `Runnable` and the new one is `Blocked`. That way we're not always wasting cycles on an expensive atomic conditional for a case (`Initing` --> `Blocked`) that will likely never occur.

I think it should be an error to attempt to block a task that is in the `Initing` runstate, since that means something went wrong with initialization and you were able to obtain a reference to a task that wasn't yet finished being initialized. I don't believe that could happen, but better safe than sorry.
For the one case where this is currently accepted (spawning a task that shouldn't run), perhaps we can add a separate code path to allow that transition in only this case.

Could you also condense the formatting a bit?

kevinaboos: Also should it be considered an error to attempt to block a task that is already blocked? Might be indicative of a logic error in the calling application; currently it's fine and doesn't hurt, but still worth raising the question.

kernel/task/src/lib.rs
```diff
@@ -678,13 +678,59 @@ impl Task {
...
+            Ok(RunState::Runnable)
+        } else if self
+            .runstate
+            .compare_exchange(RunState::Blocked, RunState::Runnable)
+            .is_ok()
+        {
+            Ok(RunState::Blocked)
+        } else {
+            Err(())
+        }
```
kevinaboos: same comments as for `block()` above

kernel/wait_queue/src/lib.rs
```diff
@@ -198,23 +199,27 @@ impl WaitQueue {
...
-        // trace!("  notify: chose task to wakeup: {:?}", tref);
-        if let Some(t) = tref {
-            // trace!("WaitQueue::notify():  unblocked task on waitqueue\n    --> WQ: {:?}", &*wq_locked);
-            t.unblock();
-            true
-        } else {
-            // trace!("WaitQueue::notify():  did nothing");
-            false
+        
+        loop {
```
kevinaboos, line 209: i think this could cause an infinite loop; any options to avoid that problem?

kevinaboos (changes requested):
Thanks for the PR, I agree this can help avoid potential weird task behavior if someone is able to get ahold of an exited task and mark it as unblocked. 

I left a few comments; aside from that, I think it'd be nice to return `Result`s wherever possible instead of panicking on an `unwrap`. If there are places where you're unsure how best to bubble up those `Result`s, just leave a comment denoting where and I can handle them.

Thoughts for a future PR: eventually I'd like to completely redesign task blocking to have some kind of "`BlockGuard`", with the idea being that when a task `A` has blocked another task `B`, if task `A` crashes and panics we can then drop that `BlockGuard` during unwinding, which will then unblock task `B`. Right now without that feature, task `B` will never be unblocked if task `A` fails, which is sort of a less-than-ideal outcome, at least from the mindset of fault tolerance.

kernel/wait_queue/src/lib.rs
tsoutsman: I don't think it could because on every loop an item gets removed from `wq_locked` and so eventually `wq_locked` will run out of items and the function will return `false`.

kernel/serial_port/src/lib.rs
tsoutsman: The best I can do is [this](https://github.com/theseus-os/Theseus/pull/657/commits/00880ec54bdd6e04ac021ce41524c8fe1cab34af). Your suggestion won't work because if the closure is more than one line it must be delimited using curly braces.

kernel/sleep/src/lib.rs
tsoutsman: should we log something here? is a task being killed while it is sleeping a failure?

kernel/task/src/lib.rs
tsoutsman: By separate code path do you mean a function?

> Also should it be considered an error to attempt to block a task that is already blocked?

Maybe? The only legitimate use case I can currently think of is the shell processing a Ctrl+Z signal while running a task that is already blocked by something (e.g. Mutex). But in my rework of the `tty`, I add a separate `Suspended` task state similar to Linux's `STOPPED` which ensures the two states aren't mixed.

applications/test_wait_queue/src/lib.rs
tsoutsman: I like it. It's like a historical artifcat.

kernel/task/src/lib.rs
kevinaboos: > By separate code path do you mean a function?

yes, sorry, i meant exposing a new dedicated function within `task` for this purpose alone, i.e., something like `block_new_task()` or `block_initing_task()` which only allows a Runstate transition from `Initing` --> `Blocked`.

kevinaboos: As for blocking an already-blocked task, for now let's just log a warning and see where/when it occurs.

kernel/wait_queue/src/lib.rs
kevinaboos: good enough for me :smile:

kernel/serial_port/src/lib.rs
kevinaboos: ah i see. It's not because it's multiple lines, it's because we don't wan't to return the value of the last expression in the block, we want the return to be void `()`.

anyway, good enough.

kernel/sleep/src/lib.rs
kevinaboos: yep, it's worth logging and then unwrapping. The whole `sleep` crate will be redesigned once we have better sync primitives, e.g., the ones you're working on as part of `std` stuff.

Another eventual goal is to force a killed task to unwind (and thus, release its resources) rather than just stopping its execution immediately. That'd be relevant here too, but for now I think an `unwrap` call is warranted.

tsoutsman:
Should I remove the unwraps from the test applications as well?

Also FYI, I originally encountered this issue when killing a task waiting on a mutex. The task would remain in the wait queue and hence be unblocked.

kevinaboos:
> Should I remove the unwraps from the test applications as well?

If you want, but IMO it's fine if application code panics.

> Also FYI, I originally encountered this issue when killing a task waiting on a mutex. The task would remain in the wait queue and hence be unblocked.

I see. That sort of odd behavior is also a good motivating factor for why we need the above `BlockGuard`-esque solution.

kevinaboos:
Looks good as of now, I made a few small changes and am testing some things.

@tsoutsman btw, when running this branch did you ever encounter any of the new `warn!()` log statements that you added about blocked tasks being blocked again or unblocked tasks being unblocked again? I wouldn't want to flood the log with warnings if that's common behavior (but afaik i try to avoid doing spurious wakeups like that).

kevinaboos (approved):
I made a few small changes and removed the one case of a blocked task being re-blocked (in the shell). Ready to merge, thanks!

tsoutsman:
> @tsoutsman btw, when running this branch did you ever encounter any of the new warn!() log statements that you added about blocked tasks being blocked again or unblocked tasks being unblocked again?

yea the shell doesn't yield after blocking itself leading to a couple of logs at startup. I've fixed it in #675 but the scope of that PR is much larger.

Issue #482, opened 2022-01-26
Remove `owning_ref` to avoid (unused) unsoundness


kevinaboos:
See: https://github.com/Kimundi/owning-ref-rs/pull/78

We don't currently use the unsound functions, but it's a good idea to patch this nonetheless (in the root `Cargo.toml`'s `[patch]` section).

kevinaboos:
Could also consider replacing `owning_ref` with https://crates.io/crates/reffers

kevinaboos:
I think we'll need to replace `owning_ref` entirely. Good candidates are:
* `self_cell`: good but a bit limited
* `ouroboros`: huge and perhaps too much build-time overhead
* `reffers`: not as well-tested as others

kevinaboos:
Note that all other RustSec advisory database warnings have been addressed. `owning_ref` is all that remains

kevinaboos:
As an alternative and better solution, we no longer support using `owning_ref` anywhere in Theseus. All of the crates that use it have now been removed; all that's left is to remove it from various lock implementations that offer compatibility with it by implementing `stable_deref_trait` (e.g., `MutexIrqSafe`, `MutexSleep` and similar).

PR #664, opened 2022-10-13
Use `BorrowedMappedPages` to replace the usage of `owning_ref` to stably-deref types overlaid atop `MappedPages`

kevinaboos:
We no longer need `owning_ref` (which is unsound, see #482) for this purpose with the `MappedPages` type. This also avoid needing to wrap a deref'd `MappedPages` object in a `Box`, which is a wasteful heap allocation pattern.


kevinaboos:
Note: this is not yet complete and is blocked on #648

kevinaboos:
Reminder to properly utilize the sealed trait for the marker structs that implement the `Mutability` trait

kevinaboos:
@tsoutsman requesting your review since we discussed the initial PR #648. I've extended and modified things quite a bit since that first PR. 

@Ramla-I requesting your review since this makes several changes to NIC drivers.

kernel/memory/src/paging/mapper.rs
```diff
@@ -919,100 +920,221 @@ pub fn mapper_from_current() -> Mapper {
...
+    ///
+    /// Upon failure, returns an error containing the unmodified `MappedPages` and a string
+    /// describing the error.
+    pub fn try_into_borrowed_slice(
+        mp: MappedPages,
+        byte_offset: usize,
+        length: usize,
+    ) -> Result<BorrowedSliceMappedPages<T>, (MappedPages, &'static str)> {
+        let borrowed_mp = Self {
+            ptr: match mp.as_slice::<T>(byte_offset, length) {
```
tsoutsman, line 1126: From the [`slice::from_raw_parts`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html) docs:
> The total size len * mem::size_of::<T>() of the slice must be no larger than isize::MAX. See the safety documentation of [pointer::offset](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset).

`as_slice` uses `usize::checked_mul` to compute `size_in_bytes`. We should ensure that it also fits in an `isize`.

I can't imagine a case where it wouldn't but we may as well if we're checking for overflows.

kernel/memory/src/paging/mapper.rs
```diff
@@ -919,100 +920,221 @@ pub fn mapper_from_current() -> Mapper {
...
         mut mp: MappedPages,
-        offset: usize,
-    ) -> Result<BorrowedMutMappedPages<T>, (MappedPages, &'static str)> {
-        let borrowed_mp = BorrowedMutMappedPages {
-            ptr: match mp.as_type_mut::<T>(offset) {
+        byte_offset: usize,
+        length: usize,
+    ) -> Result<Self, (MappedPages, &'static str)> {
+        let borrowed_mp = Self {
+            ptr: match mp.as_slice_mut::<T>(byte_offset, length) {
```
tsoutsman, line 1153: Same as above but for `as_slice_mut`.

tsoutsman (changes requested):
Aside from some small problems, it looks good.

Can we shorten the function names to just `new` or something rather? The structs already have long names and I think the long initialiser names are unnecessary.

The structs should implement `AsRef`, `Borrow`, and `BorrowMut` (obviously only the mutable variants). However, for the `Borrow` and `BorrowMut` implementations to be correct, we'll have to forward the `Eq`, `Hash`, and `Ord` implementations to the `T` ( or `[T]`). `Rc`, `Box`, etc. do this.

As an alternative API:
```rust
struct BorrowedMappedPages<Ref>
where
    Ref: Reference,
{
    ptr: Unique<Ref::Target>,
    mp: MappedPages,
    _marker: PhantomData<Ref>,
}

impl<T> BorrowedMappedPages<&T>
where
    T: zerocopy::FromBytes,
{
    pub fn new() -> Self {
        todo!();
    }
}

impl<T> BorrowedMappedPages<&mut T>
where
    T: zerocopy::FromBytes,
{
    pub fn new() -> Self {
        todo!();
    }
}

impl<T> BorrowedMappedPages<&[T]>
where
    T: zerocopy::FromBytes,
{
    pub fn new() -> Self {
        todo!();
    }
}

impl<T> BorrowedMappedPages<&mut [T]>
where
    T: zerocopy::FromBytes,
{
    pub fn new() -> Self {
        todo!();
    }
}
pub trait Reference: private::Sealed {
    type Target;
}

impl<T> Reference for &T
where
    T: zerocopy::FromBytes,
{
    type Target = T;
}
impl<T> Reference for &mut T
where
    T: zerocopy::FromBytes,
{
    type Target = T;
}
impl<T> Reference for &[T]
where
    T: zerocopy::FromBytes,
{
    type Target = T;
}
impl<T> Reference for &mut [T]
where
    T: zerocopy::FromBytes,
{
    type Target = T;
}

impl<T> private::Sealed for &T where T: zerocopy::FromBytes {}
impl<T> private::Sealed for &mut T where T: zerocopy::FromBytes {}
impl<T> private::Sealed for &[T] where T: zerocopy::FromBytes {}
impl<T> private::Sealed for &mut [T] where T: zerocopy::FromBytes {}

mod private {
    pub trait Sealed {}
}
```
I don't like it because you can't generalise over mutability and it's confusing to have the same struct represent both references and slices, but I figured I might as well throw it out there for _inspiration_ 💡. ✨. 💭. 🤩. 💫.

kernel/memory/src/paging/mapper.rs
kevinaboos: thanks. Resolved by #674

kernel/memory/src/paging/mapper.rs
kevinaboos: thanks. Resolved by #674

kevinaboos:
> Can we shorten the function names to just `new` or something rather? The structs already have long names and I think the long initialiser names are unnecessary.

Yep, good idea. I changed them to `from()` and `from_mut()`, and I added some convenience methods that operate on `MappedPages`:
* `into_borrowed()`
* `into_borrowed_mut()`
* `into_borrowed_slice()`
* `into_borrowed_slice_mut()`

Those function names are a bit long but it's unavoidable since we need to disambiguate between the 8 related variants, the four `as*` functions and the four `into_borrrowed*` functions.

> The structs should implement `AsRef`, `Borrow`, and `BorrowMut` (obviously only the mutable variants). However, for the `Borrow` and `BorrowMut` implementations to be correct, we'll have to forward the `Eq`, `Hash`, and `Ord` implementations to the `T` ( or `[T]`). `Rc`, `Box`, etc. do this.

done.

tsoutsman (approved):

PR #665, opened 2022-10-17
Removed wildcard imports from memory & paging code

NathanRoyer:
Tiny PR to make some wildcard imports explicit in the `memory` crate.

Some of these imports were only required by other crates so I initially missed them.

### DIFF
--- a/kernel/memory/src/lib.rs
+++ b/kernel/memory/src/lib.rs
@@ -32,9 +32,7 @@ mod paging;
 #[cfg(mapper_spillful)]
 pub mod paging;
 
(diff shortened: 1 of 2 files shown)


kevinaboos (changes requested):
This changeset is okay with me, but would you mind combining the `use` import statements into one per crate/module hierarchy instead of one item per line?

We'll eventually do this in a `rustfmt` step, but until we customize the configuration of `rustfmt` i'd like to format it in this way.

kevinaboos (approved):

PR #666, opened 2022-10-17
Handle white space and | in shell

### LINKED ISSUE #567: [applications/shell] Shell panics in Shell::eval_cmdline().
https://github.com/theseus-os/Theseus/blob/585ab93e0505f9ac8385764ed4353c8271b80d83/applications/shell/src/lib.rs#L688-L721

```Rust
let command = args.remove(0); 
```
This panics if args.len()==0.
So shell input(cmdline) like
```
|
```
and

``` 
X | 
```
  cause panic at args.remove().

I think the priority for this is low for now.
So I just leave an issue here.

ouz-a:
Fixes #567

### DIFF
--- a/applications/shell/src/lib.rs
+++ b/applications/shell/src/lib.rs
@@ -684,6 +684,12 @@ impl Shell {
         let cmdline = self.cmdline.clone();
         let mut task_refs = Vec::new();
 
+        if cmdline.trim().is_empty(){
+            return Err(AppErr::NotFound((cmdline)))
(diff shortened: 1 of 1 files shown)


kevinaboos:
Thanks for the PR @ouz-a! I made a few minor changes, can you confirm that everything is still correct in your view? Then I will merge it in.

ouz-a:
it looks good to me :+1:

kevinaboos (approved):

Issue #373, opened 2021-04-15
PS2 keyboard doesn't work properly in default VirtualBox environment


kevinaboos:
There's probably an issue with the `ps2` crate and how its initialization routine works.

Relevant log excerpts:
```
[I] kernel/ps2/src/lib.rs:92: ps2 controller test pass!!!
[I] kernel/ps2/src/lib.rs:101: first ps2 port test pass!!!
[I] kernel/ps2/src/lib.rs:114: first ps2 port interrupt enabled
[W] kernel/ps2/src/lib.rs:121: first ps2 data port disabled
[I] kernel/keyboard/src/lib.rs:48: MF2 Keyboard with translator enabled in PS/2 Controller
[I] kernel/ps2/src/lib.rs:160: ps2 controller test pass!!!
[I] kernel/ps2/src/lib.rs:168: second ps2 port test pass!!!
[I] kernel/ps2/src/lib.rs:180: second ps2 port interrupt enabled
[I] kernel/ps2/src/lib.rs:185: second ps2 data port enabled
[I] kernel/mouse/src/lib.rs:39: the initial mouse ID is: 4
```

with this line being the key:
```
[W] kernel/ps2/src/lib.rs:121: first ps2 data port disabled
```

kevinaboos:
Fixed in #884

PR #667, opened 2022-10-17
PS/2 Overhaul

hecatia-elegua:
No more magic values, less errors, more docs, simpler functions, structs for all the response-bits, so no bit ops.
In general, saner code.

Some notes/questions:
- error handling is still not very good, this seems to be because the Error trait is not no_std and some other problems, so for now we have strings, which ~~I would still like to improve a bit~~ I have improved a bit
- why is there a pattern of `.map_err(|_| { warn!("..."); "disable mouse streaming failed"})`? it might make sense because it shows a kind of backtrace-like route up the error stack, but still seems/feels bad? I have read the recommended [reading](https://blog.burntsushi.net/rust-error-handling/), but maybe I should look into some error-handling/logging crates (though I dislike the topic) -> some map_err calls were obsolete, but I'm still not comfortable with some of the error handling, again because we can't impl Error and I wonder in what cases we should just shutdown the system or reset the device/s
- how should I improve my commit messages?
- I chose modular-bitfield because it comes closest to what I would want, [this](https://internals.rust-lang.org/t/pre-rfc-arbitrary-bit-width-integers/15603), or [this](https://therealprof.github.io/blog/roadmap-2021-arbitrary-size-primitives/) -> feedback: I think this is more approachable than bitflags and some others
- not sure if I did `KBD_MODIFIERS` right -> there is a note about "remove unsafe static mut" on it, not sure what to do about it (access is unsafe but there also is a note on access like "it should never race"
- there's a lot of code in the ps2 crate now, any thoughts on where/how to split it up? some functions could also be enum/struct methods instead -> I wasted a lot of time scrolling up and down, still not sure how to divide it up (mod mouse, keyboard, controller?)

Feel free to critique!

### DIFF
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -147,15 +147,15 @@ fn handle_keyboard_input(scan_code: u8, extended: bool) -> Result<(), &'static s
         // The "*Lock" keys are toggled only upon being pressed, not when released.
         x if x == Keycode::CapsLock as u8 => {
             modifiers.toggle(KeyboardModifiers::CAPS_LOCK);
-            set_keyboard_led(&modifiers);
+            set_keyboard_led(modifiers);
         }
         x if x == Keycode::ScrollLock as u8 => {
             modifiers.toggle(KeyboardModifiers::SCROLL_LOCK);
-            set_keyboard_led(&modifiers);
+            set_keyboard_led(modifiers);
         }
         x if x == Keycode::NumLock as u8 => {
             modifiers.toggle(KeyboardModifiers::NUM_LOCK);
-            set_keyboard_led(&modifiers);
+            set_keyboard_led(modifiers);
         }
 
         _ => { } // do nothing
@@ -171,7 +171,7 @@ fn handle_keyboard_input(scan_code: u8, extended: bool) -> Result<(), &'static s
     };
 
     if let Some(keycode) = Keycode::from_scancode(adjusted_scan_code) {
-        let event = Event::new_keyboard_event(KeyEvent::new(keycode, action, modifiers.clone()));
+        let event = Event::new_keyboard_event(KeyEvent::new(keycode, action, *modifiers));
         if let Some(producer) = KEYBOARD_PRODUCER.get() {
             producer.push(event).map_err(|_e| "keyboard input queue is full")
         } else {
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -10,7 +10,7 @@ use once_cell::unsync::Lazy;
 use spin::Once;
 use mpmc::Queue;
 use event_types::Event;
-use ps2::{init_ps2_port1, test_ps2_port1, keyboard_detect, KeyboardType, read_scancode, LEDState};
+use ps2::{init_ps2_port1, test_ps2_port1, keyboard_detect, KeyboardType, read_scancode, LEDState, keyboard_scancode_set, ScancodeSet};
 use x86_64::structures::idt::InterruptStackFrame;
 
 /// The first PS2 port for the keyboard is connected directly to IRQ 1.
@@ -44,7 +44,7 @@ pub fn init(keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
         }
     }
 
-    // TODO: set keyboard to scancode set 1, since that's the only one we support (?)
+    keyboard_scancode_set(ScancodeSet::Set2)?;
 
     // Register the interrupt handler
     interrupts::register_interrupt(PS2_KEYBOARD_IRQ, ps2_keyboard_handler).map_err(|e| {
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -5,7 +5,7 @@
 
 use core::sync::atomic::{AtomicBool, Ordering};
 use keycodes_ascii::{Keycode, KeyboardModifiers, KEY_RELEASED_OFFSET, KeyAction, KeyEvent};
-use log::{info, error, warn};
+use log::{error, warn, debug};
 use once_cell::unsync::Lazy;
 use spin::Once;
 use mpmc::Queue;
@@ -32,14 +32,15 @@ pub fn init(keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
     init_ps2_port1();
     // Test the first port.
     test_ps2_port1()?;
+
     // Detect which kind of keyboard is connected.
     // TODO: actually do something with the keyboard type.
     match keyboard_detect() {
-        Ok(KeyboardType::AncientATKeyboard) => info!("Ancient AT Keyboard with translator enabled in the PS/2 Controller"),
-        Ok(KeyboardType::MF2Keyboard) => info!("MF2Keyboard"),
-        Ok(KeyboardType::MF2KeyboardWithPSControllerTranslator) => info!("MF2 Keyboard with translator enabled in PS/2 Controller"),
+        Ok(KeyboardType::AncientATKeyboard) => debug!("The PS/2 keyboard type is: Ancient AT Keyboard with translator enabled in the PS/2 Controller"),
+        Ok(KeyboardType::MF2Keyboard) => debug!("The PS/2 keyboard type is: MF2Keyboard"),
+        Ok(KeyboardType::MF2KeyboardWithPSControllerTranslator) => debug!("The PS/2 keyboard type is: MF2 Keyboard with translator enabled in PS/2 Controller"),
         Err(e) => {
-            error!("Failed to detect the Ps2 keyboard type, error: {} ", e);
+            error!("Failed to detect the PS/2 keyboard type: {e}");
             return Err("Failed to detect the PS2 keyboard type");
         }
     }
@@ -48,10 +49,8 @@ pub fn init(keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
 
     // Register the interrupt handler
     interrupts::register_interrupt(PS2_KEYBOARD_IRQ, ps2_keyboard_handler).map_err(|e| {
-        error!("PS2 keyboard IRQ {:#X} was already in use by handler {:#X}! Sharing IRQs is currently unsupported.", 
-            PS2_KEYBOARD_IRQ, e,
-        );
-        "PS2 keyboard IRQ was already in use! Sharing IRQs is currently unsupported."
+        error!("PS/2 keyboard IRQ {PS2_KEYBOARD_IRQ:#X} was already in use by handler {e:#X}! Sharing IRQs is currently unsupported.");
+        "PS/2 keyboard IRQ was already in use! Sharing IRQs is currently unsupported."
     })?;
 
     // Final step: set the producer end of the keyboard event queue.
@@ -70,12 +69,12 @@ extern "x86-interrupt" fn ps2_keyboard_handler(_stack_frame: InterruptStackFrame
     // 0xE0 indicates an extended scancode, so we must wait for the next interrupt to get the actual scancode
     if scan_code == 0xE0 {
         if extended {
-            error!("PS2_PORT interrupt: got two extended scancodes (0xE0) in a row! Shouldn't happen.");
+            error!("ps2_keyboard_handler: got two extended scancodes (0xE0) in a row! Shouldn't happen.");
         }
         // mark it true for the next interrupt
         EXTENDED_SCANCODE.store(true, Ordering::SeqCst);
     } else if scan_code == 0xE1 {
-        error!("PAUSE/BREAK key pressed ... ignoring it!");
+        error!("ps2_keyboard_handler: PAUSE/BREAK key pressed ... ignoring it!");
         // TODO: handle this, it's a 6-byte sequence (over the next 5 interrupts)
         EXTENDED_SCANCODE.store(true, Ordering::SeqCst);
     } else { // a regular scancode, go ahead and handle it
@@ -183,13 +182,11 @@ fn handle_keyboard_input(scan_code: u8, extended: bool) -> Result<(), &'static s
         if let Some(producer) = KEYBOARD_PRODUCER.get() {
             producer.push(event).map_err(|_| "keyboard input queue is full")
         } else {
-            warn!("handle_keyboard_input(): KEYBOARD_PRODUCER wasn't yet initialized, dropping keyboard event {:?}.", event);
+            warn!("handle_keyboard_input(): KEYBOARD_PRODUCER wasn't yet initialized, dropping keyboard event {event:?}.");
             Err("keyboard event queue not ready")
         }
     } else {
-        error!("handle_keyboard_input(): Unknown scancode: {:?}, adjusted scancode: {:?}",
-            scan_code, adjusted_scan_code
-        );
+        error!("handle_keyboard_input(): Unknown scancode: {scan_code:?}, adjusted scancode: {adjusted_scan_code:?}");
         Err("unknown keyboard scancode")
     }
 }
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -69,43 +69,31 @@ pub fn init(keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
 
 /// The interrupt handler for a ps2-connected keyboard, registered at IRQ 0x21.
 extern "x86-interrupt" fn ps2_keyboard_handler(_stack_frame: InterruptStackFrame) {
-    // see this: https://forum.osdev.org/viewtopic.php?f=1&t=32655
+    // see this: [How do 0xE0 keyboard scancodes (PS/2) work?](https://forum.osdev.org/viewtopic.php?f=1&t=32655)
     static EXTENDED_SCANCODE: AtomicBool = AtomicBool::new(false);
 
-    let indicator = ps2_status_register();
+    let scan_code = read_scancode();
+    let extended = EXTENDED_SCANCODE.load(Ordering::SeqCst);
 
-    // whether there is any data on the port 0x60
-    if indicator & 0x01 == 0x01 {
-        // Skip this if the PS2 event came from the mouse, not the keyboard
-        if indicator & 0x20 != 0x20 {
-            // in this interrupt, we must read the PS2_PORT scancode register before acknowledging the interrupt.
-            let scan_code = read_scancode();
-            // trace!("PS2_PORT interrupt: raw scan_code {:#X}", scan_code);
-
-
-            let extended = EXTENDED_SCANCODE.load(Ordering::SeqCst);
-
-            // 0xE0 indicates an extended scancode, so we must wait for the next interrupt to get the actual scancode
-            if scan_code == 0xE0 {
-                if extended {
-                    error!("PS2_PORT interrupt: got two extended scancodes (0xE0) in a row! Shouldn't happen.");
-                }
-                // mark it true for the next interrupt
-                EXTENDED_SCANCODE.store(true, Ordering::SeqCst);
-            } else if scan_code == 0xE1 {
-                error!("PAUSE/BREAK key pressed ... ignoring it!");
-                // TODO: handle this, it's a 6-byte sequence (over the next 5 interrupts)
-                EXTENDED_SCANCODE.store(true, Ordering::SeqCst);
-            } else { // a regular scancode, go ahead and handle it
-                // if the previous interrupt's scan_code was an extended scan_code, then this one is not
-                if extended {
-                    EXTENDED_SCANCODE.store(false, Ordering::SeqCst);
-                }
-                if scan_code != 0 {  // a scan code of zero is a PS2_PORT error that we can ignore
-                    if let Err(e) = handle_keyboard_input(scan_code, extended) {
-                        error!("ps2_keyboard_handler: error handling PS2_PORT input: {:?}", e);
-                    }
-                }
+    // 0xE0 indicates an extended scancode, so we must wait for the next interrupt to get the actual scancode
+    if scan_code == 0xE0 {
+        if extended {
+            error!("PS2_PORT interrupt: got two extended scancodes (0xE0) in a row! Shouldn't happen.");
+        }
+        // mark it true for the next interrupt
+        EXTENDED_SCANCODE.store(true, Ordering::SeqCst);
+    } else if scan_code == 0xE1 {
+        error!("PAUSE/BREAK key pressed ... ignoring it!");
+        // TODO: handle this, it's a 6-byte sequence (over the next 5 interrupts)
+        EXTENDED_SCANCODE.store(true, Ordering::SeqCst);
+    } else { // a regular scancode, go ahead and handle it
+        // if the previous interrupt's scan_code was an extended scan_code, then this one is not
+        if extended {
+            EXTENDED_SCANCODE.store(false, Ordering::SeqCst);
+        }
+        if scan_code != 0 {  // a scan code of zero is a PS2_PORT error that we can ignore
+            if let Err(e) = handle_keyboard_input(scan_code, extended) {
+                error!("ps2_keyboard_handler: error handling PS2_PORT input: {:?}", e);
             }
         }
     }
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -60,7 +60,8 @@ pub fn init(keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
 
 /// The interrupt handler for a ps2-connected keyboard, registered at IRQ 0x21.
 extern "x86-interrupt" fn ps2_keyboard_handler(_stack_frame: InterruptStackFrame) {
-    // see this: [How do 0xE0 keyboard scancodes (PS/2) work?](https://forum.osdev.org/viewtopic.php?f=1&t=32655)
+    // Some of the scancodes are "extended", which means they generate two different interrupts,
+    // the first handling the E0 byte, the second handling their second byte.
     static EXTENDED_SCANCODE: AtomicBool = AtomicBool::new(false);
 
     let scan_code = read_scancode();
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -6,6 +6,7 @@
 use core::sync::atomic::{AtomicBool, Ordering};
 use keycodes_ascii::{Keycode, KeyboardModifiers, KEY_RELEASED_OFFSET, KeyAction, KeyEvent};
 use log::{info, error, warn};
+use once_cell::unsync::Lazy;
 use spin::Once;
 use mpmc::Queue;
 use event_types::Event;
@@ -16,8 +17,8 @@ use x86_64::structures::idt::InterruptStackFrame;
 /// Because we perform the typical PIC remapping, the remapped IRQ vector number is 0x21.
 const PS2_KEYBOARD_IRQ: u8 = interrupts::IRQ_BASE_OFFSET + 0x1;
 
-// TODO: avoid unsafe static mut using the following: https://www.reddit.com/r/rust/comments/1wvxcn/lazily_initialized_statics/cf61im5/
-static mut KBD_MODIFIERS: KeyboardModifiers = KeyboardModifiers::new();
+// TODO: avoid unsafe static mut
+static mut KBD_MODIFIERS: Lazy<KeyboardModifiers> = Lazy::new(KeyboardModifiers::new);
 
 static KEYBOARD_PRODUCER: Once<Queue<Event>> = Once::new();
 
@@ -150,7 +151,7 @@ fn handle_keyboard_input(scan_code: u8, extended: bool) -> Result<(), &'static s
     };
 
     if let Some(keycode) = Keycode::from_scancode(adjusted_scan_code) {
-        let event = Event::new_keyboard_event(KeyEvent::new(keycode, action, *modifiers));
+        let event = Event::new_keyboard_event(KeyEvent::new(keycode, action, **modifiers));
         if let Some(producer) = KEYBOARD_PRODUCER.get() {
             producer.push(event).map_err(|_e| "keyboard input queue is full")
         } else {
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -45,7 +45,8 @@ pub fn init(keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
         }
     }
 
-    keyboard_scancode_set(ScancodeSet::Set2)?;
+    // TODO: figure out what we should do
+    // keyboard_scancode_set(ScancodeSet::Set2)?;
 
     // Register the interrupt handler
     interrupts::register_interrupt(PS2_KEYBOARD_IRQ, ps2_keyboard_handler).map_err(|e| {
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -84,8 +84,9 @@ extern "x86-interrupt" fn ps2_keyboard_handler(_stack_frame: InterruptStackFrame
         if extended {
             EXTENDED_SCANCODE.store(false, Ordering::SeqCst);
         }
-        // a scan code of zero is a PS2_PORT error that we can ignore
-        if scan_code != 0 {
+        // a scan code of zero is a PS2_PORT error that we can ignore,
+        // a scan code of 0xFA is a command ACK response, already handled in polling (when sending a command, see ps2 crate)
+        if scan_code != 0 && scan_code != 0xFA {
             if let Err(e) = handle_keyboard_input(scan_code, extended) {
                 error!("ps2_keyboard_handler: error handling PS2_PORT input: {e:?}");
             }
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -10,7 +10,7 @@ use once_cell::unsync::Lazy;
 use spin::Once;
 use mpmc::Queue;
 use event_types::Event;
-use ps2::{init_ps2_port1, test_ps2_port1, keyboard_detect, KeyboardType, read_scancode, LEDState, keyboard_scancode_set, ScancodeSet};
+use ps2::{init_ps2_port1, test_ps2_port1, keyboard_detect, KeyboardType, read_scancode, LEDState};
 use x86_64::structures::idt::InterruptStackFrame;
 
 /// The first PS/2 port for the keyboard is connected directly to IRQ 1.
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -9,7 +9,7 @@ use log::{info, error, warn};
 use spin::Once;
 use mpmc::Queue;
 use event_types::Event;
-use ps2::{init_ps2_port1, test_ps2_port1, keyboard_led, keyboard_detect, KeyboardType, read_scancode, ps2_status_register};
+use ps2::{init_ps2_port1, test_ps2_port1, keyboard_led, keyboard_detect, KeyboardType, read_scancode};
 use x86_64::structures::idt::InterruptStackFrame;
 
 /// The first PS2 port for the keyboard is connected directly to IRQ 1.
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -9,7 +9,7 @@ use log::{info, error, warn};
 use spin::Once;
 use mpmc::Queue;
 use event_types::Event;
-use ps2::{init_ps2_port1, test_ps2_port1, keyboard_led, keyboard_detect, KeyboardType, read_scancode, LEDState};
+use ps2::{init_ps2_port1, test_ps2_port1, keyboard_detect, KeyboardType, read_scancode, LEDState};
 use x86_64::structures::idt::InterruptStackFrame;
 
 /// The first PS2 port for the keyboard is connected directly to IRQ 1.
@@ -169,10 +169,12 @@ fn handle_keyboard_input(scan_code: u8, extended: bool) -> Result<(), &'static s
 
 
 fn set_keyboard_led(modifiers: &KeyboardModifiers) {
-    keyboard_led(
+    if let Err(e) = ps2::set_keyboard_led(
         LEDState::new()
             .with_scroll_lock(modifiers.is_scroll_lock())
             .with_number_lock(modifiers.is_num_lock())
             .with_caps_lock(modifiers.is_caps_lock()),
-    );
+    ) {
+        error!("{e}");
+    }
 }
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -38,8 +38,7 @@ pub fn init(keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
     // Init the first ps2 port, which is used for the keyboard.
     init_ps2_port1();
     // Test the first port.
-    // TODO: return an error if this test fails.
-    test_ps2_port1();
+    test_ps2_port1()?;
     // Detect which kind of keyboard is connected.
     // TODO: actually do something with the keyboard type.
     match keyboard_detect() {
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -3,10 +3,9 @@
 #![no_std]
 #![feature(abi_x86_interrupt)]
 
-#[macro_use] extern crate log;
-
 use core::sync::atomic::{AtomicBool, Ordering};
 use keycodes_ascii::{Keycode, KeyboardModifiers, KEY_RELEASED_OFFSET, KeyAction, KeyEvent};
+use log::{info, error, warn};
 use spin::Once;
 use mpmc::Queue;
 use event_types::Event;
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -9,7 +9,7 @@ use log::{info, error, warn};
 use spin::Once;
 use mpmc::Queue;
 use event_types::Event;
-use ps2::{init_ps2_port1, test_ps2_port1, keyboard_led, keyboard_detect, KeyboardType};
+use ps2::{init_ps2_port1, test_ps2_port1, keyboard_led, keyboard_detect, KeyboardType, ps2_read_data, ps2_status_register};
 use x86_64::structures::idt::InterruptStackFrame;
 
 /// The first PS2 port for the keyboard is connected directly to IRQ 1.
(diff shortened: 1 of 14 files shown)


kevinaboos:
Thanks for this awesome PR @hecatia-elegua! Much appreciated.

I'm a bit pressed for time this week, but I did a quick review and things generally look quite good! I will do a more detailed review next week, but in the meantime feel free to finish up the various to-do items and clean up tasks that you had noted, since I definitely plan to merge this PR in. I will also answer your questions above when time permits.

hecatia-elegua:
This is done.
(or rather, I seriously need to stop :smile:)

kevinaboos:
> This is done. (or rather, I seriously need to stop smile)

haha, thanks so much, it's really an awesome contribution. I've been a bit backlogged but am slowly catching up; I will definitely do a full review this week. Much appreciated, and sorry for the delays!

kernel/keyboard/src/lib.rs
```diff
@@ -39,28 +31,26 @@ pub fn init(keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
...
+        Ok(KeyboardType::MF2KeyboardWithPSControllerTranslator) => debug!("The PS/2 keyboard type is: MF2 Keyboard with translator enabled in PS/2 Controller"),
         Err(e) => {
-            error!("Failed to detect the Ps2 keyboard type, error: {} ", e);
+            error!("Failed to detect the PS/2 keyboard type: {e}");
             return Err("Failed to detect the PS2 keyboard type");
         }
     }
 
-    // TODO: set keyboard to scancode set 1, since that's the only one we support (?)
+    keyboard_scancode_set(ScancodeSet::Set2)?;
```
kevinaboos: I'm vaguely familiar with scancode sets, but from what I can tell we were previously assuming set 1 (i think?). I don't mind forcing Set 2, but we should probably:
1. justify that decision -- is it the most common case? is it otherwise superior to Set 1? Citing something in a comment would be a good idea here.
2. If it fails, we just bail and return an error. Is there something else we ought to be doing, such as trying to use Set 1 as a fallback option?

kernel/keyboard/src/lib.rs
```diff
@@ -70,43 +60,31 @@ pub fn init(keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
 
 /// The interrupt handler for a ps2-connected keyboard, registered at IRQ 0x21.
 extern "x86-interrupt" fn ps2_keyboard_handler(_stack_frame: InterruptStackFrame) {
-    // see this: https://forum.osdev.org/viewtopic.php?f=1&t=32655
+    // see this: [How do 0xE0 keyboard scancodes (PS/2) work?](https://forum.osdev.org/viewtopic.php?f=1&t=32655)
```
kevinaboos: FYI, formatting like this only works in doc comments.
Also we can probably remove this link, I just looked at it again and it doesn't really offer any information. Instead, let's  replace it with a short comment describing what an extended scancode is and what it means (how it applies to the next scancode, etc) and how we detect it (i.e., `0xE0`)

kernel/mouse/src/lib.rs
```diff
@@ -31,136 +26,84 @@ pub fn init(mouse_queue_producer: Queue<Event>) -> Result<(), &'static str> {
...
 
     // Set Mouse ID to 4, and read it back to check that it worked.
-    let _e = set_mouse_id(4);
-    match check_mouse_id() {
-        Ok(id) => info!("the initial mouse ID is: {}", id),
-        Err(_e) => {
-            error!("Failed to read the initial PS2 mouse ID, error: {:?}", _e);
-            return Err("Failed to read the initial PS2 mouse ID");
+    if let Err(e) = set_mouse_id(MouseId::Four) {
+        error!("Failed to set the mouse id to four: {e}");
```
kevinaboos, line 36: I notice we just continue on if this error occurs; is this an error that can be ignored? If not, let's return an error. If so, is there a fallback option that we need to attempt, such as choosing another mouse ID?

kernel/mouse/src/lib.rs
```diff
@@ -31,136 +26,84 @@ pub fn init(mouse_queue_producer: Queue<Event>) -> Result<(), &'static str> {
...
-        }
+    let mouse_packet = read_mouse_packet();
+    if mouse_packet.x_overflow() || mouse_packet.y_overflow() {
+        //NOTE: overflow should maybe just be handled by using the max value on overflow, though it's apparently impossible to trigger it
+        //and often the hardware clamps it anyways (https://forum.osdev.org/viewtopic.php?f=1&t=31176)
+        error!("The overflow bits in the mouse data packet's first byte are set! Discarding the whole packet.");
+    } else if mouse_packet.always_one() != 1 {
+        // it's very likely that the PS/2 controller send us an [interrupt](https://wiki.osdev.org/%228042%22_PS/2_Controller#Interrupts),
+        // if it did, the whole 32 bits of the mouse_packet are zero (at least on my end)
+        // checking the ps/2 controller status register's mouse_output_buffer_full might work, but this current solution does also work
```
kevinaboos: so what is the actual current solution described here? Just ignoring this packet, sending an EOI, and returning from the interrupt handler? Let's clarify what the options are here, which one we decided to take, and why.

kernel/mouse/src/lib.rs
```diff
@@ -31,136 +26,84 @@ pub fn init(mouse_queue_producer: Queue<Event>) -> Result<(), &'static str> {
...
 
-/// return a Mouse Event according to the data
-fn handle_mouse_input(readdata: u32) -> Result<(), &'static str> {
-    let action = unsafe { &mut BUTTON_ACT };
-    let mmove = unsafe { &mut MOUSE_MOVE };
-    let dis = unsafe { &mut DISPLACEMENT };
+/// enqueue a Mouse Event according to the data
+fn handle_mouse_input(mouse_packet: MousePacketBits4) -> Result<(), &'static str> {
+    let action = button_action_from(&mouse_packet);
+    let mmove = mouse_movement_from(&mouse_packet);
```
kevinaboos: nice job with this! very clean

kernel/mouse/src/lib.rs
```diff
@@ -31,136 +26,84 @@ pub fn init(mouse_queue_producer: Queue<Event>) -> Result<(), &'static str> {
...
+// NOTE: This crate depends on mouse_data and ps2, so I'm doing this here
+fn button_action_from(mouse_packet: &MousePacketBits4) -> ButtonAction {
+    ButtonAction::new(
+        mouse_packet.button_left(),
+        mouse_packet.button_right(),
+        mouse_packet.button_middle(),
+        mouse_packet.button_4(),
+        mouse_packet.button_5(),
+    )
+}
```
kevinaboos: For these you may want to implement the `core::convert::Into` or `From` traits, in order to be as canonically Rust-y, if possible.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+use DeviceToHostResponse::*;
+
+/// Port used by PS/2 Controller and devices
+static PS2_DATA_PORT: Mutex<Port<u8>> = Mutex::new(Port::new(0x60));
+/// Port used to send commands to and receive status from the PS/2 Controller
+static PS2_COMMAND_AND_STATUS_PORT: Mutex<Port<u8>> = Mutex::new(Port::new(0x64));
+
+// https://wiki.osdev.org/%228042%22_PS/2_Controller#PS.2F2_Controller_Commands
+/// Command types which can be sent to the PS/2 Controller
+enum HostToControllerCommand {
```
kevinaboos, line 25: Quite a few of these are commented out. Would you mind adding a comment describing why they're commented out? e.g., are they not supported, untested, deprecated, etc

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
-pub fn ps2_clean_buffer() -> Vec<u8> {
-    let mut buffer_data = Vec::new();
-    loop {
-        // if no more data in the output buffer
-        if ps2_status_register() & 0x01 == 0 {
-            break;
-        } else {
-            buffer_data.push(ps2_read_data());
-        }
+/// Write a command to the PS/2 command port/register
```
kevinaboos, line 109: formatting nit -- if you insert a newline after this, it gets formatted properly by rustdoc. Kindly check this (and all of the other excellent docs you added) by running `make view-docs` to make sure they're displayed nicely and the doclinks are working correctly.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
 }
 
-/// write command to the command ps2 port (0x64)
-pub fn ps2_write_command(value: u8) {
-    unsafe {
-        PS2_COMMAND_PORT.lock().write(value);
+#[allow(non_camel_case_types)]
+type u1 = B1;
+#[allow(non_camel_case_types)]
+type bool1 = bool;
```
kevinaboos: small nit: it'd be a bit easier to read the code if we did stick to Rust's expected camel case formatting for typedefs, maybe `U1` or `SingleBit` or `OneBit` or something for `B1` or and `Bool1Bit` or `OneBitBool` for `bool1`.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+
+// see https://wiki.osdev.org/%228042%22_PS/2_Controller#Status_Register
+// and https://users.utcluj.ro/~baruch/sie/labor/PS2/PS-2_Keyboard_Interface.htm
+// `pub` because of `dead_code` warnings, see https://github.com/Robbepop/modular-bitfield/issues/56
+#[bitfield(bits = 8)]
+pub struct ControllerToHostStatus {
```
kevinaboos, line 124: for this and all other structs that must have a specific size, it'd be great to use the `static_assertions` crate's `const_assert` macros to ensure that the size of this struct is what we expect. See #658 for some examples on how we do this elsewhere.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+    #[allow(dead_code)]
+    parity_error: bool1,
+}
+
+/// Read the PS/2 status port/register
+fn status_register() -> ControllerToHostStatus {
+    ControllerToHostStatus::from_bytes([PS2_COMMAND_AND_STATUS_PORT.lock().read()])
+}
+
+//NOTE: if we are _not_ in an interrupt handler, it might be necessary to check for `output_buffer_full` first
```
kevinaboos: i think this NOTE (and probably a lot of other NOTEs) should be included as doc comments, e.g., something like:
```rust
/// Read data from the PS/2 data port.
///
/// NOTE: this is expected to be called from within a PS/2 device's interrupt handler.
/// If not, it may be necessary to check for `output_buffer_full` being set first.
fn read_data() -> u8 {
    PS2_DATA_PORT.lock().read()
}
```

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+                    }
+                    HostToDevice::Mouse(value) => {
+                        match value {
+                            HostToMouseCommandOrData::MouseCommand(c) => c as u8,
+                            HostToMouseCommandOrData::MouseResolution(r) => r as u8,
+                            HostToMouseCommandOrData::SampleRate(s) => s as u8,
+                        }
+                    }
+                }
+            }
```
kevinaboos: Can reduce indentation and improve legibility like so:

```suggestion
            WritableData::Configuration(value) => value.into_bytes()[0],
            WritableData::HostToDevice(value) => match value {
                HostToDevice::Keyboard(value) => match value {
                    HostToKeyboardCommandOrData::KeyboardCommand(c) => c as u8,
                    HostToKeyboardCommandOrData::LEDState(l) => l.into_bytes()[0],
                    HostToKeyboardCommandOrData::ScancodeSet(s) => s as u8,
                }
                HostToDevice::Mouse(value) => match value {
                        HostToMouseCommandOrData::MouseCommand(c) => c as u8,
                        HostToMouseCommandOrData::MouseResolution(r) => r as u8,
                        HostToMouseCommandOrData::SampleRate(s) => s as u8,
                    }
                }
            }
```

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+enum WritableData {
+    Configuration(ControllerConfigurationByte),
+    HostToDevice(HostToDevice)
+}
+
+// NOTE: could also use a macro / reminds me of serde
+impl From<WritableData> for u8 {
+    fn from(value: WritableData) -> Self {
+        match value {
+            WritableData::Configuration(value) => value.into_bytes()[0],
```
kevinaboos: is there a better way to do this without converting it into a byte array and then indexing that array? This introduces panic checking branches, which is quite heavyweight for such a trivial conversion.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+            WritableData::HostToDevice(value) => {
+                match value {
+                    HostToDevice::Keyboard(value) => {
+                        match value {
+                            HostToKeyboardCommandOrData::KeyboardCommand(c) => c as u8,
+                            HostToKeyboardCommandOrData::LEDState(l) => l.into_bytes()[0],
```
kevinaboos: is there a better way to do this without converting it into a byte array and then indexing that array? This introduces panic checking branches, which is quite heavyweight for such a trivial conversion.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+        }
     }
 }
 
-/// read the ps2 status register
-pub fn ps2_status_register() -> u8 {
-    PS2_COMMAND_PORT.lock().read()
+//NOTE: it might be necessary to check for `!input_buffer_full` first
+/// Write data to the PS/2 data port.
+/// This is the lowest level interface to the port.
```
kevinaboos: same comment as above: move this `NOTE` into the doc comments

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+    One,
+    Two,
+}
 
-        // set the new configuration
-        let new_config = (old_config & 0xEF) | 0x01;
-        ps2_write_config(new_config);
+// see https://wiki.osdev.org/%228042%22_PS/2_Controller#Initialising_the_PS.2F2_Controller
+fn init_ps2_port(port: PS2Port) {
+    // Step 3: Disable Devices
```
kevinaboos, line 273: probably should mention why we skip step 1 (obvious, no USB support yet) and step 2 (which I'm not too sure about, but we do actually support the FADT so we could technically check this first -- maybe in a future PR)

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+        }
     }
+    write_config(config);
+
+    // Step 9: Enable Devices
+    write_command(EnablePort1);
+    match port {
+        PS2Port::Two => write_command(EnablePort2),
+        PS2Port::One => (),
+    };
```
kevinaboos: can you add a comment explaining why we unconditionally enable port 1 but only enable port 2 when initializing port 2? Seems a bit strange, and a future reader of this code may interpret it as an accidental oversight

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+}
 
-    ps2_write_command(0xAE);
+/// Clean the [PS2_DATA_PORT] output buffer, skipping the [ControllerToHostStatus] `output_buffer_full` check
+fn flush_output_buffer() {
+    //NOTE(hecatia): on my end, this is always 250 for port 1 and 65 for port 2, even if read multiple times
+    debug!("ps2::flush_output_buffer: {}", read_data());
```
kevinaboos: is this note something we should be aware of? I'm not sure what to think when I read it, is it a to-do item or a form of unexpected behavior?

Also probably should remove this debug statement since it's quite minor and seems to be unimportant (?)

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+    test_ps2_port(PS2Port::One)
+}
+
+// test the second ps2 data port
+pub fn test_ps2_port2() -> Result<(), &'static str> {
+    test_ps2_port(PS2Port::Two)
+}
+
+// see https://wiki.osdev.org/%228042%22_PS/2_Controller#Initialising_the_PS.2F2_Controller
+fn test_ps2_port(port: PS2Port) -> Result<(), &'static str> {
```
kevinaboos, line 321: also really well done here! super nice to read

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+        PS2Port::One => config.port1_interrupt_enabled(),
+        PS2Port::Two => config.port2_interrupt_enabled(),
+    };
+    let clock_enabled = match port {
+        PS2Port::One => !config.port1_clock_disabled(),
+        PS2Port::Two => !config.port2_clock_disabled(),
+    };
+
+    if port_interrupt_enabled {
+        debug!("ps2 port {} interrupt enabled", port as u8 + 1)
```
kevinaboos: tiny nit: sometimes we use "ps2" and sometimes we use "PS/2". Let's make sure everything is consistent so it makes it easy to grep the code, the entire log, and error msgs.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
-    //disable PS2 ports first
-    {
-        ps2_write_command(0xA7);
-        ps2_write_command(0xAD);
-    }
+#[derive(TryFromPrimitive)]
+#[repr(u8)]
+enum ControllerTestResult {
+    Passed = 0x55,
+    //NOTE: can just be removed
```
kevinaboos: why can this be removed? If so, should we just remove it now?

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+pub enum DeviceToHostResponse {
+    KeyDetectionErrorOrInternalBufferOverrun1 = 0x00,
+    SelfTestPassed = 0xAA, //sent after "0xFF (reset)" command or keyboard power up
+    ResponseToEcho = 0xEE,
+    Acknowledge = 0xFA,
+    SelfTestFailed1 = 0xFC, //sent after "0xFF (reset)" command or keyboard power up
+    SelfTestFailed2 = 0xFD, //sent after "0xFF (reset)" command or keyboard power up
+    ResendCommand = 0xFE,
+    KeyDetectionErrorOrInternalBufferOverrun2 = 0xFF,
+}
```
kevinaboos: all of these comments would be better off as doc comments such that rustdoc can display them, otherwise they're invisible to developers and folks browsing the docs. Same with the other struct/enum definitions herein.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+
+// https://wiki.osdev.org/%228042%22_PS/2_Controller#Sending_Bytes_To_Device.2Fs
+// and https://wiki.osdev.org/PS/2_Mouse#Set_Sample_Rate_Example
+// AND https://wiki.osdev.org/%228042%22_PS/2_Controller#Polling, which still has some _maybe relevant information_:
+// We might still need to disable one of the devices every time we send so we can read_data reliably.
+// As it currently works, it might only be a problem for older devices, so I won't overoptimize.
+/// write data to the data port and return the response
+fn send(value: HostToDevice) -> Result<DeviceToHostResponse, &'static str> {
+    const VERY_ARBITRARY_TIMEOUT_VALUE: u8 = 3;
```
kevinaboos: :rofl:

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
-                if response == 0xFA {
-                    return Ok(());
-                } else if response == 0xfe {
-                    return Err("Please resend the command to the keyboard");
-                }
-            }
-            Err("command is not acknowledged")
+/// write command to the keyboard and return the result
+fn command_to_keyboard(value: HostToKeyboardCommandOrData) -> Result<(), &'static str> {
+    for _ in 0..3 {
```
kevinaboos: why do we try 3 times? i assume this is similar to the hilarious "VERY_ARBITRARY..." thing above?

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
-    //    info!("mouse command is responded !!!!");
-    Ok(())
+    Err("mouse doesn't support the command or there has been a hardware failure")
 }
 
-/// write data to the second ps2 output buffer
-pub fn write_to_second_output_buffer(value: u8) {
-    ps2_write_command(0xD3);
-    unsafe { PS2_PORT.lock().write(value) };
+//TODO: different MousePacketBits for different MouseId?
```
kevinaboos: yea, i agree it'd be nice to confirm this before merging in this PR. If it's possible, i'd greatly appreciate that.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+    _60 = 60,
+    _80 = 80,
+    _100 = 100,
+    _200 = 200
+}
+
+/// set ps2 mouse's sampling rate
+fn set_mouse_sampling_rate(value: SampleRate) -> Result<(), &'static str> {
+    command_to_mouse(HostToMouseCommandOrData::MouseCommand(SampleRate))
+        .and(command_to_mouse(HostToMouseCommandOrData::SampleRate(value)))
```
kevinaboos: i think this needs to be `and_then`, not `and`, because I assume you only want to invoke the second command if the first command succeeds. 

Currently this will result in both commands being run regardless of the first one's success.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
-                warn!("fail to stop streaming, before trying to read mouse id");
-                return Err("fail to set the mouse id!!!");
-            }
-            if let Err(_e) = set_sampling_rate(100) {
-                warn!("fail to stop streaming, before trying to read mouse id");
-                return Err("fail to set the mouse id!!!");
+        //(maybe) TODO: apparently this needs to check after the first 80 if mouse_id is 3
+        MouseId::Four => {
+            for rate in [_200, _100, _80, _200, _200, _80] {
+                set_mouse_sampling_rate(rate)?;
```
kevinaboos, line 803: i love this way of doing this! very clever.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
-            // begin streaming again
-            if let Err(e) = mouse_packet_streaming(true) {
-                warn!("the streaming of mouse is disabled,please open it manually");
-                return Err(e);
-            }
-            return Ok(id_num);
-        }
-    }
+
+pub fn mouse_id() -> Result<u8, &'static str> {
```
kevinaboos: public functions should include docs.

Also, is it potentially a problem if this is called randomly by another crate? e.g., before the ps/2 port or mouse has been initialized?
If so, we should make it a member method of some PS/2 struct or something else that prevents it from being arbitrarily invoked.

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
-pub fn keyboard_led(value: u8) {
-    if let Err(_e) = command_to_keyboard(0xED) {
-        warn!("failed to set the keyboard led");
-    } else if let Err(_e) = command_to_keyboard(value) {
-        warn!("failed to set the keyboard led");
-    }
+/// set LED status of the keyboard
+pub fn set_keyboard_led(value: LEDState) -> Result<(), &'static str> {
+    command_to_keyboard(HostToKeyboardCommandOrData::KeyboardCommand(SetLEDStatus))
+        .and(command_to_keyboard(HostToKeyboardCommandOrData::LEDState(value)))
```
kevinaboos: same comment about `and_then` instead of `and`

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
-pub fn keyboard_scancode_set(value: u8) -> Result<(), &'static str> {
-    if let Err(_e) = command_to_keyboard(0xF0) {
-        return Err("failed to set the keyboard scancode set");
-    } else if let Err(_e) = command_to_keyboard(value) {
-        return Err("failed to set the keyboard scancode set");
-    }
-    Ok(())
+pub fn keyboard_scancode_set(value: ScancodeSet) -> Result<(), &'static str> {
+    command_to_keyboard(HostToKeyboardCommandOrData::KeyboardCommand(SetScancodeSet))
+        .and(command_to_keyboard(HostToKeyboardCommandOrData::ScancodeSet(value)))
```
kevinaboos: same comment about `and_then` instead of `and`

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
-            warn!("the resolution value is not accepted");
-            Err("set mouse resolution failed!!!")
-    } else {
-            // info!("set mouse resolution succeeded!!!");
-            Ok(())
-    }
+#[allow(dead_code)]
+fn mouse_resolution(value: MouseResolution) -> Result<(), &'static str> {
+    command_to_mouse(HostToMouseCommandOrData::MouseCommand(SetResolution))
+        .and(command_to_mouse(HostToMouseCommandOrData::MouseResolution(value)))
```
kevinaboos: same comment about `and_then` instead of `and`

kernel/ps2/src/lib.rs
```diff
@@ -1,484 +1,704 @@
...
+        .map_err(|_| {
+            "failed to set the mouse resolution"
+        })
```
kevinaboos:
```suggestion
        .map_err(|_| "failed to set the mouse resolution")
```

kernel/ps2/src/lib.rs
```diff
@@ -487,25 +707,25 @@ pub enum KeyboardType {
...
-        return Err(e);
-    } else {
-        let reply = ps2_read_data();
-        match reply {
-            0xAB => Ok(KeyboardType::MF2KeyboardWithPSControllerTranslator),
-            0x41 => Ok(KeyboardType::MF2KeyboardWithPSControllerTranslator),
-            0xC1 => Ok(KeyboardType::MF2KeyboardWithPSControllerTranslator),
-            0x83 => Ok(KeyboardType::MF2Keyboard),
-            _ => Ok(KeyboardType::AncientATKeyboard),
+    command_to_keyboard(HostToKeyboardCommandOrData::KeyboardCommand(DisableScanning))?;
```
kevinaboos, line 903: minot nit: i know you weren't the original author of this, but could you improve the formatting & content of this function's docs? thanks.

kernel/window_manager/src/lib.rs
```diff
@@ -684,30 +684,26 @@ fn window_manager_loop(
...
                                     && next_mouse_event.buttonact.fourth_button_hold
                                         == mouse_event.buttonact.fourth_button_hold
                                     && next_mouse_event.buttonact.fifth_button_hold
-                                        == mouse_event.buttonact.fifth_button_hold
-                                {
-                                    x += (next_mouse_event.displacement.x as i8) as isize;
-                                    y += (next_mouse_event.displacement.y as i8) as isize;
+                                        == mouse_event.buttonact.fifth_button_hold {
+                                    x += next_mouse_event.mousemove.x_movement as isize;
+                                    y += next_mouse_event.mousemove.y_movement as isize;
```
kevinaboos: above this, you mentioned being concerned about overflow. I agree, and have recently been auditing other parts of the Theseus codebase that use arithmetic operations without handling overflow (e.g., #663).

Thus, we should probably use either `wrapping_add` or `saturating_add` here, whichever is appropriate according to the semantics of these values.

kevinaboos: or, if you want to disallow overflow, `checked_add` is also available. But that likely isn't the correct choice here.

libs/keycodes_ascii/src/lib.rs
```diff
@@ -1,13 +1,11 @@
...
-
-#[macro_use] extern crate bitflags;
-
-// use core::cell::RefCell;
+use bitflags::bitflags;
+use num_enum::TryFromPrimitive;
 
 // TODO: use these tables and tips:
 // https://sourceforge.net/p/oszur11/code/ci/master/tree/Chapter_06_Shell/04_Makepp/arch/i386/arch/devices/i8042.c
+// ^link dead; https://stackoverflow.com/questions/219120/protected-mode-keyboard-access-on-x86-assembly
```
kevinaboos: thanks, you can remove this whole comment and/or replace it with that link.

libs/keycodes_ascii/src/lib.rs
```diff
@@ -133,6 +130,7 @@ impl KeyEvent {
     }
 }
 
+// FIXME: this is only true for scancode set 1
```
kevinaboos: let's discuss this based on my other question about unconditionally setting scancode Set 2. Shall we change this to be compatible with Set 2? Seems like a pretty important consideration.

libs/keycodes_ascii/src/lib.rs
```diff
@@ -241,138 +242,35 @@ pub enum Keycode {
...
     /// Obtains the ascii value for a keycode under the given modifiers
     pub fn to_ascii(&self, modifiers: KeyboardModifiers) -> Option<char> {
         // handle shift key being pressed
         if modifiers.is_shift() {
-            // if shift is pressed and caps lock is on, give a regular lowercase letter
             if modifiers.is_caps_lock() && self.is_letter() {
+            // if shift is pressed and caps lock is on, give a regular lowercase letter
                 return self.as_ascii();
-            }
+            } else {
```
kevinaboos: no need to reorder these. It's more readable (imo) to place the commend above the condition that it applies to, just like how we place doc comments right above the item they apply to.

libs/mouse_data/src/lib.rs
```diff
@@ -1,164 +1,62 @@
...
+            x_movement,
+            y_movement,
+            scroll_movement,
+        }
     }
 }
-#[derive(Debug, Copy, Clone)]
+
+//NOTE: could be reduced to 8-bit
+#[derive(Debug, Clone)]
```
kevinaboos: I agree, let's use the `bitflags` crate for this, as with `KeyboardModifiers`.

kevinaboos (changes requested):
Ok phew, I got through all the changes! In general everything looks great! I did leave a lot of comments, some of which are a bit nitpicky, and others which can be applied to many other parts of the code.

Thanks again, and kindly do let me know if there's anything you disagree with; I'm more than open to discussion.

kernel/keyboard/src/lib.rs
hecatia-elegua: I tried writing up an answer here but realized I don't really know.
Previously we just didn't set it, so I first tried setting it to Set 1, which gave back the wrong scancodes, so then I tried Set 2 and it worked.
I have some ideas on where to look, so I will fix

kernel/keyboard/src/lib.rs
hecatia-elegua: I know and yes that's a good idea.
Markdown-esque formatting was for describing what the link contains, not really for docs

kernel/ps2/src/lib.rs
hecatia-elegua: I think I will just remove these, I kind of wanted to emulate what the above mentioned arbitrary bit-width integer people proposed, but if that ever hits rust it would be a simple search & replace anyways.

kernel/ps2/src/lib.rs
hecatia-elegua: nit: I dislike how much one needs to write for these, not sure why there's no attribute, like `#[size(8)]` or something, which would feel similar to the `repr` attributes.
`modular_bitfield` also asserts the size already.

oh and btw. `const _: () = assert!(core::mem::size_of::<ControllerToHostStatus>() == 8);` also works and might be better since `static_assertions` last commit was in 2020

kernel/ps2/src/lib.rs
hecatia-elegua: Not sure about the low-level rust handling this. Do you know why e.g. a [u8; 1] can't just be turned into a u8 / why they aren't handled the same? Where can I learn more about it?
Anyways, this would need changes to the `modular_bitfield` crate.

kernel/ps2/src/lib.rs
hecatia-elegua: I just left it the way it was, because I'm not sure either, but will test.
(Just some rambling which might be useful later:
In wiki.osdev they initialize and test both keyboard and mouse at once, which we currently do separately, probably because of modularity.
I can just guess that we probably init the keyboard first, which means port 1 needs to be reenabled and then we init the mouse, so afterwards both need to be reenabled. This would mean we can't change the order of initialization anyways, because it would leave the mouse disabled.)

kernel/ps2/src/lib.rs
hecatia-elegua: probably missed that one, here the spec says "Failed is 0xFC", but it doesn't matter because
`read_data().try_into().map_err(|_| "failed to read controller test result")`
rejects anything besides "Passed" anyways.

kernel/ps2/src/lib.rs
hecatia-elegua: that actually gives me some input on how to modularize all the functions
most of these mustn't be called before init

kernel/window_manager/src/lib.rs
hecatia-elegua: thank you! forgot about those

libs/keycodes_ascii/src/lib.rs
hecatia-elegua: I will have to read up on it. Set 2 can get converted to Set 1 by the PS/2 controller, I'm a bit confused about what it's doing in our setup.

libs/keycodes_ascii/src/lib.rs
hecatia-elegua: the `} else {` formatting OCD was too strong

kernel/mouse/src/lib.rs
hecatia-elegua: iirc the problem was that the types are in different crates.
Workaround would be to define newtypes somewhere, will see.

kernel/keyboard/src/lib.rs
kevinaboos: thanks! I briefly searched arbut couldn't find a definitive answer

kernel/mouse/src/lib.rs
kevinaboos: right, when they're in different crates, can't you still implement `Into`? I mean it's not a big deal, but if you can't do it then we can just ignore & resolve this comment thread.

kernel/ps2/src/lib.rs
kevinaboos: yea it's a bit verbose, true, but it's important for structures that must have a fixed, correct size in order to match hardware expectations.

Interesting, i didn't know Rust now supported regular assertions in const contexts now, thanks for sharing!

kernel/ps2/src/lib.rs
kevinaboos: oh ok, didn't realize that was part of `modular_bitfield`. Makes sense though, since the bitfield could be larger than a single byte.

> Do you know why e.g. a [u8; 1] can't just be turned into a u8 / why they aren't handled the same? 

Yes, good news! there is a function for that: https://doc.rust-lang.org/std/primitive.u8.html#method.from_ne_bytes

kernel/ps2/src/lib.rs
kevinaboos: oh that is odd. thanks for the clarification!

I think it's completely fine to initialize them together as part of a larger `ps2::init()` function or something like that, in which we follow your advice above and attempt to init keyboard, then mouse. That should be possible to directly invoke from the device manager or captain, wherever appropriate.

kernel/ps2/src/lib.rs
kevinaboos: got it, makes sense. Feel free to remove it.

kernel/ps2/src/lib.rs
kevinaboos: yep, that's kind of the typestate logic that we like to use, and a similar effort to implement NIC drivers in this way is ongoing at Yale. If it's easier, you can add that redesign in a follow-up PR too, since this one is already pretty much done (except for a few minor nits here & there).

libs/keycodes_ascii/src/lib.rs
kevinaboos: either one is ok in this scenario cuz the basic blocks within the conditional branches are very short. If you keep it as is, just fix the indentation to line up with the statement within the braces. Your call.

hecatia-elegua:
Some questions about `make view-docs`:
can I somehow only build docs for one crate?
why does it create `Cargo.lock` files for all `libs/`?
would we want the option to also include all the docs on private items?
thanks

hecatia-elegua:
wait, why did this break only now?
I mean, I built often, without error. How strange

kevinaboos:
> Some questions about `make view-docs`: can I somehow only build docs for one crate?

Sure, you can call rustdoc on a specific crate, like I do here. But there's no real reason to do that (unless I'm missing something?)

> why does it create `Cargo.lock` files for all `libs/`? 

The crates in `libs/` are not part of the main Cargo workspace, so they don't get documented because we don't want to document all third-party dependencies.
Instead, we manually call rustdoc on the `libs/` crates that we _do_ want to include in the docs, which currently includes only the first-party crates in `libs/`.
The only reason you see those `.lock` files is because I haven't added them to the `.gitignore` in each libs crate. You can safely delete/ignore them though.

> would we want the option to also include all the docs on private items? thanks

On some things, yes. That's mostly so intra-doc links on private items are checked and processed correctly, but technically it's not required.

kevinaboos:
@hecatia-elegua thanks for addressing my comments! Everything looks pretty good, but it's a bit hard to tell if I missed any significant changes due to the fact that the branch has been force-pushed, which causes github to lose track of what I had previously reviewed.

Anyway, are there any other remaining to-do items that you intend to work on, or do you consider this PR ready to be tested and merged in?

hecatia-elegua:
> would we want the option to also include all the docs on private items?

What I meant by this: would we want another build-target with `--document-private-items` specified, so e.g. newcomers could read even the private function docs in the browser? Sorry for going off-topic.

> which causes github to lose track of what I had previously reviewed

Does it? It says "outdated" on most of your review comments, but on the lines I didn't change it doesn't.

Changes started from [this commit](https://github.com/theseus-os/Theseus/pull/667/commits/8d2c8c4568e10d8671fe97d70a72ba6c9485a7ea), so you could go through them from there, but
if you want to save time, the only "new new" part is `enum MousePacket` and its sub-structs, like: https://github.com/theseus-os/Theseus/blob/6693e355d48a570ab31e6b292a561aaa564a852a/kernel/ps2/src/lib.rs#L566


The PR already had CI runs btw. and can be merged.

kevinaboos:
> > would we want the option to also include all the docs on private items?
> 
> What I meant by this: would we want another build-target with `--document-private-items` specified, so e.g. newcomers could read even the private function docs? Sorry for going off-topic.

Oh, I see. In that case, yes, potentially. Feel free to open up a new PR with that change; if it's good and only improves the docs then we can enable it unconditionally. 

> > which causes github to lose track of what I had previously reviewed
> 
> Does it? It says "outdated" on most of your review comments, but on the lines I didn't change it doesn't.

Yeah it breaks the nice github feature that shows "changes since your last review". But no big deal, since you told me the relevant starting commit too. Thanks!

> The PR already had CI runs btw. and can be merged.

Great, but those CI passes don't actually perform any runtime tests (yet... one day we'll get around to it). I test it myself, so I'll do that and merge things in if it goes well. I assume you've tested it thoroughly on QEMU at least.

hecatia-elegua:
I have tested on QEMU, hope I didn't miss anything

kevinaboos (changes requested):
I tested everything and it's working well, yay! I made a few small changes and merged in the latest changes from `theseus_main` to ensure the branch still builds on our more recent nightly Rust version.

There's only one very minor issue - when pressing the Caps lock key, I see this error on the log:
```
[E] kernel/keyboard/src/lib.rs:192: handle_keyboard_input(): Unknown scancode: 250, adjusted scancode: 122
[E] kernel/keyboard/src/lib.rs:90: ps2_keyboard_handler: error handling PS2_PORT input: "unknown keyboard scancode"
```

If you have a moment to address that error, that'd be great. If not, let me know and I'm happy to look into it. Aside from that, I'd say this PR is ready to merge!

hecatia-elegua:
Thank you.
I had problems with the caps lock key too.
in set 1, it would be 58 (which is what it should be)
in set 2, it would be 88
250 is FA, so that's an ACK "command acknowledged" response.

If you have an idea what it could be, you can fix or tell me, otherwise I'll look into it later

hecatia-elegua:
In our code, it looks like a caps lock press first triggers an interrupt with scancode 58 (make caps), then immediately another one with 250 (ack), a release only triggers an interrupt with 186 (break caps).

I'm guessing the `set_keyboard_led` command has something to do with it, although the ack is already getting handled there...
Also not sure why the keyboard command ack would trigger an interrupt, because afaik these should only happen when `output_buffer_full`, which should not be full after `polling_receive` in `set_keyboard_led`.
Maybe we should wait a bit inside the handler after setting the LED? Otherwise we could also just ignore values of 250.

update: waiting works, but I just printed some stuff in a loop to wait, so instead, what's the right way to do it?

kevinaboos:
> I'm guessing the `set_keyboard_led` command has something to do with it, although the ack is already getting handled there... Also not sure why the keyboard command ack would trigger an interrupt, because afaik these should only happen when `output_buffer_full`, which should not be full after `polling_receive` in `set_keyboard_led`. Maybe we should wait a bit inside the handler after setting the LED? Otherwise we could also just ignore values of 250.

I'm not sure if the keyboard works this way, but a lot of legacy devices on x86 (e.g., [like this in the serial port](https://github.com/theseus-os/Theseus/blob/b36e50d41ccaab3de13c1ddad637ec2a49c2ce42/kernel/serial_port_basic/src/lib.rs#L292-L296)) require you to spin in a loop while polling a status register in order to determine when the device is ready to receive another command. Perhaps you need to do that here too.

> update: waiting works, but I just printed some stuff in a loop to wait, so instead, what's the right way to do it?

If there's not a status register (as I mentioned above) with a bit flag that indicates whether the keyboard is ready to receive a command, then we can insert a wait loop (but not by using print statements). This shouldn't be necessary, but a common way to do that in minimal legacy environments [is this](https://github.com/theseus-os/Theseus/blob/b36e50d41ccaab3de13c1ddad637ec2a49c2ce42/kernel/pic/src/lib.rs#L249-L257).

kevinaboos:
Ok I took a deeper look, and fortunately this is probably a lot simpler than what we were discussing. You can ignore what i wrote above; i double-checked that your code paths are all properly checking/waiting for status registers. 

Your guess about `set_keyboard_led` is correct. The interrupt to acknowledge your "set kbd LEDs" command going to fire immediately before you can poll for it. There's no real way for you to avoid that interrupt, so I think it's fine for you to ignore any ACK value (0xFA) when handling scancodes, for now.
The real issue is that you're mixing interrupt-based data receive and polling-based data receive. So the byte that you read in from port 0x60 is being treated as a keypress action from the user instead of an acknowledgment or error response from the keyboard.  Generally you should only do one or the other, i.e., only polling or only interrupts, but it works here for the keyboard because the keyboard is a simple/dumb device.

hecatia-elegua:
> The real issue is that you're mixing interrupt-based data receive and polling-based data receive.

A lot of wording online sounded like you always need to use polling for commands, e.g. when initializing the devices.
Now reading [here](https://wiki.osdev.org/%228042%22_PS/2_Controller#Interrupts) again, I guess this is what I got wrong: The last paragraph is only about commands to the PS/2 Controller, so kind of irrelevant for mouse and keyboard commands.

We also at least need polling-send, as the PS/2 Controller does not support interrupt-driven sending to devices.

I think we can merge now :)

kevinaboos (approved):
Looks good, thanks so much for this amazing contribution!

kevinaboos:
Did you also intend to tackle #685 and #686 yourself, or were those just reminders for myself or a future contributor?

Also I wonder if this fixes #373, i should test that sometime.

Thanks again!

hecatia-elegua:
I intend to come back to these, yes.

PR #668, opened 2022-10-24
Rework shell and console connections

tsoutsman:
This PR is the first step in reworking the shell, adding a `shell_2` crate that is used by `console` for serial connections. It works and can be connected to using `screen` or a similar application. A couple of problems with the shell:

- [ ] I've added history deduplication to `noline` but haven't merged it into upstream yet. There's currently a patch in `Cargo.toml`.
- [ ] Autocompletion will need to be implemented upstream.
- [ ] Theseus handles Ctrl+C in a non-standard way and `noline` doesn't clear the current line when it encounters a Ctrl+C.
- [ ] If multiple keys are pressed in quick succession after connecting to the serial port, `noline` will fail. `noline` queries the terminal emulator for screen size using ANSII escape codes. However, the key press will be interspersed with the response from the terminal if the `port_to_tty_loop` isn't scheduled between the `noline` request and user key press. This isn't a major issue but I should probably document it somewhere.

I've started on a terminal emulator that uses `shell_2` but decided to first submit this PR as it's already massive as it is.

The PR is dependent on #657.

I've also done an ad hoc rework of `app_io` which I'm not a particular fan of. It leads to a number of `Arc<Arc<T>>` sandwiches. I also need to port over `immediate_input_echo`, `input_echo`, and `keyboard_echo` to use ANSII escape codes and the line discipline.

The PR also removes the `print` and `terminal_print` crates which were both legacy interfaces. They've been replaced with `app_io`. Most of the file changes are due to this.


Cargo.lock
```diff
@@ -2866,18 +2886,34 @@ dependencies = [
...
 ]
 
+[[package]]
+name = "shell_2"
+version = "0.1.0"
+dependencies = [
+ "app_io",
+ "core2",
+ "embedded-hal",
+ "hashbrown 0.12.3",
```
kevinaboos: is there a strong need for this hashmap version? if not, i'd prefer to keep all the versions consistent (we can upgrade later) at v0.11.*

Cargo.toml
```diff
@@ -126,6 +126,7 @@ bincode = { git = "https://github.com/bincode-org/bincode" }
 wasmparser = { git = "https://github.com/theseus-os/wasm-tools", branch = "no-std-wasmparser" }
 backtrace = { path = "ports/backtrace" }
 region = { path = "ports/region" }
+noline = { git = "https://github.com/tsoutsman/noline", branch = "history-dedup" }
```
kevinaboos: please move this to a `theseus-os`-owned repo and change this patch to use that

applications/immediate_input_echo/src/lib.rs
```diff
@@ -1,47 +1,49 @@
 //! An application to test instant stdin flush.
 #![no_std]
 
+// FIXME
```
kevinaboos: do you intend to fix this for this PR, or just remove the entire app?

applications/shell/src/lib.rs
```diff
@@ -628,13 +615,6 @@ impl Shell {
                     // to the cmdline.
                     if let Some(fg_job_num) = self.fg_job_num {
                         self.insert_char_to_input_buff(c, true)?;
-                        if let Some(job) = self.jobs.get(&fg_job_num) {
-                            if app_io::is_requesting_instant_flush(&job.task_ids[0])? {
-                                job.stdin_writer.lock().write_all(self.input_buffer.as_bytes())
-                                    .or(Err("shell failed to write to stdin"))?;
-                                self.input_buffer.clear();
-                            }
-                        }
```
kevinaboos, line 638: quick question: this appears to be for handling line-buffered stdin vs unbuffered stdin; is that now handled elsewhere? If not, we should leave a todo or some other kind of comment (or file an issue as a reminder)

applications/shell_2/src/internal.rs
```diff
@@ -0,0 +1,109 @@
+use crate::{Error, Result, Shell};
+use alloc::{borrow::ToOwned, vec::Vec};
+use path::Path;
+
+// TODO: Decide which internal commands we don't need.
```
kevinaboos: i think the typical term for these are "shell builtins", so we ought to use that term for this stuff instead of `internal`
https://en.wikipedia.org/wiki/Shell_builtin

applications/shell_2/src/error.rs
```diff
@@ -0,0 +1,11 @@
+use alloc::string::String;
+
+pub type Result<T> = core::result::Result<T, Error>;
+
+#[derive(Clone, Debug)]
+pub enum Error {
+    ExitRequested,
+    CurrentTaskUnavailable,
+    CommandNotFound(String),
+    Command(isize),
```
kevinaboos: add docs describing each enum variant, especially the `Command` one and what the internal `isize` represents

applications/shell_2/src/internal.rs
```diff
@@ -0,0 +1,109 @@
...
+        let task = task::get_my_current_task().unwrap();
+
+        match task.get_env().lock().chdir(&path) {
+            Ok(()) => Ok(()),
+            Err(_) => todo!(),
+        }
+    }
+
+    pub(crate) fn exec(&self, _args: Vec<&str>) -> Result<()> {
+        todo!();
```
kevinaboos: for all of these commands, instead of panicking, let's either:
1. print a message (and return an error) explaining that this builtin isn't yet implemented, we can optionally print a link to the Theseus github issue tracker with the suggestion to file an issue asking for it to be supported
2. print a message (and return an error) explaining that this builtin is not relevant or applicable to Theseus and will never be supported

applications/shell_2/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+use alloc::{borrow::ToOwned, format, string::String, sync::Arc, vec, vec::Vec};
+use app_io::println;
+use core::fmt::Write;
+use hashbrown::HashMap;
+use job::Job;
+use noline::{builder::EditorBuilder, sync::embedded::IO as Io};
+use path::Path;
+use tty::{Event, LineDiscipline};
+
+// FIXME: export main function rather than shell struct
```
kevinaboos: ? clarify please

applications/shell_2/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    pub fn new() -> Self {
+        Self {
+            discipline: app_io::line_discipline().unwrap(),
+            jobs: HashMap::new(),
+            stop_order: Vec::new(),
+        }
+    }
+
+    /// Runs the shell, consuming it in the process.
+    pub fn run(mut self) -> Result<()> {
```
kevinaboos: is there a use case where creating a new `Shell` and `run`ning it need to be separate actions?

applications/shell_2/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    fn set_shell_discipline(&self) {
+        self.discipline.set_raw();
+    }
+
+    /// Configures the line discipline for use by applications.
+    fn set_app_discipline(&self) {
+        self.discipline.set_sane();
+    }
+
+    fn _run(&mut self) -> Result<()> {
```
kevinaboos: Rust typically uses a leading underscore to indicate something that isn't used. If that's not appropriate for this, i suggest something like `run_inner` or the like

applications/shell_2/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+            }
+            Err(Error::CommandNotFound(command)) => {
+                println!("{}: command not found", command);
+                Ok(())
+            }
+            _ => Ok(()),
+        }
+    }
+
+    // TODO: Use guards?
```
kevinaboos: for what, setting the line discipline? Yea, i think that'd be a nice feature.

applications/shell_2/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    fn execute_external(&mut self, cmd: &str, args: Vec<&str>) -> Result<()> {
+        self.set_app_discipline();
+        let result = self._execute_external(cmd, args);
+        self.set_shell_discipline();
+        result
+    }
+
+    fn _execute_external(&mut self, cmd: &str, args: Vec<&str>) -> Result<()> {
```
kevinaboos: same comment as above about the leading underscore. When looking at just this function in isolation, I can't tell if this is supposed to be an alternative to "commenting out" this method, or if it's an internal fn.

applications/shell_2/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+                    .collect::<Vec<_>>(),
+            )
+            .block()
+            .spawn()
+            .unwrap();
+
+        let id = task.id;
+        // TODO: Double arc :(
+        app_io::insert_child_streams(id, app_io::streams().unwrap());
+        // TODO: set environment
```
kevinaboos: bump; address these before merge or delay for later?

applications/shell_2/src/wrapper.rs
```diff
@@ -0,0 +1,32 @@
...
+    fn write(&mut self, byte: u8) -> nb::Result<(), Self::Error> {
+        assert_eq!(self.stdout.write(&[byte])?, 1);
+        Ok(())
+    }
+
+    fn flush(&mut self) -> nb::Result<(), Self::Error> {
+        // TODO: Forward to slave?
+        Ok(())
+    }
+}
```
kevinaboos: this is quite strange with the assert statements. Can you shed some light on this?

kernel/app_io/Cargo.toml
```diff
@@ -3,36 +3,31 @@ name = "app_io"
 version = "0.1.0"
 authors = ["Zhiyao Ma <zm16@pku.edu.cn>"]
 description = "Offers applications the ability to read from or print to the terminal."
+edition = "2021"
 
 
 [dependencies]
-spin = "0.9.0"
+hashbrown = "0.12.3"
```
kevinaboos: use v0.11 for consistency

kernel/async_channel/src/lib.rs
```diff
@@ -349,14 +370,21 @@ impl <T: Send> Receiver<T> {
...
 }
 
 
 /// Drop implementation marks the channel state and notifys the `Sender`
 impl<T: Send> Drop for Receiver<T> {
     fn drop(&mut self) {
         // trace!("Dropping the receiver");
-        self.channel.channel_status.store(ChannelStatus::ReceiverDisconnected);
+        // FIXME
+        // self.channel.channel_status.store(ChannelStatus::ReceiverDisconnected);
```
kevinaboos: what's with these changes to the async_channel; are they necessary for the new shell/console design?

Note that there's also the `mpmc` queue or crossbeam's various `mpsc` queue types that might be better for your needs, depending on what features you actually need.

If the channel needs to change, i'd prefer to separate that out into a distinct PR so it's easier to track why things changed, rather than hiding it within a huge PR like this

kernel/async_channel/src/lib.rs
```diff
@@ -365,7 +393,8 @@ impl<T: Send> Drop for Receiver<T> {
 impl<T: Send> Drop for Sender<T> {
     fn drop(&mut self) {
         // trace!("Dropping the sender");
-        self.channel.channel_status.store(ChannelStatus::SenderDisconnected);
+        // FIXME
+        // self.channel.channel_status.store(ChannelStatus::SenderDisconnected);
```
kevinaboos: same question as above

kernel/console/Cargo.toml
```diff
@@ -1,33 +1,45 @@
 [package]
-authors = ["Kevin Boos <kevinaboos@gmail.com>"]
 name = "console"
-description = "Creation and management of virtual consoles/terminals"
 version = "0.1.0"
+authors = ["Kevin Boos <kevinaboos@gmail.com>"]
```
kevinaboos: should probably add yourself as an author

kernel/console/src/lib.rs
```diff
@@ -3,156 +3,166 @@
...
+        }
+
+        let _ = spawn::new_task_builder(shell_loop, (serial_port, serial_port_address, receiver))
+            .name(format!("{:?}_manager", serial_port_address))
+            .spawn()
+            .unwrap();
+    }
 }
 
+// TODO: Remove unwraps
```
kevinaboos: indeed

kernel/console/src/lib.rs
```diff
@@ -3,156 +3,166 @@
...
+        .spawn()
+        .unwrap();
+    let writer_task = spawn::new_task_builder(port_to_tty_loop, (receiver, tty.master()))
+        .name(format!("{:?}_to_tty", address))
+        .spawn()
+        .unwrap();
+
+    let new_app_ns = mod_mgmt::create_application_namespace(None).unwrap();
+
+    // TODO: Replace shell with shell_2
```
kevinaboos: here? it's already using shell_2 below. Or is this a general todo for the future? If so, we can file an issue to track that

kernel/console/src/lib.rs
```diff
@@ -3,156 +3,166 @@
...
+            stdout: stream.clone(),
+            stderr: stream,
+        },
+    );
+
+    task.unblock();
+    task.join().unwrap();
+
+    reader_task.kill(KillReason::Requested).unwrap();
+    writer_task.kill(KillReason::Requested).unwrap();
```
kevinaboos, line 140: Currently the `kill` functionality doesn't work properly (it doesn't unwind the task and clean up its resources), so we definitely should not use it here. Instead, use some kind of shared state or message to inform those two tasks that they should exit cleanly, just as you would do with classic userspace pthreads.

kernel/console/src/lib.rs
```diff
@@ -3,156 +3,166 @@
...
+
+    // Flush the tty in case the reader task didn't run between the last time the
+    // shell wrote something to the slave end and us killing the task.
+    let mut data = [0; 256];
+    if let Ok(len) = tty.master().try_read(&mut data) {
+        port.lock().write(&data[..len]).unwrap();
+    };
+
+    // TODO: Close port?
```
kevinaboos, line 151: so, we've reached here when the remote client endpoint quits? (e.g., when i close the terminal emulator on the host side)
At that point, yes, we should clean up everything and return this to its initial state, in preparation for a brand new connection

kernel/console/src/lib.rs
```diff
@@ -3,156 +3,166 @@
...
-		}
-	}	
-
-	// Err("console_entry task returned unexpectedly")
+fn port_to_tty_loop((receiver, master): (Receiver<DataChunk>, tty::Master)) {
+    loop {
+        let DataChunk { data, len } = receiver.receive().unwrap();
+        trace!("read data from serial port: {:?}", &data[..len.into()]);
+        master.write(&data[..len as usize]).unwrap();
+    }
```
kevinaboos: i know we haven't done a great job of this in the past, but are these two tasks blocking? It's pretty heavyweight to spawn so many tasks to handle each console connection, so i want to make sure they're not Runnable and endlessly spinning while waiting for new data

kevinaboos: (looks like they're using the async_channel, but i just want to be sure)

kernel/task/src/lib.rs
```diff
@@ -206,6 +206,10 @@ pub enum RunState {
     Runnable,
     /// blocked on something, like I/O or a wait event
     Blocked,
+    /// The `Task` has been suspended from proceding further.
+    ///
+    /// Currently, this is only triggered by a Ctrl + Z in the terminal.
+    Suspended,
```
kevinaboos: I'm hesitant to introduce new runstates unnecessarily, as it adds lots of complexity to all the places where task runstates are used and checked. Is this absolutely mandatory, and if so, why?  I understand it's slightly semantically different than `Blocked`, but is it a true problem to reuse `Blocked` for this purpose?

kernel/scheduler/src/lib.rs
```diff
@@ -27,9 +27,12 @@ use task::{get_my_current_task, TaskRef};
 ///    meaning the current task will continue running.
 pub fn schedule() -> bool {
     let preemption_guard = preemption::hold_preemption();
+
```
kevinaboos: undo changes to this file before merge

kernel/tty/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "tty"
+version = "0.1.0"
+edition = "2021"
```
kevinaboos, line 6: add author & description

kernel/tty/src/channel.rs
```diff
@@ -0,0 +1,95 @@
+// TODO: Use more efficient data structure?
+// TODO: A sender and receiver are the same under the hood. Hypothetically we
```
kevinaboos: technically true, but they're separated in order to limit the capabilities of the caller, so let's not overload that

kernel/tty/src/channel.rs
```diff
@@ -0,0 +1,95 @@
...
+// could store just one.
+// FIXME: async_channel is not a proper mpmc
+// FIXME: Error handling
+
+use async_channel::{Receiver, Sender};
+use core2::io::{ErrorKind, Result};
+
+#[derive(Clone)]
+pub(crate) struct Channel {
```
kevinaboos, line 5: this whole module isn't really tty-specific, right?. It looks like you want a channel effectively does what `read_to_end` does? 

If so, let's separate out all of these channel-specific changes into a different PR so we can analyze them in a standalone manner

kernel/tty/src/discipline.rs
```diff
@@ -0,0 +1,266 @@
...
+///
+/// The line discipline can be configured based on what application is using the
+/// slave end. Most applications should use the [`sane`](Self::sane) setting,
+/// which handles line editing and echoing to the terminal. Applications that
+/// require more control over the display should use the [`raw`](Self::raw)
+/// setting.
+///
+/// The line discipline's behaviour is documented in terms of Linux `termios`
+/// flags. For more information, visit the [`cfmakeraw`
+/// documentation][cfmakeraw].
```
kevinaboos, line 21: Q: does this render properly in rustdoc? i haven't seen this syntax before

kernel/tty/src/discipline.rs
```diff
@@ -0,0 +1,266 @@
...
+            *self.canonical.lock().unwrap() = Some(Vec::new());
+        } else {
+            // TODO: Flush buffer?
+            *self.canonical.lock().unwrap() = None;
+        }
+    }
+
+    fn clear_input_buf(&self) {
+        if let Some(ref mut input_buf) = *self.canonical.lock().unwrap() {
+            *input_buf = Vec::new();
```
kevinaboos: nit: maybe a bit better to use `Vec::clear()` ?

kernel/tty/src/discipline.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    pub(crate) fn process_input_buf(
+        &self,
+        buf: &[u8],
+        master: &Channel,
+        slave: &Channel,
+    ) -> Result<()> {
+        for byte in buf {
+            // TODO: This locks internal channels on every byte.
```
kevinaboos: the serial port impl already attempts to read multiple bytes (more specifically, `u16` input events) on every interrupt and send larger batches (up to 256 `u16`s for a total max of 512 bytes) per msg, so you could take a similar batching approach here

kernel/tty/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+impl Read for Slave {
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
+        let immutable: &Self = self;
+        immutable.read(buf)
+    }
+}
+
+impl Write for Slave {
+    fn write(&mut self, buf: &[u8]) -> Result<usize> {
+        let immutable: &Self = self;
```
kevinaboos, line 177: i see what's happening, but i guess i'm confused by the usage of the term "immutable" here. hmm....

kernel/wait_queue/src/lib.rs
```diff
@@ -125,7 +125,8 @@ impl WaitQueue {
                 if !wq_locked.contains(curr_task) {
                     wq_locked.push_back(curr_task.clone());
                 } else {
-                    warn!("WaitQueue::wait_until():  task was already on waitqueue (potential spurious wakeup?). {:?}", curr_task);
+                    // FIXME: This continually triggers after the shell is shutdown. #657 fixes it.
```
kevinaboos: 657 has been merged in now :tada:

theseus_features/Cargo.toml
```diff
@@ -29,7 +29,7 @@ pmu_sample_stop = { path = "../applications/pmu_sample_stop", optional = true }
 ps = { path = "../applications/ps", optional = true }
 pwd = { path = "../applications/pwd", optional = true }
 rm = { path = "../applications/rm", optional = true }
-shell = { path = "../applications/shell", optional = true }
+shell = { package = "shell_2", path = "../applications/shell_2", optional = true }
```
kevinaboos: add a separate feature here, this removes the existing shell as an optional component.

kernel/app_io/src/lib.rs
```diff
@@ -1,329 +1,226 @@
...
-pub fn get_my_terminal() -> Option<Arc<Mutex<Terminal>>> {
-    task::get_my_current_task_id()
-        .and_then(|id| shared_maps::lock_stream_map()
-            .get(&id)
-            .map(|property| property.terminal.clone())
-        )
+/// Stores the stdio queues, key event queue and the pointer to the terminal
+/// for applications. This structure is provided for application's use and only
+/// contains necessary one-end readers/writers to queues. On the shell side, we
+/// have full control to queues.
```
kevinaboos: "full control to queues"? small typo?

kevinaboos (changes requested):
Top-level comments:
* For specific test apps like `immediate_input_echo`, `keyboard_echo`, `less`, and `input_echo`, do you plan to fix them as part of this PR, remove them entirely, or re-implement them later?
* Add crate-level or module-level docs to the new files, e.g., `shell_2` and its modules
* There are _quite_ a few `unwrap` operations in the `shell_2`. Normally I don't mind being lazy and introducing panicking instructions in application code, but in this case the `shell` is kind of the fundamental interactive application since Theseus lacks a true desktop environment. So if the shell panics, it's quite hard to display something indicating that to the user. 
* [optional] rename `shell_2` to something more descriptive? "ng" or something, or perhaps a new name entirely. I had some ideas related to the "Ship of Theseus" theme, e.g., "hull" -- it's the _shell_ of a ship... lol
* `ImmutableWrite`/`Read` sound a bit like my concept of `LockableIo` -- can we combine those concepts or perhaps adapt `LockableIo` for your needs?

* Last but not least, how does this changeset interact with the "legacy" feature that allows kernel crates (without an app context) to also print to the "default" terminal? I'm primarily concerned about printing exceptions and other important messages both before and after the shell registers app_io streams. Without a desktop environment, there's no other way to do that, considering that the logger isn't available on most real hardware.

kernel/tty/src/discipline.rs
tsoutsman: A: Yes.
image

They are called [reference-style links](https://www.markdownguide.org/basic-syntax/#reference-style-links) and are [commonly used](https://github.com/rust-lang/rust/blob/e221616639fb87de9dca21e252ee8a2565ec51d0/library/std/src/thread/mod.rs#L136-L148) in `std`.

kernel/app_io/src/lib.rs
tsoutsman: I didn't write that but I updated it to be correct.

kernel/console/src/lib.rs
tsoutsman: Yes, the call to `read`/`receive` eventually boils down to `WaitQueue::wait_until`. We need separate tasks because there currently isn't a way to wait on two things at once.

kernel/console/src/lib.rs
tsoutsman: This was more of an "I'm not sure if dropping `port` is enough to return it to its initial state".

kernel/console/src/lib.rs
tsoutsman: I'm not sure how we should do this since they're blocked on a channel.

1. Spin in a loop using `try_receive` and checking an `AtomicBool`. This is terrible as it means `tty_to_port_loop` and `port_to_tty_loop` are always `Runnable`.
2. 
```rust
let mut data = [0; 256];
loop {
    let len = master.read(&mut data).unwrap();
    if should_exit.load(Ordering::SeqCst) {
        break;
    }
    port.lock().write(&data[..len]).unwrap();
}
```
and then write something to the buffer from `shell_loop` after setting `should_exit` to `true` to ensure it is tested. I'm not sure how we would do this for `port_to_tty_loop` as it's waiting on a serial port. Could we expose a function from `serial_port` that returns a write handle so that we can imitate input from the serial port?

applications/shell/src/lib.rs
tsoutsman: yep that's now handled by the canonical flag on the line discipline.

applications/shell_2/src/lib.rs
tsoutsman: I think this can be delayed till later. The double `Arc` is unfortunate but not too important. I haven't set the environment because currently, it behaves in a weird way where the child can modify the parent's environment. I think this should be changed but left it for another PR.

applications/shell_2/src/wrapper.rs
tsoutsman: Since `stdout.read` and `stdout.write` block, they should never return `Ok(0)`.

kernel/async_channel/src/lib.rs
tsoutsman: They are because we use `async_channel` as an `mpmc`. `async_channel` will stop working as soon as a single sender or receiver is dropped which seems wrong.

kernel/tty/src/channel.rs
tsoutsman: Not quite. This makes `read` blocking when operating on buffers. It will block on the first byte, and then use `try_read`/`try_write` to operate on the remainder of the bytes.

kernel/task/src/lib.rs
tsoutsman: If we press Ctrl+Z (suspend) while a task is blocked on something, it will be unblocked by the thing it is waiting for unless we have a separate run state.

FWIW (nothing), [Linux has a separate STOPPED state](https://www.cbtnuggets.com/blog/certifications/open-source/what-are-the-5-linux-process-states) (equivalent to suspended).

tsoutsman: There may be other reasons Linux has a separate state but I can't think of any.

tsoutsman:
@kevinaboos 
> For specific test apps like immediate_input_echo, keyboard_echo, less, and input_echo, do you plan to fix them as part of this PR, remove them entirely, or re-implement them later?

I don't think it's possible to reimplement keyboard_echo. I can reimplement immediate_input_echo and input_echo however, this means they won't work with the current shell and GUI, only with `hull` over serial. Once we have a GUI terminal emulator that runs `hull` this won't be an issue. less can be implemented, but it will require a fair bit of work so I'd rather leave that for another PR.

> Add crate-level or module-level docs to the new files, e.g., shell_2 and its modules

:+1: I've added some basic docs. I'm not sure what else to add.

> There are quite a few unwrap operations in the shell_2. Normally I don't mind being lazy and introducing panicking instructions in application code, but in this case the shell is kind of the fundamental interactive application since Theseus lacks a true desktop environment. So if the shell panics, it's quite hard to display something indicating that to the user.

Should be done. There are still a couple of unwraps/excepts but they are for some pretty basic stuff like having `stdin` and `stdout`.

> rename shell_2 to something more descriptive? "ng" or something, or perhaps a new name entirely. I had some ideas related to the "Ship of Theseus" theme, e.g., "hull" -- it's the shell of a ship... lol

I like it. Goes in line with `captain`.

> ImmutableWrite/Read sound a bit like my concept of LockableIo -- can we combine those concepts or perhaps adapt LockableIo for your needs?

I'm not sure. For example `Slave` implements `ImmutableWrite/Read`, but internally it uses a queue not locks.

> Last but not least, how does this changeset interact with the "legacy" feature that allows kernel crates (without an app context) to also print to the "default" terminal? I'm primarily concerned about printing exceptions and other important messages both before and after the shell registers app_io streams. Without a desktop environment, there's no other way to do that, considering that the logger isn't available on most real hardware.

I'll add that functionality to `app_io`.

Aside from that, most of the "changes requested" are complete. I've left comments on the ones that haven't yet been resolved.

applications/shell_2/src/lib.rs
kevinaboos: right, the environment is shared with that of the "parent" task. If you don't want that behavior, you can clone the environment, set the child task's env to that new cloned instance, and then modify it.

applications/shell_2/src/wrapper.rs
kevinaboos: I see. Returning an error (e.g., "short write") might be a bit clearer than a panicking assert without a message.

kernel/async_channel/src/lib.rs
kevinaboos: ah i see, thanks for identifying that. Looks like that was added in #277 (during a paper deadline crunch) so the author was likely only testing it with a single endpoint of each variety. Should be a simple enough fix: only mark the channel as disconnected when the _last_ Sender or Receiver gets dropped, not just any sender or receiver.

kevinaboos: I filed #703 about this. For now you can leave this change in there as part of this PR.

kernel/console/src/lib.rs
kevinaboos: hmm, then what happens currently if you close the PTY on the linux host side and then try to re-open it?

kernel/task/src/lib.rs
kevinaboos: I see. So we either need a separate runstate or a "blocked semaphore" that counts how many times a task has been blocked (kind of like how we do preemption holds). 

Q: even with a separate runstate, consider how the following scenario plays out:
1. task is running as normal
2. task blocks itself upon something, e.g., waiting for I/O. 
    --> task is now Blocked
3. user suspends the task using the shell
    --> task is now Suspended (?)
4. the I/O operation completes, unblocking the task
   --> now what should happen? 

The task is no longer blocked but it's also not runnable... we'd need a sort of "stack" to track prior runstate transitions to accomplish this.

Alternatively, should "Suspended" be a completely separate bit of information rather than a runstate? That's potentially even more annoying to handle, but it's also more correct.

kernel/tty/src/channel.rs
kevinaboos: oh right, i skimmed it too quickly. 

Could be useful on its own too, right? Perhaps consider moving it into a separate crate since it's essentially just a wrapper around a channel.

kernel/tty/src/discipline.rs
kevinaboos: neat, TIL you can combine those two forms of hyperlinking

kernel/app_io/src/lib.rs
kevinaboos: oh my apologies. Thanks for fixing it anyway.

kevinaboos:
> I don't think it's possible to reimplement keyboard_echo. I can reimplement immediate_input_echo and input_echo however, this means they won't work with the current shell and GUI, only with `hull` over serial. Once we have a GUI terminal emulator that runs `hull` this won't be an issue. less can be implemented, but it will require a fair bit of work so I'd rather leave that for another PR.

Totally fine with me. Instead of commenting out their content, we can just move them to the `old_crates/` directory (and update. Alternatively, you could create an `app_io_ng` or whatever to service your new shell, such that the original `app_io` would still support those old applications. But I don't personally care about those old apps, only `less` is actually useful; it'd be easy to re-do it later if needed.

> > ImmutableWrite/Read sound a bit like my concept of LockableIo -- can we combine those concepts or perhaps adapt LockableIo for your needs?
> 
> I'm not sure. For example `Slave` implements `ImmutableWrite/Read`, but internally it uses a queue not locks.

np, just ignore it for now.

> Aside from that, most of the "changes requested" are complete. I've left comments on the ones that haven't yet been resolved.

Cool. I've left a few responses to comments as well. Let me know when things are ready to review again. 👍

kernel/console/src/lib.rs
tsoutsman: the host's terminal hangs

kevinaboos: ok yeah, so then we need to do a more thorough cleanup/reset procedure to allow another host terminal to connect.

applications/shell_2/src/lib.rs
tsoutsman: Turns out this is already done in `Task::new`. The old shell was intentionally overwriting the child's environment with a reference to the parent's environment.

kernel/task/src/lib.rs
tsoutsman: I added a separate `suspended` field to `Task` in 2f0d95c.

kernel/tty/src/channel.rs
tsoutsman: Moved the methods to `async_channel` in 595948c. Now the module is just to convert the `async_channel::Error` to `core2::io::Error`.

kernel/console/src/lib.rs
tsoutsman: `console` doesn't detect the second connection, and since this isn't a pressing issue I'd rather leave it for a future PR.

kevinaboos:
@tsoutsman Now that #707 is merged in and #703 is closed, hopefully the issues with async_channel here are resolved. (no rush)

tsoutsman:
> @tsoutsman Now that https://github.com/theseus-os/Theseus/pull/707 is merged in and https://github.com/theseus-os/Theseus/issues/703 is closed, hopefully the issues with async_channel here are resolved. (no rush)

Done.

kevinaboos (approved):
Quite a few to-do items left, but we can go ahead and merge this in order to move forward and not let such a large changeset linger and get stale.

Issue #655, opened 2022-10-07
GPF on x2apic systems when trying to disable the LVT timer


kevinaboos:
Not sure why this happens, but a General Protection Fault occurs on x2apic systems sometimes when trying to disable the Local APIC's LVT timer, which happens when disabling preemption on a CPU. As far as I can tell, the rest of the x2apic code is completely fine.

## Steps to Reproduce
If your host system has an x2apic, which is quite likely, then simply run:
```
make run host=yes
```

Output:
```
[D] kernel/apic/src/lib.rs:132: is x2apic? true. IA32_APIC_BASE (phys addr): FEE00900
[D] kernel/acpi/src/lib.rs:45: RXSDT is located in Frame 0x1FFE1685
[T] kernel/acpi/acpi_table/src/lib.rs:82: [0] Dropping original MappedPages { page_table_p4: Frame(p0x0), pages: AllocatedPages(Page(v0x1000)..=Page(v0x0)), flags: GLOBAL }
[D] kernel/acpi/hpet/src/lib.rs:191: Initialized HPET, period: 10000000, counter val: 1009, num timers: 3, vendor_id: 32902
[D] kernel/ioapic/src/lib.rs:96: Created new IoApic, id: 0, gsi_base: 0, phys_addr: 0xFEC00000
[T] kernel/apic/src/lib.rs:350: size of LocalApic: 24
[I] kernel/apic/src/lib.rs:436: LAPIC x2 ID 0x0, version: 0x50014, is_bsp: true
[I] kernel/apic/src/lib.rs:447: x2LAPIC ID 0x0, version 0x50014, (cluster 0x0 logical 0x1), is_bsp: true
[T] kernel/apic/src/lib.rs:528: LocalApic 0, timer period count: 502656 (0x7AB80)
[I] kernel/apic/src/lib.rs:389: Found new processor core (LocalApic { type: x2apic, apic_id: 0, processor: 0, is_bsp: true })
[T] kernel/acpi/madt/src/lib.rs:347: MadtIntSrcOverride (bus: 0, irq: 0, gsi: 2, flags 0x0) handled by IoApic 0
[T] kernel/acpi/madt/src/lib.rs:347: MadtIntSrcOverride (bus: 0, irq: 5, gsi: 5, flags 0xD) handled by IoApic 0
[...]

PR #670, opened 2022-10-26
Set initial sentry value for IA32_TSC_AUX MSR

kevinaboos:
Ensures that it can't be misinterpreted as a valid APIC ID by early code before APIC initialization has occurred. 

Partial fix for #655; should trigger the issue on x2apic systems, but it's not completely addressed yet.


kevinaboos:
Closed in favor of #676

PR #674, opened 2022-10-27
Clarify safety requirements for `slice::from_raw_parts[_mut]()`

kevinaboos:
Currently this is relevant only in `MappedPages::as_slice()` and `MappedPages::as_slice_mut()`

Credit to @tsoutsman for identifying this in #664

### DIFF
--- a/kernel/memory/src/paging/mapper.rs
+++ b/kernel/memory/src/paging/mapper.rs
@@ -893,6 +893,7 @@ impl MappedPages {
         // SAFETY:
         // ✅ same as for `MappedPages::as_slice()`, plus:
         // ✅ The underlying memory is not accessible through any other pointer, as we require a `&mut self` above.
+        // ✅ The underlying memory can be mutated because it is mapped as writable (checked above).
         let slc: &mut [T] = unsafe {
             slice::from_raw_parts_mut(start_vaddr as *mut T, length)
         };
--- a/kernel/memory/src/paging/mapper.rs
+++ b/kernel/memory/src/paging/mapper.rs
@@ -805,8 +805,8 @@ impl MappedPages {
             );
         }
 
-        if size_in_bytes >= isize::MAX as usize {
-            return Err("MappedPages::as_slice(): length * size_of::<T>() must be less than isize::MAX");
+        if size_in_bytes > isize::MAX as usize {
+            return Err("MappedPages::as_slice(): length * size_of::<T>() must be no larger than isize::MAX");
         }
 
         if byte_offset % mem::align_of::<T>() != 0 {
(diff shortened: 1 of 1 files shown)


kernel/memory/src/paging/mapper.rs
```diff
@@ -877,7 +891,9 @@ impl MappedPages {
             return Err("MappedPages::as_slice_mut(): requested slice length and byte_offset would not fit within the MappedPages bounds");
         }
 
-        // SAFE: we guarantee the bounds and lifetime are within that of this MappedPages object
+        // SAFETY:
+        // ✅ same as for `MappedPages::as_slice()`, plus:
+        // ✅ The underlying memory is not accessible through any other pointer, as we require a `&mut self` above.
```
tsoutsman, line 896: We've also checked that the underlying memory is writable. That's mentioned in `BorrowedMappedPages::deref_mut`.

kernel/memory/src/paging/mapper.rs
```diff
@@ -805,6 +805,10 @@ impl MappedPages {
             );
         }
 
+        if size_in_bytes >= isize::MAX as usize {
+            return Err("MappedPages::as_slice(): length * size_of::<T>() must be less than isize::MAX");
```
tsoutsman:
```suggestion
        if size_in_bytes > isize::MAX as usize {
            return Err("MappedPages::as_slice(): length * size_of::<T>() must be no larger than isize::MAX");
```

kernel/memory/src/paging/mapper.rs
```diff
@@ -847,6 +857,10 @@ impl MappedPages {
                 length, size_in_bytes, byte_offset, self.size_in_bytes()
             );
         }
+
+        if size_in_bytes >= isize::MAX as usize {
+            return Err("MappedPages::as_slice_mut(): length * size_of::<T>() must be less than isize::MAX");
+        }
```
tsoutsman:
```suggestion
        if size_in_bytes > isize::MAX as usize {
            return Err("MappedPages::as_slice_mut(): length * size_of::<T>() must be no larger than isize::MAX");
```

tsoutsman (changes requested):

kernel/memory/src/paging/mapper.rs
kevinaboos: yup, thought about it and forgot to commit it. thanks!

kernel/memory/src/paging/mapper.rs
kevinaboos: wow, must be tired... i misread the req from the `slice::from_raw_parts` docs

tsoutsman (approved):

PR #675, opened 2022-10-28
Redesign `TaskRef`

tsoutsman:
This builds on #657 to ensure blocked tasks are unblocked during the unwinding of the blocker. As a side effect, it makes spurious wakeups impossible, as a task can only be unblocked by the `TaskRef` that blocked it in the first place.

However, rather than use a conventional guard, I've implemented this using generics on `TaskRef`. I also merged `TaskRef` and `JoinableTaskRef`.

There are three unavoidable problems that arise from using guards, be that in the form of generics or a conventional guard struct:
The `run_and_block` shenanigans. `sleep` and `wait_queue` both need to add the guard to their internal queues before blocking the task to avoid race conditions.
The shell enum needing to store different variants. In the case of a conventional guard, the guard struct would need to be optionally stored in the list of tasks.
`serial_port` having to recreate the guard. The task is blocked either on creation or in `deferred_interrupt_tasks::deferred_task_entry_point()`, but it must be unblocked by the interrupt handler. Unlike the preemption guard stored across context switches, there isn't a good place to store the unblockable `TaskRef`. I currently recreate the unblockable `TaskRef` in the interrupt handler, but this is obviously an awful solution.

An issue specific to generics is the `TaskRefKind::Temp` enum variant found in `shell`. It's a workaround for a common problem with APIs that consume `self` in their methods. Moving out of a mutable reference without replacing it with some value is impossible, even if temporary. `*self = f(self)` does not work; if `f` panics, `self' will have an invalid value during unwinding as it has been moved out of but not reassigned to.

Having said that, I feel the benefits outweigh the costs compared to a conventional guard. Not having to store two separate structs (both of which would be an `Arc` pointing to the same task) is convenient. For example, spawning a task with conventional guards would require `spawn()` to return either a `TaskRef` or `(TaskRef, BlockGuard)`, depending on whether the task was initially blocked. The additional compile-time information granted by generics is very powerful, as I'll describe below.

As a side note, I'm not a big fan of using const generic booleans the way I did. I think expressing the generic parameters using sealed marker types would be better, i.e. `Joinable`/`Unjoinable` and `Blockable`/`Unblockable`, as they are more descriptive and let us expand the potential states (e.g. `Killed`).

Const generics cannot accurately track the run state of the task as they are a compile-time concept. However, they let us reduce the set of possible run states as we know the set of all possible state transitions, encoded as methods on `TaskRef`.

Using the sealed types I suggested above for clarity, if we have a `TaskRef<_, Unblockable>` (created by `task.block()`), we know that the task is either blocked (as only we can unblock it), killed, or reaped.

Similarly, if we were to add a `Killed` variant, we know that a `TaskRef<_, Killed>` could only be killed or reaped as no state transition would turn a killed task back into a runnable or blocked one.

`TaskRef<_, Unblockable>` does not have a `block` method because the `TaskRef`'s existence guarantees that it is not in the `Runnable` state. Currently, there is a way to get a `TaskRef<_, Blockable>` from a `TaskRef<_, Unblockable>` through `clone`, but that's for ergonomic reasons to avoid requiring generics in the scheduler. Ideally, we would prevent this state transition, but even as it is, it requires an additional function call to suggest to the consumer that what they are doing does not make sense. And, even though this invalid state transition is possible, performing it at runtime and trying to block the task will result in an error.

The generics are just additional compile-time information that can be used to better restrict interfaces.

I haven't added much documentation yet because I'm not sure if this is something we would want to merge in, but obviously, everything will be documented if we do decide to go ahead.

### DIFF
--- a/applications/channel_eval/src/lib.rs
+++ b/applications/channel_eval/src/lib.rs
@@ -96,8 +96,8 @@ fn test_multiple(iterations: usize) -> Result<(), &'static str> {
         .spawn()?;
 
     info!("test_multiple(): Finished spawning the sender and receiver tasks");
-    t2.unblock().unwrap();
-    t1.unblock().unwrap();
+    let t2 = t2.unblock().unwrap();
+    let t1 = t1.unblock().unwrap();
 
     t1.join()?;
     t2.join()?;
--- a/applications/channel_eval/src/lib.rs
+++ b/applications/channel_eval/src/lib.rs
@@ -96,7 +96,8 @@ fn test_multiple(iterations: usize) -> Result<(), &'static str> {
         .spawn()?;
 
     info!("test_multiple(): Finished spawning the sender and receiver tasks");
-    t2.unblock(); t1.unblock();
+    t2.unblock().unwrap();
+    t1.unblock().unwrap();
 
     t1.join()?;
     t2.join()?;
--- a/applications/rq_eval/src/lib.rs
+++ b/applications/rq_eval/src/lib.rs
@@ -157,7 +157,7 @@ fn run_single(iterations: usize) -> Result<(), &'static str> {
         |_, _| loop { }, // dummy failure function
     )?;
     task.name = String::from("rq_eval_single_task_unrunnable");
-    let taskref = TaskRef::new(task);
+    let taskref = task.init();
     
     let hpet = get_hpet().ok_or("couldn't get HPET timer")?;
     let start = hpet.get_counter();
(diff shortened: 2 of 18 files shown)


tsoutsman:
I've marked `into_kind` and `run_and_block` because calling them with invalid inputs won't violate memory safety. There's some discussion about "expert" functions [here](https://www.reddit.com/r/rust/comments/tphnqf/is_there_a_standard_way_to_mark_some_functions_in/), but I don't like the fact that these functions are exposed without any checks. Here's an [interesting discussion](https://www.reddit.com/r/rust/comments/dqfvkd/statically_checkable_function_propertiesmultiple/) about custom unsafe blocks. but I don't think we can feasibly use that approach. Ideally, this would be implemented in Rust itself...

PR #676, opened 2022-10-29
Redesign Local APIC init to not assume the BSP's APIC ID

kevinaboos:
Part of the solution for #655, not yet complete.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,7 +90,6 @@ dependencies = [
  "scheduler",
  "spawn",
  "stack",
- "tlb_shootdown",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


kevinaboos:
This solves the GPF issue in #655 but the full shell still won't boot on KVM using `make run host=yes`, so something else must be wrong. However, this change does let the boot process finish. There could be issues with `spin_loop_hint` or with timeslice interval calculations (might be too short?), but whatever problem is causing this is definitely beyond the scope of this PR.

kevinaboos:
The problem is that the APIC timer calibration is wrong, which results in the timeslice period being VERY short, and effectively just constantly context switching on all CPUs incessantly.

With a fixed large APIC LVT timer calibration value (e.g., `cfg(apic_timer_fixed)` setting a value of 0x10000000), the system works perfectly well.

Issue #587, opened 2022-08-05
Remove patches from `Cargo.toml`


tsoutsman:
Our eventual goal is to have `std` depend on `libtheseus`. This means getting rid of the patch tables in `Cargo.toml`, as they only work within a workspace:
- [x] `spin`

  ```toml
  ### Patch `spin` to use the `pause` asm instruction in busy-wait loops,
  ### because the `core::hint::spin_loop()` only uses it if "sse2" is enabled.
  spin = { git = "https://github.com/theseus-os/spin-rs" }
  ```
- [ ] `volatile`

  ```toml
  ### use our own version of volatile which supports zerocopy
  volatile = { git = "https://github.com/theseus-os/volatile" }
  ```
- [ ] `getopts`

  ```toml
  ### use our own no_std-compatilbe getopts
  getopts = { git = "https://github.com/theseus-os/getopts" }
  ```
- [ ] `smoltcp` - we can probably just depend on `v0.5.0`
  
  ```toml
  ### use the latest version of smoltcp from github; the one on crates.io is out of date
  smoltcp = { git = "https://github.com/m-labs/smoltcp", rev = "0fedb1db9aa26712830822dd61f065deaa34d611" }
  ```
- [ ] `libc` - either upstream or remove entirely

  ```toml
  ### Patch `libc` so we can use libc-specific types when using `cfg(target_os = "theseus")`.
  libc = { path = "ports/libc" }
  ```
- [ ] `core2`

  ```toml
  ### Patch `core2` with newer functions from `std::io`, e.g., additional `Seek` trait functions
  core2 = { path = "libs/core2" }
  ```
- [ ] `bincode` - This will eventually be fixed upstream

  ```toml
  ### Patch `bincode` because the version on crates.io doesn't handle no_std features correctly.
  bincode = { git = "https://github.com/bincode-org/bincode" }
  ```
### Wasmtime-related crates
- [ ] `wasmparser`
- [ ] `backtrace`
- [ ] `region`
- [ ] `thiserror_core2`

kevinaboos:
- `spin`:  we can probably just use the upstream directly; that needs some testing though to ensure system perf isn't affected.

- `volatile`:  our fork of `volatile` has diverged significantly from the original crate; we should simply rename that and publish it as a completely different crate, with `zerocopy` being an optional feature/dependency.
 
- `getopts`:  not upstreamable because this project is a rustc dependency. Similar to volatile we could republish it as a new package.
 
[...]

PR #677, opened 2022-10-31
Add new networking API

tsoutsman:
`smoltcp`'s API has significantly improved since `0.5` (the version used by Theseus). I'm working on networking as part of the `std` port and so it seems like a good time to update Theseus' networking stack. The eventual goal of the new `net` crate would be to replace `ethernet_smoltcp_device`, `network_interface_card`, `network_manager`, and `smoltcp_helper`.

As part of the PR, I also modified the networking buffers to return an error on creation rather than on access. I think the buffer abstractions could potentially be improved but that would require significant modifications to the NIC code.

I haven't yet implemented any time-related functionality because #615 will soon be merged.

To limit the changes in this PR, I've implemented the new trait only for the E1000 driver. I've added a basic `ping_2` crate that uses the new networking stack.


kernel/nic_buffers/src/lib.rs
```diff
@@ -68,14 +77,35 @@ pub struct ReceiveBuffer {
...
+        self.mp.as_slice(0, usize::from(self.length)).unwrap() 
+    }
+    
+    pub fn as_slice_mut(&mut self) -> &mut [u8] {
+        // We checked that the mapped pages are >= to self.length during initialisation
+        // and that they are writable. There can be no overflows since length is
+        // a u16, nor can there be alignment issues because we are operating on
+        // u8s.
+        self.mp.as_slice_mut(0, usize::from(self.length)).unwrap()
+    }
```
kevinaboos: This is a great idea for a redesign! I like it and it fits well with the theme of early error detection.

The only other thing I'd possibly be concerned with is whether one can obtain a mutable reference to the `MappedPages` object within a `ReceiveBuffer`. This would allow someone to swap in a different `MappedPages` with a completely different size and location.
We don't want to let that happen; I don't _think_ it's possible but I haven't personally gone through the code to verify that. Aside from that, everything here looks good.

kernel/net/src/device.rs
```diff
@@ -0,0 +1,113 @@
...
+/// Standard maximum transition unit for ethernet cards.
+const STANDARD_MTU: usize = 1500;
+
+/// A network device.
+///
+/// Devices implementing this trait can then be registered using
+/// [`register_device`].
+///
+/// [`register_device`]: crate::register_device
+pub trait Device: Send + Sync + Any {
```
kevinaboos: Let's call this `NetworkDevice` in order to disambiguate and improve searchability/discoverability.

kernel/net/src/device.rs
```diff
@@ -0,0 +1,113 @@
...
+    where
+        F: FnOnce(&mut [u8]) -> smoltcp::Result<R>,
+    {
+        if self.inner.0.len() > 1 {
+            error!(
+                "DeviceWrapper::receive(): received frame with {} buffers",
+                self.inner.0.len()
+            );
+        }
+        // TODO: Which error variant?
```
kevinaboos: `Exhausted` i believe.

It could also be `Finished`, though I don't think we have sufficient information here to determine whether the remote end was closed

kernel/net/src/interface.rs
```diff
@@ -0,0 +1,67 @@
...
+use irq_safety::MutexIrqSafe;
+use smoltcp::{iface, phy::DeviceCapabilities, wire};
+
+pub use smoltcp::iface::SocketSet;
+pub use wire::{IpAddress, IpCidr};
+
+#[derive(Clone)]
+pub struct Interface {
+    // FIXME: Can this be a regular mutex?
+    inner: Arc<MutexIrqSafe<iface::Interface<'static>>>,
```
kevinaboos: if/when we use deferred interrupt handlers here, then yes, we'd be able to partition the NIC type into things that need to be accessed from the interrupt handler and things that only need to be accessed in the deferred handler (or any other normal execution context). 

But for now it needs to be wrapped in this because it's jointly accessed from the NIC's interrupt handler.

kernel/net/src/interface.rs
```diff
@@ -0,0 +1,67 @@
+use crate::{device::DeviceWrapper, Device, Result};
+use alloc::{collections::BTreeMap, sync::Arc};
+use irq_safety::MutexIrqSafe;
+use smoltcp::{iface, phy::DeviceCapabilities, wire};
+
+pub use smoltcp::iface::SocketSet;
+pub use wire::{IpAddress, IpCidr};
+
+#[derive(Clone)]
+pub struct Interface {
```
kevinaboos: add some docs to describe the role of this struct and why we need it

kernel/net/src/interface.rs
```diff
@@ -0,0 +1,67 @@
...
+                .routes(routes)
+                .neighbor_cache(iface::NeighborCache::new(BTreeMap::new()))
+                .finalize(&mut wrapper),
+        ));
+
+        Self { inner, device }
+    }
+
+    /// Transmit packets queued in the given sockets, and receive queued
+    /// packets.
```
kevinaboos:
```suggestion
    /// Transmit and receive any packets queued in the given `sockets`.
```

kernel/net/src/interface.rs
```diff
@@ -0,0 +1,67 @@
...
+    ///
+    /// Returns a boolean value indicating whether any packets were processed or
+    /// emitted, and thus, whether the readiness af any socket might have
```
kevinaboos:
```suggestion
    /// emitted, and thus, whether the readiness of any socket might have
```

kernel/net/src/interface.rs
kevinaboos: oh wait on second thought, this doesn't appear to be a type that accesses any NIC-specific h/w stuff during an interrupt handler. 
Hmm, what type was this derived from in the existing main branch?

kernel/net/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+
+/// A randomly chosen IP address that must be outside of the DHCP range.
+///
+/// The default QEMU user-slirp network gives IP address of `10.0.2.*`.
+const DEFAULT_LOCAL_IP: &str = "10.0.2.15/24";
+
+/// Standard home router address.
+///
+/// `10.0.2.2` is the default QEMU user-slirp networking gateway IP.
+const DEFAULT_GATEWAY_IP: IpAddress = IpAddress::Ipv4(Ipv4Address::new(10, 0, 2, 2));
```
kevinaboos, line 27: when we first implemented networking, smoltcp didn't support DHCP. I think it does now, so it'd be nice to replace these with those obtained from the DHCP reservation request. Could also save it for another PR but it's definitely something we'd need to do prior to supporting `std` in full.

kernel/net/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+
+// TODO: Make mutex rwlock?
+// TODO: Use atomic append-only vec?
+static NETWORK_INTERFACES: Mutex<Vec<Interface>> = Mutex::new(Vec::new());
```
kevinaboos: wrapping the inner registered interfaces in an Arc will relieve the multi-locking contention issues too.

kernel/net/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+///
+/// The function will convert the device to an interface and it will then be
+/// accessible using [`get_interface`].
+pub fn register_device<T>(device: &'static MutexIrqSafe<T>)
+where
+    T: 'static + Device + Send,
+{
+    let interface = Interface::new(
+        device,
+        // TODO: use DHCP to acquire an IP address and gateway.
```
kevinaboos, line 43: ah yea, just saw this. :+1:

kernel/net/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+pub fn get_interfaces() -> &'static Mutex<Vec<Interface>> {
+    &NETWORK_INTERFACES
+}
+
+/// Gets the interface with the specified `index`.
+///
+/// If `index` is `None` the first interface is returned.
+pub fn get_default_interface() -> Option<Interface> {
+    NETWORK_INTERFACES.lock().get(0).cloned()
+}
```
kevinaboos: mismatch between docs and functionality

kernel/nic_buffers/src/lib.rs
```diff
@@ -29,17 +29,26 @@ impl TransmitBuffer {
...
+        self.mp.as_slice(0, self.length.into()).unwrap() 
+    }
+    
+    pub fn as_slice_mut(&mut self) -> &mut [u8] {
+        // We checked that the mapped pages are >= to self.length during initialisation
+        // and that they are writable. There can be no overflows since length is
+        // a u16, nor can there be alignment issues because we are operating on
+        // u8s.
+        self.mp.as_slice_mut(0, self.length.into()).unwrap()
+    }
```
kevinaboos: same comments apply here as with `ReceiveBuffer` -- we must ensure the inner `MappedPages` isn't accessible. Again, I don't think it is but we ought to double check and then explicitly state it in the comments.

kevinaboos (changes requested):
Nice work on this PR! 

I've recently merged in the changeset that migrates from `owning_ref` types to `BorrowedMappedPages`, so you'll need to grab those changes here too since they're directly relevant to `ethernet_smoltcp_device`. I think those changes also hold the answers to some of your `TODO`s about which error type to return -- typically it's going to be `Exhausted` unless you know some part of the packet is malformed (which is rare at these layers of the net stack).

It'd be a good idea to add some top crate-level docs describing the interaction/relationship between each type and trait in the new `net` crate, just something to aid understanding at a glance. All other changes look quite good, and the comments I left are relatively minor in scope.

kernel/net/src/lib.rs
tsoutsman: I'm not sure what the contention issues are. Cloning an `Interface` is basically cloning an `Arc` because its two fields are an `Arc` and a reference. I'm not sure wrapping it in an additional `Arc` will help.

Having said that I'd rather use an `Arc<Interface>>`, rather than have `interface.inner` be an `Arc` as currently, it isn't immediately clear that `Interface` is essentially a reference.

kernel/nic_buffers/src/lib.rs
tsoutsman: should be addressed by https://github.com/theseus-os/Theseus/pull/677/commits/e5dde86deb35d9639afa89bb4d7ba95d207630cd

kernel/nic_buffers/src/lib.rs
tsoutsman: should be addressed by https://github.com/theseus-os/Theseus/pull/677/commits/e5dde86deb35d9639afa89bb4d7ba95d207630cd

kernel/net/src/lib.rs
tsoutsman: I quickly tried implementing it but it seems trickier than just polling a socket. I'd rather do it in a separate PR so I'll create an issue for it once this PR lands.

tsoutsman:
I've merged in `theseus_main` but it now fails to build on both my machine and CI with the following error:
```
thread 'main' panicked at 'not currently active!?', src/tools/cargo/src/cargo/core/resolver/mod.rs:832:20
```
https://github.com/rust-lang/cargo/issues/11336

I haven't yet properly merged the two branches (e.g. there are still calls to `as_slice_mut`) but that's because I couldn't test it.

kevinaboos:
> I've merged in `theseus_main` but it now fails to build on both my machine and CI with the following error:
> 
> ```
> thread 'main' panicked at 'not currently active!?', src/tools/cargo/src/cargo/core/resolver/mod.rs:832:20
> ```
> 
> [rust-lang/cargo#11336](https://github.com/rust-lang/cargo/issues/11336)
> 
> I haven't yet properly merged the two branches (e.g. there are still calls to `as_slice_mut`) but that's because I couldn't test it.

Yikes, that's no good. If you downgrade to an earlier nightly, does this issue disappear? Sorry for the troubles randomly introduced in newer Rust versions, that rarely happens. :disappointed:

kevinaboos:
Hmm, on that cargo issue you linked, there is a comment mentioning that this bug may occur when

> Your package depends on two SemVer-compatible versions of a single package, which is not currently possible in Cargo.

In your current branch, this does occur with `smoltcp`. If you remove the old version entirely, does this go away?

tsoutsman:
> In your current branch, this does occur with smoltcp. If you remove the old version entirely, does this go away?

@kevinaboos would that involve removing the old network crates?

kevinaboos:
> > In your current branch, this does occur with smoltcp. If you remove the old version entirely, does this go away?
> 
> @kevinaboos would that involve removing the old network crates?

Yeah, I don't believe there's another way. You can try it out in a new branch/PR too, in order to keep this one intact.
Perhaps a newer nightly version fixes it? Though it seems to be a fundamental limitation within cargo

tsoutsman:
Removing the `spin` patch fixes it. In #587, you mentioned that the patch could probably be removed.

I compiled the following crate in release mode:
```rust
#![no_std]

extern crate panic_entry;
extern crate captain;

#[no_mangle]
pub fn temp(mutex: &spin::Mutex<usize>) {
    let x = mutex.lock();
    black_box(*x);
}

#[no_mangle]
#[inline(never)]
pub fn black_box(a: usize) {
    log::debug!("{a}");
}
```

Both the basic and SSE targets result in the exact same disassembly:
```asm
0000000000000000 <temp>:
   0:	53                   	push   rbx
   1:	48 83 ec 10          	sub    rsp,0x10
   5:	48 89 fb             	mov    rbx,rdi
   8:	b1 01                	mov    cl,0x1
   a:	31 c0                	xor    eax,eax
   c:	f0 0f b0 0f          	lock cmpxchg BYTE PTR [rdi],cl
  10:	75 10                	jne    22 <temp+0x22>
  12:	eb 1d                	jmp    31 <temp+0x31>
  14:	66 66 66 2e 0f 1f 84 	data16 data16 cs nop WORD PTR [rax+rax*1+0x0]
  1b:	00 00 00 00 00 
  20:	f3 90                	pause
  22:	0f b6 03             	movzx  eax,BYTE PTR [rbx]
  25:	84 c0                	test   al,al
  27:	75 f7                	jne    20 <temp+0x20>
  29:	31 c0                	xor    eax,eax
  2b:	f0 0f b0 0b          	lock cmpxchg BYTE PTR [rbx],cl
  2f:	75 f1                	jne    22 <temp+0x22>
  31:	48 8d 43 08          	lea    rax,[rbx+0x8]
  35:	48 89 1c 24          	mov    QWORD PTR [rsp],rbx
  39:	48 89 44 24 08       	mov    QWORD PTR [rsp+0x8],rax
  3e:	48 8b 7b 08          	mov    rdi,QWORD PTR [rbx+0x8]
  42:	48 b8 00 00 00 00 00 	movabs rax,0x0
  49:	00 00 00 
  4c:	ff d0                	call   rax
  4e:	c6 03 00             	mov    BYTE PTR [rbx],0x0
  51:	48 83 c4 10          	add    rsp,0x10
  55:	5b                   	pop    rbx
  56:	c3                   	ret
  57:	48 89 c3             	mov    rbx,rax
  5a:	48 b8 00 00 00 00 00 	movabs rax,0x0
  61:	00 00 00 
  64:	48 89 e7             	mov    rdi,rsp
  67:	ff d0                	call   rax
  69:	48 b8 00 00 00 00 00 	movabs rax,0x0
  70:	00 00 00 
  73:	48 89 df             	mov    rdi,rbx
  76:	ff d0                	call   rax
  78:	0f 0b                	ud2
```
which has the `pause` instruction, so I think the patch is safe to remove.

kernel/net/src/interface.rs
tsoutsman: I think this is equivalent to `smoltcp::EthernetInterface<'_, '_, '_, EthernetDevice<N>>` of `EthernetNetworkInterface.iface` in `ethernet_smoltcp_device`. The new versions of `smoltcp` separate the interface and device. The interface doesn't store the device, but polling it requires a reference to the device.

kevinaboos:
> Removing the `spin` patch fixes it. 

Really? wow, I did not foresee that. So, it's currently ready to merge, and the multiple versions of smoltcp _don't_ cause any issues?

> In #587, you mentioned that the patch could probably be removed.

Yes, I believe so. I would have to do some significant testing with this first, but if that's true then I will **gladly** remove that dumb patch for `spin`.

tsoutsman:
> So, it's currently ready to merge

Yep. Aside from the `spin` testing.

kernel/net/src/interface.rs
kevinaboos: Interesting! Since I'm not intimately familiar with the new smoltcp design, can you state with certainty that this is never accessed or accessible from any NIC interrupt handler (just e1000 for now i guess)?  If so, meaning that it's only ever accessed from a regular execution (non-interrupt handler) context, then yes we can reduce this to a regular `Mutex` instead of an interrupt-safe one.

tsoutsman: Yep. The implementor crates only access the concrete device type e.g.:
```rust
static E1000_NIC: Once<MutexIrqSafe<E1000Nic>> = Once::new();
```

The interface is a layer on top of the device that the interrupt handler never needs to touch.

I changed it to use a `MutexSleep` in [d9ab2e7](https://github.com/theseus-os/Theseus/pull/677/commits/d9ab2e75c0d23c70dc2d10cd2f86e3e324369b40).

kevinaboos (approved):
LGTM, great work! Thanks.

PR #688, opened 2022-11-07
Add initial `aarch64` crates for basic boot-up

NathanRoyer:
This adds the `aarch64` directory, which contains code for Arm and has the same structure as the root directory.

### DIFF
--- a/aarch64/Makefile
+++ b/aarch64/Makefile
@@ -1,6 +1,6 @@
 TARGET=aarch64-unknown-uefi
 
-CARGOFLAGS += -p nano_core
+CARGOFLAGS += --workspace
 CARGOFLAGS += --release
 
 BUILD_STD_CARGOFLAGS += -Z unstable-options
@@ -41,7 +41,7 @@ create-build-structure:
 cargo-build: create-build-structure
 	@echo -e "\n=================== BUILDING ALL CRATES ==================="
 	@echo -e "\t TARGET: \"$(TARGET)\""
-	cargo +nightly build $(CARGOFLAGS) $(BUILD_STD_CARGOFLAGS) --target $(TARGET)
+	cargo build $(CARGOFLAGS) $(BUILD_STD_CARGOFLAGS) --target $(TARGET)
 
 ## This builds the nano_core binary itself, which is the fully-linked code that first runs right after the bootloader
 $(nano_core_binary): cargo-build
--- a/aarch64/Makefile
+++ b/aarch64/Makefile
@@ -1,0 +1,53 @@
+TARGET=aarch64-unknown-uefi
+
+CARGOFLAGS += -p nano_core
+CARGOFLAGS += --release
+
+BUILD_STD_CARGOFLAGS += -Z unstable-options
+BUILD_STD_CARGOFLAGS += -Z build-std=core,alloc
+BUILD_STD_CARGOFLAGS += -Z build-std-features=compiler-builtins-mem
+
+AARCH64_DIR = .
+BUILD_DIR = $(AARCH64_DIR)/build
+QEMU_EFI_FD = $(AARCH64_DIR)/resources/qemu-efi.fd
+NATIVE_OUTPUT = $(AARCH64_DIR)/target/aarch64-unknown-uefi/release/nano_core.efi
+nano_core_binary = $(BUILD_DIR)/efi
+
+# Configure the machine
+QEMU_FLAGS += -machine virt
+QEMU_FLAGS += -cpu cortex-a72
+QEMU_FLAGS += -smp 4
+QEMU_FLAGS += -m 1000
+QEMU_FLAGS += -net none
+# QEMU_FLAGS += -nographic
+
+# Include that file which enables efi support in qemu
+QEMU_FLAGS += -drive if=pflash,format=raw,file=$(QEMU_EFI_FD)
+
+# Include the build dir as an emulated fat drive
+QEMU_FLAGS += -drive file=fat:rw:$(BUILD_DIR)
+
+
+
+
+
+
+all: run
+
+create-build-structure:
+	mkdir -p $(BUILD_DIR)
+
+## This target invokes the actual Rust build process via `cargo`.
+cargo-build: create-build-structure
+	@echo -e "\n=================== BUILDING ALL CRATES ==================="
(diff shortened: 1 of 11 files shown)


aarch64/Makefile
```diff
@@ -0,0 +1,53 @@
+TARGET=aarch64-unknown-uefi
+
+CARGOFLAGS += -p nano_core
```
kevinaboos: what's this for, just to build only the one `nano_core` package? I'm not intimately familiar with all the requirements of building for UEFI, so perhaps this is wrong, but another approach that might fit in better with the existing build system is to define a `default-members` and `members` set such that:
* `default-members` includes just the `nano_core` (which will bring in all of its dependencies)
* `members` includes all crates in `kernel/`
That way we can build more than just one package in the future

aarch64/Makefile
```diff
@@ -0,0 +1,53 @@
...
+all: run
+
+create-build-structure:
+	mkdir -p $(BUILD_DIR)
+
+## This target invokes the actual Rust build process via `cargo`.
+cargo-build: create-build-structure
+	@echo -e "\n=================== BUILDING ALL CRATES ==================="
+	@echo -e "\t TARGET: \"$(TARGET)\""
+	cargo +nightly build $(CARGOFLAGS) $(BUILD_STD_CARGOFLAGS) --target $(TARGET)
```
kevinaboos: just specifying `+nightly` can unfortunately lead to nonrepeatable builds (ask me how I know.... 😓). 

Do we need to pin the aarch64 build to a different nightly Rust version than the main repo? If so, let's add a `rust-toolchain` file to the `/aarch64` directory to ensure a repeatable build. If not, i _think_ the `rust-toolchain` file in the root should apply to this build too, but I'm not 100% sure about that.

aarch64/kernel/nano_core/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+    unsafe { asm!("mov x1, #0xbeef") };
+    loop {}
+}
+
+#[entry]
+fn main(
+    handle: Handle,
+    mut system_table: SystemTable<Boot>,
+) -> Status {
+    init(&mut system_table).unwrap();
```
kevinaboos: nit: for `init` or similarly vague function calls, don't import it and instead preface it with the crate/module name such that it's clearer and more readable.

aarch64/kernel/nano_core/src/main.rs
```diff
@@ -0,0 +1,51 @@
...
+extern "C" fn inf_loop_0xbeef() -> ! {
+    unsafe { asm!("mov x1, #0xbeef") };
+    loop {}
+}
+
+#[entry]
+fn main(
+    handle: Handle,
+    mut system_table: SystemTable<Boot>,
+) -> Status {
```
kevinaboos: Instead of unwrapping all the Results in this function, return them. It might be easiest to move everything below to a separate function that returns a Result such that you can use the try `?` operator, and then convert that `Result` into `Status` in this top-level `main` entry point function.

kevinaboos (changes requested):
Looks good so far! I suggest adding a small `README.md` file that gives a basic overview of things so far, at a bare minimum the required packages that one needs to install and the command(s) needed to build & run.

One blocker to merging this in is the huge `qemu-efi.fd` file. Can you describe what that is and why it's necessary? Perhaps we could obtain it in a separate build step or host system setup step, because it's too large and opaque to be appropriate for version control.

aarch64/Makefile
NathanRoyer: Indeed, I will remove this to just use our rust-toolchain file. By the way, I think this file should now be a TOML file, based on https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file; that new format also allows specifying a toolchain by path, as well as the required components.

NathanRoyer:
> One blocker to merging this in is the huge `qemu-efi.fd` file. Can you describe what that is and why it's necessary? Perhaps we could obtain it in a separate build step or host system setup step, because it's too large and opaque to be appropriate for version control.

I'm not entirely sure what it is; it's a file which seems to enable UEFI in Qemu. From what I read online, it apparently comes from the UEFI development kit ("OVMF EDK2"). To me it's not a priority to trace its origins but that binary blob could raise security concerns, to be honest.

I agree with you, it's big. As a temporary measure I added a downloading step to the `run` make target, so that it's downloaded from my home server. I'd like you to choose an online location where we could properly host it, if you agree.

kevinaboos:
@NathanRoyer can you undo the last 2 commits from this PR such that it only includes the commits through fb38c9f? I'd like to keep it small, and the other 2 mem-related commits aren't related to this, nor have i reviewed them.

You can create a new PR for the page/frame allocator stuff that branches off from this one, and then when you're ready to move past that, you'd create yet another PR on yet another branch. That way you can continue developing without changing the content of PRs that are under review.

NathanRoyer:
@kevinaboos no problem; done :ok_hand: I'll just add that Makefile change which is more related to other changes in this PR than memory stuff

aarch64/Makefile
kevinaboos: oh neat, i didn't know about that new `rust-toolchain` file format!

aarch64/kernel/nano_core/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+git = "https://github.com/theseus-os/uefi-rs/"
+branch = "until-upstream-pr-merged"
+default-features = false
+features = [ "alloc", "exts" ]
+
+[dependencies.uefi-services]
+git = "https://github.com/theseus-os/uefi-rs/"
+branch = "until-upstream-pr-merged"
+default-features = false
+features = [ "panic_handler" ]
```
kevinaboos: FYI your PR was merged in and made it in the upstream ([proof](https://github.com/rust-osdev/uefi-rs/blob/main/CHANGELOG.md#uefi-services---0150-2022-11-15)) so I think we can use that instead.

kevinaboos (changes requested):
one small comment, otherwise ready to go.

aarch64/kernel/nano_core/Cargo.toml
NathanRoyer: That's right :)

kevinaboos (approved):

PR #689, opened 2022-11-09
Use TLS-based `with_current_task()` in all crates

kevinaboos:
* Refactor `schedule()` and `task_switch()` to more clearly handle
  the ownership/borrowing of `TaskRef`s for the current and next tasks.

* `get_my_current_task_id()` is now infallible, since we can guarantee
   that the task's ID is known before it can possibly begin executing.

* `get_my_current_task()` now returns an owned `TaskRef`, because
   the current task is now stored in a TLS variable which cannot be 
   returned as a reference since its thread lifetime is undefinable.

* Introduce new `with_current_task<F>(closure: F)` function, which is 
  more efficient because it avoids cloning the current task's `TaskRef`.

* Completely remove `TaskLocalData`, as it is no longer used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -946,6 +946,7 @@ dependencies = [
  "acpi_table",
  "memory",
  "sdt",
+ "static_assertions",
  "zerocopy",
 ]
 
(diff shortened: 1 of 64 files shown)


kevinaboos:
Note: this is waiting on #692 to be merged

PR #691, opened 2022-11-11
Update `ppv-lite86`

### LINKED ISSUE #654: `rand_chacha` crate fails to build on debug mode
The `random` crate depends on `rand_chacha`, which apparently fails to build correctly in debug mode.

## Steps to reproduce
```
make run BUILD_MODE=debug
```
You'll see an LLVM error about split operands. 

Note that this doesn't happen in regular release builds, which is odd.

## Related
I fixed a similar LLVM error in the `crc32fast` crate here: https://github.com/srijs/rust-crc32fast/pull/22, which makes me think that this is related to how `rand_chacha` specific its cfg gates for enabling SSE instructions.

tsoutsman:
Closes #654.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2362,9 +2362,9 @@ version = "0.2.1"
 
 [[package]]
 name = "ppv-lite86"
-version = "0.2.16"
+version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


kevinaboos (commented):
Thanks for confirming this and for filing the original issues!

kevinaboos (approved):

PR #693, opened 2022-11-15
Basic page_allocator and frame_allocator init for aarch64

NathanRoyer:
This PR builds on #688 and adds the following kernel crates in aarch64:

`page_allocator` and `frame_allocator` contained duplicate modules which I moved to a separate crate: `static_array_collections`. Apart from that, no modification was required in these.

`memory_struct` was modified for the removal of ELF- and multiboot-related code; the local `page_allocator` and `frame_allocator` depend on that local `memory_struct`.


aarch64/kernel/nano_core/main.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    let _callback = frame_allocator::init(
+        free_regions.iter().flatten(),
+        reserved_regions.iter().flatten(),
+    )?;
+
+    // for now, virtual addresses will be above 4GB
+    page_allocator::init(VirtualAddress::new_canonical(0x100_000_000))?;
```
kevinaboos: why, is this just a random choice?

Also, small nit: typically octal numbers are grouped in 3, while hex numbers are grouped in 4: `0x1_0000_0000`

kevinaboos (changes requested):
Looks pretty good to me! I left a small comment, but have a couple more things here:

1. can you merge in the changes from `theseus_main` into here? that way github will only show the differences since your last PR #688, which is easier to read. You'll actually want to do that for all PRs on an ongoing basis too - just a heads up.
2. did you change anything in the `page_allocator` or `frame_allocator`, or were they already sufficiently arch-independent? I'd like to have that knowledge as part of the PR comments / commit msg.

aarch64/kernel/nano_core/main.rs
NathanRoyer: Basically; I wanted to test the init function and chose a value which I was sure wouldn't cause issues; once we have an ELF aarch64 kernel we'll be able to use the original limit defined with KERNEL_OFFSET and the ELF layout.

NathanRoyer:
> can you merge in the changes from `theseus_main` into here? that way github will only show the differences since your last PR #688, which is easier to read. You'll actually want to do that for all PRs on an ongoing basis too - just a heads up.

Yes

> did you change anything in the `page_allocator` or `frame_allocator`, or were they already sufficiently arch-independent? I'd like to have that knowledge as part of the PR comments / commit msg.

`page_allocator` and `frame_allocator` contained duplicate modules which I moved to a separate crate: `static_array_collections`; apart from that, no modification was required; `memory_struct` was modified for the removal of ELF- and multiboot-related code; the local `page_allocator` and `frame_allocator` depend on that local `memory_struct`.

aarch64/kernel/nano_core/main.rs
kevinaboos: sounds good, we'll accept it for now.

aarch64/kernel/nano_core/main.rs
```diff
@@ -30,16 +36,53 @@ fn main(
...
+    let mut free_index = 0;
+    let mut reserved_regions: [Option<PhysicalMemoryRegion>; 32] = Default::default();
+    let mut reserved_index = 0;
+
+    for descriptor in mem_iter {
+        let size = descriptor.page_count as usize * PAGE_SIZE;
+        if size > 0 {
+            let (dst, index, region_type) = match descriptor.ty {
+                MemoryType::CONVENTIONAL => (&mut free_regions, &mut free_index, MemoryRegionType::Free),
+                _ => (&mut reserved_regions, &mut reserved_index, MemoryRegionType::Reserved),
```
kevinaboos: There's an important distinction here. The concept of "reserved" memory in the frame allocator is likely different than the concept of what UEFI/bootloaders consider to be "reserved".

In the frame allocator, "reserved" means that it is still available to be used and allocated, but that it is memory dedicated for a special purpose (e.g., device MMIO). Thus, it has to be specifically requested by address. A general allocation request (where no address is requested) cannot allocate from reserved memory.

That doesn't necessarily seem the be the same thing as the `uefi::boot::MemoryType` enum, so I don't think this `_` catch-all match variant is correct. You need to carefully go through each `MemoryType` to see which ones can be considered as "reserved" in Theseus's terminology.

I also think you need to do this for the `free_regions` too, since `MemoryType::CONVENTIONAL` probably isn't the only type of memory that can be considered as free for general purposes by Theseus's frame allocator.

We can discuss this further too in person or on discord if you're unsure about each `MemoryType` variant.

kevinaboos (changes requested):
Ok thanks, now that you've merged in the latest changes, I can more easily identify some issues.

Currently I see just one major issue, see my comment about `MemoryType`s in UEFI and how they can be translated into the `MemoryRegionType` used in Theseus's frame allocator. Once that's resolved, we can merge this.

aarch64/kernel/nano_core/main.rs
```diff
@@ -30,16 +40,61 @@ fn main(
...
+            };
+
+            let start_addr = descriptor.phys_start as usize;
+            let start_addr = PhysicalAddress::new_canonical(start_addr);
+            let end_addr = start_addr + size;
+
+            let first_frame = Frame::containing_address(start_addr);
+            let last_frame = Frame::containing_address(end_addr - 1);
+
+            let range = FrameRange::new(first_frame, last_frame);
```
kevinaboos: there's a convenience function for this that does what you want: https://www.theseus-os.com/Theseus/doc/memory_structs/struct.FrameRange.html#method.from_phys_addr

```suggestion
            let range = FrameRange::from_phys_addr(start_addr, size);
```

kevinaboos (changes requested):
Awesome, this looks very nice and clean now! 

I made one small suggestion for the `FrameRange` stuff used for `frame_allocator::init()`.

The `uefi::MemoryType` conversion routines are okay for now, but we'll need to change that once we have real UEFI support with a separate UEFI loader/runner and an ELF kernel image. Some of those areas will become re-usable after we've exited boot services, so we'll need to save their details for later (but obviously you cannot do that yet in this PR).

Finally, regarding the canonical addresses on aarch64, I don't know if what you wrote is correct. I'm not 100% certain of this, but wouldn't the physical addresses at least have some requirements, since we're forcibly only supporting 48-bit physical addresses? Thus, wouldn't we need to check that the most significant 16 bits are either `0` or `1` or some other specific set of bit values?

I'm not sure about the virtual address canonicality either, but if you have a source for your claim that there's no concept of that on aarch64, then we should include it here in a comment.

NathanRoyer:
> I made one small suggestion for the `FrameRange` stuff used for `frame_allocator::init()`.

Thanks

> The `uefi::MemoryType` conversion routines are okay for now, but we'll need to change that once we have real UEFI support with a separate UEFI loader/runner and an ELF kernel image. Some of those areas will become re-usable after we've exited boot services, so we'll need to save their details for later (but obviously you cannot do that yet in this PR).

Ok

> Finally, regarding the canonical addresses on aarch64, I don't know if what you wrote is correct. I'm not 100% certain of this, but wouldn't the physical addresses at least have some requirements, since we're forcibly only supporting 48-bit physical addresses? Thus, wouldn't we need to check that the most significant 16 bits are either `0` or `1` or some other specific set of bit values?

I agree; I'll make sure the most significant bits are always zero so that paddrs which are >= `(1 << 48)` will be invalid.

> I'm not sure about the virtual address canonicality either, but if you have a source for your claim that there's no concept of that on aarch64, then we should include it here in a comment.

Well, the most significant bits of vaddrs are for ASID, which is either 8 or 16 bits (8 as I configured it); as our ASID is zero, I can also put that check in these utility functions, however it's not really related to the x86_64 meaning of canonical addresses. I'll prepare the commit, ping me when you're online and I'll push them if you want.

kevinaboos (approved):

PR #694, opened 2022-11-15
Uefi aarch64 paging_entry_flags

NathanRoyer:
This PR builds on #693 and adds the `paging_entry_flags` kernel crate in aarch64.

This crate supports both `aarch64` and `x86_64`; The following flags are implemented:

- valid entry
- writeable
- executable
- cacheable
- global
- userland
- exclusive

I was also wondering if "paging_slot_flags" wouldn't be a better name, as a slot can be empty; I'm not a native english speaker so I don't really know what's best.

I could also document each flag in the crate's documentation, what do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.so
 *.rlib
 *.dll
+*.fd
 
 # Executables
 *.exe
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,9 @@ exclude = [
 	"build",
 	"target",
 
+	## Exclude the aarch64 experiment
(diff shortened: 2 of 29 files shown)


kevinaboos:
I think this PR would be better as a direct contribution to the main Theseus workspace, since it's relatively self-contained and is easy to integrate alongside other crates, even those that pertain to x86. Then, in a future PR, we can change all existing crates to use the new PTE flag abstraction.

Thus, could you include only the changes to that new crate, and move it into `/kernel/pte_flags`? Thanks, then I can more easily review. (Or you can just close this and open another PR)

PR #698, opened 2022-11-18
Use canonical target triple names (`x86_64-unknown-theseus`)

kevinaboos:
Previously, the `vendor` component was skipped; now it is explicitly `unknown`.


tshepang:
Am curious why this change was done. I see there is a target named `wasm32-wasi`, meaning vendor can be skipped. Is there a plan that vendor in this case could be anything but `unknown`?

kevinaboos:
It was done as part of an upcoming effort to upstream a Theseus-specific target spec into rustc. This is the typical way that target specs are named, and it should have been named in this manner from the get-go. 

It's debatably required because some crates even rely on this format in order to work (see [target_spec](https://crates.io/crates/target-spec))

PR #699, opened 2022-11-19
Architecture-dependent Page Table Entry (PTE) flags for `x86_64` and `aarch64`

kevinaboos:
For purposes of comparison with #696

~~Remaining TODO: add `#[test]` cases for the four conversion routines.~~ Done


kernel/pte_flags/src/pte_flags_aarch64.rs
```diff
@@ -0,0 +1,168 @@
...
+        const OUTER_SHAREABLE    = 2 << 8;
+        /// Multiple cores from the same
+        /// cluster can access this page.
+        /// Indicates that multiple CPUs from only a single cluster may access this page.
+        const _INNER_SHAREABLE   = 3 << 8;
+
+        /// * The hardware will set this bit when the page is accessed.
+        /// * The OS can then clear this bit once it has acknowledged that the page was accessed,
+        ///   if it cares at all about this information.
+        const ACCESSED           = 1 << 10;
```
NathanRoyer: I think this is slightly incorrect: without setting the bit, qemu's emulated CPU refused to insert the descriptor in the TLB, and this resulted in page faults. By [manually setting the bit](https://github.com/NathanRoyer/aarch64-uefi/blob/main/uefi/src/memory.rs#L79), the descriptor was accepted into the TLB and I encountered no page fault. Maybe this was because of my configuration of translation control registers though? I'll check. It could also be qemu not implementing FEAT_HAFDBS.

kernel/pte_flags/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// Also, this contains only the flags that are common to both `x86_64` and `aarch64`.
+    ///
+    /// This type can be losslessly converted into `PteFlagsX86_64` and `PteFlagsAarch64`
+    /// with the typical [`From`] and [`Into`] traits.
+    /// This makes it to set general architecture-indepedent flags first,
+    /// and then convert it in order to set more architecture-specific flags.
+    /// 
+    /// This type can also be converted *from* `PteFlagsX86_64` and `PteFlagsAarch64`,
+    /// but it may be lossless as only the bit flags defined herein are preserved.
+    pub struct PteFlags: u64 {
```
NathanRoyer, line 65: I think it lacks a way to set PAGE_DESCRIPTOR & ACCESSED; I needed to set this bit for my descriptors to be used correctly by qemu's emulated CPU. As I wrote below, the CPU refused to use the descriptors I was creating if they didn't have the ACCESSED bit set.

NathanRoyer (commented):
Looks good to me, other than the two other comments I left on specific lines.

Not so important: `PteFlags` has the flags that best match x86_64 & aarch64; if we want to support risc-v in the future, would we change it based on flags common to aarch64, x86_64 & risc-v? I guess we'll see in the future

kevinaboos:
ok this should be completely ready now. @NathanRoyer please re-review and also see how I did the cool cfg stuff with docs

kernel/pte_flags/src/pte_flags_aarch64.rs
kevinaboos: should be fixed now

kernel/pte_flags/src/lib.rs
kevinaboos: hmm, i could add a function for setting ACCESSED, but for the PAGE_DESCRIPTOR that's something you should need to manually set by doing something like `flags.insert( PteFlagsAarch64::PAGE_DESCRIPTOR)` or remove, etc.

NathanRoyer:
> ok this should be completely ready now. @NathanRoyer please re-review

~~Your solution for ACCESSED will work. However, regarding PAGE_DESCRIPTOR, I'm not sure I understand why we shouldn't put it by default like ACCESSED because it's also a requirement for things to work... Without PAGE_DESCRIPTOR, all the other flags will be interpreted as if they were in a block descriptor~~

edit: we solved this during our meeting

> also see how I did the cool cfg stuff with docs

Very nice, didn't know about this trick

kernel/pte_flags/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+}}
+cfg_if!{ if #[cfg(any(target_arch = "aarch64", doc))] {
+    mod pte_flags_aarch64;
+    pub use pte_flags_aarch64::PteFlagsAarch64;
+}}
+
+#[cfg(target_arch = "x86_64")]
+use pte_flags_x86_64::PteFlagsX86_64 as PteFlagsArch;
+#[cfg(target_arch = "aarch64")]
+use pte_flags_aarch64::PteFlagsAarch64 as PteFlagsArch;
```
NathanRoyer: Could these be re-exported publicly? So that I can do:
```rust
impl PageTableEntry {
    pub fn flags(&self) -> PteFlags {
        PteFlagsArch::from_bits_truncate(self.0).into()
    }
}
```
in `page_table_entry`

NathanRoyer:
Also, should we add a dummy `PteFlags::is_huge(self) -> bool` method? one that would always returns false?
(For the moment I commented out the calls to that method in `memory`)

NathanRoyer:
> By [manually setting the bit](https://github.com/NathanRoyer/aarch64-uefi/blob/main/uefi/src/memory.rs#L79), the descriptor was accepted into the TLB and I encountered no page fault. Maybe this was because of my configuration of translation control registers though? I'll check. It could also be qemu not implementing FEAT_HAFDBS.

Something I forgot: `TCR_EL1` (the translation control register) has two flags that control the behaviour of the CPU regarding the ACCESSED and DIRTY flags: [HA](https://docs.rs/cortex-a/latest/cortex_a/registers/TCR_EL1/HA/index.html) (for ACCESSED) and [HD](https://docs.rs/cortex-a/latest/cortex_a/registers/TCR_EL1/HD/index.html) (for DIRTY). However, even with these enabled, qemu triggers an access fault when the bit is initially cleared.

Regarding the output address size: I was using 48-bit OA mode in my experimental code; I assumed we could only address up to 48 bits (4 levels * 9 bits + 12 bits).

However, actually one of the following configurations can be selected (using `TCR_ELn::IPS` and `TCR_ELn::DS` - weirdly, the latter is not exposed by `cortex_a`... I assumed the DS field was just not present in TCR_EL1):

image

image

As we can see, the 52-bit output address mode is a bit more complex as some bits of the OA are shifted to the middle of lower attributes. This allows the dirty bit to always occupy the 51st bit. In 48-bit OA mode, bits 48 & 49 are reserved and bit 50 is the Guarded Page bit - implementation is optional.

In 52-bit OA mode, the shareability attributes are stored in the TCR_EL1 register, specifcially in field SH0 for TTBR0 page table walks and in field SH1 for TTBR1 page table walks. This limits us to two shareability profiles and also complexifies page table management a bit since that requires us to maintain two page tables at all time.

Apparently long mode x86_64 also supports [52-bit OA mode](https://wiki.osdev.org/Paging#Page_Map_Table_Entries); I don't know how this is configured however, but maybe you know.

kernel/pte_flags/src/pte_flags_aarch64.rs
```diff
@@ -0,0 +1,208 @@
...
+    ///
+    /// **Note:** items beginning with an underscore `_` are not used in Theseus.
+    ///
+    /// The designation of bits in each `PageTableEntry` is as such:
+    /// * Bits `[0:11]` (inclusive) are reserved by hardware for access flags, cacheability flags,
+    ///   shareability flags, and TLB storage flags.
+    /// * Bits `[12:50]` (inclusive) are reserved by hardware to hold the physical frame address.
+    /// * Bits `[51:54]` (inclusive) are reserved by hardware for more access flags.
+    /// * Bits `[55:58]` (inclusive) are available for custom OS usage.
+    /// * Bits `[59:63]` (inclusive) are reserved by hardware for extended access flags.
```
NathanRoyer: If we want to keep 48-bit OA mode, the correct documentation would be:
```rust
    /// * Bits `[0:11]` (inclusive) are reserved by hardware for access flags, cacheability flags,
    ///   shareability flags, and TLB storage flags.
    /// * Bits `[12:47]` (inclusive) are reserved by hardware to hold the physical frame address.
    /// * Bits `[48:49]` (inclusive) are ignored.
    /// * Bits `[50:54]` (inclusive) are reserved by hardware for more access flags.
    /// * Bits `[55:58]` (inclusive) are available for custom OS usage.
    /// * Bits `[59:63]` (inclusive) are reserved by hardware for extended access flags.
```

kernel/pte_flags/src/pte_flags_aarch64.rs
```diff
@@ -0,0 +1,208 @@
...
+//! has a "DEVICE nGnRE" entry, and [MAIR] index 1 has a Normal + Outer Shareable entry.
+//!
+//! [MAIR]: https://docs.rs/cortex-a/latest/cortex_a/registers/MAIR_EL1/index.html
+
+use crate::PteFlags;
+use bitflags::bitflags;
+use static_assertions::const_assert_eq;
+
+/// A mask for the bits of a page table entry that contain the physical frame address.
+pub const PTE_FRAME_MASK: u64 = 0x000_7FFFFFFFFF_000;
```
NathanRoyer: In 48-bit OA mode, this would be:
```rust
pub const PTE_FRAME_MASK: u64 = 0x000_0FFFFFFFFF_000;
```

kernel/pte_flags/src/lib.rs
kevinaboos: hmm so i considered exporting this publicly but couldn't think of a good use case for it. My primary concern was that users would be tempted to use `PteFlagsArch` everywhere when they really should be using `PteFlagsX86_64` or `PteFlagsAarch64` explicitly, which would then lead to really confusing compilation errors on a different platform than the one they initially were testing with, because the definition of `PteFlagsArch` would change depending on target platform. Generally that's indicative of a bad design.

Your use case here is definitely valid, but I think that you'll need a fair bit of aarch64-specific code in the `page_table_entry` crate. For example: several things have to change:
* `frame_value()` needs a different impl on aarch64 than x86 because more bits need to be masked out
* `set_entry()` needs to change (depending on how you implement it), e.g., you may choose to set the `PAGE_DESCRIPTOR` bit there (or elsewhere, higher up the call stack)
* `set_flags()` will be different too; we actually need to mask all of the bits in the arch-specific `PteFlags` type in order to make sure they're correctly set. The existing code only works for x86

Thus, since you'll need tons of arch-specific cfg, I think it's best to keep things explicit, which is also more Rust-y in nature. As such, we'll leave `PteFlagsArch` typedef private.

kevinaboos:
> Also, should we add a dummy `PteFlags::is_huge(self) -> bool` method? one that would always returns false? (For the moment I commented out the calls to that method in `memory`)

whoa, definitely not. That would be invalid and could lead to us completely misinterpreting what PTE values mean. For a concrete example of why, in the initial bootstrap assembly code (boot.asm) we actually do use huge pages to efficiently and easily map large regions of the very lowest and very highest addresses before jumping to Rust code. If we attempted to call `translate()` on one of those addresses early on, it would completely fail because it would unconditionally consider it a non-huge page PTE.

In any case, why do something like that which could confuse the caller? It also acts as a sanity check if we ever do come across a mapping that uses huge pages, or in the future when we add proper support for them.

kevinaboos:
> > By [manually setting the bit](https://github.com/NathanRoyer/aarch64-uefi/blob/main/uefi/src/memory.rs#L79), the descriptor was accepted into the TLB and I encountered no page fault. Maybe this was because of my configuration of translation control registers though? I'll check. It could also be qemu not implementing FEAT_HAFDBS.
> 
> Something I forgot: `TCR_EL1` (the translation control register) has two flags that control the behaviour of the CPU regarding the ACCESSED and DIRTY flags: [HA](https://docs.rs/cortex-a/latest/cortex_a/registers/TCR_EL1/HA/index.html) (for ACCESSED) and [HD](https://docs.rs/cortex-a/latest/cortex_a/registers/TCR_EL1/HD/index.html) (for DIRTY). However, even with these enabled, qemu triggers an access fault when the bit is initially cleared.

That's fine, it seems that those bits just allow hardware to set the access flag (as is the default behavior on x86) instead of the OS always doing it (?). Looks like  the Access Flag Fault always occurs regardless of HA/HD value, so let's just always set the `ACCESSED` bit for now, but also implement an Access Flag Fault handler such that we know when it happens (and can handle those faults in the future if/when we support swapping pages to disk).

kernel/pte_flags/src/pte_flags_aarch64.rs
kevinaboos: Yes, let's stick with 48-bit OA size and also assume an equivalent 48-bit IPA size. Thanks for digging into this!

kevinaboos:
> Regarding the output address size: I was using 48-bit OA mode in my experimental code; I assumed we could only address up to 48 bits (4 levels * 9 bits + 12 bits).

48-bit is fine with me, we can use that for now and if it presents problems we can move to 52-bit.

> As we can see, the 52-bit output address mode is a bit more complex as some bits of the OA are shifted to the middle of lower attributes. This allows the dirty bit to always occupy the 51st bit. In 48-bit OA mode, bits 48 & 49 are reserved and bit 50 is the Guarded Page bit - implementation is optional.

I was wondering about the DIRTY bit, that makes sense. Glad we resolved that.

> In 52-bit OA mode, the shareability attributes are stored in the TCR_EL1 register, specifcially in field SH0 for TTBR0 page table walks and in field SH1 for TTBR1 page table walks. This limits us to two shareability profiles and also complexifies page table management a bit since that requires us to maintain two page tables at all time.

We'd only ever use `OUTER_SHAREABLE` anyway, so that's not a problem. We could just use the same page tables for both TTBRs, i think.

> Apparently long mode x86_64 also supports [52-bit OA mode](https://wiki.osdev.org/Paging#Page_Map_Table_Entries); I don't know how this is configured however, but maybe you know.

We don't use it, but it's much simpler on x86 -- the 52-bit physical address can pretty much always be used.

kevinaboos:
@NathanRoyer I have addressed all of our comments, please double-check everything again.

**Important Note**: in the `memory_structs` crate, you will need to implement `aarch64`-specific canonicalization for 48-bit physical addresses, since they now differ from the 52-bit physical addresses we use on x86_64. 
Also, I assume that virtual addresses on aarch64 are also canonicalized in a similar fashion to x86; if not, you'll need to handle that too. 
This stuff is on L129-159 of `memory_structs/src/lib.rs`

NathanRoyer:
> @NathanRoyer I have addressed all of our comments, please double-check everything again.

About `PteFlagsAarch64::_CONTIGUOUS`:

> If not set, this translation table is not contiguous with the previous one in memory.

I think it's not necessarily the previous one; from the DDI0487:

> The Contiguous bit identifies a descriptor as belonging to a group of adjacent translation table entries that point to a contiguous OA range.

But anyway, the nuance is small and we don't use it so I'll leave it as is.

> **Important Note**: in the `memory_structs` crate, you will need to implement `aarch64`-specific canonicalization for 48-bit physical addresses, since they now differ from the 52-bit physical addresses we use on x86_64. Also, I assume that virtual addresses on aarch64 are also canonicalized in a similar fashion to x86; if not, you'll need to handle that too. This stuff is on L129-159 of `memory_structs/src/lib.rs`

I responded to this on Discord: the most significant bits of an address are typically used for tagged addresses (ASID) or ignored, when that mechanism is disabled.

I will approve this, you can merge it in.

NathanRoyer (approved):

kevinaboos:
> About `PteFlagsAarch64::_CONTIGUOUS`:
> 
> > If not set, this translation table is not contiguous with the previous one in memory.
> 
> I think it's not necessarily the previous one; from the DDI0487:
> 
> > The Contiguous bit identifies a descriptor as belonging to a group of adjacent translation table entries that point to a contiguous OA range.

Right, thanks for pointing that out. I'll fix it.

PR #700, opened 2022-11-21
Add ps2_controller crate

### LINKED ISSUE #685: Init both keyboard and mouse in one ps2 controller driver instead
I think it's completely fine to initialize them together as part of a larger `ps2::init()` function or something like that, in which we follow your advice above and attempt to init keyboard, then mouse. That should be possible to directly invoke from the device manager or captain, wherever appropriate.

_Originally posted by @kevinaboos in https://github.com/theseus-os/Theseus/pull/667#discussion_r1013392822_

hecatia-elegua:
Closes #685

This needed some more things to be public, which will be made private again in another PR.
Step 2 with the ACPI check sounds to me like it should be done in the device_manager instead, if you want I can put it there.
I only left it in init to keep all the steps together.

### DIFF
--- a/kernel/ps2/src/lib.rs
+++ b/kernel/ps2/src/lib.rs
@@ -1,6 +1,5 @@
 #![no_std]
 
-use log::debug;
 use num_enum::TryFromPrimitive;
 use port_io::Port;
 use spin::Mutex;
@@ -22,7 +21,7 @@ static PS2_COMMAND_AND_STATUS_PORT: Mutex<Port<u8>> = Mutex::new(Port::new(0x64)
 /// 
 /// Naming perspective: Output means "out of the device into the host".
 /// e.g. [WriteByteToPort2InputBuffer] means "write byte out of the host into port 2"
-enum HostToControllerCommand {
+pub enum HostToControllerCommand {
     /// returns [ControllerConfigurationByte]
     ReadFromInternalRAMByte0 = 0x20,
     
@@ -110,7 +109,7 @@ enum HostToControllerCommand {
 /// 
 /// Note: Devices attached to the controller should be disabled
 /// before sending commands that return data, otherwise the output buffer could get overwritten
-fn write_command(value: HostToControllerCommand) {
+pub fn write_command(value: HostToControllerCommand) {
     unsafe {
         PS2_COMMAND_AND_STATUS_PORT.lock().write(value as u8);
     }
@@ -207,14 +206,14 @@ fn write_data(value: WritableData) {
 // wiki.osdev.org/%228042%22_PS/2_Controller#PS.2F2_Controller_Configuration_Byte
 /// Used for [HostToControllerCommand::ReadFromInternalRAMByte0] and [HostToControllerCommand::WriteToInternalRAMByte0]
 #[bitfield(bits = 8)]
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct ControllerConfigurationByte {
     /// interrupt on [ControllerToHostStatus] `output_buffer_full`
-    port1_interrupt_enabled: bool,
+    pub port1_interrupt_enabled: bool,
     /// interrupt on [ControllerToHostStatus] `mouse_output_buffer_full`
     /// 
     /// Note: only if 2 PS/2 ports supported
-    port2_interrupt_enabled: bool,
+    pub port2_interrupt_enabled: bool,
     /// Cleared on reset; set when the system passed Power-on self-test
     #[allow(dead_code)]
     system_passed_self_test: bool,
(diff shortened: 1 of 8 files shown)


kernel/ps2/src/lib.rs
```diff
@@ -207,168 +206,53 @@ fn write_data(value: WritableData) {
 // wiki.osdev.org/%228042%22_PS/2_Controller#PS.2F2_Controller_Configuration_Byte
 /// Used for [HostToControllerCommand::ReadFromInternalRAMByte0] and [HostToControllerCommand::WriteToInternalRAMByte0]
 #[bitfield(bits = 8)]
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
kevinaboos: this (and probably all the other bitfield types) can also derive `Copy`, since they're just plain old data types.

kernel/ps2_controller/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+    write_config(config.clone());
+
+    // Step 6: Perform Controller Self Test
+    write_command(TestController);
+    read_controller_test_result()?;
+    debug!("passed PS/2 controller test");
+    // as this can reset the controller on some hardware, we restore the config
+    write_config(config.clone());
+
+    // Step 7 not needed, already done above
```
kevinaboos: can you explicitly state what Step 7 is, even though it's already been done? It'd be nice to preserve it if the osdev wiki goes down

kernel/ps2_controller/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+use ps2::{write_command, flush_output_buffer, read_config, write_config, read_controller_test_result, read_port_test_result, PortTestResult, HostToControllerCommand::*, reset_keyboard, reset_mouse};
+
+
+// see https://wiki.osdev.org/%228042%22_PS/2_Controller#Initialising_the_PS.2F2_Controller
+/// initialize the first and second (if it exists) PS/2 port
+pub fn init() -> Result<(), &'static str> {
+    // Step 1: Initialise USB Controllers
+    // no USB support yet
+
+    // Step 2: Determine if the PS/2 Controller Exists
```
kevinaboos, line 13: so if there isn't a `FADT` table, is the general advice just to continue trying to init ps2, kbd, and mouse?

kevinaboos (changes requested):
Thanks, this looks great! I left a few small comments.

> This needed some more things to be public, which will be made private again in another PR.

Sounds good, no problem.

> Step 2 with the ACPI check sounds to me like it should be done in the device_manager instead, if you want I can put it there. I only left it in init to keep all the steps together.

I actually think it belongs in the ps2 controller as you have currently done it, so let's keep it there.

This PR gave me ann idea: I think we should move the keyboard and mouse init calls from `device_manager` into your new `ps2_controller` crate. That way, eventually the device manager will contain only a sequence of high-level calls to initialize certain device classes/subsystems, and all the details can be moved out of the device manager into each subsystem. What do you think?

hecatia-elegua:
> move the keyboard and mouse init calls from `device_manager` into your new `ps2_controller` crate.

I think naming would be wrong then, as they're not really _part of_ the controller. It would also be less modular, since there's at least one ARM-specific controller which could be switched in instead. But...

> eventually the device manager will contain only a sequence of high-level calls to initialize certain device classes/subsystems

I think that's a good idea, though when someone tackles the "Load device driver crates upon discovery of matching device" draft issue, I guess it will be restructured a lot anyways.

kernel/ps2_controller/src/lib.rs
hecatia-elegua: "if the bit is set, or the system doesn't support ACPI (no ACPI tables and no FADT) then there is a PS/2 Controller"
What happens with "ACPI yes" and "FADT no"?
Looking at our acpi code, we currently always assume the system to support ACPI, it otherwise shuts down.
The FADT is mandatory (or our acpi code says so).

If we were to support systems without ACPI, I assume `get_acpi_tables` would return an empty table and `Fadt::get` would return `None`. I'm adding a comment.

kevinaboos:
> > move the keyboard and mouse init calls from `device_manager` into your new `ps2_controller` crate.
> 
> I think naming would be wrong then, as they're not really _part of_ the controller. It would also be less modular, since there's at least one ARM-specific controller which could be switched in instead. But...

I see your argument, and yes I suppose they're not technically within the hierarchy of the PS2 controller itself, they just talk to it. Indeed we don't need to refactor that now.

Not familiar with the ARM controller, but I think you're getting at the fact that* the keyboard & mouse drivers can be used independently of the specific controller incarnation (though we'd have to refactor them to support an abstract controller type). 

> I think that's a good idea, though when someone tackles the "Load device driver crates upon discovery of matching device" draft issue, I guess it will be restructured a lot anyways.

yep 👍

hecatia-elegua:
I was getting at that, yes, and you're right, it would need refactoring 😄 
Is this ready for merging?

kevinaboos (changes requested):
Well, while the code does look good, it actually doesn't work. When testing on my machine (a laptop using WSL), I don't get any input from the keyboard nor the mouse. Here's the only relevant log output:
```
[D] kernel/ps2_controller/src/lib.rs:44: passed PS/2 controller test
[D] kernel/ps2_controller/src/lib.rs:93: ControllerConfigurationByte { port1_interrupt_enabled: true, port2_interrupt_enabled: true, system_passed_self_test: false, should_be_zero: 0, port1_clock_disabled: true, port2_clock_disabled: true, port1_translation_enabled: false, must_be_zero: 0 }
[D] kernel/keyboard/src/lib.rs:35: The PS/2 keyboard type is: MF2Keyboard
[D] kernel/mouse/src/lib.rs:37: The PS/2 mouse ID is: Four
```

Does the PR actually work on your machine? I'm not sure what the problem is, but I don't have time to dig too deep into it right now. All I can see is that you did change several init steps, both the order of them and what actual operations were done.

hecatia-elegua:
Now it's getting interesting. It works for me (laptop Linux), WSL works the same way, right? So I'm not sure if there's some other difference. Maybe it's a timing-related issue.
I will push another change so it actually shows the port success logs and another config log after the device reset.
After reset, the second `ControllerConfigurationByte` should show both `_clock_disabled` as false.

kevinaboos:
Thanks, I've just tested it on my linux desktop (Ubuntu 20) and it still isn't working. To be fair, I've never experienced a difference when running Theseus on QEMU on linux vs QEMU on linux via WSL.

I pulled down your latest version and ran it, here are the relevant logs:
```
[D] kernel/ps2_controller/src/lib.rs:44: passed PS/2 controller test
[D] kernel/ps2_controller/src/lib.rs:81: passed PS/2 port 1 test
[D] kernel/ps2_controller/src/lib.rs:89: passed PS/2 port 2 test
[D] kernel/ps2_controller/src/lib.rs:95: ControllerConfigurationByte { port1_interrupt_enabled: true, port2_interrupt_enabled: true, system_passed_self_test: false, should_be_zero: 0, port1_clock_disabled: true, port2_clock_disabled: true, port1_translation_enabled: false, must_be_zero: 0 }
[D] kernel/ps2_controller/src/lib.rs:104: ControllerConfigurationByte { port1_interrupt_enabled: true, port2_interrupt_enabled: true, system_passed_self_test: false, should_be_zero: 0, port1_clock_disabled: true, port2_clock_disabled: true, port1_translation_enabled: false, must_be_zero: 0 }
[D] kernel/keyboard/src/lib.rs:35: The PS/2 keyboard type is: MF2Keyboard
[D] kernel/mouse/src/lib.rs:37: The PS/2 mouse ID is: Four
```

kevinaboos:
In case it's relevant, here is the QEMU version I'm testing with:
```sh
$ qemu-system-x86_64  --version
QEMU emulator version 4.2.1 (Debian 1:4.2-3ubuntu6.23)
Copyright (c) 2003-2019 Fabrice Bellard and the QEMU Project developers
```

hecatia-elegua:
Ok so as I thought, the second log still shows both `_clock_disabled` to be `true`.
My qemu version is `6.2.0 (Debian 1:6.2+dfsg-2ubuntu6.5`, so I wonder what to do now. I skimmed through some qemu changes, didn't really find anything.

Please give it one more try with the new change.
If it still doesn't work, maybe try a newer qemu version (unless this would take much time, then I would just set the config again after reset).

kevinaboos:
> Please give it one more try with the new change. 

Yep, that did the trick! It's working correctly now. Nice job figuring that out, and thanks for the fix! I will also test it on my WSL setup and on real hardware, but you can assume it works there too unless you hear otherwise from me.

> If it still doesn't work, maybe try a newer qemu version

Requiring a new QEMU version wouldn't be the right solution since it still wouldn't work on real hardware, but that's no longer a concern since it works now.

------ 
If you'd like to do any final clean-up before i merge this PR in, please go ahead; if not, just let me know and I'll merge it in as is.

hecatia-elegua:
> since it still wouldn't work on real hardware

My train of thought was that the older qemu version didn't work right (so it would work on real hardware), which would have meant any changes would only be for conforming to old qemu versions.

I will keep the second config log in, you can merge :)

kevinaboos (approved):
Tested working on all platforms. Thanks very much!

PR #701, opened 2022-11-22
Porting `memory` to aarch64

NathanRoyer:
This adds aarch64 support to the `memory` as well as `memory_aarch64` which contains utility functions for MMU management under aarch64.


aarch64/kernel/memory/src/paging/mapper.rs
```diff
@@ -0,0 +1,1280 @@
...
+
+    /// Translates a virtual memory `Page` to a physical memory `Frame` by walking the page tables.
+    ///
+    /// This method assumes that the modified page is active
+    pub fn translate_page(&self, page: Page) -> Option<Frame> {
+        let p3 = self.p4(true).next_table(page.p4_index(), true);
+
+        // Temporarily removing these checks while we
+        // Learn how to create huge pages on aarch64.
+        /*
```
kevinaboos: some changes in the main branch should already address this.

aarch64/kernel/memory/src/paging/mapper.rs
```diff
@@ -0,0 +1,1280 @@
...
+    /// Otherwise, the code assumes we are in identity-mapping, and
+    /// the physical addresses in a table are used as virtual addresses.
+    pub(super) fn internal_map_to(
+        &mut self,
+        pages: AllocatedPages,
+        frames: AllocatedFrames,
+        flags: PteFlags,
+        active: bool,
+    ) -> Result<(MappedPages, AllocatedFrames), &'static str> {
+        // On aarch64, P4/P3/P2 should only contain valid
```
kevinaboos: some new changes in the main branch should address this. I've added a function called `PteFlagsArch::adjust_for_higher_level_pte()` that is the proper place to set these `higher_level_flags` (previously called `top_level_flags`)

aarch64/kernel/memory/src/paging/mapper.rs
```diff
@@ -0,0 +1,1280 @@
...
+
+    /// Maps the given `AllocatedPages` to randomly chosen (allocated) physical frames.
+    /// 
+    /// Consumes the given `AllocatedPages` and returns a `MappedPages` object which contains those `AllocatedPages`.
+    ///
+    /// This method assumes that the modified page is active
+    pub fn map_allocated_pages(&mut self, pages: AllocatedPages, flags: PteFlags)
+        -> Result<MappedPages, &'static str>
+    {
+        // On aarch64, P4/P3/P2 should only contain valid
```
kevinaboos: see other comment about `top_level_flags`

aarch64/kernel/memory/src/paging/table.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// Otherwise, the code assumes we are in identity-mapping, and
+    /// the physical addresses in a table are used as virtual addresses.
+    pub fn next_table_create(
+        &mut self,
+        index: usize,
+        flags: PteFlags,
+        active: bool,
+    ) -> Result<&mut Table<L::NextLevel>, &'static str> {
+        if self.next_table(index, active).is_none() {
+            // commenting until we understand how huge pages work on aarch64
```
kevinaboos, line 122: all the stuff about huge pages should be cfg-gated for x86 only in the latest changes from the main branch

aarch64/kernel/memory_aarch64/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+            // Allow the MMU to update the DIRTY flag.
+            + TCR_EL1::HD::Enable
+        );
+
+        barrier::isb(barrier::SY);
+    }
+}
+
+/// Installs a page table in the TTBR CPU register
+pub fn set_page_table_addr(page_table: PhysicalAddress) {
```
kevinaboos: 1. I think this should be called `set_as_active_page_table_root` or something like that
2. It should accept a reference to a `PageTable` to ensure that nobody can just directly call it with some random `PhysicalAddress`
3. Add a similar function to `memory_x86_64` and use it in `PageTable::switch()` so that the structure mimics that of aarch64

aarch64/kernel/nano_core/main.rs
```diff
@@ -47,6 +50,17 @@ fn main(
...
+    let mut mapped_regions = 0;
+    {
+        let (_, layout) = boot_svc.memory_map(&mut mmap).unwrap();
+        for descriptor in layout {
+            if descriptor.ty != MemoryType::CONVENTIONAL {
+                mapped_regions += 1;
+            }
+        }
+    }
+    let mut mapped_regions = Vec::with_capacity(mapped_regions + safety);
```
kevinaboos: what is this for? I'm a bit confused why we're iterating over available memory regions from UEFI multiple times

kevinaboos (changes requested):
A few things:
1. Can you incorporate the latest changes to the source files in `memory` (from the main branch) into your copied versions of them in `aarch64/kernel/memory/`? That will make it a lot easier to see what has changed.
2.  A lot of the arch-specific things in `memory` have been recently improved in `theseus_main` so you can probably avoid or move lots of those cfg(target_arch) parts.
3. In `page_table_entry`, you're directly modifying PTE flags. Instead of doing it there, move it elsewhere (e.g., into `PteFlagsArch::adjust_for_higher_level_pte()` or somewhere else in the `PteFlagsAarch64` module). Ideally the entire `page_table_entry` crate would be arch-independent, and any required flags would be handled in the `PteFlags<Arch>` module.

aarch64/kernel/nano_core/main.rs
NathanRoyer: The code counts the number of mapped regions so that it knows precisely how many slots to allocate in `Vec::with_capacity`

NathanRoyer: `layout` is only usable once (owned Iterator implementer) so the code iterates on it only once, and then later when boot services are exited we have the pre-allocated vec where we can push mapped regions

kevinaboos: I see, sounds good. One more question -- can we continue to safely use memory that was allocated by UEFI's heap allocator _after_ we've exited boot services? That is, `mapped_regions` belongs in that heap allocator's memory, which no longer exists after we exit boot services, right?

NathanRoyer: It's forbidden to call allocator functions but the memory is still mapped after boot services are exited, and the regions are present in the returned layout that we use to build the new page table.

kevinaboos: Ok, so once that heap item (the `Vec`) gets dropped it will call heap deallocation routines, so we must ensure that happens before exiting boot services, right? If not, won't undefined behavior occur by calling dealloc in a now-unusable heap allocator?

NathanRoyer: We should ensure it's never dropped I think, because that vec is used by paging::mod::init (which is called after boot services are exited); maybe we should wrap it in a NoDrop or use mem::forget after it's used

aarch64/kernel/memory/src/paging/mapper.rs
NathanRoyer: I changed the `aarch64` version of `adjust_for_higher_level_pte` so that it works; my implementation discards any flag set by the caller, which isn't great

aarch64/kernel/nano_core/main.rs
NathanRoyer: I simplified what's done in `nano_core` in the last commits

aarch64/kernel/memory_aarch64/src/lib.rs
NathanRoyer: One problem with 2: PageTable is defined in `memory`, which depends on `memory_aarch64` and `memory_x86_64`; we can't use PageTable in them. I'd keep it the way it is for simplicity (I think it would be a bit cumbersome to pull PageTable from `memory` while keeping the rest of paging there).

kevinaboos: ah, cyclic dependencies strike again. Okay, that's fine for now, we can always change it later.

kernel/pte_flags/src/pte_flags_aarch64.rs
```diff
@@ -379,9 +379,9 @@ impl PteFlagsAarch64 {
     /// * Sets the `VALID` bit, as every P4, P3, and P2 entry must be valid.
     #[must_use]
     pub fn adjust_for_higher_level_pte(self) -> Self {
-        self.executable(true)
-            .exclusive(false)
-            .valid(true)
+        Self::empty().accessed(true)
```
kevinaboos: I understand adding `.accessed(true)` and `.page_descriptor(true)`, but why start with `Self::empty()` and why remove the other parts I had there before? At the very least, we must force `exclusive` to `false`; I don't know about `executable `on aarch64 but I think what I had there is still correct.

Seems like the correct thing is:
```suggestion
        self.executable(true)
            .exclusive(false)
            .accessed(true)
            .page_descriptor(true)
            .valid(true)
```

kevinaboos (changes requested):
I left a comment about the `adjust_higher_level_pte_flags()` function on aarch64. 

One major issue is that I don't want to merge in the concept of `NextLevelAccess` to the `memory` crate in the main repo. I understand this is necessary because you're currently re-using the UEFI-provided page table, but after we finish merging in support for UEFI on x86, this won't be necessary. Therefore, I don't want to merge in such wide-sweeping, intrusive changes to all of the page translation code, and then have to undo it all later. 

So I think it's best to keep those changes in your `/aarch64/kernel` crates only. We can integrate aarch64 stuff into `/kernel/memory` after UEFI is complete, and that will also make it easier to implement `memory` in a mostly arch-agnostic manner, since I don't think most of it will need to change.

If you can remove those changes, I'm happy to accept this PR; everything else looks good.

aarch64/kernel/memory_aarch64/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+}
+
+/// Installs a page table in the TTBR CPU register
+pub fn set_as_active_page_table_root(page_table: PhysicalAddress) {
+    unsafe {
+        let page_table_addr = page_table.value() as u64;
+        TTBR0_EL1.write(
+              TTBR0_EL1::ASID.val(THESEUS_ASID as u64)
+            + TTBR0_EL1::BADDR.val(page_table_addr >> 1)
```
kevinaboos, line 193: Is this correct? I looked up the format of `TTBR0_EL1` and it doesn't seem to indicate that right-shifting it by 1 bit is what we should do.

aarch64/kernel/memory_aarch64/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        //
+        // For more information on MAIR, See section D17.2.97
+        // of [DDI0487l.a](https://l0.pm/arm-ddi0487l.a.pdf).
+        MAIR_EL1.write(
+            // Attribute 1 - Cacheable normal DRAM.
+            MAIR_EL1::Attr1_Normal_Outer::WriteBack_NonTransient_ReadWriteAlloc +
+            MAIR_EL1::Attr1_Normal_Inner::WriteBack_NonTransient_ReadWriteAlloc +
+
+            // Attribute 0 - Device.
+            MAIR_EL1::Attr0_Device::nonGathering_nonReordering_EarlyWriteAck,
```
kevinaboos: In `pte_flags_aarch64`, we expect the following:
* The [MAIR] index 0 has a Normal + Outer Shareable entry.
* The [MAIR] index 1 has a "DEVICE nGnRE" entry.

Perhaps I'm not understanding the code here, but it doesn't seem to match up. I don't see where the MAIR indices are specifically set.

kernel/page_table_entry/src/lib.rs
```diff
@@ -67,7 +68,17 @@ impl PageTableEntry {
...
-        PteFlagsArch::from_bits_truncate(self.0)
+        #[cfg(target_arch = "x86_64")]
+        {
+            PteFlagsArch::from_bits_truncate(self.0)
+        }
+
+        #[cfg(target_arch = "aarch64")]
+        {
+            let mut arch_flags = PteFlagsArch::from_bits_truncate(self.0);
+            arch_flags.remove(PteFlagsArch::PAGE_DESCRIPTOR);
```
kevinaboos: hm, I thought we wanted this bit to be set, since it indicates a page descriptor instead of a block descriptor. (I thought a block descriptor was similar to huge pages on x86)

Is this correct, or am I misunderstanding this flag's meaning? Here's what we wrote in `pte_flags_aarch64`:
```rust
        /// * If set, this represents a page descriptor.
        /// * If not set, this represents a block descriptor.
        const PAGE_DESCRIPTOR    = 1 << 1;
````

File: kernel/page_table_entry/src/lib.rs [hunk 11]
```diff
@@ -82,9 +93,19 @@ impl PageTableEntry {
...
     /// Extracts the value of the frame referred to by this page table entry.
     fn frame_value(&self) -> Frame {
-        let mut frame_paddr = self.0 as usize;
-        frame_paddr.set_bits(0 .. (PAGE_SHIFT as u8), 0);
-        Frame::containing_address(PhysicalAddress::new_canonical(frame_paddr))
+        #[cfg(target_arch = "x86_64")]
+        {
+            let mut frame_paddr = self.0 as usize;
+            frame_paddr.set_bits(0 .. (PAGE_SHIFT as u8), 0);
+            Frame::containing_address(PhysicalAddress::new_canonical(frame_paddr))
```
kevinaboos: I changed this in a recent PR to be the same as the aarch64 code, so this function should now be completely architecture independent. Go check out the latest code to see what i mean.

kernel/page_table_entry/src/lib.rs
```diff
@@ -93,15 +114,47 @@ impl PageTableEntry {
...
+        {
+            self.0 = (frame.start_address().value() as u64) | flags.bits();
+        }
+
+        #[cfg(target_arch = "aarch64")]
+        {
+            let mut flags = flags;
+
+            // telling the MMU that this is not a Block Descriptor
+            flags.insert(PteFlagsArch::PAGE_DESCRIPTOR);
```
kevinaboos: ok, now i'm even more confused, because here you _are_ always setting `PAGE_DESCRIPTOR` but in the above function you are always clearing it.

kevinaboos (changes requested):
Some of the changed files seem to be identical to those in the main repo. Can you comment here which ones didn't need to change? That'll make it easier to integrate later.

I left a few other comments. Aside from that, this PR looks good.

aarch64/kernel/memory_aarch64/src/lib.rs
NathanRoyer: My hint was the authors of `cortex-a` do that [here](https://docs.rs/cortex-a/latest/src/cortex_a/registers/ttbr0_el1.rs.html#60); From the manual, the BADDR field is defined like so:

image

the least significant bit is omitted; this is explained by the presence of the CnP field (which we don't use) which corresponds to the least significant bit of TTBR0_EL1; that way the MMU doesn't have to shift TTBR0_EL to get BADDR, it's only masked.

aarch64/kernel/memory_aarch64/src/lib.rs
NathanRoyer: Wow I didn't notice that, you're totally right I'll swap them

kernel/page_table_entry/src/lib.rs
NathanRoyer: It looks like I introduced a regression in that branch while I've been reverting the workspace merging commit... I'll double check the revert to see if this is my only mistake.

kernel/page_table_entry/src/lib.rs
NathanRoyer: see previous reply

kernel/page_table_entry/src/lib.rs
NathanRoyer: see previous reply

kernel/pte_flags/src/pte_flags_aarch64.rs
NathanRoyer: Ok, it works too so I'll push that

kevinaboos (approved):
Looking all good now! Thanks. 

One thing for the future: the `set_as_active_page_table_root()` fn being both public _and_ safe is not sound, since it allows anyone to change the root page table to any arbitrary `PhysicalAddress`. All other pub functions in `memory_<arch>` are safe to expose publicly. 

There are a few easy solutions to this:
* move `memory_x86_64` and `memory_aarch64` crates into modules within the `memory` crate and make that fn private or `pub(crate)`
* move the logic back into `PageTable::switch()` and have arch-specific cfg blocks directly in that function
  * This is currently what I have done in this PR
* any other idea you may have, etc

PR #705, opened 2022-12-01
Fix `DEVICE_MEMORY` PTE flags on aarch64

NathanRoyer:
On AArch64:
- `PteFlags::DEVICE_MEMORY` → `PteFlagsAarch64::DEVICE_MEMORY`
- `PteFlagsAarch64::DEVICE_MEMORY` → `PteFlagsAarch64::MAIR_INDEX_0`
- `PteFlagsAarch64::MAIR_INDEX_0` = 0

This makes every page table entry appear as if it described device memory because a zero flag is always present.

This PR makes `PteFlags::DEVICE_MEMORY` work like `WRITABLE` & `GLOBAL`: it has an inverted value on aarch64 and the bit is flipped during conversion to and from arch-specific `PteFlags`.

### DIFF
--- a/kernel/pte_flags/src/lib.rs
+++ b/kernel/pte_flags/src/lib.rs
@@ -28,6 +28,7 @@
 
 use cfg_if::cfg_if;
 use bitflags::bitflags;
+use static_assertions::const_assert_ne;
 
 cfg_if!{ if #[cfg(any(target_arch = "x86_64", doc))] {
     mod pte_flags_x86_64;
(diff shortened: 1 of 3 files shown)


kevinaboos:
Thanks. As we discussed on discord, ranges of multiple bits that are used as indices or selectors between more than 2 values cannot really be toggled like other single-bit flags. So we have to mask out the full bit range and then set the value manually.

I've pushed a commit that fixes this, kindly let me know if you see any issues with it.

kevinaboos:
I need this to be merged in for some other work I'm doing, so I'm going to go ahead and accept and merge it in.

Kindly do let me know if you see anything wrong with my fix for `DEVICE_MEMORY` conversions; we can always fix it in the future too.

kevinaboos (approved):

Issue #703, opened 2022-11-25
`async_channel` marks itself as disconnected too eagerly


kevinaboos:
Looks like this was added in #277 (during a paper deadline crunch) so the author was likely only testing it with a single endpoint of each variety. 

Should be a simple enough fix: only mark the channel as disconnected when the _last_ Sender or Receiver gets dropped, not just any sender or receiver.

_Originally posted by @kevinaboos in https://github.com/theseus-os/Theseus/pull/668#discussion_r1031940127_

amab8901:
I think I solved it. Check out my PR and let me know what you think :smile:

kevinaboos:
> I think I solved it. Check out my PR and let me know what you think smile

Thanks! I have left a review on #707.

PR #707, opened 2022-12-02
Mark `async_channel` as disconnected only when the last `Sender` or the last `Receiver` is dropped

amab8901:
Solves #703


kevinaboos:
Thanks for the PR @amab8901! Much appreciated.

I think this looks pretty good, just need to test it of course.

@tsoutsman was the one who identified this problem initially as part of #668, so I'll let him leave a review.

kevinaboos (changes requested):
Okay I had a chance to review this, and I think you may have misunderstood the condition for disconnecting the channel. In this PR, you are checking the number of remaining elements in the channel's queue and using that as the determining factor for whether it is disconnected. This would incorrectly mark the channel as disconnected any time we drop either a sender or receiver and there are no outstanding elements in the queue.

What we actually want to do is to check how many senders and receivers there are (i.e., the `strong_count` of the inner `Arc` inside `Sender`/`Receiver`), and only mark the channel as disconnected if we're dropping the last sender or the last receiver.

Also, this itself is a policy choice. This policy is fine by me, but we should then allow for reconnection by adding new methods like `Sender::obtain_receiver()` and `Receiver::obtain_sender()`, which would change the channel status back to `ChannelStatus::Connected` when we transition from having 0 --> 1 `Sender`s or `Receiver`s.

tsoutsman:
> What we actually want to do is to check how many senders and receivers there are (i.e., the strong_count of the inner Arc inside Sender/Receiver), and only mark the channel as disconnected if we're dropping the last sender or the last receiver.

Senders and receivers use the same `Arc`, so it's not quite as simple as checking the strong count. I think we need to store two additional `AtomicUsize`s tracking the number of senders and receivers. Using just one (e.g. tracking senders) and then tracking receivers by subtracting the senders from the strong count could lead to race conditions as the strong count and senders are loaded separately.

> Also, this itself is a policy choice. This policy is fine by me, but we should then allow for reconnection by adding new methods like Sender::obtain_receiver() and Receiver::obtain_sender(), which would change the channel status back to ChannelStatus::Connected when we transition from having 0 --> 1 Senders or Receivers.

Is there a point to tracking disconnected senders and receivers if one can be created? Does it make sense to fail a blocking operation because there are no senders/receivers, even if another one may be created in the future?

amab8901:
`AtomicUsize` is part of std crate, so we cannot use it directly via `use std::sync::atomic::AtomicUsize`. Idk, maybe we could create our own version of `AtomicUsize` internally? What are your thoughts on that?

> > What we actually want to do is to check how many senders and receivers there are (i.e., the strong_count of the inner Arc inside Sender/Receiver), and only mark the channel as disconnected if we're dropping the last sender or the last receiver.
> 
> Senders and receivers use the same `Arc`, so it's not quite as simple as checking the strong count. I think we need to store two additional `AtomicUsize`s tracking the number of senders and receivers. Using just one (e.g. tracking senders) and then tracking receivers by subtracting the senders from the strong count could lead to race conditions as the strong count and senders are loaded separately.
> 
> > Also, this itself is a policy choice. This policy is fine by me, but we should then allow for reconnection by adding new methods like Sender::obtain_receiver() and Receiver::obtain_sender(), which would change the channel status back to ChannelStatus::Connected when we transition from having 0 --> 1 Senders or Receivers.
> 
> Is there a point to tracking disconnected senders and receivers if one can be created? Does it make sense to fail a blocking operation because there are no senders/receivers, even if another one may be created in the future?

amab8901:
> Is there a point to tracking disconnected senders and receivers if one can be created? 

I don't see any obvious reason why we should track disconnected senders/receivers. Even if it were meaningful to do so, I think it would be outside the scope of this PR, and would be best addressed in a separate PR so that we can keep our PR:s simple and modular (many simple PR:s rather than few giant PR:s). 

> Does it make sense to fail a blocking operation because there are no senders/receivers, even if another one may be created in the future?

A sender/receiver may indeed be created in the future. But how long will it take? If it happens soon, then it's not an issue. But if the creation of sender/receiver is delayed for extended periods, or perhaps never happens at all, then the blocking operation would waste resources. So to prevent resource waste, I think it's better to fail the blocking operation in the absence of senders/receivers. The fail may cause short-term pain if it occurs, but it provides the greater long-term value of pointing out the need to solve a design flaw. 

At least that's how I see it.

kevinaboos:
> Senders and receivers use the same `Arc`, so it's not quite as simple as checking the strong count. I think we need to store two additional `AtomicUsize`s tracking the number of senders and receivers. Using just one (e.g. tracking senders) and then tracking receivers by subtracting the senders from the strong count could lead to race conditions as the strong count and senders are loaded separately.

Right, that's what I was getting at but I do appreciate the specificity and clarify. We'd have to refactor it a bit to add those counts.

-----
> > Does it make sense to fail a blocking operation because there are no senders/receivers, even if another one may be created in the future?
> 
> A sender/receiver may indeed be created in the future. But how long will it take? If it happens soon, then it's not an issue. But if the creation of sender/receiver is delayed for extended periods, or perhaps never happens at all, then the blocking operation would waste resources. So to prevent resource waste, I think it's better to fail the blocking operation in the absence of senders/receivers. The fail may cause short-term pain if it occurs, but it provides the greater long-term value of pointing out the need to solve a design flaw.


Regarding whether we want to bother with having disconnection semantics given that other endpoints may re-connect in the future, the short answer is "yes". We don't need to make it overly complex with fancy semantics, but I believe the correct/best policy is to inform a sender when there are no longer any receivers, and likewise to inform a receiver when there are no longer any senders. 

This is similar to the idea of wanting to inform a waiting task `A` that another task `B` (that `A` is waiting on) has failed; it's just the "right" thing to do because it prevents the waiting task (or in the case of the channel, the opposite endpoint) from waiting endlessly. In my view, channels typically embody semantics like this, as opposed to something that is more "long-term asynchronous" and intended for communication over a less reliable medium, like network sockets. 

Note that the original purpose of this was to facilitate fault tolerance and recovery, so notifications of failure/disconnection is indeed necessary and desirable.

kevinaboos (changes requested):
In any case, as @tsoutsman pointed out, we do need a bit more complex solution for `Sender`- and `Receiver`-specific counts. 

We'd also need the `Receiver::obtain_sender()` and `Sender::obtain_receiver()` methods that I mentioned above in order to actually allow the channel to recover from a disconnection. 

Also, adding some test scenarios to ensure things work as expected would be a good idea; see `applications/test_channel` for more, as it provides a starting point to which you could add more tests.

amab8901:
I added a counter for sender and receiver, respectively. When creating a new channel, the sender and receiver counter will each initiate with value 1. When dropping, it will subtract the value by 1. I'm not sure about adding senders/receivers beyond a single instance. Is this something you would achieve via `clone`, or will it be added in the future but doesn't exist yet?

totoroyyb:
Hi all, I just started to follow this discussion and took a brief look at the current issue we are facing.

> I added a counter for sender and receiver, respectively. When creating a new channel, the sender and receiver counter will each initiate with value 1. When dropping, it will subtract the value by 1.

Probably, we can refactor the `count` into the sender and receiver themselves. Therefore, when cloning the receiver or sender, we can delegate the receiver and sender to manage their `count` directly. From the big picture, it doesn't really make much difference compared to managing `count` in the channel.

> I'm not sure about adding senders/receivers beyond a single instance. Is this something you would achieve via clone, or will it be added in the future but doesn't exist yet?

From my observation, probably yes? Therefore, based on the refactor I mentioned, we can maybe implement the `clone` trait to customize the clone behavior and manage the internal `count`.

I might misunderstand things here; just my two cents.

amab8901:
sorry for spamming these force-pushes. I was trying to incorporate latest changes into my local repository. I ran into this cycle:
1) remote forked repo is 9 commits behind the main repo
2) I click `synchronize`
3) it says `3 commits ahead of main repo` (but no commits behind)
4) I run `git pull` followed by `git rebase`
5) `git add .`
6) `git push --force`
7) back to step # 1
Idk if I remember all details correctly but it went something like that

amab8901:
@totoroyyb 

I'm thinking about your suggestion.
> we can refactor the count into the sender and receiver themselves. Therefore, when cloning the receiver or sender, we can delegate the receiver and sender to manage their count directly. From the big picture, it doesn't really make much difference compared to managing count in the channel.

If we do this, wouldn't the count then be of the total number of receivers/senders? I think we're trying to count only those receivers/senders that are within some given channel (rather than the total across multiple channels). 

> From my observation, probably yes? Therefore, based on the refactor I mentioned, we can maybe implement the clone trait to customize the clone behavior and manage the internal count.

I tried implementing the trait, but the compiler wouldn't let me. As I understand it, implementing Clone on Sender/Receiver requires you to first implement it on the structs within the Sender/Receiver - in this case Arc. And Rust doesn't allow manual implementation of Clone on Arc (via `impl` syntax) because Arc is defined outside of the crate. Did I understand it correctly? Is there a way around this? (keep in mind that I'm kinda junior so it's possible I may have just misunderstood something)

kevinaboos:
> sorry for spamming these force-pushes. I was trying to incorporate latest changes into my local repository. I ran into this cycle:
> 
> 1. remote forked repo is 9 commits behind the main repo
> 2. I click `synchronize`
> 3. it says `3 commits ahead of main repo` (but no commits behind)
> 4. I run `git pull` followed by `git rebase`
> 5. `git add .`
> 6. `git push --force`
> 7. back to step # 1
>    Idk if I remember all details correctly but it went something like that

No problem. FYI, the reason that happens is that you are doing a rebase, which rewrites history by "replaying" your branch's 3 local commits over the top of the 9 commits that came from the main branch. Instead of a `rebase` you can just do a `merge`, which won't require a force push since you wouldn't be rewriting history.

kevinaboos:
> I added a counter for sender and receiver, respectively. When creating a new channel, the sender and receiver counter will each initiate with value 1. When dropping, it will subtract the value by 1. 

Great! That's correct so far. The next step is to actually set the channel to disconnected and notify any remaining endpoints on the other side. That would look something like this for the `Receiver`, and the `Sender` impl of `Drop` will be similar.
```rust
/// When the only remaining `Receiver` is dropped, we mark the channel as disconnected
/// and notify all of the `Sender`s.
impl<T: Send> Drop for Receiver<T> {
    fn drop(&mut self) {
        // trace!("Dropping a receiver");
        if self.channel.receiver_count.fetch_sub(1, Ordering::SeqCst) == 1 {
            self.channel.channel_status.store(ChannelStatus::ReceiverDisconnected);
            self.channel.waiting_senders.notify_all(); // we need to add this to `WaitQueue`
        }
    }
}
```

> I'm not sure about adding senders/receivers beyond a single instance. Is this something you would achieve via `clone`, or will it be added in the future but doesn't exist yet?

Yep, any time a new `Sender` or `Receiver` is created, the counts would need to be updated accordingly. That would occur on a `Clone` operation for `Sender`/`Receiver`, and also in new functions like `Sender::obtain_receiver()` or `Receiver::obtain_sender()` (if we choose to add those).

kevinaboos:
> we can refactor the count into the sender and receiver themselves. Therefore, when cloning the receiver or sender, we can delegate the receiver and sender to manage their count directly. From the big picture, it doesn't really make much difference compared to managing count in the channel.

That sounds nice in theory, but in practice it would require another `Arc` (which is a heap-alloc'd object) to ensure that the sender count is shared across all `Sender`s and that the receiver count is shared across all `Receiver`s. Putting the counts in the `Channel` object avoids the need for more `Arc`s.


> I tried implementing the trait, but the compiler wouldn't let me. As I understand it, implementing Clone on Sender/Receiver requires you to first implement it on the structs within the Sender/Receiver - in this case Arc. And Rust doesn't allow manual implementation of Clone on Arc (via `impl` syntax) because Arc is defined outside of the crate. Did I understand it correctly? Is there a way around this? (keep in mind that I'm kinda junior so it's possible I may have just misunderstood something)

No worries, this is a good learning opportunity. You're correct that you cannot implement a foreign trait on a foreign type; here, "foreign" means "defined in another crate". Check out some articles on Rust's "orphan rule" to see why this rule exists. In your example, you cannot do something like `impl<T> Clone for Arc<T>` because your crate didn't define the `Clone` trait nor the `Arc` type. 

Luckily, we do define both the `Sender` and `Receiver` types, so you should be able to implement `Clone` for them without problems. It would look something like:
```rust
impl<T: Send> Clone for Receiver<T> {
    fn clone(&self) -> Self {
        if self.channel.receiver_count.fetch_add(1, ...) == 0 {
            self.channel.channel_status.store(ChannelStatus::Connected);
        }
    }
}         
```

Note that since we're now manually implementing `Clone` to ensure the counts and channel status are correct, we must remove the auto-derived impl of `Clone`:
```diff
  /// The receiver side of a channel.
- #[derive(Clone)]
  pub struct Receiver<T: Send> {
      channel: Arc<Channel<T>>,
  }
```

Then you can apply similar logic to `Sender`.

amab8901:
ok I think I completed all tasks so far. I look forward to feedback.

kernel/async_channel/src/lib.rs
```diff
@@ -246,10 +262,22 @@ impl <T: Send> Sender<T> {
 }
 
 /// The receiver side of a channel.
-#[derive(Clone)]
 pub struct Receiver<T: Send> {
     channel: Arc<Channel<T>>,
 }
+
+impl<T: Send> Receiver<T> {
```
kevinaboos: forgot the `Clone` trait here

```suggestion
impl<T: Send> Clone for Receiver<T> {
```

kernel/async_channel/src/lib.rs
```diff
@@ -347,20 +375,28 @@ impl <T: Send> Receiver<T> {
...
     fn drop(&mut self) {
-        // trace!("Dropping the receiver");
-        self.channel.channel_status.store(ChannelStatus::ReceiverDisconnected);
-        self.channel.waiting_senders.notify_one();
+        // trace!("Dropping a receiver");
+        if self.channel.receiver_count.fetch_sub(1, Ordering::SeqCst) == 1 {
+            self.channel.channel_status.store(ChannelStatus::ReceiverDisconnected);
+            self.channel.waiting_senders.notify_all();
+        }
+        self.channel.receiver_count.fetch_sub(1, Ordering::SeqCst);
```
kevinaboos: The `receiver_count` was already decremented at the beginning of this function. Is there a reason to do it twice?

kernel/async_channel/src/lib.rs
```diff
@@ -347,20 +375,28 @@ impl <T: Send> Receiver<T> {
...
     fn drop(&mut self) {
-        // trace!("Dropping the sender");
-        self.channel.channel_status.store(ChannelStatus::SenderDisconnected);
-        self.channel.waiting_receivers.notify_one();
+        // trace!("Dropping a sender");
+        if self.channel.sender_count.fetch_sub(1, Ordering::SeqCst) == 1 {
+            self.channel.channel_status.store(ChannelStatus::SenderDisconnected);
+            self.channel.waiting_receivers.notify_all();
+        }
+        self.channel.sender_count.fetch_sub(1, Ordering::SeqCst);
```
kevinaboos: The `sender_count` was already decremented at the beginning of this function. Is there a reason to do it twice?

kevinaboos (changes requested):
I left a few small comments. 

Also, the changeset now includes some changes to two submodules, `region` and `wasmtime`. Can you undo those changes? They don't seem to be relevant to this PR.

kevinaboos:
I also just merged in a small PR #724 that implements `WaitQueue::notify_all()` in more efficient manner. Kindly take a look at it to see how I implemented that and how it differs from what was added here. You'll also want to merge that change into this PR branch.

kernel/async_channel/src/lib.rs
amab8901: oh, no I just missed the fact that it's decremented twice. Thanks for pointing out

amab8901:
I'll have a look at it, thanks

kernel/async_channel/src/lib.rs
kevinaboos: no problem, i figured

amab8901:
ok I think it's good now. 

I imitated your implementation of `WaitQueue::notify_all()` but I removed a redundant semicolon after the if-statement.

kevinaboos:
> ok I think it's good now.

Thanks, yeah it looks good! However you'll still have to merge in the actual commits from the latest `theseus_main` branch in order to properly resolve the conflicts in git's eyes. There are also still some unrelated changes in two `ports/` folders, kindly remove those.
 
> I imitated your implementation of `WaitQueue::notify_all()` but I removed a redundant semicolon after the if-statement.

Thanks for catching that, I did it real quickly 😄.  Kindly remove the now-outdated doc comments about return values on that function too.

amab8901:
this git stuff can be a bit tricky :thinking:

kevinaboos:
> this git stuff can be a bit tricky 🤔

yea, it's tricky for sure. I think the problem is likely related to the fact that you're trying to use `rebase`, which I'm assuming is the case since I just saw you do a force push. Let me try to help out a bit:

Assuming you are currently on your own branch `703-async-channel-disconnection`, you need to check out the main branch and then merge it into this branch.
```sh
$ git checkout theseus_main
$ git pull upstream theseus_main
```
now your local instance of `theseus_main` has the latest changes.
(This assumes you have a remote called `upstream` which tracks the `theseus_main` branch from `theseus-os/Theseus`. If not, you can add that remote by running `git remote add upstream git@github.com:theseus-os/Theseus.git`, and then re-run the above `pull` command.

Now that you have all the latest changes from the upstream `theseus_main` branch in your local repo, you just need to merge them into this branch:
```sh
$ git checkout 703-async-channel-disconnection
$ git merge theseus_main
```
Now you can resolve conflicts manually, add the files to mark them as resolved, commit, and the push again:
```sh
## edit your files to resolve conflicts
$ git add <the files you changed>
$ git commit
$ git push origin 703-async-channel-disconnection
```

After that, this PR page should look good.

amab8901:
Thanks, is it good now?

kevinaboos:
> Thanks, is it good now?

Whoa, not sure what happened, but there are tons of unrelated changes now. I think your merge went wrong somewhere.

Quick tip: if you take a look at the [`Files changed`](https://github.com/theseus-os/Theseus/pull/707/files) tab of this PR, you can see the exact changes you're proposing to be merged in. Right now it shows 27 files, most of which are unrelated.

amab8901:
I think those unrelated changes are due to the fact that I pulled and merged from `upstream` to keep everything updated. I merged it locally so it says locally that "everything is up to date" basically

amab8901:
So on one hand, you want me to pull and merge from the main branch. And on the other hand, you don't want unrelated changes. Are these two goals compatible?

kevinaboos:
> I think those unrelated changes are due to the fact that I pulled and merged from `upstream` to keep everything updated. I merged it locally so it says locally that "everything is up to date" basically

Sorry for the trouble, but yes, that is indeed correct. You likely made a mistake when resolving conflicts during the merge, but I don't know for sure.

> So on one hand, you want me to pull and merge from the main branch. And on the other hand, you don't want unrelated changes. Are these two goals compatible?

Yes, definitely, this is a common use case for git. Doing it that way will allow the git diff to reflect only the new changes you are proposing, all without any conflicts. 

------------
In any case, your PR now fails the CI checks, so something is definitely not right. Kindly check out the `Files changed` tab that I mentioned before to see what changes you're now proposing with this PR. What we expect to see there is just a few changes to `async_channel`, some cargo files, and perhaps `waitqueue` (if you made a small fix like removing a semicolon). Nothing else should be present.

kevinaboos:
In the spirit of creating a git learning opportunity, I took a quick look and it looks like you have _two_ merge commits. The [first one](https://github.com/theseus-os/Theseus/pull/707/commits/30143403e0b4e6ee4c212a4750acbc6e18a7c5c5) may be towards the correct path, but the [last one](https://github.com/theseus-os/Theseus/pull/707/commits/7ad680efd9477aa6f336bc0772a91d5ad4e139b3) was likely an accident because it has a lot of conflict markers in the source files. 

Try removing that latter commit and seeing what things look like then; hopefully that'll put you back on the right track.

amab8901:
wow I can't believe the checks have actually passed after I pulled and merged :sweat:

kevinaboos:
Glad to see the CI checks passing! The `async_channel` and `wait_queue` changes look good.

Unfortunately, you still have a lot of unrelated changes in this PR. Kindly remove files or changes that are irrelevant to this PR, or if you're stuck on how to do this with git, I can do it for you. Up to you. 

I suspect that you're doing something like `git add .`, which adds _all_ files and directories recursively in the current directory tree. Pro-tip: you almost never want to do that, and that's likely why all these unrelated file changes are being added.

You've accidentally committed a bunch of `.orig` files that are auto-created backups that happen when resolving conflicts incorrectly or bailing out during a merge. These should be removed.
* `Cargo.lock.orig`
* `kernel/memory_structs/src/lib.rs.orig`
* `kernel/memory_x86_64/src/lib.rs.orig`
* `kernel/page_table_entry/Cargo.toml.orig`
* `kernel/page_table_entry/src/lib.rs.orig`
* etc

Other irrelevant changes that should be removed/undone are (mostly just new lines added or removed):
* `kernel/memory_structs/src/lib.rs`
* `kernel/memory_x86_64/src/lib.rs`
* `kernel/net/src/lib.rs`
* `kernel/page_table_entry/src/lib.rs`
* `kernel/pte_flags/src/pte_flags_aarch64.rs`
* etc

Please take a look at the `Files changed` tab next time you finish some commits to double-check that everything there is correct. For this PR, based on our discussion, I would only expect changes to `async_channel` and `wait_queue`.  Once you're sure that the changeset is correct, please then go ahead and re-request a review from us.

If you have any additional specific questions on how to use git, feel free to post them here and I will try to answer them; there are also lots of great tutorials on sites like github, atlassian, and stackoverflow.

amab8901:
Here's what we have now:
* [x] no unrelated changes
* [x] all checks have passed

Did I forget anything?

tsoutsman (approved):

amab8901:
wow it's amazing that I made it all the way past approval. It was really educational to figure out this problem and receive feedback from you guys, thanks :smile:

kevinaboos:
> wow it's amazing that I made it all the way past approval. 

No problem, and thanks very much for your contribution!

> It was really educational to figure out this problem and receive feedback from you guys, thanks 😄

Great, glad to hear it. We all start somewhere; I remember the challenges I struggled through when learning git too. Hell, even today I still run into git problems. I hope I didn't come off as too pedantic 😄.

kevinaboos (approved):

PR #708, opened 2022-12-03
Use new `PteFlags` to replace old `EntryFlags`

kevinaboos:


kevinaboos:
cc @NathanRoyer on this since the changes to the `memory`-related crates may affect your other work & current PRs

PR #709, opened 2022-12-04
Add UEFI boot option

tsoutsman:
The port isn't complete. This PR doesn't implement GOP, and hence the captain loops before initialising the framebuffer when compiling for UEFI.

Summary of changes:
- Abstract over multiboot2 and UEFI boot info
- Move `nano_core` assembly compilation to build script
- Create different versions of `_start` in assembly for UEFI and BIOS. The UEFI version uses a naked function as it's quite short.
- Modify the nano_core assembly files so that the VBE card and VGA aren't accessed when booting APs on UEFI.
- Add nasm installation to docs workflow (nano_core fails to build otherwise)
- Add UEFI variable to Makefile
- Rename `iso` to `image` in Makefile
- Use RSDP provided by bootloader if available
- Modify paging initialiser not to assume that stack and boot info frames are contiguous
- Load the nano_core page table from `Cr3` rather than the `.page-table` section.
- Load the stack address by computing `initial_double_fault_stack - stack_size` rather than the `.stack` section
- Modify `AggregatedSectionMemoryBounds` not to include the `.page-table` and `.stack` sections as they do not exist in the nano_core binary when compiling for UEFI
- Change the `EXTRAF_FILES_DIRECTORY_DELIMITER` to `!` as `?` is not a valid FAT character
- Rename `start` to `_start` to follow naming conventions
- Create `uefi_builder`

I'm not sure what to do about VGA printing. For now, I've disabled it completely for both BIOS and UEFI.

[Summary of bootloader changes](https://github.com/theseus-os/bootloader/commit/6cc130e2f9dc633445846a1e2c4d961b1daddd78).


tsoutsman:
Closed in favour of #745.

PR #711, opened 2022-12-04
Add time to doc index

amab8901:
[this](https://www.theseus-os.com/Theseus/doc/___Theseus_Crates___/index.html) page doesn't mention `time` in the list. I solved it.


kevinaboos:
Thanks! That list is pretty out-of-date and should be regenerated. It'd be ideal if we could auto-generate it (I did that manually once by grepping for the `description` field in each `Cargo.toml`) as part of a CI step, but that's fairly low priority.

kevinaboos (approved):

PR #717, opened 2022-12-06
Add a basic asynchronous executor

tsoutsman:
Turns out it's easier to use one task per future rather than one executor task running all the futures.

This executor is pretty useless at the moment as there's nothing useful a future can await on (e.g. timer or a network read).


tsoutsman:
I added asynchronous `sleep` and `spawn` functions.

kevinaboos:
This is a great start! I think we can basically merge this in as is, and then improve it or expand upon it later with additional executor policies, various N-to-M `Future`-to-`Task` mappings, etc. 

My one concern is including a `Waker` object directly in the `Task` struct; I like basically everything else in this PR. 
I understand that it's being used to notify/wake a waker upon task exit, but in my mind the ideal design of the task subsystem is that it sits beneath everything else, i.e., a task is independent of anything related to any async executor or the concept of a Future. 
There are other ways to accomplish notifying wakers of task exit, e.g.:
* something more abstract like a task exit callback (similar to the current unused idea of a kill handler)
* supporting more conventional join handling such that dreadnought can be the main joiner of any task spawned for the purpose of running an async workload
* perhaps guard types on the stack of the task that get dropped as part of it running to completion or being unwound
 
If you have any other ideas we can discuss them too.

tsoutsman:
I was considering adding an asynchronous flag to spawn's task builder. It would use r14 to pass the task wrapper a pointer to the waker which it would use when the task completes.

It would require a bit of dependency finagling because spawn would have to depend on the concrete waker struct rather than the trait meaning spawn would depend on dreadnought which already depends on spawn, but that's solvable.

kevinaboos:
> I was considering adding an asynchronous flag to spawn's task builder.

Adding things to the `TaskBuilder` is fine, but typically it's only used to provide extra information that either eventually goes into the new `Task` struct itself, or is used for initial scheduling. So I'm not sure if that's really needed here.

>  It would use r14 to pass the task wrapper a pointer to the waker which it would use when the task completes.

Ah, that's odd. I already don't love my solution for using `r15` for things like this, so I definitely don't want to continue that trend. Any reason why we simply can't pass it into the new task as a regular argument?

kernel/dreadnought/src/task.rs
```diff
@@ -0,0 +1,92 @@
...
+    pin::Pin,
+    task::{Context, Poll},
+};
+use task::{ExitValue, JoinableTaskRef, KillReason, PanicInfoOwned};
+
+/// Spawn a new asynchronous task, returning a [`JoinHandle`] for it.
+///
+/// You do not need to poll the handle to make the task execute --- it will
+/// start running in the background immediately.
+pub fn spawn<F>(future: F) -> JoinHandle<F::Output>
```
kevinaboos: would it be ok to change this to `spawn_async` just for clarity? I believe `tokio` uses the same nomenclature, and it would further help us distinguish between `task::Task` and dreadnought's async tasks.

kernel/dreadnought/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+pub use futures::{future, pin_mut, select_biased, FutureExt};
+
+pub mod task;
+pub mod time;
+
+/// Executes a future to completion.
+///
+/// This runs the given future on the current thread, blocking until it is
+/// complete, and yielding its result.
+pub fn block_on<F>(future: F) -> F::Output
```
kevinaboos, line 40: should `block_on()` return a Result? Currently there's only one error case (w/ MutexSleep usage) but I can foresee others, e.g., remotely killing an async task, anything else panicking, etc

kernel/dreadnought/src/task.rs
kevinaboos: same as `block_on()`, i think this should return a `Result`

kernel/dreadnought/src/task.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+impl<T> JoinHandle<T> {
+    /// Abort the task associated with the handle.
+    ///
+    /// If the cancelled task was already completed at the time it was
+    /// cancelled, it will return the successful result. Otherwise, polling the
+    /// handle will fail with an [`Error::Cancelled`].
+    pub fn abort(&self) {
+        let _ = self.task.kill(KillReason::Requested);
```
kevinaboos, line 52: add a TODO FIXME thing here b/c task killing is hot garbage rn

kevinaboos: even better would be a huge `## Warning` section in the doc comments, lol

kernel/dreadnought/src/task.rs
```diff
@@ -0,0 +1,92 @@
...
+    type Output = Result<T>;
+
+    fn poll(self: Pin<&mut Self>, context: &mut Context<'_>) -> Poll<Self::Output> {
+        self.task.set_waker(context.waker().clone());
+        if self.is_finished() {
+            Poll::Ready(match self.task.retrieve_exit_value() {
+                Some(exit_value) => match exit_value {
+                    ExitValue::Completed(value) => {
+                        // SAFETY: The task ran block_on which returns a T.
+                        // TODO: Why doesn't T have to be Any?
```
kevinaboos: The type def of `ExitValue(Completed)` is `Box<dyn Any + Send>` so it is indeed `Any`. That's why you can downcast it in the first place.

kevinaboos (changes requested):
Left a few small comments.

In the future (after `time` is completed) we should re-do `sleep` entirely and fix the way it's used in this crate. But for now, all good. Thanks!

kernel/dreadnought/src/task.rs
tsoutsman: Tokio uses [`spawn`](https://docs.rs/tokio/latest/tokio/fn.spawn.html). We can still change it, but I'm unsure if it's necessary. A dreadnought async task __is__ a `task::Task`. Also, there is no `spawn::spawn`, so it couldn't really be confused with another task.

It is just a name though, so if you feel it should be changed I'm completely ok with it.

kernel/dreadnought/src/lib.rs
tsoutsman: Could you elaborate on the other error cases? What does remotely killing an async task mean? If something panics, then `block_on` will also panic; the point of `block_on` is to return a future as if it's a normal function.

For reference, Tokio's [`block_on`](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on) returns `F::Output`.

kernel/dreadnought/src/task.rs
tsoutsman: I don't think so, since we're passing in a `T`. Imagine if `T` didn't implement `Any`, `downcast_unchecked` wouldn't make any sense, even if the thing inside the box does implement `Any`. It would be impossible to construct a `Box<dyn Any>` from a `T` in the first place.

It's because of [this blanket implementation](https://doc.rust-lang.org/std/any/trait.Any.html#impl-Any-for-T). The `'static` bound on `T` guarantees that `Any` is implemented for `T`.

kernel/dreadnought/src/task.rs
tsoutsman: Doesn't really make sense to leave a TODO FIXME since the problem is with `Task::kill`, not `JoinHandle::abort`. I added a doc comment.

kernel/dreadnought/src/task.rs
kevinaboos: oh ok, i swear i saw a `spawn_async` function in some Rust project, but it obviously wasn't tokio. Nbd, this is fine.

kernel/dreadnought/src/lib.rs
kevinaboos: all good, i just tend to prefer avoiding panics where possible.

kernel/dreadnought/src/task.rs
kevinaboos: oh you're talking about `T`. Yes, `Any` is auto-impl'd for static types

kevinaboos (approved):
I went ahead and ended up changing some naming conventions to be more explicitly async. Will likely change it again after I rename some components in `task`.

I also removed the unsafe downcast in favor of a checked one, and removed the now-unnecessary features

PR #722, opened 2022-12-09
Add new graphics layer

ouz-a:
# Porthole
Porthole, is new graphics layer for the Theseus OS, inspired by the old graphic's stack written from ground up Porthole has several goals:
- Make it easy to use.
- KISS (E.g; We won't write code that is going increase performance by %2 while making it harder to understand)
- Performance is high priority.
- It should be modular.

## Design
Detailed design of Porthole will be written here.

## Tasks:
- [X] Display pixels
- [X] Display text
- [X] Resizable windows
- [X] Create the initial PR
- [ ]  Handle keyboard
- [X]  Handle rendering order
- [X]  Error handling
- [ ]  Replace old graphics stack entirely.

_This is very early phase code most of the it going change after reviews/discussions_

### DIFF
--- a/applications/porthole/src/lib.rs
+++ b/applications/porthole/src/lib.rs
@@ -17,7 +17,7 @@ use core::marker::PhantomData;
 use core::ops::{Add, Sub};
 use core::slice::IterMut;
 use log::{debug, info};
-use spin::{Mutex, MutexGuard, Once};
+use spin::{Mutex, MutexGuard, Once, RwLockReadGuard};
 
 use event_types::Event;
 use mpmc::Queue;
@@ -25,8 +25,10 @@ use mpmc::Queue;
 use alloc::string::{String, ToString};
 use alloc::vec::Vec;
 use font::{CHARACTER_HEIGHT, CHARACTER_WIDTH, FONT_BASIC};
-use hpet::get_hpet;
-use memory::{BorrowedSliceMappedPages, Mutable, PhysicalAddress, PteFlags, PteFlagsArch};
+use hpet::{get_hpet, Hpet};
+use memory::{
+    BorrowedMappedPages, BorrowedSliceMappedPages, Mutable, PhysicalAddress, PteFlags, PteFlagsArch,
+};
 use mouse_data::MouseEvent;
 pub static WINDOW_MANAGER: Once<Mutex<WindowManager>> = Once::new();
 static TITLE_BAR_HEIGHT: usize = 20;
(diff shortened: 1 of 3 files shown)


kernel/multicore_bringup/src/lib.rs
```diff
@@ -60,11 +60,11 @@ pub static GRAPHIC_INFO: Mutex<GraphicInfo> = Mutex::new(GraphicInfo::new());
 #[repr(packed)]
 pub struct GraphicInfo {
     /// The visible width of the screen, in pixels.
-    width: u16,
+    pub width: u16,
```
kevinaboos: these are private to avoid accidental modification; there are some accessor functions that you can use instead: https://www.theseus-os.com/Theseus/doc/multicore_bringup/struct.GraphicInfo.html#method.width

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,714 @@
...
+        }
+    }
+
+    pub fn print_string(&mut self, slice: &str, window: &mut MutexGuard<Window>) {
+        let rect = window.rect;
+        let buffer_width = rect.width / CHARACTER_WIDTH;
+        let buffer_height = rect.height / CHARACTER_HEIGHT;
+        let (x, y) = (rect.x, rect.y);
+
+        let some_slice = slice.as_bytes();
```
kevinaboos: Is there a particular reason why you're converting the string to a byte slice here? I would think it's better to keep it as a string so you can more easily handle wide or multi-byte characters (e.g., emoji, other unicode chars). Reducing it to a byte slice removes important information.

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,714 @@
...
+    }
+}
+
+pub struct FrameBuffer {
+    width: usize,
+    height: usize,
+    buffer: BorrowedSliceMappedPages<u32, Mutable>,
+}
+impl FrameBuffer {
+    fn init_front_buffer() -> Result<FrameBuffer, &'static str> {
```
kevinaboos: I think it's a good idea to use two separate types for this:
1. A type to represent the real physical framebuffer (or framebuffers, in the case of double buffering) that is mapped to actual graphics memory
2. A type to represent virtual framebuffers that are just arbitrary chunks of memory

The idea is that, by design, we can prevent accidentally reading from a real framebuffer by not implementing functions that do that, like `get_pixel()`. Only the virtual framebuffer type would allow reading from it, while the real/physical framebuffer type would only allow writing to it.

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,714 @@
...
+
+    pub fn get_pixel(&self, x: isize, y: isize) -> u32 {
+        self.buffer[(self.width * y as usize) + x as usize]
+    }
+
+    pub fn draw_rectangle(&mut self, rect: &Rect) {
+        for y in rect.start_y()..rect.end_y() {
+            for x in rect.start_x()..rect.end_x() {
+                if x > 0 && x < self.width as isize && y > 0 && y < self.height as isize {
+                    self.draw_something(x, y, 0xF123999);
```
kevinaboos: one of the main goals of an API for drawing things should be to allow the callers to easily avoid the expensive bounds checks and calculations of pixel offsets on a per-pixel basis. Here, you're forcing that to occur not just once, but _twice_ per pixel, which is a lot of overhead. 

Instead of using indexing (which requires bound checking), use iterators.

kevinaboos (changes requested):
Thanks! I left some comments; the biggest issue is the way that offset calculations and bounds checking happens on a per-pixel basis, sometimes more than once per pixel.

kernel/multicore_bringup/src/lib.rs
ouz-a: fixed that

applications/porthole/src/lib.rs
ouz-a: That's how old api worked to support basic text displaying, supporting multi-byte characters is future work.

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,952 @@
...
+            DEFAULT_TEXT_COLOR,
+            DEFAULT_WINDOW_COLOR,
+            0,
+            1,
+        )
+    }
+}
+
+pub fn display_window_title(window: &mut MutexGuard<Window>, fg_color: Color, bg_color: Color) {
+    if let Some(title) = window.title.clone() {
```
kevinaboos: this eagerly clones the whole `title` String every time. Instead move the `.clone()` operation into the inside of the branch

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,952 @@
...
+        let slice = title.as_str();
+        let mut border = window.return_title_border();
+        let middle = (border.width - (slice.len() * CHARACTER_WIDTH)) / 2;
+        border.x = middle as isize;
+        print_string(window, &border, slice, fg_color, bg_color, 0, 0);
+    }
+}
+
+pub fn print_string(
+    window: &mut MutexGuard<Window>,
```
kevinaboos: I assume here you mean `&mut Window`, since there's no real reason to take a mutable reference to a `MutexGuard`. The only reason you'd pass a guard is if you needed the lock to be released within this function, and if that was the case, you'd need to pass it by ownership

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,952 @@
...
+            let gap = (self.y + self.height as isize) - screen_height;
+            end_y = self.height as isize - gap;
+        }
+        let f = Rect::new(end_x as usize, end_y as usize, start_x, start_y);
+        f
+    }
+}
+
+/// FrameBuffer with no actual physical memory mapped,
+/// used for Window and WindowManager's backbuffer.
```
kevinaboos: Note that in the future when we support true double buffering using graphics memory, this will not be true (and thus this type shouldn't be used to represent that backbuffer in graphics memory).

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,952 @@
...
+        })
+    }
+
+    fn copy_window_from_iterators(&mut self, window: &mut MutexGuard<Window>) {
+        let window_screen = window.on_screen_window(self.width as isize, self.height as isize);
+
+        let w_it = window.return_framebuffer_iterator();
+        let f = buffer_indexer(&mut self.buffer, self.width, window_screen);
+
+        for (w_color, pixel) in w_it.zip(f) {
```
kevinaboos: this is the right idea and, ultimately, the end point for what i was hoping to see once you have obtained an iterator over the window region (source) and an iterator over the framebuffer region (destination).

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,952 @@
...
+        }
+    }
+
+    // I'm not exactly sure if using `unsafe` is right bet here
+    // but since we are dealing with arrays/slices most of the time
+    // we need to only prove they are within bounds once and this let's us safely call `unsafe`
+    fn draw_unchecked(&mut self, x: isize, y: isize, col: Color) {
+        let x = x;
+        let y = y;
+        unsafe {
```
kevinaboos: definitely no need to use unsafe here or anywhere else in this entire crate.

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,952 @@
...
+    }
+
+    pub fn return_dynamic_border_pos(&self) -> Rect {
+        let mut rect = self.rect;
+        rect.height = TITLE_BAR_HEIGHT;
+        rect
+    }
+
+    // We don't want user to draw on top a border
+    pub fn return_drawable_area(&self) -> Rect {
```
kevinaboos: consider caching the values of rectangular bounding boxes for things like the title rectangle, drawable areas, etc, since they only need to change rarely upon a window resize action.
Then, when the window actually gets resized, you can recalculate them. A great design pattern for this is to use an `Option<Rect>` with an accessor function like `get_or_insert()`

kevinaboos (changes requested):
leaving a partial review to share notes

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,1042 @@
...
+        Ok(VirtualFrameBuffer {
+            width,
+            height,
+            buffer: mapped_buffer
+                .into_borrowed_slice_mut(0, width * height)
+                .map_err(|(_mp, s)| s)?,
+        })
+    }
+
+    fn copy_windows_into_main_vbuffer(&mut self, window: &mut MutexGuard<Window>) {
```
m-hugo: this should really be a Window method
```suggestion
    fn copy_into_vbuffer(&mut self, vbuf: &mut VirtualFrameBuffer){
```

m-hugo (commented):
and you have many methods and fields that should be pub but aren't
or you could remove every pub if you don't care about modularity

applications/porthole/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+            relative_visible_rect,
+            window_stride,
+        )
+        .unwrap();
+        for (screen_row, window_row) in screen_rows.zip(window_rows) {
+            screen_row.copy_from_slice(window_row);
+        }
+    }
+
+    fn draw_unchecked(&mut self, x: isize, y: isize, col: Color) {
```
kevinaboos: remove unsafety

kevinaboos:
Closing in favor of #823

PR #723, opened 2022-12-12
Change `extra_files` delimiter

tsoutsman:
Changed the `extra_files` delimiter from `?` to `!` so that the bootloader names are compatible with a FAT filesystem for the upcoming UEFI port.


kevinaboos (approved):
Thanks!

PR #727, opened 2022-12-13
Timer, GIC, IRQ support for aarch64

NathanRoyer:
This PR will help us discuss the support of these features for aarch64:

- ARMv8 Generic Timer
- ARMv8 Generic Interrupt Controller (GIC)
- the use of these devices in the context of interrupt management & handling via the `interrupts` crate (and maybe more crates)


aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+
+/// 8-bit unsigned integer
+pub type IntNumber = u8;
+
+/// 8-bit unsigned integer
+pub type Priority = u8;
+
+bitflags! {
+    /// Which CPU to route an interrupt to
+    pub struct TargetCpu: u8 {
```
kevinaboos: this is odd, it only supports 8 CPUs?

kevinaboos: if it matters, the relevant type on x86 is `LapicIpiDestination`

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+const GICC_CTLR: usize = 0x00 / U32BYTES;
+const GICC_PMR:  usize = 0x04 / U32BYTES;
+const GICC_IAR:  usize = 0x0C / U32BYTES;
+const GICC_RPR:  usize = 0x14 / U32BYTES;
+const GICC_EOIR: usize = 0x10 / U32BYTES;
+
+const GICD_CTLR:      usize = 0x000 / U32BYTES;
+const GICD_ISENABLER: usize = 0x100 / U32BYTES;
+const GICD_ICENABLER: usize = 0x180 / U32BYTES;
+const GICD_ITARGETSR: usize = 0x800 / U32BYTES;
```
kevinaboos: This is an unusual way to define the format of the struct. Can you use the more typical design that already exists in Theseus, like this: https://github.com/theseus-os/Theseus/blob/b29196d385788cd8bd454597c8e020165019f802/kernel/e1000/src/regs.rs#L18

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        TargetCpu::from_bits_truncate(flags as u8)
+    }
+
+    /// Sets which CPU cores to forward that
+    /// interrupt to, when it's received
+    pub fn set_int_target(&mut self, int: IntNumber, target: TargetCpu) {
+        self.write_gicd_array::<4>(GICD_ITARGETSR, int, target.bits as u32);
+    }
+
+    /// Performs priority drop for the specified interrupt
```
kevinaboos: this doc comment doesn't make sense to me, can you clarify it? In particular, what is "priority drop"?

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub fn end_of_interrupt(&mut self, int: IntNumber) {
+        self.write_gicc(GICC_EOIR, int as u32);
+    }
+
+    /// Acknowledge the currently serviced interrupt
```
kevinaboos: how does this differ from `end_of_interrupt()`?

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        let mut reg = self.read_gicd(offset);
+        reg &= !(mask << shift);
+        reg |= (value & mask) << shift;
+        self.write_gicd(offset, reg);
+    }
+
+    /// Interrupts have a priority; if their priority
+    /// is lower or equal to this one, they're discarded
+    pub fn get_minimum_int_priority(&self) -> Priority {
+        255 - (self.read_gicc(GICC_PMR) as u8)
```
kevinaboos: if the value `255` is based on `u8::MAX`, then it'd be better to use `u8::MAX` directly. Alternatively, you could define a const with a more descriptive name.

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// is lower or equal to this one, they're discarded
+    pub fn set_minimum_int_priority(&mut self, priority: Priority) {
+        self.write_gicc(GICC_PMR, (255 - priority) as u32)
+    }
+
+    /// Is the distributor enabled or disabled?
+    ///
+    /// When it's disabled, interrupts are not forwarded.
+    pub fn get_gicd_state(&self) -> Enabled {
+        (self.read_gicd(GICD_CTLR) & 1) > 0
```
kevinaboos: use a const or bitflags or something to define what the bits of `GICD_CTLR` mean.

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    /// Enables or disables interrupt forwarding in the distributor
+    pub fn set_gicd_state(&mut self, enabled: Enabled) {
+        let mut reg = self.read_gicc(GICD_CTLR);
+        reg &= !1;
```
kevinaboos: use a const or bitflags or something to define what the bits of `GICD_CTLR` mean.

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        reg |= enabled as u32;
+        self.write_gicd(GICD_CTLR, reg);
+    }
+
+    /// Is the cpu interface enabled or disabled?
+    ///
+    /// When it's disabled, interrupts are not forwarded.
```
kevinaboos: there are a few places where the term "forwarding" or "forwarded" interrupts is used. Can you elaborate what that means? I presume that forwarding means that an interrupt is or is not delivered from the GIC to the actual CPU, but it'd be nice to clarify it since it's not a standard term (it could also be interpreted as forwarding an interrupt from one CPU to another)

aarch64/kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,411 @@
...
+// meant to handle IRQs. Synchronous
+// Exceptions (syscalls) are not IRQs on
+// aarch64; this crate doesn't expose any
+// way to handle them at the moment.
+static IRQ_HANDLERS: RwLockIrqSafe<[HandlerFunc; 256]> = RwLockIrqSafe::new([default_irq_handler; 256]);
+
+/// Wrapper structs for memory copies of registers.
+#[repr(transparent)]
+struct SpsrEL1(InMemoryRegister<u64, SPSR_EL1::Register>);
+struct EsrEL1(InMemoryRegister<u64, ESR_EL1::Register>);
```
kevinaboos: did you intend to apply `transparent` to this struct too?

aarch64/kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,411 @@
...
+
+/// The exception context as it is stored on the stack on exception entry.
+#[repr(C)]
+pub struct ExceptionContext {
```
kevinaboos: if the layout of this struct is accessed from your asm code, you should make a very clear note/comment warning a developer that it must be kept in sync with that code.

aarch64/kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,411 @@
...
+
+    /// Exception syndrome register.
+    esr_el1: EsrEL1,
+}
+
+type HandlerFunc = extern "C" fn(&ExceptionContext) -> bool;
+
+// called for all exceptions other than interrupts
+fn default_exception_handler(exc: &ExceptionContext, origin: &'static str) {
+    log::error!("Kernel Panic: Unhandled Exception ({})\r\n{}", origin, exc);
```
kevinaboos: nit: unless you intend to invoke a `panic!()` here, remove the "Kernel Panic:" language (and everywhere else it's used).

aarch64/kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,411 @@
...
+/// Please call (once) this before using this crate.
+///
+/// This initializes the Generic Interrupt Controller
+/// using the addresses which are valid on qemu's "virt" VM.
+pub fn init(page_table: &mut PageTable) -> Result<(), &'static str> {
+    let mut gic = GIC.lock();
+
+    let inner = ArmGic::map(page_table, qemu_virt_addrs::GICD, qemu_virt_addrs::GICC)?;
+    let mut result = Err("The GIC has already been initialized!");
+    gic.call_once(|| { result = Ok(()); inner });
```
kevinaboos: this is a strange way to use `call_once()`. I believe there is a `try_call_once()` that supports results; that's probably what you want to do here.

aarch64/kernel/interrupts/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+//! Basic interrupt handling structures and simple handler routines.
+
+#![no_std]
+
+#[cfg(target_arch = "x86_64")]
+#[path = "x86_64/mod.rs"]
+mod arch;
```
kevinaboos: no need to duplicate the x86 code in this repo folder, we'll combine it later.

aarch64/kernel/interrupts/src/x86_64/mod.rs
```diff
@@ -0,0 +1,390 @@
+#![feature(abi_x86_interrupt)]
```
kevinaboos: if this is unchanged from the version in the main repo, you can delete this file

aarch64/kernel/interrupts/src/lib.rs
kevinaboos: alternatively, if you want to keep it in, set it's `path` to the file that already exists in the top-level repo.

aarch64/kernel/interrupts/src/x86_64/rtc.rs
```diff
@@ -0,0 +1,166 @@
+use port_io::Port;
```
kevinaboos: same comment as `x86_64/mod.rs`

aarch64/kernel/nano_core/multiprocessor.rs
```diff
@@ -0,0 +1,26 @@
...
+///
+/// - Affinity level 0 in bits [0:7]
+/// - Affinity level 1 in bits [8:15]
+/// - Affinity level 2 in bits [16:23]
+/// - Affinity level 3 in bits [24:31]
+pub fn get_core_num() -> u32 {
+    let reg: u64 = MPIDR_EL1.get();
+
+    let aff_0_1_2 =    reg & 0x00_ff_ff_ff;
+    let aff_3 = (reg >> 8) & 0xff_00_00_00;
```
kevinaboos: either i'm misunderstanding the doc comments, or there's a small mistake here -- these operations don't seem to match your comments. 
Specifically, why are you right-shifting the register by 8 bits to get `aff_3`?

aarch64/kernel/gic/src/redist.rs
```diff
@@ -0,0 +1,74 @@
...
+    // Wake the redistributor
+    reg &= !RD_WAKER_PROCESSOR_SLEEP;
+    registers[offset::RD_WAKER] = reg;
+
+    // then poll ChildrenAsleep until it's cleared
+
+    /* TODO
+    // this could lock the entire kernel...
+    // but it's what we're supposed to do
+    // maybe we should put a loop run limit?
```
kevinaboos: glad you're thinking about these kind of things, I would agree that you can place a timeout on the loop, though I'm not sure how long something like this might take. We could do something very long like 2-3 seconds, and then panic or return an error if it fails.

aarch64/kernel/gic/src/redist.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    // Another problem with this: it loops forever
+    // I don't understand why yet, could this be the result
+    // of non-volatile access to mmio?
```
kevinaboos: definitely could be. Make sure this and all other MMIO accesses are volatile. You can see how I use our fork of the `volatile` crate in various other crates to accomplish this.

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// any PE that is not busy with another,
+    /// more important task
+    AnyCpuAvailable,
+    GICv2TargetList(TargetList),
+}
+
+// = 4
+const U32BYTES: usize = core::mem::size_of::<u32>();
+// = 32
+const U32BITS: usize = U32BYTES * 8;
```
kevinaboos: there's a built-in value for this: https://doc.rust-lang.org/std/primitive.u32.html#associatedconstant.BITS

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+    pub const GICR: PhysicalAddress = PhysicalAddress::new_canonical(0x080A0000);
+}
+
+/// Boolean
+pub type Enabled = bool;
+
+/// 24-bit unsigned integer
+///
+/// An u32 is used because there is no u24.
+pub type IntNumber = u32;
```
kevinaboos: i'm a bit confused by `IntNumber`. It's 24 bits, but then some functions say it cannot be greater than 16?

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+use bitflags::bitflags;
+
+/// Physical addresses of the CPU and Distributor
+/// interfaces as exposed by the qemu "virt" VM.
+pub mod qemu_virt_addrs {
+    use super::*;
+
+    pub const GICD: PhysicalAddress = PhysicalAddress::new_canonical(0x08000000);
+    pub const GICC: PhysicalAddress = PhysicalAddress::new_canonical(0x08010000);
+    pub const GICR: PhysicalAddress = PhysicalAddress::new_canonical(0x080A0000);
```
kevinaboos: is there a way to discover these addresses dynamically? I assume they won't always be fixed. If they are, you should describe where the values come from (perhaps something like a board config file or device tree file?)

aarch64/kernel/gic/src/cpu3.rs
```diff
@@ -0,0 +1,97 @@
...
+            // aff3 in bits [48:55]
+            let aff3 = (cpu & 0xff000000) << 24;
+            // aff2 in bits [32:39]
+            let aff2 = cpu & 0xff0000 << 16;
+            // aff1 in bits [16:23]
+            let aff1 = cpu & 0xff00 << 8;
+            // aff0 as a GICv2-style target list
+            let aff0 = cpu & 0xff;
+            let target_list = if aff0 >= 16 {
+                log::error!("[GIC driver] cannot send an IPI to a core with Aff0 >= 16");
```
kevinaboos: this should be a hard error or panic

aarch64/kernel/gic/src/cpu3.rs
```diff
@@ -0,0 +1,97 @@
...
+        TargetCpu::Specific(cpu) => {
+            let cpu = cpu as usize;
+            // aff3 in bits [48:55]
+            let aff3 = (cpu & 0xff000000) << 24;
+            // aff2 in bits [32:39]
+            let aff2 = cpu & 0xff0000 << 16;
+            // aff1 in bits [16:23]
+            let aff1 = cpu & 0xff00 << 8;
+            // aff0 as a GICv2-style target list
+            let aff0 = cpu & 0xff;
```
kevinaboos: the comments don't match what the code is doing here. For ex, you say `aff3` is in bits [48:55] but then you only left-shift it by 24 bits.

aarch64/kernel/gic/src/dist.rs
```diff
@@ -0,0 +1,204 @@
...
+// bit 15: which interrupt group to target
+const SGIR_NSATT_GRP0: u32 = 0 << 15;
+
+fn assert_cpu_bounds(target: &TargetCpu) {
+    if let TargetCpu::Specific(cpu) = target {
+        assert!(*cpu < 8, "affinity routing is disabled; cannot target a CPU with id >= 8");
+    }
+}
+
+/// Initializes the distributor
```
kevinaboos: more detail would be nice, e.g., "Initializes the distributor by doing x, y, z"

aarch64/kernel/gic/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// = 4
+const U32BYTES: usize = core::mem::size_of::<u32>();
+// = 32
+const U32BITS: usize = U32BYTES * 8;
+
+pub type MmioPageOfU32 = [u32; 0x400];
+
+type BorrowedPage = BorrowedMappedPages<MmioPageOfU32, Mutable>;
```
kevinaboos: not really a big fan of these typedefs, i think they make it harder to read the code, if anything.

kevinaboos (changes requested):
Avoid using "magic" numbers as much as possible; instead, you should define `const`s for things like affinity bitmaps and bitshifts.

Kindly add higher-level comments to each module and each function so we know what the module is for and what the function generally does. It's a bit hard to understand things with all of the acronyms; ideally the code (with comments) should be self-explanatory, such that the reader wouldn't have to fully know all details of the aarch64 interrupt spec to understand what the code is doing. 
For example, someone reading the file `gic/src/dist.rs` or `cpu2` or `cpu3` probably wouldn't know what it's for unless they were already intimately familiar with aarch64 spec terminology.

aarch64/kernel/gic/src/lib.rs
NathanRoyer: > this is odd, it only supports 8 CPUs?

GICv2, yes

aarch64/kernel/nano_core/multiprocessor.rs
NathanRoyer: image
aff3 isn't next to aff2 in MPIDR; I'm putting it next to aff2 so that the whole thing fits in a u32

aarch64/kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: yep, both structs should have it

aarch64/kernel/gic/src/lib.rs
NathanRoyer: Some functions are meant for SGIs (IPIs) and Private Peripheral Interrupts (i.e. timer = per-core); if you're using functions specific to those, it's invalid to specify an interrupt number above 15 or 31 (depending on the function)

For SPIs (shared peripheral interrupts like PCIe ints), the interrupt number must lie between 32 & 1020.

There are also extensions if this is not enough, in which case the theoretical maximum interrupt number is u24::MAX (u24 type doesn't exist)

aarch64/kernel/gic/src/lib.rs
NathanRoyer: they should be read from the device tree blob, which we don't do yet.
I'll add a comment indicating where they're from

aarch64/kernel/gic/src/cpu3.rs
NathanRoyer: I'm moving Aff3 from `cpu`[24:32] to `aff3`[48:55], by shifting 24b to the left; I'll clarify the comments

aarch64/kernel/gic/src/lib.rs
NathanRoyer: I did it that way because it's defined like that in the spec
I'll do it your way

kevinaboos: Update: we will leave this for a future PR.

aarch64/kernel/gic/src/cpu3.rs
kevinaboos: Got it, thanks for clarifying. Comments are good, even better would be to use dedicated types, e.g., one for the affinity value that has aff3 in [24:32), and a different type that has it in [48:56), and then expressly converting between the types. More types = more Rusty power!

aarch64/kernel/nano_core/multiprocessor.rs
kevinaboos: got it -- as with my other comment about bit-shifting aff3, it'd be best to use dedicated types for each variant here, and then use a typical `From` conversion to make this both explicit and canonically Rusty.

aarch64/kernel/gic/src/lib.rs
kevinaboos: Sounds like an enum with a variant for each of these possibilities would be the best way to represent this, e.g., (pseudocode):
```rust
enum IntNumbers {
    PrivatePeripheralInterrupt { <the appropriate value> },
    #[doc(alias("ipi"))]
    SGI(<some value up to 15 or 31>),
    SharedPeripheralInterrupt( <data from 32-1020 > ),
    Extended ( ... ),
}
``` 
lmk if there's some additional detail or hardware-based reason why we can't or shouldn't do this

aarch64/kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: Actually I'm not sure how to ensure it's the first time without doing this ugly thing.

Here's a [playground snippet](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=293a35d21a96a42b2baf1e676ed3943b) showing that try_call_once doesn't fail if the once has already been initialized.

Should we silently discard when this is run twice? (we'd just need to make sure the allocations are done in the body of the closure)

aarch64/kernel/gic/src/lib.rs
NathanRoyer: Ok, I replaced the comment with:
```rust
/// Zeros the current priority level of this CPU,
/// Meaning that the CPU is ready to process interrupts
/// again.
```

aarch64/kernel/gic/src/lib.rs
NathanRoyer: New comment:
```rust
/// Acknowledge the currently serviced interrupt
/// and fetches its number; this tells the GIC that
/// the requested interrupt is being handled by
/// this CPU.
```

aarch64/kernel/gic/src/lib.rs
NathanRoyer: image

As you can see, a generic interrupt controller is composed of multiple units which are configured to either "forward" interrupts or forget about them. The path is:

1. distributor, which receives interrupt requests from peripherals and CPUs
2. redistributor (absent on GICv2)
3. CPU interface

aarch64/kernel/interrupts/src/lib.rs
NathanRoyer: Note: both workspaces have been merged now (#817)

aarch64/kernel/nano_core/multiprocessor.rs
NathanRoyer: I'll remove this from this PR and wait for one which is actually related to multiprocessor features to re-introduce it.

aarch64/kernel/gic/src/redist.rs
NathanRoyer: I no longer find this to be blocking in practice, now that a volatile access is made. Should I still add a timeout?

aarch64/kernel/gic/src/redist.rs
NathanRoyer: I found that this is the only place where rustc makes assumptions about this memory mapping, so I didn't bother to ensure volatile accesses in the whole crate

aarch64/kernel/gic/src/lib.rs
NathanRoyer: It looks difficult to me to make sure that the inner numbers match their category, and superfluous since we can detect the category based on this number; I'm also already ensuring appropriate IntIds are passed to the various methods of the ArmGic object

aarch64/kernel/gic/src/cpu3.rs
NathanRoyer: It's the only place where this conversion will ever happen... what if I improve the doc?
```rust
            let cpu = cpu as usize;

            // level 3 affinity is expected in cpu[24:31]
            // we want it in bits [48:55]
            let aff3 = (cpu & 0xff000000) << 24;

            // level 2 affinity is expected in cpu[16:23]
            // we want it in bits [32:39]
            let aff2 = cpu & 0xff0000 << 16;

            // level 1 affinity is expected in cpu[8:15]
            // we want it in bits [16:23]
            let aff1 = cpu & 0xff00 << 8;

            // level 0 affinity is expected in cpu[0:7]
            // we want it as a GICv2-style target list
            let aff0 = cpu & 0xff;
            let target_list = if aff0 >= 16 {
                panic!("[GIC driver] cannot send an IPI to a core with Aff0 >= 16");
            } else {
                1 << aff0
            };
            aff3 | aff2 | aff1 | target_list
```

kernel/gic/lib.rs
```diff
@@ -0,0 +1,9 @@
+//! Allows configuring the Generic Interrupt Controller
+#![no_std]
+#![feature(doc_cfg)]
+
+#[cfg(any(target_arch = "aarch64", doc))]
+mod src;
+
+#[cfg(any(target_arch = "aarch64", doc))]
+pub use src::*;
```
kevinaboos: this is an odd non-standard setup. Let's follow standard practice: put this file in `gic/src/lib.rs` and remove the `path` part in `gic/Cargo.toml`.

Also, `src` is a vague name for a module, as is `cpu2` and `cpu3`. Give them accurate names that reflect what they actually implement.

kernel/gic/src/cpu2.rs
```diff
@@ -0,0 +1,59 @@
+use super::MmioPageOfU32;
```
kevinaboos: add module-level comments describing what this module does at a high level.

kernel/gic/src/cpu2.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// enable group 0
+// const CTLR_ENGRP0: u32 = 0b01;
+
+// enable group 1
+const CTLR_ENGRP1: u32 = 0b10;
+
+/// Enables routing of group 1 interrupts
+/// in the for the current CPU
```
kevinaboos: Please expand on what "group 0" and "group 1" interrupts are.  Are there any other groups that exist? do we only care about particular groups?

kernel/gic/src/cpu2.rs
```diff
@@ -0,0 +1,59 @@
...
+pub fn init(registers: &mut MmioPageOfU32) {
+    let mut reg = registers[offset::CTLR];
+    reg |= CTLR_ENGRP1;
+    registers[offset::CTLR] = reg;
```
kevinaboos: there are a lot of register operations like this that read and write to it as if it's a normal value in memory. Is there some hidden mechanism that forces these register accesses to be `volatile`? If not, we must do that to ensure the compiler doesn't optimize out such read/write operations.

kernel/gic/src/cpu2.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/// Interrupts have a priority; if their priority
+/// is lower or equal to this one, they're discarded
+pub fn get_minimum_int_priority(registers: &MmioPageOfU32) -> Priority {
```
kevinaboos: all of these public functions have an odd interface in that they expect the caller to have obtained mutable/immutable access to the underlying mapped pages that represent the GIC registers. That's not really an ideal interface since it (1) leaks the inner details of the abstraction, and (2) forces us to expose access (especially mutable access!) to the raw memory contents, which allows the caller to do anything they want.

You should refactor this to look more like the `LocalApic` type, in which a function like `get_min_int_priority()` wouldn't take any arguments (except an implicit &self)` and functions like `end_of_interrupt()` would only take `IntNumber`.

kernel/gic/src/cpu2.rs
```diff
@@ -0,0 +1,59 @@
...
+    u8::MAX - (registers[offset::PMR] as u8)
+}
+
+/// Interrupts have a priority; if their priority
+/// is lower or equal to this one, they're discarded
```
kevinaboos: "Discarded" is used a few times here. Let's clarify this -- does that mean the interrupt is masked but will be redelivered when it is unmasked, or does it mean that the interrupt simply never happens (I interpret it as the latter)?

kernel/gic/src/cpu2.rs
```diff
@@ -0,0 +1,59 @@
...
+pub fn set_minimum_int_priority(registers: &mut MmioPageOfU32, priority: Priority) {
+    registers[offset::PMR] = (u8::MAX - priority) as u32;
+}
+
+/// Zeros the current priority level of this CPU,
+/// Meaning that the CPU is ready to process interrupts
+/// again.
```
kevinaboos: This doc comment makes it sound like this is strictly about setting priorities, and therefore can be used outside of an interrupt handler context. Is this true? If so, we should probably rename or clarify this.

If not, add some doc comments describing the typical use case of this, e.g., on x86 an EOI must be performed at the end of each interrupt handler in order to ensure another interrupt of that same type can occur in the future.

kernel/gic/src/cpu3.rs
```diff
@@ -0,0 +1,110 @@
+use core::arch::asm;
```
kevinaboos: add module-level doc comments

kernel/gic/src/cpu3.rs
```diff
@@ -0,0 +1,110 @@
...
+pub fn end_of_interrupt(int: IntNumber) {
+    let reg_value = int as usize;
+    unsafe { asm!("msr ICC_EOIR1_EL1, {}", in(reg) reg_value) };
+}
+
+/// Acknowledge the currently serviced interrupt
+/// and fetches its number; this tells the GIC that
+/// the requested interrupt is being handled by
+/// this CPU.
+pub fn acknowledge_int() -> (IntNumber, Priority) {
```
kevinaboos: I can't tell when I should use `end_of_interrupt` and when I should use `acknowledge_int` -- what's the difference, and when is it appropriate to use each one?

Also, tiny nit: let's be consistent with "int" and "interrupt"

kernel/gic/src/dist.rs
```diff
@@ -0,0 +1,205 @@
...
+    reg |= CTLR_ENGRP1;
+    registers[offset::CTLR] = reg;
+
+    // Return value: whether or not affinity routing is
+    // currently enabled for both secure and non-secure
+    // states.
+    reg & CTLR_ARE_NS > 0
+}
+
+/// Will that interrupt be forwarded by the distributor?
```
kevinaboos: nit: you should phrase doc comments in an active, positive-assertive manner, i.e.,

"Returns whether the given `interrupt` will be forwarded by the distributor"

Also, what does forwarded mean, and what is it forwarding the interrupt to?

kernel/gic/src/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+            let mapped = page_table.map_allocated_pages_to(pages, frames, mmio_flags)?;
+            mapped.into_borrowed_mut(0).map_err(|(_, e)| e)
+        };
+
+        match version {
+            Version::InitV2 { dist, cpu } => {
+                let mut distributor = map_dist(dist)?;
+
+                let mut processor: BorrowedPage = {
+                    let pages = allocate_pages(1).ok_or("couldn't allocate pages for GICC interface")?;
```
kevinaboos: is "GICC" a typo for "GIC" or does it mean something different?

aarch64/kernel/gic/src/lib.rs
kevinaboos: Ok, I left another comment about this but I'll respond here instead -- this information needs to be reflected in the documentation such that someone reading the code doesn't need to refer back to the aarch64 manual in order to understand it. The goal is that the code and docs are self-sufficient.

aarch64/kernel/gic/src/redist.rs
kevinaboos: a timeout would still be good, but you should document what you choose to do and why. For ex, if you try it 50 times or for 10 milliseconds or for 1 million iterations, then explain why.

aarch64/kernel/gic/src/redist.rs
kevinaboos: As i said in another comment, we absolutely have to use volatile instructions here. It's required to ensure that the compiler doesn't omit instructions that read from or write to MMIO registers.

aarch64/kernel/gic/src/cpu3.rs
kevinaboos: yep, that's much better. Types would still be better as I suggested in my Jan 12 comment, but docs are a good intermediary improvement.

kevinaboos (changes requested):
A couple high-level comments:
* When writing docs and code, you sometimes limit the number of columns/characters to something very small (maybe 50-60). You can use wider sentences with fewer line breaks, up to 100 characters per line.
* This PR does a lot of different things, which I understand was necessary for you to test it. Now that it's getting ready to merge, can we break it up into a few different PRs? The first can be one that simply changes the existing `interrupts` crate into one with an x86_64-specific module, which will make it easier to track changes since it preserves the change history of that source file, whereas this PR doesn't. Then, after that, we can merge in all the aarch64- related changes.
* The `gic` crate is aarch64-specific, and although I agree it should be it's own crate, I'm not sure how that'll work with the current build system. If I build this PR for x86 as normal, what happens? I assume that crate would fail to build. We can explore the package-based solution for including/excluding crates in the workspace too, if need be.

kernel/gic/lib.rs
NathanRoyer: > If I build this PR for x86 as normal, what happens?

Thanks to this setup, this crate appears as empty during an x86_64 build. As you can see, it keeps the standard src folder for the real code, and the top-level lib.rs is just an arch gate.

kernel/gic/src/cpu2.rs
NathanRoyer: Group 1 is available at both EL0 & EL1 whereas group 0 is only available at EL1. Linux always uses group 1 so I thought I'd do the same. Apart from the availability, they're the same.

kernel/gic/src/cpu2.rs
NathanRoyer: The modules are privately imported:
```rust
// mod.rs, line 71
mod cpu3;
mod cpu2;
mod dist;
mod redist;
```

So these functions aren't available to crate users

kernel/gic/src/cpu3.rs
NathanRoyer: I agree with the consistency nit.

When an interrupt reaches a CPU, the CPU first have to ackowledge the interrupt to the controller, then it processes the interrupt, and finally when the interrupt has been processed it sends an "end of interrupt" signal to the controller, lowering the current priority, meaning the controller can send other interrupts again.

kernel/gic/src/mod.rs
NathanRoyer: I forgot to rename it to CPU interface (GICC is how the manual references that)

aarch64/kernel/gic/src/lib.rs
NathanRoyer: I added this to the crate's doc:
```rust
//! The term "Forwarding" is sometimes used in this crate.
//! This is because the Distributor, Redistributor and CPU interface are
//! chained in the controller. The distributor and the redistributor are
//! configured by the code of this crate to either allow (forward)
//! interrupts or disallow (discard) them.
```

kernel/gic/src/cpu2.rs
NathanRoyer: Done for all interfaces

kernel/gic/src/cpu2.rs
NathanRoyer: You're right and I'm wrong for having documented it this way; I'll fix

NathanRoyer: New doc:
```rust
    /// Interrupts have a priority; if their priority
    /// is lower or equal to this one, they're queued
    /// until this CPU or another one is ready to handle
    /// them
```

kernel/gic/src/cpu2.rs
NathanRoyer: > This doc comment makes it sound like this is strictly about setting priorities, and therefore can be used outside of an interrupt handler context.

I don't understand :thinking: What would you use it for?

kernel/gic/src/dist.rs
NathanRoyer: Used your doc + documented the use of "Forwarding" as so:
```rust
//! The term "Forwarding" is sometimes used in this crate.
//! This is because the Distributor, Redistributor and CPU interface are
//! chained in the controller. The distributor and the redistributor are
//! configured by the code of this crate to either allow (forward)
//! interrupts or disallow (discard) them.
```
As this is distributor code, ints are forwarded to the redistributor.

kernel/gic/src/cpu2.rs
kevinaboos: to clarify, i meant that typically you send an EOI at the end of an interrupt handler before it returns. That's what one would assume by reading the _name_ of this function, but when i read the docs it makes it sound like it does something else entirely related to setting interrupt priorities, which implies you can call it in any context.

aarch64/kernel/gic/src/lib.rs
kevinaboos: great!
thanks

kernel/gic/lib.rs
kevinaboos: that's good, but you can still achieve that with the conventional source code organization that i mentioned above, right?

kernel/gic/src/cpu2.rs
kevinaboos: That's good. It's okay for now -- I think we already discussed this previously, but we should move to a design more similar to LocalAPIC for a variety of reasons.

kernel/gic/src/cpu2.rs
kevinaboos: nice!

kernel/gic/src/cpu3.rs
kevinaboos: ah, that's a very interesting usage paradigm. That should be very clearly stated everywhere such that people know to call this at the beginning of an interrupt handler (right?)

kernel/gic/src/dist.rs
kevinaboos: great!

kernel/gic/lib.rs
NathanRoyer: Ok. how should I name the folder with all the actual source though?
```
kernel/gic
├── Cargo.toml
└── src
    ├── lib.rs
    └── ?
        ├── cpu_interface_gicv2.rs
        ├── cpu_interface_gicv3.rs
        ├── dist_interface.rs
        ├── mod.rs
        └── redist_interface.rs
```

As an alternative, I could put the following in lib.rs, rename `mod.rs` → `aarch64.rs` and place it next to every other module in src
```
kernel/gic
├── Cargo.toml
└── src
    ├── cpu_interface_gicv2.rs
    ├── cpu_interface_gicv3.rs
    ├── dist_interface.rs
    ├── aarch64.rs
    ├── lib.rs
    └── redist_interface.rs
```
```rust
// lib.rs
#![feature(doc_cfg)]

#[cfg(any(target_arch = "aarch64", doc))]
include_str!("aarch64.rs")
```

I still prefer the current structure as it feels more usual to me but that's personal.

NathanRoyer:
> This PR does a lot of different things, which I understand was necessary for you to test it. Now that it's getting ready to merge, can we break it up into a few different PRs? The first can be one that simply changes the existing interrupts crate into one with an x86_64-specific module, which will make it easier to track changes since it preserves the change history of that source file, whereas this PR doesn't. Then, after that, we can merge in all the aarch64- related changes.

I don't understand why github shows `kernel/interrupts/lib.rs` as "changed", [this commit](https://github.com/theseus-os/Theseus/pull/727/commits/0590e80cabc7040259309c66b2a3663deb15fcee#diff-161c680b99f53bbd69472a931ffcf91501934a43e61b88fecf3a846cc1920c70) clearly moved it properly to `kernel/interrupts/x86_64/mod.rs`

Github seems to forget it's been moved and interprets the PR as having changed the entirety of lib.rs

Will mod.rs inherit the history of lib.rs if we merge this, or will the content really teleport and lose the file's history?

kernel/gic/src/cpu2.rs
NathanRoyer: Oh ok, will fix the doc
edit: done

kernel/gic/src/cpu3.rs
NathanRoyer: Don't we have to ack interrupt requests on x86_64? I'm currently doing it in my "gateway" interrupt handler (interrupts::aarch64::current_elx_irq)

It's the only way to get the interrupt number, so it's pretty much required for my gateway function to read it, thus acknowledging the int

aarch64/kernel/gic/src/lib.rs
NathanRoyer: Just changed that [here](https://github.com/theseus-os/Theseus/pull/727/commits/37004eddfd046dfacc430710e7ade992d9325e56); this also ensures reads and writes are volatile

aarch64/kernel/gic/src/redist.rs
NathanRoyer: Is [this](https://github.com/theseus-os/Theseus/pull/727/commits/20a18fce01edade56a303ebfc4d9dc4d8e48d336) good enough?

aarch64/kernel/gic/src/redist.rs
NathanRoyer: Yep; the GicMappedPage ensures volatile accesses now

aarch64/kernel/interrupts/src/aarch64/mod.rs
kevinaboos: Let's use `Option` here instead of `Once` since it's already in a mutex

NathanRoyer: Yep; done

aarch64/kernel/gic/src/redist.rs
kevinaboos: well it's all commented out right now, so i suppose we can accept it as is. I wonder if real hw will require looping here.

kernel/gic/lib.rs
kevinaboos: Either one of the above structures is fine, including the latter. We just want all the source code to be inside `src/`. I suppose my main critique of the latter option is that `aarch64` is a vague name for a module, especially since the entire GIC crate is already aarch64-specific.

We don't need "include_str" here, that's not appropriate for Rust source code. Target-specific `mod` statements are preferred to include/exclude code.

kernel/gic/src/redist_interface.rs
```diff
@@ -0,0 +1,93 @@
+//! Redistributor Interface
+//!
+//! Included functionnality:
```
kevinaboos: typo

kernel/gic/src/redist_interface.rs
```diff
@@ -0,0 +1,93 @@
+//! Redistributor Interface
```
kevinaboos: can you describe what the Redistributor is for? Or if it is described elsewhere, link to that here.

kernel/gic/src/redist_interface.rs
```diff
@@ -0,0 +1,93 @@
...
+        registers.read_volatile(offset::RD_WAKER) & RD_WAKER_CHLIDREN_ASLEEP > 0
+    };
+
+    let mut counter = 0;
+    let mut timed_out = || {
+        counter += 1;
+        counter >= TIMEOUT_ITERATIONS
+    };
+
+    while children_asleep() && !timed_out() { }
```
kevinaboos: so we use this timeout here but not elsewhere? (it was commented out in the other file).

Shouldn't we either use it in all relevant places, or never at all?

kernel/gic/src/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// = 4
+const U32BYTES: usize = core::mem::size_of::<u32>();
+// = 32
+const U32BITS: usize = U32BYTES * 8;
+
+#[repr(C)]
+#[derive(zerocopy::FromBytes)]
+pub struct GicMappedPage {
```
kevinaboos: this is a misnomer, when reading this type name i'd expect it to contain `MappedPages` itself or a derivative of that like `BorrowedMappedPages`.

This is actually more like the GIC's registers.

kernel/gic/src/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+
+const_assert_eq!(core::mem::size_of::<GicMappedPage>(), 0x1000);
+
+const REDIST_SGIPPI_OFFSET: usize = 0x10000;
+const DIST_P6_OFFSET: usize = 0x6000;
+
+mod cpu_interface_gicv3;
+mod cpu_interface_gicv2;
+mod dist_interface;
+mod redist_interface;
```
kevinaboos: nit: move `mod` statements to the top (right after `use ...` statements) so we can more easily figure out the module layout.

kernel/gic/src/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+            Self::V3( _) => cpu_interface_gicv3::send_ipi(int_num, target),
+        }
+    }
+
+    /// Acknowledge the currently serviced interrupt
+    /// and fetches its number
+    /// 
+    /// Note: this constructor accesses the
+    /// interfaces; their addresses have to be
+    /// readable and writable.
```
kevinaboos: outdated comment? this isn't a constructor, and i'm not sure how the caller can determine whether this condition is already met when invoking this method.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+}
+
+/// Please call this (only once) before using this crate.
+///
+/// This initializes the Generic Interrupt Controller
+/// using the addresses which are valid on qemu's "virt" VM.
+pub fn init() -> Result<(), &'static str> {
+    extern "Rust" {
+        // in assembly file
+        static __exception_vector_start: UnsafeCell<()>;
```
kevinaboos: this is an odd way to do FFI with assembly -- is this recommended by some guide somewhere? typically you use raw types that are compatible with C; I don't think `UnsafeCell` qualifies, a `usize` is probably more correct here.

kernel/interrupts/src/aarch64/table.s
```diff
@@ -0,0 +1,128 @@
...
+
+.macro FIQ_SUSPEND
+1:  wfe
+    b   1b
+.endm
+
+.section .text
+
+// Align by 2^11 bytes, as demanded by ARMv8-A. Same as ALIGN(2048) in an ld script.
+.align 11
```
kevinaboos: it's not clear to me whether `align` works like this on ARM assembly; i think it matters which assembler we use.

To be more clear, perhaps you should use `.balign 2048`.

kevinaboos: also it'd be nice to have a link to some docs that explain why such a large alignment is required.

kernel/gic/src/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    // Writes one slot of an array spanning across
+    // multiple u32s.
+    //
+    // - `int` is the index
+    // - `offset` tells the beginning of the array
+    // - `INTS_PER_U32` = how many array slots per u32 in this array
+    // - `value` is the value to write
+    fn write_array_volatile<const INTS_PER_U32: usize>(&mut self, offset: usize, int: IntNumber, value: u32) {
```
kevinaboos: why do these functions take a const generic argument? isn't the value always `32`? In general i find these 2 functions quite confusing.

kernel/gic/src/redist_interface.rs
```diff
@@ -0,0 +1,93 @@
...
+// (if the value works for your SoC, please add it to this list.)
+//
+// If the redistributor's initialization times out, it means either:
+// - that your ARM SoC is not GICv3 compliant (try initializing it as GICv2)
+// - that the timeout value is too low for your ARM SoC. Try increasing it
+// to see if the booting sequence continues.
+//
+// If it wasn't enough for your machine, reach out to the Theseus
+// developers (or directly submit a PR).
+const TIMEOUT_ITERATIONS: usize = 0x10_000;
```
kevinaboos: hex numbers are typically written as 4-digit groups, e.g., `0x1_0000`

kernel/gic/src/redist_interface.rs
```diff
@@ -0,0 +1,93 @@
...
+pub fn get_sgippi_state(registers: &GicMappedPage, int: IntNumber) -> Enabled {
+    registers.read_array_volatile::<32>(offset::SGI_ISENABLER, int) > 0
+    &&
+    // part of group 1?
+    registers.read_array_volatile::<32>(offset::IGROUPR, int) == GROUP_1
+}
+
+/// Enables or disables the forwarding of
+/// a particular SGI or PPI
+pub fn set_sgippi_state(registers: &mut GicMappedPage, int: IntNumber, enabled: Enabled) {
```
kevinaboos: `set_sgippi_state` might make sense for someone who has read the manual or is an armv8 expert, but for the rest of us, why not use a clearer name that describes what's actually happening? for ex, "enable_forwarding_...`

kevinaboos: And for the `get_sgippi_state` function, something like "is_interrupt_forwarded...". Whatever you think is most descriptive

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+use tock_registers::interfaces::Readable;
+use tock_registers::registers::InMemoryRegister;
+
+use gic::{qemu_virt_addrs, ArmGic, IntNumber, Version as GicVersion};
+use irq_safety::{RwLockIrqSafe, MutexIrqSafe};
+use memory::get_kernel_mmi_ref;
+use log::{info, error};
+
+// Raw/Early exception handling is defined
+// in this assembly file.
```
kevinaboos: isn't this file just assembly-level exception handler trampolines / entry points? it's not actually early handling, right?

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+// meant to handle IRQs. Synchronous
+// Exceptions (syscalls) are not IRQs on
+// aarch64; this crate doesn't expose any
+// way to handle them at the moment.
+static IRQ_HANDLERS: RwLockIrqSafe<[HandlerFunc; 256]> = RwLockIrqSafe::new([default_irq_handler; 256]);
+
+/// Wrapper structs for memory copies of registers.
+#[repr(transparent)]
+struct SpsrEL1(InMemoryRegister<u64, SPSR_EL1::Register>);
+struct EsrEL1(InMemoryRegister<u64, ESR_EL1::Register>);
```
kevinaboos: are both of these supposed to be transparent? the doc comment and `repr` attribute only apply to the first one right now.
You should also provide better docs for each of these structs, i.e., describe what Spsr and Esr stand for and what they're used for.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+
+/// The exception context as it is stored on the stack on exception entry.
+///
+/// Warning: the assembly file next to this one assumes this exact
```
kevinaboos: be specific
```suggestion
/// Warning: the assembly file `table.s` assumes this exact
```

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+        let kernel_mmi_ref = get_kernel_mmi_ref()
+            .ok_or("logger_aarch64: couldn't get kernel MMI ref")?;
+
+        let mut mmi = kernel_mmi_ref.lock();
+        let page_table = &mut mmi.deref_mut().page_table;
+
+        log::info!("Configuring the GIC");
+        let mut inner = ArmGic::init(
+            page_table,
+            GicVersion::InitV3 {
```
kevinaboos, line 108: is there a way to determine whether we should use GIC v3 or v2? Right now it's hard-coded as v3 but there's no mention of that anywhere in the docs.

I think perhaps we need a board config file that specifies all of the assumptions and choices we've made about our QEMU environment. This can be a separate crate for ease of use.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+        inner.set_minimum_priority(0);
+        *gic = Some(inner);
+
+        log::info!("Done Configuring the GIC");
+
+        Ok(())
+    }
+}
+
+pub fn enable_timer_interrupts() -> Result<(), &'static str> {
```
kevinaboos: the indentation on this function is 8 spaces instead of 4

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+
+#[no_mangle]
+extern "C" fn current_el0_irq(_e: &mut ExceptionContext) {
+    panic!("Should not be here. Use of SP_EL0 in EL1 is not supported.")
+}
+
+#[no_mangle]
+extern "C" fn current_el0_serror(_e: &mut ExceptionContext) {
+    panic!("Should not be here. Use of SP_EL0 in EL1 is not supported.")
+}
```
kevinaboos: use "BUG: ..." instead of "Should not be here". It's clearer and more consistent with the rest of Theseus.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+/// Send an "end of interrupt" signal, notifying the interrupt chip that
+/// the given interrupt request `irq` has been serviced.
+pub fn eoi(irq_num: IntNumber) {
+    let mut gic = GIC.lock();
+    let gic = gic.as_mut().expect("GIC is uninitialized");
+    gic.end_of_interrupt(irq_num);
+}
+
+#[rustfmt::skip]
+impl fmt::Display for SpsrEL1 {
```
kevinaboos: These 3 display blocks should implement `Debug`, not `Display`. This is far too much inner details for a `Display`impl, imo.

kevinaboos (changes requested):
I think the latter half of `kernel/interrupts/src/aarch64/mod.rs` should be moved into an "exception"-related crate or module, since they're not strictly about interrupt handling; they're  mostly early/default exception handlers.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: Wow I fixed this in the past, why wasn't it commited I'm losing my mind

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: image

This PR configures virt as a GICv3 machine

kernel/gic/src/mod.rs
NathanRoyer: No, sometimes it's not 32, for instance in dist_interface::get_spi_target it's 4.

Well, these functions are complex... but I tried my best with the doc, what's confusing?

kernel/gic/lib.rs
NathanRoyer: > We don't need "include_str" here, that's not appropriate for Rust source code. Target-specific `mod` statements are preferred to include/exclude code.

Well that excludes the second solution, as far as I know you can't have multiple levels of modules in the same folder.

Ok, I'll add another directory. I'll do that after we settle on the rest of the to-be-closed PR, in the mergeable one. Please leave this unresolved.

aarch64/kernel/gic/src/redist.rs
NathanRoyer: What do you mean, it's all commented out? the github preview is outdated, the PR is not including any commented code right now except in interrupts::aarch64::enable_timer_interrupts.

kernel/gic/src/redist_interface.rs
NathanRoyer: Which other file? The redistributor is the only component that has such polling initialization requirements.

kernel/gic/src/redist_interface.rs
NathanRoyer: I added this to dist_interface.rs:
```rust
//! The Distributor forwards or discards SPIs (shared peripheral interrupts)
//! to:
//! - The redistributor, in GICv3,
//! - The CPU core, in GICv2.
//! There's one Distributor per GIC chip.
```
and this to redist_interface.rs:
```rust
//! The Redistributor forwards or discards PPIs (private peripheral interrupts)
//! & SGIs (software generated interrupts) to the CPU core, in GICv3. There's
//! one redistributor per CPU core.
```

kernel/gic/src/mod.rs
NathanRoyer: I disagree with naming it `MemoryMappedGicRegisters` as it doesn't mention that the mapping is one page long, which to me is important as some interfaces (dist, redist for example) are more than one page long; they use multiple instances of this struct.

kernel/gic/src/mod.rs
NathanRoyer: Indeed, outdated comment, I'm removing it.

kernel/interrupts/src/aarch64/table.s
NathanRoyer: `align` is well supported: https://doc.rust-lang.org/nightly/reference/inline-assembly.html

image

The Arm manual doesn't explain the 2048b alignment but I'll add my own explanation if you want:
```
// Align by 2^11 bytes, as demanded by ARMv8-A. Same as ALIGN(2048) in an ld script.
//
// This is required because upon taking an exception, the CPU adds an offset (based
// on the context and type of exception) to the aligned vector address (which is
// stored in VBAR_ELx) and jumps there. If the vector address wasn't aligned, a real
// addition would be required, but as it is aligned, a simple and cheaper OR operation
// is used.
//
.align 11
```

kevinaboos: Thanks. The extra docs you provide there are great, super clear.

I'd still prefer using ".balign" since 99% of developers reading this in the future will be very puzzled by an `.align 11` statement. Kindly change it to that for the sake of better readability.

kernel/gic/src/mod.rs
kevinaboos: The issue with `GicMappedPage` is two-fold:
1. it implies that it is the singular mapped page for the GIC, which may not be true (since there are multiple sets of single-page MappedPages objects for a GIC
2. It's not a MappedPage, it's the type that you are overlaying atop a MappedPages, i.e., what you are reinterpreting that memory as. Calling it a MappedPage in this context is actually incorrect, not just my preference.

Why does it matter how many pages it is? Take a look at the `ApicRegisters` type, which is 100% analogous to this type. If you really want to include that unnecessary detail in the type name, you could go with `GicRegistersPage` or something similar.

kernel/gic/src/mod.rs
kevinaboos: ok, good to know that the const arg is needed.

I think perhaps the first line of the comment is confusing. It writes one "slot" of an array, which implies one element, but then it goes on to say that something spans multiple `u32`s. Does the element itself span multiple `u32`s, or does the whole array span multiple `u32`s`, or do you interpret the contents of the array differently based on the arguments, meaning that it's not really an array with a standard fixed layout? 

In either case you're just reading or writing a single `u32`, so the concept of spanning multiple `u32`s doesn't really make sense, unless it's an unaligned read/write that touches different parts of two adjacent `u32`s?

Oh, and I also just noticed you're using a code comment `//` instead of a doc comment `///` with 3 slashes.

kernel/interrupts/src/aarch64/mod.rs
kevinaboos: right, I understand that. I was asking if there's a way to determine that dynamically, or if it's yet another characteristic of the system that we have to obtain from a board config file. 

For example, on x86 you can query a certain system register to determine whether the system supports xapic or x2apic (similar to GIC v2 vs v3).

kernel/gic/src/mod.rs
NathanRoyer: > In either case you're just reading or writing a single `u32`, so the concept of spanning multiple `u32`s`doesn't really make sense, unless it's an unaligned read/write that touches different parts of two adjacent`u32`s?

The maximum item size is 32 bits, and the items are always aligned to 2**n bits. The array spans multiple adjacent u32s but there is always a integer number of items in a single u32.

→ I'll add this to the doc.

> Oh, and I also just noticed you're using a code comment `//` instead of a doc comment `///` with 3 slashes.

I did that intentionally as it's a private function.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: > right, I understand that. I was asking if there's a way to determine that dynamically, or if it's yet another characteristic of the system that we have to obtain from a board config file.

We have no way to detect the version at runtime that I know of; nothing is told in the spec about that.

But as there are multiple interrupt controllers in the ARMv8 world, it's probably good to read the version info from a board config file as we'll probably have to read the interrupt controller type from such a config file too, when we support multiple ones.

kernel/gic/src/mod.rs
NathanRoyer: Ok, I renamed it GicRegisters

kernel/gic/lib.rs
NathanRoyer: I just pushed a commit that migrates to this layout:
```
kernel/gic
├── Cargo.toml
└── src
    ├── lib.rs
    └── gic
        ├── cpu_interface_gicv2.rs
        ├── cpu_interface_gicv3.rs
        ├── dist_interface.rs
        ├── mod.rs
        └── redist_interface.rs
```

kernel/interrupts/src/aarch64/table.s
NathanRoyer: Ok, done

kernel/gic/src/redist_interface.rs
NathanRoyer: I went for `is_sgippi_enabled` / `enable_sgippi` & documented the acronyms in the functions' documentation.

I did this for distributor equivalents too.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: Converted it to Debug.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: I'm not an expert of `UnsafeCell`; but using `usize` here would probably make people think that the label points to data (a usize) whereas it's actually code. We can however do:
```rust
    extern "Rust" {
        // in assembly file
        static __exception_vector_start: ();
    }
    unsafe { VBAR_EL1.set(&__exception_vector_start as *const _ as u64) };
```
If you prefer

PR #733, opened 2022-12-14
Bootstrap initial page table from `CR3`, not `.page_table` section

tsoutsman:
Change the paging initialiser to bootstrap the bootloader page table from the `CR3` register rather than the `.page_table` section as that won't be present in the upcoming UEFI port.

### DIFF
--- a/kernel/memory/src/paging/mod.rs
+++ b/kernel/memory/src/paging/mod.rs
@@ -234,7 +234,7 @@ pub fn init(
     debug!("{:X?}\n{:X?}", aggregated_section_memory_bounds, _sections_memory_bounds);
     
     // bootstrap a PageTable from the currently-loaded page table
-    let current_active_p4 = frame_allocator::allocate_frames_at(aggregated_section_memory_bounds.page_table.start.1, 1)?;
+    let current_active_p4 = frame_allocator::allocate_frames_at(get_current_p4().start_address(), 1)?;
     let mut page_table = PageTable::from_current(current_active_p4)?;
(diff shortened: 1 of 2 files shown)


kernel/memory/src/paging/mod.rs
```diff
@@ -234,7 +234,7 @@ pub fn init(
     debug!("{:X?}\n{:X?}", aggregated_section_memory_bounds, _sections_memory_bounds);
     
     // bootstrap a PageTable from the currently-loaded page table
-    let current_active_p4 = frame_allocator::allocate_frames_at(aggregated_section_memory_bounds.page_table.start.1, 1)?;
+    let current_active_p4 = frame_allocator::allocate_frames_at(get_current_p4().start_address(), 1)?;
```
kevinaboos: while we're changing this, let's add a more detailed error since it's possible that the new page table `AllocatedFrame` might not be properly separated from the rest of the sections and therefore not available (this would indicate a build/link mistake on our part).
```suggestion
    let current_active_p4 = frame_allocator::allocate_frames_at(get_current_p4().start_address(), 1)
        .map_err(|_| "Failed to allocate frame for initial page table; is it merged with another section?")?;
```

kevinaboos (changes requested):
Looks great, thanks! Just one tiny improvement for error msg clarity.

kevinaboos (approved):
I went ahead and applied my one tiny suggestion to save us time. Plus I want to improve the `PageTable::from_current()` function since it's doing a lot of dumb redundant checks.

PR #734, opened 2022-12-14
Add option to use RSDP address provided by bootloader

tsoutsman:
I wasn't able to actually test the changes since GRUB doesn't provide the RSDP address.

Signed-off-by: Klim Tsoutsman <klim@tsoutsman.com>

PR #735, opened 2022-12-14
Add boot information abstraction

tsoutsman:
Necessary for the upcoming UEFI port.

### DIFF
--- a/kernel/boot_info/src/lib.rs
+++ b/kernel/boot_info/src/lib.rs
@@ -98,5 +98,5 @@ pub trait BootInformation: 'static {
     fn rsdp(&self) -> Option<PhysicalAddress>;
 
     /// Returns the stack size in bytes.
-    fn stack_size(&self) -> usize;
+    fn stack_size(&self) -> Result<usize, &'static str>;
 }
--- a/kernel/boot_info/src/lib.rs
+++ b/kernel/boot_info/src/lib.rs
@@ -1,0 +1,102 @@
+//! This crate provides an abstraction over multiboot2 and UEFI boot
+//! information.
+//!
+//! It allows the kernel's initialisation to be the same, regardless of whether
+//! it was booted using BIOS or UEFI.
+
+#![feature(type_alias_impl_trait)]
+#![no_std]
+
+pub mod multiboot2;
+
+use core::{iter::Iterator, ops::Range};
+use memory_structs::{PhysicalAddress, VirtualAddress};
+
+pub trait MemoryRegion {
+    /// Returns the region's starting physical address.
+    fn start(&self) -> PhysicalAddress;
+
+    /// Returns the region's length.
+    fn len(&self) -> usize;
+
+    /// Returns whether the region can be used by the frame allocator.
+    fn is_usable(&self) -> bool;
+}
+
+pub trait ElfSection {
+    /// Returns the section's name.
+    fn name(&self) -> &str;
+
+    /// Returns the section's starting virtual address.
+    fn start(&self) -> VirtualAddress;
+
+    /// Returns the section's length.
+    fn len(&self) -> usize;
+
+    /// Returns the section's flags.
+    fn flags(&self) -> ElfSectionFlags;
+}
+
+bitflags::bitflags! {
+    /// ELF section flags.
+    pub struct ElfSectionFlags: u64 {
+        /// The section contains data that should be writable during program execution.
+        const WRITABLE = 0x1;
+
+        /// The section occupies memory during the process execution.
+        const ALLOCATED = 0x2;
+
+        /// The section contains executable machine instructions.
+        const EXECUTABLE = 0x4;
+    }
+}
+
+pub trait Module {
+    /// Returns the module's name.
+    fn name(&self) -> Result<&str, &'static str>;
+
+    /// Returns the module's starting physical address.
+    fn start(&self) -> PhysicalAddress;
+
+    /// Returns the module's length.
+    fn len(&self) -> usize;
+}
+
+pub trait BootInformation: 'static {
+    type MemoryRegion<'a>: MemoryRegion;
+    type MemoryRegions<'a>: Iterator<Item = Self::MemoryRegion<'a>>;
+
+    type ElfSection<'a>: ElfSection;
+    type ElfSections<'a>: Iterator<Item = Self::ElfSection<'a>>;
+
+    type Module<'a>: Module;
+    type Modules<'a>: Iterator<Item = Self::Module<'a>>;
+
+    /// Returns the boot information's starting virtual address.
+    fn start(&self) -> Option<VirtualAddress>;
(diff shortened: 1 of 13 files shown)


kernel/boot_info/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Returns the range of physical addresses at which the modules are
+    /// located.
+    fn modules_memory_range(&self) -> Result<Range<PhysicalAddress>, &'static str>;
+
+    /// Returns memory regions describing the physical memory.
+    fn memory_regions(&self) -> Result<Self::MemoryRegions<'_>, &'static str>;
+    /// Returns the kernel's ELF sections.
+    fn elf_sections(&self) -> Result<Self::ElfSections<'_>, &'static str>;
+    /// Returns the modules found in the kernel image.
+    fn modules(&self) -> Self::Modules<'_>;
```
kevinaboos, line 95: i haven't tried this, but I _think_ rustc lets you omit the `<'_>` inferred lifetime part to where you can just write `Result<Self::MemoryRegions, ...>`. Nbd either way.

kernel/boot_info/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+
+pub trait MemoryRegion {
+    /// Returns the region's starting physical address.
+    fn start(&self) -> PhysicalAddress;
+
+    /// Returns the region's length.
+    fn len(&self) -> usize;
+
+    /// Returns whether the region can be used by the frame allocator.
+    fn is_usable(&self) -> bool;
```
kevinaboos, line 23: "usable" here means that it's available for general-purpose usage is not reserved for any other specific use, right? 

I'm fine with a bool as long as it's sufficient for our purpose, but I think we should add a bit more detail about what "usable" really means, and what it means if it's not usable. Perhaps an enum (non-exhaustive?) would be better, and then we could more easily add different types if we need to disambiguate in the future. 

Open to hearing your thoughts here.

kernel/boot_info/src/multiboot2.rs
```diff
@@ -0,0 +1,176 @@
...
+    fn start(&self) -> VirtualAddress {
+        VirtualAddress::new_canonical(self.start_address() as usize)
+    }
+
+    fn len(&self) -> usize {
+        multiboot2::ElfSection::size(self) as usize
+    }
+
+    fn flags(&self) -> ElfSectionFlags {
+        ElfSectionFlags::from_bits_truncate(multiboot2::ElfSection::flags(self).bits())
```
kevinaboos: this implicitly relies on the `ElfSectionFlags` bit layout to be identical to that of `multiboot2::ElfSection`. 

I think this is okay, but that requirement is sneakily hidden on only expressed here if you look closely.
For ex, I didn't know that until I studied this line.

One way to encode this explicitly would be to directly use the `multiboot2::ElfSection` const flag definition's bit values in the definition of `ElfSectionFlags`, i.e.,:
```rust
    pub struct ElfSectionFlags: u64 {
        /// The section contains data that should be writable during program execution.
        const WRITABLE = multiboot2::ElfSectionFlags::WRITABLE.bits();

        /// The section occupies memory during the process execution.
        const ALLOCATED = multiboot2::ElfSectionFlags::ALLOCATED.bits();

        /// The section contains executable machine instructions.
        const EXECUTABLE = multiboot2::ElfSectionFlags::EXECUTABLE.bits();
    }
```

That way, an implicitly incorrect conversion could never happen. Alternatively, you could code this using a more readable series of three conditionals, one per flag. That might mesh better with the upcoming UEFI definition of `ElfSectionFlags`, which I assume likely differs from multiboot2.

kernel/boot_info/src/multiboot2.rs
```diff
@@ -0,0 +1,176 @@
...
+            .ok_or("no elf sections tag")?
+            .sections())
+    }
+
+    fn modules(&self) -> Self::Modules<'_> {
+        self.module_tags()
+    }
+
+    fn rsdp(&self) -> Option<PhysicalAddress> {
+        // FIXME: Implement. Depends on #734.
```
kevinaboos: #734 has been merged in, thanks!

kernel/boot_info/src/multiboot2.rs
```diff
@@ -0,0 +1,176 @@
...
+        self.elf_sections()
+            .expect("couldn't get elf sections")
+            .filter(|section| section.name() == ".stack")
+            .map(|section| {
+                let start = section.start();
+                let end = start + section.len();
+                (end - start).value()
+            })
+            .next()
+            .expect("no stack section")
```
kevinaboos: why does only this function panic? Might as well return a `Result` here too, just like for all the other fallible functions.

kevinaboos (changes requested):
Great work on this, the code is super clean and very nice to read!

🎉 for using GATs, lol

Just left a few minor comments, should be quick & easy to address.

kernel/boot_info/src/multiboot2.rs
tsoutsman: The ELF section flags are the same regardless of the boot specification used. [Table 7--8](https://docs.oracle.com/cd/E19120-01/open.solaris/819-0690/6n33n7fcj/index.html). I do agree that using three conditionals would be more readable, so I can do that if you think we should.

kernel/boot_info/src/lib.rs
tsoutsman: Not for associated types.

kernel/boot_info/src/lib.rs
tsoutsman: I initially had an enum, but the memory startup code only checks whether or not the memory is usable, so I figured it would be easier to use a boolean.

It's only ever used in this loop:
```rust
    for area in boot_info.memory_regions()? {
        let frames = FrameRange::from_phys_addr(PhysicalAddress::new_canonical(area.start().value()), area.len());
        if area.is_usable() {
            free_regions[free_index] = Some(PhysicalMemoryRegion::new(frames, MemoryRegionType::Free));
            free_index += 1;
        } else {
            reserved_regions[reserved_index] = Some(PhysicalMemoryRegion::new(frames, MemoryRegionType::Reserved));
            reserved_index += 1;
        }
    }
```
where it seems natural to be expressed in terms of a boolean.

kernel/boot_info/src/lib.rs
kevinaboos: Thanks for clarifying, i figured you knew better 😄

kernel/boot_info/src/multiboot2.rs
kevinaboos: > The ELF section flags are the same regardless of the boot specification used.

Oh true. For some reason I was thinking of bootloader-defined flag sets.

> I do agree that using three conditionals would be more readable, so I can do that if you think we should.

Sure, go for it.

PR #736, opened 2022-12-14
Fix typo

hecatia-elegua:


kevinaboos (approved):
Thanks for catching that! Much appreciated.

PR #737, opened 2022-12-14
Added `multicore`: An abstraction over the discovery of processor cores

NathanRoyer:
That muticore crate re-exports `apic::get_current_core_id` and related stuff:

- CoreId -> the type of a core id = u8 on x86_64, u32 on aarch64
- cores_count -> previously `cpu_count`
- get_bootstrap_core_id -> previously `get_bsp_id`
- is_bootstrap_core -> previously `is_bsp`
- get_my_core_id -> previously `get_my_apic_id`

As you can see I also renamed some of them to avoid confusion to future users. Please notice me if I made mistakes there.

There were crates which only depended on `apic`, in which cases I replaced the `apic` dependency with a `multicore` dependency.

Feel free to refactor `multicore` to another name if you want.


kernel/apic/src/lib.rs
```diff
@@ -39,22 +41,22 @@ pub enum InterruptChip {
 const_assert!(AtomicCell::<InterruptChip>::is_lock_free());
 
 /// The set of system-wide `LocalApic`s, one per CPU core.
-static LOCAL_APICS: AtomicMap<u8, RwLockIrqSafe<LocalApic>> = AtomicMap::new();
+static LOCAL_APICS: AtomicMap<CoreId, RwLockIrqSafe<LocalApic>> = AtomicMap::new();
 
 /// The number of CPUs currently initialized in the system.
 /// This must match the number of Local APICs initialized in the system.
-static CPU_COUNT: AtomicU8 = AtomicU8::new(0);
+static CPU_COUNT: AtomicUsize = AtomicUsize::new(0);
```
kevinaboos: u32 is the max size of a CPU ID
```suggestion
static CPU_COUNT: AtomicU32 = Atomic32::new(0);
```

kernel/apic/src/lib.rs
```diff
@@ -423,7 +425,7 @@ impl LocalApic {
             inner,
             processor_id,
             apic_id: u8::MAX, // placeholder, is replaced below.
```
kevinaboos: also change this to `CpuId::MAX` for consistency

kevinaboos (changes requested):
Thanks for this PR, looks good.

In considering naming and terminology choices, here is my view:
* In general, use the term `cpu` instead of `core`.
  * I have been trying to gradually migrate to this terminology in other parts of the repo. 
* `multicore` crate can now be renamed to `cpu` due to #746
* `get_my_core_id()` --> `current_cpu()`
  * I also plan to do the same for tasks, e.g., `get_my_current_task()` --> `current_task()`
* `get_bootstrap_core_id()` --> `bootstrap_cpu()`
* `is_bootstrap_core()` --> `is_bootstrap_cpu()`
* `cores_count()` --> `cpu_count()`
* `CoreId` --> `CpuId`

Since x86_64's APIC also supports 32-bit CPU IDs, we can always use 32-bit values there; however, you can leave it as is for this PR and make that change in a future PR.

kevinaboos (approved):
Thanks, looks great! I did a bit of cleanup; feel free to review it when merging this changeset into your other aarch64 PRs.

PR #739, opened 2022-12-15
Implement `time::ClockSource` for `HpetTimer`

tsoutsman:
Implementing the trait revealed some flaws in the design of the `time` crate:
- Using the period is easier than using the frequency
- The `duration_to_instant` and `instant_to_duration` functions aren't necessary because we know the period/frequency.
- `Instant` was missing a `new` function.

This commit also adds `seconds_counter` as a simple test that the monotonic clock is working correctly.


kernel/acpi/src/lib.rs
```diff
@@ -79,6 +79,14 @@ pub fn init(page_table: &mut PageTable) -> Result<(), &'static str> {
         let acpi_tables = ACPI_TABLES.lock();
         if let Some(hpet_table) = hpet::HpetAcpiTable::get(&acpi_tables) {
             hpet_table.init_hpet(page_table)?;
+            let period = time::Period::new(
+                hpet::get_hpet()
+                    // Initialisation succedeed so this cannot fail.
+                    .expect("couldn't get HPET")
```
kevinaboos: the above call to `init_hpet()` returns a wrapped `Hpet` object, so you can just use that directly instead of making a fallible call to `get_hpet()` again.

kernel/time/src/lib.rs
```diff
@@ -44,6 +31,10 @@ pub struct Instant {
 impl Instant {
     const ZERO: Self = Self { counter: 0 };
 
+    pub fn new(counter: u64) -> Self {
+        Self { counter }
+    }
+
     /// Returns the amout of time elapsed from another instant to this one, or
```
kevinaboos: unrelated, but just saw a small typo here "amout"

kernel/time/src/lib.rs
```diff
@@ -53,26 +44,29 @@ impl Instant {
                 None => return Duration::ZERO,
             },
         };
-        instant_to_duration(instant)
+        let femtos = u128::from(instant.counter) * u128::from(MONOTONIC_PERIOD.load());
+        Duration::from_nanos((femtos / FEMTOS_TO_NANOS) as u64)
     }
 
     pub fn checked_duraction_since(&self, earlier: Self) -> Option<Duration> {
```
kevinaboos: another type "duraction"

kernel/time/src/lib.rs
```diff
@@ -100,26 +95,35 @@ impl core::ops::Sub<Instant> for Instant {
...
 
-impl Frequency {
-    /// Creates a new frequency with the specified hertz.
+impl Period {
+    const MAX: Self = Self(u64::MAX);
+
+    /// Creates a new frequency with the specified femtoseconds.
     pub fn new(frequency: u64) -> Self {
         Self(frequency)
     }
```
kevinaboos: should change `frequency` terms here to `period`

kernel/time/src/lib.rs
```diff
@@ -100,26 +95,35 @@ impl core::ops::Sub<Instant> for Instant {
...
-    /// Creates a new frequency with the specified hertz.
+impl From<Period> for u128 {
+    /// Returns the period in femtoseconds.
+    fn from(f: Period) -> Self {
+        f.0.into()
+    }
+}
+
+impl From<u64> for Period {
+    /// Creates a new frequency with the specified femtoseconds.
```
kevinaboos:
```suggestion
    /// Creates a new period with the specified femtoseconds.
```

kernel/acpi/hpet/src/lib.rs
```diff
@@ -106,6 +107,13 @@ impl Hpet {
     }
 }
 
+impl time::ClockSource for Hpet {
+    type ClockType = time::Monotonic;
+
+    fn now() -> Instant {
+        Instant::new(get_hpet().expect("couldn't get HPET").get_counter())
+    }
```
kevinaboos, line 115: I understand why it works like this, but I still find this API supremely weird. Typically when you impl a trait for a type, you use that type's members in some way.

Obviously for the `now()` fn (and that function only), you'll call it without any args/context, so this is a bit more sensical. However, I would argue that there should be an "inner now" fn that _does_ use the clock type's fields/members along with the rest of the `Clocksource` methods, and then an "outer now" fn that can depend on those.

In any case we can keep this as is for now, but I think the entire `time` API needs to change such that you implement `Clocksource` for specific clock types in a typical way that actually uses their fields & member functions. Then you can have a standalone, contextless/argumentless `now()` adapter around that.

kevinaboos: Follow-up: the "outer now" fn would basically just hold a reference to the current "best" impl for each clocktype. This is more in line with the idea of registering all clocksources and then ranking them as they are registered. 

The key design concept is that there would never exist a way to invoke `now()` on an `Hpet` clocksource if the `Hpet` did not exist and hadn't been registered; this is how the other device registration APIs work. 

With the current impl, I could invoke `now()` on an `Hpet` device directly and cause a panic even without obtaining a reference to an initialized Hpet device first; that's really what I find strange about this API
```rust
fn main() {
    use time::Clocksource;
    let now = Hpet::now(); // panic
}
```
I realize it wouldn't be called if you go through the `time` API (since it wouldn't've been registered), but it's still pretty odd.

kevinaboos (changes requested):
In `kernel/time` make sure to change all mentions of "frequency" to "period", and also that we prioritize smaller periods (same as previously with larger frequencies).

In general things are fine, but I believe we'll need to fundamentally change the `Clocksource` API to handle clock device registration in a more typical way. See my large comment about that in the `impl Clocksource for Hpet` block.

kernel/acpi/hpet/src/lib.rs
tsoutsman: We want to minimise the number of atomic function pointers we use. If `Clocksource::now` has a `&self` parameter, then we also need a `Clocksource::get_ref` function that returns a type that dereferences into a `&self`. But then you're back to where you started because the `get_ref` function would have to assume that the static is initialised.

We can't pass a `&'static impl Lockable<'static, dyn ClockSource>` into `time::register_clock_source` because we can't cast a `&'static RwLock<BorrowedMappedPages<Hpet, Mutable>>` into a trait object as the layout of `Hpet` is important.

At some point, the `Clocksource` trait needs to interact with the static, but we can never guarantee that the `static` is initialised. I don't see a way around at least one of the `ClockSource` functions panicking, and although using `now` as we do is confusing, it minimises the number of atomic function pointers we use.

kernel/time/src/lib.rs
```diff
@@ -44,7 +31,11 @@ pub struct Instant {
...
     const ZERO: Self = Self { counter: 0 };
 
-    /// Returns the amout of time elapsed from another instant to this one, or
+    pub fn new(counter: u64) -> Self {
+        Self { counter }
+    }
+
+    /// Returns the amount of time elapsed from another instant to this one, or
     /// zero duration if that instant is later than this one.
     pub fn duration_since(&self, earlier: Self) -> Duration {
```
kevinaboos, line 40: kindly deduplicate this function. Something like `checked_duration_since(...).unwrap_or(Duration::ZERO)`

kernel/time/src/lib.rs
```diff
@@ -100,46 +95,55 @@ impl core::ops::Sub<Instant> for Instant {
...
 /// Returns whether the early sleeper was overwritten.
-pub fn register_early_sleeper<T>(frequency: Frequency) -> bool
+pub fn register_early_sleeper<T>(period: Period) -> bool
 where
     T: EarlySleeper,
 {
-    let old_frequency = EARLY_SLEEPER_FREQUENCY.load(Ordering::SeqCst);
-    if frequency > Frequency::new(old_frequency) {
+    let old_period = EARLY_SLEEPER_PERIOD.load();
+    if period < old_period {
```
kevinaboos: _technically_ we ought to use compare and swap (or whatever Rust calls it.... `compare_exchange()`) to do this.
Same goes for `register_clock_source()`.

kevinaboos (changes requested):
Did a second (heh) review with a few more comments. Nothing major.

kernel/time/src/lib.rs
tsoutsman: Compare and swap won't work because it compares equality not less than/greater than. I did change the implementation to use `fetch_update` though.

kevinaboos:
oops merged that without fixing the commit msg. oh well

PR #742, opened 2022-12-16
Add deferred interrupt task for E1000 NIC

tsoutsman:
This is the first step in supporting an asynchronous networking interface. In a future PR, `net::Socket`'s interface will be changed to be more like `std::net::*Socket` rather than dereferencing to a smoltcp socket. However, that depends on #717 being merged, as we need wakers to implement blocking and asynchronous send/receive methods. The `Interface::poll_socket` method and rampant mutex locking in `ping_2` are temporary workarounds before then.

The PR can be reviewed, but before being merged, the git dependency for smoltcp needs to be changed back to the main smoltcp repo, not my fork. This can be done after smoltcp-rs/smoltcp#718 is merged. I didn't bother moving the fork to the Theseus organisation as the smoltcp PR is very small and should be merged relatively soon.


kevinaboos:
Now waiting for https://github.com/smoltcp-rs/smoltcp/pull/721 to be merged

tsoutsman:
Converted to draft as per Dirbaio's (don't want to ping) comment in the smoltcp pull request.

tsoutsman:
From smoltcp-rs/smoltcp#721:
> One thing I noticed from your snippet and looking at https://github.com/theseus-os/Theseus/pull/742 is that you're creating one SocketSet for each socket and doing one poll() call on the interface for each socket. This will NOT work. smoltcp assumes poll() is called with a SocketSet containing ALL sockets.

This means our legacy interface is incorrect.

There are two ways to solve this:
1. Store each socket behind a mutex and when polling, lock all the sockets and add them to a socket set.
2. Store a single socket set behind a mutex.

I initially went for option 1 ([`73dbb63`](https://github.com/theseus-os/Theseus/pull/742/commits/73dbb63679cc2e519a2fcc73f8529e69f65a325e)) but I was getting out of bounds indexing panics from within `smoltcp`, so I switched to option 2.

Since we poll after a send or receive, option 2 is better when sending or receiving because it doesn't need to lock a bunch of mutexes and create a new socket set. However, option 1 is better for other operations because it doesn't need to lock all other sockets to access a specific socket, e.g., socket.can_send()`.

Either option isn't great for performance but is necessary for correctness.

kevinaboos (approved):

PR #744, opened 2022-12-16
Context Switching Primitives for AArch64

NathanRoyer:
This builds on #701.

This draft PR contains "regular" context switching primitives for aarch64:

- the `context_switch_regular` function (same signature as on x86_64):
    - saves general pupose regs,
    - saves the current SP to an arbitrary address expected to be in x0 (= parameter `_prev_stack_pointer`),
    - installs the new one expected to be in x1 (= parameter `_next_stack_pointer_value`),
    - restores GP regs from the just-installed stack
    - jumps to address in x30
- `create_stack` is a utility function which allocates a 16-page long initial task stack; you have to give it a function pointer (such as the address of `landing_pad`) which will be executed if you use `context_switch_regular` with that new stack.
- `landing_pad` is a "task" function which simply re-triggers a context switch to return to the original main task.

These are used in `main` just before infinite looping as a demonstration.


aarch64/kernel/context_switch_regular/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+/// * The `x0` register must contain a pointer to the previous task's stack pointer.
+/// * The `x1` register must contain the value of the next task's stack pointer.
+#[macro_export]
+macro_rules! switch_stacks {
+    () => [
+        // switch the stack pointers
+        r#"
+            // Save current stack pointer to address in 1st argument.
+            mov x2, sp
+            str x2, [x0, 0]
```
kevinaboos: i'm not an expert on aarch64 asm, but can this just be `str x2, [x0]`?

aarch64/kernel/context_switch_regular/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+            x14: 0, x15: 0,
+            x16: 0, x17: 0,
+            x18: 0, x19: 0,
+            x20: 0, x21: 0,
+            x22: 0, x23: 0,
+            x24: 0, x25: 0,
+            x26: 0, x27: 0,
+            x28: 0, x29: 0,
+
+            // x30 stores the return address
```
kevinaboos, line 42: let's also mention that `x30` is a synonym for `LR`, the link register.
Add that here and everywhere else `x30` is referenced. Alternatively would it be better to use `LR` directly since it more clearly denotes that it's different than all other `x*` general-purpose regs?

aarch64/kernel/context_switch_regular/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+authors = ["Kevin Boos <kevinaboos@gmail.com>", "Nathan Royer <nathan.royer.pro@gmail.com>"]
```
kevinaboos: it's all you, haha
```suggestion
authors = ["Nathan Royer <nathan.royer.pro@gmail.com>"]
```

aarch64/kernel/context_switch_regular/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /// Sets the value of the first register to the given `value`.
+    /// 
+    /// This is useful for storing a value (e.g., task ID) in that register
+    /// and then recovering it later with [`read_first_register()`].
+    /// 
+    /// On aarch64, this sets the `x2` register.
+    pub fn set_first_register(&mut self, value: usize) {
+        self.x2 = value;
```
kevinaboos: This is fine for now, but i think once we change the context struct to hold only callee-saved registers, we'd want to use the first one there (`x18`).

Also, I only called that function `set_first_register()` because the first register is `r15` on x86, which is actually the _last_ register possibly used in the calling convention on x86. Since aarch64 has so many registers, it doesn't really matter which one we use, but technically the _last_ one would be `x29`. Here, "last" means "the least likely one to be ever overwritten".

kevinaboos: So, really, we ought to use `x29` i think. Doesn't matter much between `x18` and `x29`, but yeah.

kevinaboos (changes requested):
Looks good! 

Linux only saves/restores `x18` - `x29` on aarch64, which apparently corresponds to the set of callee-saved registers. If you change your implementation to that, does everything still work correctly?
https://github.com/torvalds/linux/blob/69b41ac87e4a664de78a395ff97166f0b2943210/arch/arm64/kernel/entry.S#L820

aarch64/kernel/context_switch_regular/src/lib.rs
NathanRoyer: https://wiki.cdot.senecacollege.ca/wiki/AArch64_Register_and_Instruction_Quick_Start

I forgot about this but x29 is the "frame register" and apparently serves some special purpose. I'll use x28.

NathanRoyer:
> If you change your implementation to that, does everything still work correctly?

Indeed, it still works correctly :ok_hand:

aarch64/kernel/context_switch_regular/src/lib.rs
```diff
@@ -101,27 +85,15 @@ macro_rules! save_registers_regular {
...
-            stp x18, x19, [sp, #8 *  8 * 2]
-            stp x20, x21, [sp, #8 *  9 * 2]
-            stp x22, x23, [sp, #8 * 10 * 2]
-            stp x24, x25, [sp, #8 * 11 * 2]
-            stp x26, x27, [sp, #8 * 12 * 2]
-            stp x28, x29, [sp, #8 * 13 * 2]
-
-            // x30 stores the return address.
-            str x30,      [sp, #8 * 14 * 2]
+            sub sp,  sp,  #8 * 6 * 2
```
kevinaboos: quite nitpicky, but i think it's clearer to say `#8 * 2 * N`, since it more clearly represents that you are saving an 8-byte register twice at the offset `N`.

aarch64/kernel/context_switch_regular/src/lib.rs
```diff
@@ -101,27 +85,15 @@ macro_rules! save_registers_regular {
         // Save all general purpose registers into the previous task.
         r#"
             // Make room on the stack for the exception context.
```
kevinaboos: is this an exception context? i think that's not quite accurate, it's a task's _execution_ context.

aarch64/kernel/nano_core/context_switch.rs
```diff
@@ -41,48 +40,25 @@ pub fn switch_to_task(new_stack: usize) {
...
-                    0, 0, // x24, x25,
-                    0, 0, // x26, x27,
-                    0, 0, // x28, x29,
-        start_address, 0, // x30, <nothing; kept for alignment>
-
-        // [higher address]
-        // [top of stack]
-    ];
+    let artificial_ctx = ContextRegular::new(start_address as *const () as _);
+    let artificial_ctx_src = artificial_ctx.as_bytes();
```
kevinaboos: we don't need `as_bytes()` here -- we can directly write the object itself to the region of memory as long as it implements `FromBytes`. Not only is that faster, but it's also more readable in terms of the code's intent.

Look at how we do it in `spawn::setup_context_trampoline()`

aarch64/kernel/context_switch_regular/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+authors = ["Nathan Royer <nathan.royer.pro@gmail.com>"]
+name = "context_switch_regular"
+description = "Context switching routines and structs for regular (non-SSE) register sets, aarch64 version"
```
kevinaboos: nit
```suggestion
description = "Context switching routines and structs for regular (non-SIMD) tasks on aarch64"
```

kevinaboos (changes requested):
Only a few nitpicks left, other than that it looks excellent.

kevinaboos (approved):

PR #745, opened 2022-12-16
Implement UEFI boot

tsoutsman:


tools/uefi_builder/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+
+[package]
+name = "uefi_builder"
+version = "0.1.0"
+edition = "2021"
+authors = ["Klim Tsoutsman <klim@tsoutsman.com>"]
+
+[dependencies]
+bootloader = { git = "https://github.com/theseus-os/bootloader", branch = "theseus" }
+ovmf-prebuilt = "0.1.0-alpha.1"
```
kevinaboos, line 13: this is an interesting albeit weird way to pull in prebuilt dependencies. We are currently already using https://github.com/theseus-os/prebuilt-aavmf to store the aarch64 ovmf files; it'd be best to unify these somehow. Perhaps we can fork and expand this, or simply host our own under a `theseus-os`-owned repo.

tools/uefi_builder/src/main.rs
```diff
@@ -0,0 +1,43 @@
+use std::path::PathBuf;
+
+fn main() {
+    let mut args = std::env::args();
+    let kernel: PathBuf = args.nth(1).expect("kernel input path not provided").into();
```
kevinaboos, line 5: i'd prefer a getopts/clap impl of arg parsing here, with a std description of each arg. 

At the bare minimum we need to at least provide a `-h`/`--help` option that lists the arguments and their expected order, if order matters.

kernel/nano_core/src/util.rs
```diff
@@ -0,0 +1,32 @@
...
+    // TODO: handle shutdowns properly with ACPI commands
+    panic!("{}", msg);
+}
+
+// FIXME: This shouldn't be necessary
+#[macro_export]
+macro_rules! println_raw {
+    ($($tail:tt)*) => {
+        #[cfg(feature = "bios")]
+        vga_buffer::println_raw!($($tail)*);
```
kevinaboos, line 30: for now just apply this change directly to the `kernel/vga_buffer` impl of this macro, that way nothing else will break on BIOS.

Add a UEFI cargo feature to `kernel/vga_buffer/Cargo.toml` and have the nano_core's UEFI feature activate it.

kernel/vga_buffer/src/lib.rs
```diff
@@ -39,14 +39,15 @@ static EARLY_VGA_WRITER: Mutex<VgaBuffer> = Mutex::new(
 #[macro_export]
 macro_rules! print_raw {
     ($($arg:tt)*) => ({
-        let _ = $crate::print_args_raw(format_args!($($arg)*));
+        // FIXME
+        // let _ = $crate::print_args_raw(format_args!($($arg)*));
```
kevinaboos, line 43: see my other comment about this

Makefile
```diff
@@ -29,7 +34,11 @@ IS_WSL = $(shell grep -is 'microsoft' /proc/version)
 ###################################################################################################
 BUILD_DIR               := $(ROOT_DIR)/build
 NANO_CORE_BUILD_DIR     := $(BUILD_DIR)/nano_core
-iso                     := $(BUILD_DIR)/theseus-$(ARCH).iso
+ifeq ($(uefi), yes)
+	iso					:= $(BUILD_DIR)/theseus-$(ARCH).efi
+else
+	iso					:= $(BUILD_DIR)/theseus-$(ARCH).iso
+endif
```
kevinaboos, line 41: i'd prefer defining a variable like `ISO_EXTENSION` or file type or something in the above "if UEFI = yes" blocks, and then using it here, e.g., 
```mk
iso := $(BUILD_DIR)/theseus-$(ARCH).$(ISO_EXTENSION)
```

just a bit cleaner

Makefile
```diff
@@ -18,7 +18,12 @@ include cfg/Config.mk
 debug ?= none
 net ?= none
 merge_sections ?= yes
-bootloader ?= grub
+uefi ?= no
+ifeq ($(uefi), yes)
```
kevinaboos, line 22: i'm thinking it may be cleaner to define a `boot_spec` variable or something, with possible values of `uefi` or `bios`, that can be used in the rest of the makefile; that'd remove a lot of the many hard-to-read conditionals throughout.

Makefile
```diff
@@ -76,6 +85,10 @@ else ifeq ($(bootloader),limine)
 	else
 $(error Error: missing '$(LIMINE_DIR)' directory! Please follow the limine instructions in the README)
 	endif
+else ifeq ($(bootloader),none)
+	ifneq ($(uefi), yes)
```
kevinaboos, line 89: and then, here, the boot_spec var could be checked instead.

kernel/nano_core/src/lib.rs
```diff
@@ -16,125 +16,63 @@
...
-use boot_info::BootInformation;
+
+mod util;
+cfg_if::cfg_if! {
+    if #[cfg(feature = "bios")] {
+        mod bios;
+    } else if #[cfg(feature = "uefi")] {
+        mod uefi;
+    }
+    // If neither is enabled, and build script wolud have already emitted an
```
kevinaboos, line 35: might as well throw a `compile_error!()` here anyway just in case some noob modifies the build script accidentally

kernel/boot_info/src/uefi.rs
```diff
@@ -0,0 +1,241 @@
+use crate::ElfSectionFlags;
+use bootloader_api::info;
+use core::{
+    iter::{Iterator, Peekable},
+    ops::Range,
+};
+use kernel_config::memory::{KERNEL_OFFSET, KERNEL_STACK_SIZE_IN_PAGES, PAGE_SIZE};
+use memory_structs::{PhysicalAddress, VirtualAddress};
+
+// TODO: Ideally this would be defined in nano_core.
```
kevinaboos, line 10: explain why we can't do it that way 
(cuz of cyclic deps, right? the nano_core must depend on boot_info)

kernel/memory/src/lib.rs
```diff
@@ -140,7 +140,8 @@ pub fn set_broadcast_tlb_shootdown_cb(func: fn(PageRange)) {
 ///  7. the kernel's list of *other* higher-half MappedPages that needs to be converted to a vector after heap initialization, and which should be kept forever,
 ///  8. the kernel's list of identity-mapped MappedPages that needs to be converted to a vector after heap initialization, and which should be dropped before starting the first userspace program. 
 pub fn init(
-    boot_info: &impl BootInformation
+    boot_info: &impl BootInformation,
+    kernel_stack_start: VirtualAddress,
```
kevinaboos, line 144: can `kernel_stack_start` virt addr be added to `boot_info`? I think it makes sense but I may be missing something.

kernel/nano_core/Cargo.toml
```diff
@@ -58,19 +56,22 @@ path = "../panic_entry"
...
+
 [lib]
 # staticlib is required to build a self-contained, fully-linked .a file 
 # that we can jump to from the bootloader.
 crate-type = ["staticlib"]  
-
-
-## Currently not used; see the `build.rs` script.
-# [build-dependencies]
-# built = "0.5"
```
kevinaboos, line 76: leave this in since the code that uses it is still there.

kevinaboos (changes requested):
One issue regarding the asm files -- maintaining the history there is very important (to me), and this PR loses some of that, e.g., for `common.asm`, `defines.asm`, and esp `ap_realmode.asm`. Touching those files is always tricky and can often lead to problems, so I'd prefer to have all their history maintained to make changes easier to track. 

Can you submit a separate PR for changing/moving _just_ the asm files? Or two. Whatever it takes to appease git and have the history be maintained. You can include the change to `linker_higher_half.ld` as part of those PRs.
EDIT: ok wait, i see that some of them have been duplicated. Let's dedup them where possible; you can use typical ifdefs.


It'd also be nice to have a separate PR for just the changes to add UEFI support to `kernel/boot_info`.

I like the refactoring of the `nano_core`, nice job with that! I'd only request that some crate-level and mod-level comments be added to help describe the structure of `nano_core`, in particular a basic overview of the entry points in the bios and uefi modules and the code flow starting from each.

kernel/memory/src/lib.rs
tsoutsman: Because the UEFI boot info does not contain the kernel stack start. The boot code immediately checks `rsp` and passes its value to `rust_entry`.

tsoutsman:
Closed in favour of #764.

PR #747, opened 2022-12-16
Modify paging initialisation to be compatible with UEFI

tsoutsman:
The kernel stack's start is now acquired from `nano_core` and the initialiser no longer assumes the stack or boot info frames are contiguous. `find_section_memory_bounds` now ignores `.stack` and `.bootloader-config` sections.

The stack start can't be provided by the boot information because on UEFI, `_start` passes the initial value of `rsdp` to `nano_core`, so the boot information doesn't contain the stack start.


kernel/memory/src/paging/mod.rs
```diff
@@ -254,6 +256,22 @@ pub fn init(
     let mut higher_half_mapped_pages: [Option<NoDrop<MappedPages>>; 32] = Default::default();
     let mut identity_mapped_pages:    [Option<NoDrop<MappedPages>>; 32] = Default::default();
 
+    let stack_end_virt = stack_start_virt + boot_info.stack_size()?;
+    // Stack frames are not guaranteed to be contiguous.
+    let mut stack_mappings = [None; 34];
+    // + 1 accounts for the guard page which does not have a corresponding frame.
+    for (i, page) in ((Page::containing_address(stack_start_virt) + 1)..=Page::containing_address(stack_end_virt - 1)).enumerate() {
```
kevinaboos: for convenience, you can use [`PageRange::from_virt_addr()`](https://www.theseus-os.com/Theseus/doc/memory_structs/struct.PageRange.html#method.from_virt_addr)

kernel/memory/src/paging/mod.rs
```diff
@@ -254,6 +256,22 @@ pub fn init(
...
+        let frame = page_table.translate_page(page).ok_or("couldn't translate stack page")?;
+        stack_mappings[i] = Some((page, frame));
+    }
+
+    // Boot info frames are not guaranteed to be contiguous.
+    let mut boot_info_mappings = [None; 10];
+    for (i, page) in (Page::containing_address(boot_info_start_vaddr)..=Page::containing_address(boot_info_start_vaddr + boot_info_size - 1)).enumerate() {
```
kevinaboos: for convenience, you can use [PageRange::from_virt_addr()](https://www.theseus-os.com/Theseus/doc/memory_structs/struct.PageRange.html#method.from_virt_addr)

kernel/memory/src/paging/mod.rs
```diff
@@ -317,18 +333,20 @@ pub fn init(
...
+        let mut iter = stack_mappings.iter();
+        while let Some(Some((page, frame))) = iter.next() {
+            let allocated_page = page_allocator::allocate_pages_at(page.start_address(), 1)?;
+            let allocated_frame = frame_allocator::allocate_frames_at(frame.start_address(), 1)?;
+            let mapped_pages = mapper.map_allocated_pages_to(allocated_page, allocated_frame, data_flags)?;
+            if let Some(ref mut stack_mapped_pages) = stack_mapped_pages {
+                stack_mapped_pages.merge(mapped_pages).map_err(|_| "failed to merge stack mapped pages")?;
+            } else {
+                stack_mapped_pages = Some(mapped_pages);
+            }
```
kevinaboos, line 348: this is a new and semi-interesting use case for `MappedPages`. 

Should I perhaps add a new API to `Mapper` or change the existing one to something like
```rust
    pub fn map_allocated_pages_to(
        &mut self,
        pages: AllocatedPages,
        frames: impl IntoIter<Item = AllocatedFrames>,
        flags: impl Into<PteFlagsArch>,
    ) -> Result<MappedPages, &'static str> { ... }
```
such that the caller can provide a series of `AllocatedFrames` objects that would be mapped by the `AllocatedPages` in sequential order of iteration

(not for this PR, just a general thought)

kernel/memory_x86_64/src/lib.rs
```diff
@@ -182,14 +176,15 @@ pub fn find_section_memory_bounds(boot_info: &impl BootInformation) -> Result<(A
                 data_flags = Some(flags);
                 "nano_core .bss"
             }
-            ".page_table" => {
-                // We bootstrap the page table from the CR3 register.
+            // This appears when compiling for BIOS.
+            ".page_table" | ".stack" => {
                 continue;
```
kevinaboos, line 182: for the sake of completeness to 
```suggestion
                debug!("     no need to map this section, it is mapped separately later");
                continue;
```

kernel/memory_x86_64/src/lib.rs
```diff
@@ -182,14 +176,15 @@ pub fn find_section_memory_bounds(boot_info: &impl BootInformation) -> Result<(A
...
             }
-            ".stack" => {
-                stack_start = Some((start_virt_addr, start_phys_addr));
-                stack_end   = Some((end_virt_addr, end_phys_addr));
-                "initial stack"
+            // This appears when compiling for UEFI.
+            ".bootloader-config" => {
+                // TODO: Ideally we'd mark .bootloader-config as not allocated
+                // so the bootloader doesn't load it.
+                continue;
```
kevinaboos, line 189: Please adjust the below debug message to be better/more correct/more specific
```suggestion
                debug!("     no need to map this section, it is only used by the bootloader for config.");
                continue;
```

kevinaboos (changes requested):
Just a few small comments

kernel/memory_x86_64/src/lib.rs
tsoutsman: I think that's as correct as it can be. Only the bootloader uses it for config; hence, Theseus doesn't need it.

kernel/memory/src/paging/mod.rs
tsoutsman: Probably? I do think it's strange that we mostly only map contiguous pages to contiguous frames, despite one of the benefits of paging being that contiguous pages can be mapped to non-contiguous frames.

The frames should probably be an [`ExactSizeIterator`](https://doc.rust-lang.org/core/iter/trait.ExactSizeIterator.html) though so that we can ensure there are as many frames as pages prior to mapping anything. `ExactSizeIterator` is implemented for `array::IntoIter`.

kevinaboos: `ExactSizeIterator` would be nice to have, but I don't know if it would always be possible to have an array or vec (or whatever form of known-sized storage) in other use cases. But we could certainly specialize on that with a function that accepted an `ExactSizeIterator`.

tsoutsman: If the user doesn't know how many frames they have, how could they allocate the correct number of pages?

kevinaboos: It's a rare case for sure. It's possible that a user could know how many frames they have, but might not be able to place them all in a single array or vec. 

That doesn't really matter though, what matters is the fact that `ExactSizeIterator` technically cannot be trusted, that's the purpose of `TrustedLen`, which is an unsafe trait. We could use `TrustedLen` but then it would require a custom unsafe impl for non-array/vec cases, which again is rare an unimportant. 

Either way, in the end we have to iterate over all of the pages and frames anyway, so accepting any `Iterator` would be sufficient.

PR #748, opened 2022-12-16
Add `boot_info::uefi`

tsoutsman:


kernel/boot_info/src/uefi.rs
```diff
@@ -0,0 +1,242 @@
...
+impl crate::Module for Module {
+    fn name(&self) -> Result<&str, &'static str> {
+        Ok(info::Module::name(&self.inner))
+    }
+
+    fn start(&self) -> PhysicalAddress {
+        PhysicalAddress::new_canonical(
+            self.regions
+                .iter()
+                .find(|region| region.kind == info::MemoryRegionKind::UnknownUefi(0x80000000))
```
kevinaboos: is this our own custom definition, or a standard definition? 

Either way, it'd be best to define a const describing what this value represents rather than having a magic number here

kernel/boot_info/src/uefi.rs
```diff
@@ -0,0 +1,242 @@
...
+                .ok_or("invalid bootloader info end address")?;
+            Ok(start..end)
+        }
+    }
+
+    fn modules_memory_range(&self) -> Result<Range<PhysicalAddress>, &'static str> {
+        let area = self
+            .memory_regions
+            .iter()
+            .find(|region| region.kind == info::MemoryRegionKind::UnknownUefi(0x80000000))
```
kevinaboos: same comment here, use something like 
```rust
const UefiModuleMemoryKind: ... = info::MemoryRegionKind::UnknownUefi(0x80000000);
```
or whatever you feel is the best name for it

kernel/boot_info/src/lib.rs
```diff
@@ -7,7 +7,10 @@
 #![feature(type_alias_impl_trait)]
 #![no_std]
 
+#[cfg(feature = "multiboot2")]
 pub mod multiboot2;
+#[cfg(feature = "uefi")]
+pub mod uefi;
```
kevinaboos: not necessary, but maybe a small quality of life improvement: shall we throw a compile error if neither multiboot2 nor uefi feature is activated?

kevinaboos (changes requested):
just a few small suggestions/questions

kernel/boot_info/src/lib.rs
tsoutsman: No, because consumer crates don't depend on either feature e.g. `memory`

kevinaboos (approved):

PR #750, opened 2022-12-17
Reorganise `nano_core` assembly files

tsoutsman:
None of the content was modified, which should preserve the history of the files. I think `uefi/ap_realmode.asm` may not have its history preserved, but that's because it technically doesn't have a history despite being a copy of `bios/ap_realmode.asm`. I'll make the content changes in a different PR.


kevinaboos (approved):
Thanks, appreciate you trying to help me preserve the history. 

I'll accept this as is, but I think in the future it'd be nice to further dedup the asm code shared between bios and UEFI builds.

PR #753, opened 2022-12-17
Fix some Clippy lints

tsoutsman:
I was initially going to fix all the Clippy complexity lints, but quickly realised that would be a very big PR, so I only fixed it for some crates. In future, I'll fix lints vertically (i.e. fix one lint in all crates) rather than horizontally (i.e. fix all lints (or at least all complexity lints) in some crates).


kernel/memory/src/lib.rs
```diff
@@ -124,33 +124,34 @@ pub fn set_broadcast_tlb_shootdown_cb(func: fn(PageRange)) {
     BROADCAST_TLB_SHOOTDOWN_FUNC.call_once(|| func);
 }
 
-
+/// Information returned after initialising the memory subsystem.
+pub struct MemoryMappings {
```
kevinaboos: Thanks, this is way overdue lol. Things were out of hand.

Small request -- can we call it `InitialMemoryMappings` or something to imply bootstrap/init relevance?

kevinaboos (changes requested):
Thanks a lot! Any method of addressing clippy is fine by me.

Left one small request comment.

kevinaboos:
Note to self: remember to update #526 after this is merged

kevinaboos (approved):

PR #754, opened 2022-12-17
Deduplicate nano core assembly files

tsoutsman:


kernel/nano_core/build.rs
```diff
@@ -131,14 +131,21 @@ fn compile_asm() {
...
                 .arg("-i")
                 .arg(&include_path)
                 .arg("-o")
                 .arg(&output_path)
                 .arg(file.path())
-                .args(cflags.split(' '))
+                .args(cflags.split(' '));
+
+            if BOOT_SPECIFICATION == "bios" {
+                command.arg("-DBIOS");
```
kevinaboos: For consistency's sake, let's add this to the cflags variable above since cflags are only really used for configuring asm compilation.

kevinaboos (changes requested):
Awesome work on this, very clear & easy to read.

One minor request -- there are now a few very "long" `ifdef` blocks, which can be a bit hard to read. Can you add comments to each `endif` statement describing what `ifdef` it corresponds to? I typically see this in C/C++ header files, and it's quite an improvement to readability because it disambiguates nested `ifdef`s:
```asm
%ifdef BIOS   
    ...
    ...
%endif  ; BIOS
```

kevinaboos (approved):

PR #755, opened 2022-12-20
Switch to storing run queues in boxed array

tsoutsman:
Removes the dependency on `atomic_linked_list` which we're trying to phase out. However, the change means we must to initialise the runqueues all at once. We do this after initialising the APs and so we can't add the APs' bootstrap tasks to the runqueue as it doesn't yet exist.

I'm not sure bootstrap tasks are strictly necessary any more, but I'll leave that for another PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -376,6 +376,7 @@ dependencies = [
  "ota_update_client",
  "page_attribute_table",
  "print",
+ "runqueue",
  "scheduler",
  "simd_personality",
  "spawn",
@@ -2888,10 +2889,10 @@ dependencies = [
 name = "runqueue_priority"
 version = "0.1.0"
 dependencies = [
- "atomic_linked_list",
  "log",
  "mutex_preemption",
  "single_simd_task_optimization",
+ "spin 0.9.4",
  "task",
 ]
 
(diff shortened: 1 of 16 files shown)


tsoutsman:
Using the new benchmark in #759 shows this change is a regression. I have no clue why performing multiple pointer dereferences to traverse the `AtomicMap` is faster than a single index into the `Box<[_]>`. Using `get_unchecked` to remove the performance penalty of the `Once` didn't have a noticeable impact.

I'll wait for #759 and #760 to be merged in, and I'll probably split this PR into two parts: modifying the run queue initialisation and then modifying the run queue data structure.

kevinaboos:
Beyond the regression in perf, which I can overlook for a cleaner, safer design, one flaw in this design choice is that we cannot dynamically bring up or shut down CPUs easily if the set of runqueues is fixed after first initialization. 

Another consideration here is that a CPU cannot begin adding tasks to its runqueue (and thus, actually scheduling them in and executing them) until all CPUs are booted; that can be a significant (unnecessary) delay especially on a manycore system.

kevinaboos:
Anyway, I think we need to keep runqueue creation/removal dynamic, so I'm electing to close this.

PR #756, opened 2022-12-20
Add unbounded (growable) `mpmc_queue` library

tsoutsman:
Unlike `mpmc`, this queue is unbounded. An unbounded `mpmc` queue is commonly used as a global fallback runqueue in schedulers. This PR is the first step in improving the round robin scheduler.

### DIFF
--- a/libs/mpmc_queue/src/lib.rs
+++ b/libs/mpmc_queue/src/lib.rs
@@ -24,7 +24,7 @@ pub struct Queue<T> {
 }
 
 struct Pointers<T> {
-    head: Option<Box<Node<T>>>,
+    head: Option<NonNull<Node<T>>>,
     tail: Option<NonNull<Node<T>>>,
 }
 
@@ -32,7 +32,7 @@ unsafe impl<T> Send for Pointers<T> {}
 
 struct Node<T> {
     item: T,
-    next: Option<Box<Node<T>>>,
+    next: Option<NonNull<Node<T>>>,
 }
 
 impl<T> Node<T> {
@@ -64,8 +64,8 @@ impl<T> Queue<T> {
 
     /// Appends an item to the queue.
     pub fn push(&self, item: T) {
-        let (node, node_pointer) = box_pointer(Node::new(item));
-        unsafe { self.push_inner(node, node_pointer, 1) };
+        let node = box_pointer(Node::new(item));
+        unsafe { self.push_inner(node, node, 1) };
     }
 
     /// Appends several items to the queue.
@@ -73,21 +73,21 @@ impl<T> Queue<T> {
     where
         I: Iterator<Item = T>,
     {
-        let (first, first_pointer) = match iterator.next() {
+        let first = match iterator.next() {
             Some(item) => box_pointer(Node::new(item)),
             None => return,
         };
 
-        let mut tail = first_pointer;
+        let mut tail = first;
         let mut len = 1;
 
         for next in iterator {
-            let (next, next_pointer) = box_pointer(Node::new(next));
+            let next = box_pointer(Node::new(next));
             // SAEFTY: The only other pointer to the tail is stored in the second-to-last
             // node. We own first, hence we own all the nodes, hence that reference is not
             // being used.
             unsafe { tail.as_mut() }.next = Some(next);
-            tail = next_pointer;
+            tail = next;
             len += 1;
         }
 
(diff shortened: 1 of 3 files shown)


tsoutsman:
Was watching <https://www.youtube.com/watch?v=EY7Wi9fV5bk> in my spare time, and turns out it's directly relevant to the PR I just submitted :)

Miri was able to detect this so I guess there's another tool I should learn:
```
running 2 tests
test tests::test_mpmc_stress ... error: Undefined Behavior: trying to retag from <415901> for Unique permission at alloc150534[0x0], but that tag does not exist in the borrow stack for this location
    --> /home/klim/.rustup/toolchains/nightly-2022-11-03-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:2297:18
     |
2297 |             Some(v) => ControlFlow::Continue(v),
     |                  ^
     |                  |
     |                  trying to retag from <415901> for Unique permission at alloc150534[0x0], but that tag does not exist in the borrow stack for this location
     |                  this error occurs as part of retag at alloc150534[0x0..0x10]
     |
     = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
     = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
```

libs/mpmc_queue/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+        Some(unsafe { Box::from_raw(current_head.as_ptr()) }.item)
+    }
+}
+
+/// Boxes the item and returns a non-null pointer to the box.
+fn box_pointer<T>(item: T) -> NonNull<T> {
+    let item = Box::new(item);
+    let item_pointer = Box::into_raw(item);
+    // SAFETY: We just constructed the box.
+    unsafe { NonNull::new_unchecked(item_pointer) }
```
kevinaboos: I thought this pattern seemed unnecessarily unsafe, so I skimmed a few PRs on the rust-lang/rust repo and discovered that there used to be a `Box::into_raw_non_null()`. But it was never stabilized, and then deprecated and removed. The consensus was that the proper way to do this is:
```rust
    NonNull::from(Box::leak(item))
```

libs/mpmc_queue/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+    ///
+    /// # Safety
+    ///
+    /// `head` must be the start of the batch, and `tail` must point to the end
+    /// of the batch. The batch must be `len` nodes long.
+    unsafe fn push_inner(&self, head: NonNull<Node<T>>, tail: NonNull<Node<T>>, len: usize) {
+        let mut pointers = self.pointers.lock();
+
+        if let Some(mut tail_pointer) = pointers.tail {
+            // SAEFTY: The only other pointer to the tail is stored in the second-to-last
```
kevinaboos:
```suggestion
            // SAFETY: The only other pointer to the tail is stored in the second-to-last
```

libs/mpmc_queue/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+            Some(item) => box_pointer(Node::new(item)),
+            None => return,
+        };
+
+        let mut tail = first;
+        let mut len = 1;
+
+        for next in iterator {
+            let next = box_pointer(Node::new(next));
+            // SAEFTY: The only other pointer to the tail is stored in the second-to-last
```
kevinaboos:
```suggestion
            // SAFETY: The only other pointer to the tail is stored in the second-to-last
```

libs/mpmc_queue/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+        // SAFETY: current_head is a valid pointer.
+        pointers.head = unsafe { current_head.as_ref() }.next;
+
+        // If we are the last node in the list:
+        if pointers.head.is_none() {
+            pointers.tail = None;
+        }
+
+        self.len.fetch_sub(1, Ordering::Release);
+        // SAFETY: current_head is a valid pointer.
```
kevinaboos: i think `Box::from_raw()` also requires that the pointer was obtained from a `Box` originally, too. Obviously you ensure that via usage of `box_pointer()`, but I still think it merits an explicit comment here.

kevinaboos: specific details here: https://doc.rust-lang.org/std/boxed/index.html#memory-layout

so it's not that it had to come from a `Box`, but the fact that it came from a `Box` does satisfy the `Layout`-related invariants.

kevinaboos (changes requested):
functionally it looks good. Left a few minor comments.

kevinaboos (approved):

PR #757, opened 2022-12-20
Switch to relaxed ordering for the task ID counter

tsoutsman:
Each task ID is guaranteed to see a unique value.

<https://en.cppreference.com/w/cpp/atomic/memory_order#Relaxed_ordering>
<https://www.reddit.com/r/rust/comments/aeii1i/is_memory_orderingrelaxed_sufficient_to_generate/>


kevinaboos (approved):
Thanks!

Yeah there are definitely some other places where we use sequential consistency when it's not required. In my newer code I've been careful to select the best consistency model, but the older code doesn't since it's never incorrect (just less performant) to use sequential consistency.

PR #760, opened 2022-12-21
Add run queue access benchmark

tsoutsman:
Depends on #739.

I disable interrupts for the duration of the benchmark to minimise external impacts. It's a blunt solution, but I think it's ok since a benchmark wouldn't be run in regular Theseus use.


applications/rq_access_eval/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "rq_access_eval"
+version = "0.1.0"
+authors = ["Klim Tsoutsman <klim@tsoutsman.com>"]
+description = "Run queue access benchmark"
+edition = "2021"
+
+[dependencies]
+app_io = { path = "../../kernel/app_io" }
+apic = { path = "../../kernel/apic" }
```
kevinaboos: this has now been replaced with the `cpu` crate, so we should use that instead to be arch-agnostic

kevinaboos (changes requested):
all good! One request: `apic` --> `cpu`

kevinaboos (approved):

PR #761, opened 2022-12-21
Refactor `ExitValue` out of the `Task` struct

kevinaboos:
* Refactored task management functions into two new wrapper types:
  * `ExitableTaskRef`, a new wrapper around `TaskRef`, allows the task
    to be marked as exited or killed.
    * This allows Theseus to restrict who can mark a task as exited
      or killed -- only itself. More types = more better.
    * This is now used in (and required by) all task lifecycle functions.
  * `JoinableTaskRef` is augmented, but still primarily exists to 
    allow the caller of `spawn()`, i.e., the "parent" that spawned a new task,
    to monitor and retrieve the exit status of that task.

* Removed the `ExitValue` from the `Task` struct; it is now only
  transferable directly via the `TaskRef` type that shares a
  `Mutex<Option<ExitValue>>` with the `JoinableTaskRef`.
  * `ExitValue` is still an erased type (`Box<Any>`); that will be changed
    to a generic type argument in a future PR.


kevinaboos:
@tsoutsman requesting your review since you have been working on some projects related to tasking.

Currently the only thing I don't like is that we still need a hidden "escape hatch" for obtaining an `ExitableTaskRef` in the unwinder. See `ExitableTaskRef::obtain_for_unwinder()`.

Also I don't love the name `ExitableTaskRef`, if you have any ideas I'm all ears.  I thought about:
* `FirstTaskRef`, since it's the first reference to the current task obtained from the context of the current task
* `MyTaskRef`, since it's guaranteed to always point to the current task and is not `Send`

but I don't love any of those. Basically, I want the name to convey that this is the only type that allows one to mark a task as exited, and that it only works for yourself. It doesn't matter _that_ much since it's an internal type that isn't available for use outside of `task` & `spawn`, but I still think I can do better.

kernel/task/src/lib.rs
```diff
@@ -1181,88 +1159,152 @@ impl Drop for JoinableTaskRef {
...
+/// Due to the above drop-based behavior, this type does not implement `Clone`
+/// because it assumes there is only ever one `ExitableTaskRef` per task.
+///
+/// However, this type auto-derefs into an inner [`TaskRef`],
+/// which *can* be cloned, so you can easily call `.clone()` on it.
+pub struct ExitableTaskRef {
+    task: TaskRef,
+}
+// Ensure this cannot be moved (sent) to a different task.
+impl !Send for ExitableTaskRef { }
```
tsoutsman:
```suggestion
// Ensure that another thread cannot mark the task as exited.
impl !Send for ExitableTaskRef { }
impl !Sync for ExitableTaskRef { }
```
The `ExitableTaskRef`'s methods can be called on an `&ExitableTaskRef`, so sending references (`Sync`) is also not ok.

kernel/task/src/lib.rs
```diff
@@ -1181,88 +1159,152 @@ impl Drop for JoinableTaskRef {
...
-            Err("`mark_as_exited()` can only be invoked on the current task, not on another task.")
+        self.internal_exit(ExitValue::Killed(reason))
+    }
+
+    /// Reaps this task (if orphaned) by taking and dropping its exit value and removing it
+    /// from the system task list.
+    ///
+    /// If this task has *not* been orphaned, meaning it is still joinable,
+    /// then this function does nothing.
+    pub fn reap_if_orphaned(&self) {
```
tsoutsman, line 1283: This should technically be implemented for `Task`, not `ExitableTaskRef` since any task should be able to reap any other task.

kernel/task/src/lib.rs
```diff
@@ -1181,88 +1159,152 @@ impl Drop for JoinableTaskRef {
...
-        JoinableTaskRef { task: taskref }
+/// A wrapper around `TaskRef` that allows this task to mark itself as exited.
+///
+/// This is only obtainable when a task is first switched to, specifically while
+/// it is executing the `spawn::task_wrapper()` function
+/// (before it proceeds to running its actual entry function).
+///
+/// ## Not `Clone`-able
+/// Due to the above drop-based behavior, this type does not implement `Clone`
+/// because it assumes there is only ever one `ExitableTaskRef` per task.
```
tsoutsman: The copied docs from `JoinableTaskRef` are incorrect.

kernel/task/src/lib.rs
```diff
@@ -1181,88 +1159,152 @@ impl Drop for JoinableTaskRef {
...
-        // Mark this task as joinable, now that it has been wrapped in the proper type.
-        taskref.joinable.store(true, Ordering::Relaxed);
-        JoinableTaskRef { task: taskref }
+/// A wrapper around `TaskRef` that allows this task to mark itself as exited.
+///
+/// This is only obtainable when a task is first switched to, specifically while
+/// it is executing the `spawn::task_wrapper()` function
+/// (before it proceeds to running its actual entry function).
+///
+/// ## Not `Clone`-able
```
tsoutsman, line 1214: I think `ExitableTaskRef` can implement `Clone` and still be correct.

kernel/task/src/lib.rs
```diff
@@ -1181,88 +1159,152 @@ impl Drop for JoinableTaskRef {
...
+    pub fn post_context_switch_action(&self) -> PreemptionGuard {
+        self.task.post_context_switch_action()
+    }
+
+    /// Allows the unwinder to obtain an `ExitableTaskRef` in order for it to
+    /// be able to invoke this task's [`FailureCleanupFunction`].
+    #[doc(hidden)]
+    pub fn obtain_for_unwinder(current_task: TaskRef) -> Self {
+        Self { task: current_task }
+    }
```
tsoutsman, line 1311: This should use `get_my_current_task` (or compare `current_task` with `get_my_current_task`) to ensure that we are constructing an `ExitableTaskRef` for the current task.

kernel/task/src/lib.rs
```diff
@@ -1181,88 +1159,152 @@ impl Drop for JoinableTaskRef {
...
+impl fmt::Debug for TaskRef {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("TaskRef")
+            .field("task", &self.0.task)
+            .finish_non_exhaustive()
+    }
+}
+struct TaskRefInner {
+    task: Task,
+    exit_value_mailbox: Mutex<Option<ExitValue>>,
```
tsoutsman, line 1335: Why is moving the exit value out `Task` beneficial?

tsoutsman (changes requested):
Just some small questions and suggestions.

> Also I don't love the name ExitableTaskRef

I like it. It's descriptive and fits in with the `JoinableTaskRef` naming convention. I don't like `FirstTaskRef` or `MyTaskRef` because they don't describe what you can do with this special task reference. Also, the name `MyTaskRef` implies that `get_my_current_task` should return a `MyTaskRef`, which could lead to confusion.

kernel/task/src/lib.rs
kevinaboos: 1. I eventually want to make it statically typed (with generic type params), but I don't want to have to parameterize the `Task` struct
2. Design-wise, I dislike huge monolithic task structs
3. The smaller the `Task` struct, the cheaper it becomes to allocate and keep a million of em in memory

I realize this doesn't truly move it out of `Task` since it's still within `TaskRef`, but it's a partial step for now.

kernel/task/src/lib.rs
kevinaboos: Agreed, good call.

kernel/task/src/lib.rs
kevinaboos: Hmm, I don't necessarily agree with that. The point of the new `ExitableTaskRef` type is to ensure that only a task itself can clean itself up in certain ways (e.g., marking itself as exited to indicate a regular return). I'd argue that if the task has been orphaned (it's `JoinableTaskRef` has been dropped elsewhere) then it should be the one to clean itself up, not any other random task.

kevinaboos: If we have a future use case that requires a random foreign task to reap another random task, then we can move it.

kernel/task/src/lib.rs
kevinaboos: Hmm interesting thought. Adding `!Sync` seems good, though I know this does mess with the way that bootstrap tasks are handled.

The main thing I wanted to convey here is that a task should only be able to acquire its own `ExitableTaskRef` instance, i.e., `ExitableTaskRef` should _not_ be passed around to another task.

This negative trait impl isn't actually necessary, it's just a nice thing to enforce. `ExitableTaskRef` is practically an internal implementation detail within `task` and `spawn`, so it's not even required to have this.

kevinaboos:
> I like it. It's descriptive and fits in with the `JoinableTaskRef` naming convention. 

Glad to hear that.

> I don't like `FirstTaskRef` or `MyTaskRef` because they don't describe what you can do with this special task reference. Also, the name `MyTaskRef` implies that `get_my_current_task` should return a `MyTaskRef`, which could lead to confusion.

That's true, but I do plan to change `get_my_current_task()` to just `current_task()`, so that might be less of an issue. But we can keep the name for now.

kernel/task/src/lib.rs
kevinaboos: sure, that may be true. As I said in the other convo, it's an internal type so it doesn't really matter. 

I guess the question is under what scenario would we need or want to clone an `ExitableTaskRef`?

kernel/task/src/lib.rs
kevinaboos: they seem correct? It doesn't implement `Clone`.

I just didn't add a static assertion that prevents implementing `Clone`, but will do that now.

kevinaboos: ah wait, you were talking about the "drop-based behavior" part. Thanks, fixed.

kevinaboos:
I've responded to all the comments, thanks for leaving feedback. Feel free to re-respond or mark anything that you consider resolved as such.

kevinaboos:
I made some minor changes to `hull` and `dreadnought` in https://github.com/theseus-os/Theseus/pull/761/commits/64b257e55e2b46201d6858bf5d4e217576653646, please review

kernel/task/src/lib.rs
tsoutsman: 1. Does that mean all task refs would have to have generics? I don't see how that's any better than `Task` having generics, as pretty much all crates use `TaskRef`s rather than `Task`s. A compromise would be only to have `JoinableTaskRef` parameterised like in `dreadnought`.
2. As I said above, `TaskRef`s are pretty much synonymous with `Task`s.
3.
```rust
pub static TASKLIST: MutexIrqSafe<BTreeMap<usize, TaskRef>> = MutexIrqSafe::new(BTreeMap::new());
```
We allocate a `TaskRef`, not `Task`, so moving fields into `TaskRef` doesn't change anything regarding allocation cost.

It just feels weird for some fields to be stored in the `Task` and others to be stored in the `TaskRef`. The only difference with `exit_value` is that `spawn` does not modify it during initialisation e.g. `joinable` could also be moved into the `TaskRef`.

kernel/dreadnought/src/task.rs
```diff
@@ -91,6 +91,7 @@ pub type Result<T> = core::result::Result<T, Error>;
 pub enum Error {
     Cancelled,
     Panic(PanicInfoOwned),
-    Reaped,
+    /// A `JoinError` should not occur; this indicates a BUG in Theseus's task mgmt.
+    JoinError(&'static str),
```
tsoutsman: Should probably be `Join`, not `JoinError`.

<https://rust-lang.github.io/rust-clippy/master/#enum_variant_names>

kernel/task/src/lib.rs
kevinaboos: As you said, we could parameterize only `Task` and not `TaskRef`, but I'm not 100% settled on a specific design. If you notice in the current main branch, there is a commented-out type called `TaskJoiner` which was my original idea to parameterize a task/join-related type with those generic types. The key design point is that (at least currently) only the `spawn`-related functions know the generic types `F`, `A`, and `R` related to a task's entry point function, and I don't currently see a reason to push that knowledge into `task`.

>  It just feels weird for some fields to be stored in the Task and others to be stored in the TaskRef. The only difference with exit_value is that spawn does not modify it during initialisation e.g. joinable could also be moved into the TaskRef.

Agreed, `joinable` will be removed entirely soon since there are other ways to realize that feature.

kernel/dreadnought/src/task.rs
kevinaboos: good point.

kernel/task/src/lib.rs
kevinaboos: Another guiding design principle here is the principle of least privilege. Essentially, I don't want any old `TaskRef` to have the privilege (ability) to access or modify an exit value; the easiest way to achieve this is to move it out of the `TaskRef` struct.
As I said before, while this PR alone doesn't achieve that, it's a stepping stone -- at least `Task` cannot access it. Eventually it will be entirely removed from `TaskRef` too.

PR #763, opened 2022-12-23
Check if task is suspended in `Task::is_runnable`

tsoutsman:


kevinaboos (approved):
Thanks!

PR #764, opened 2022-12-23
Add support for UEFI

tsoutsman:
Mostly changes to the build.

The only substantial addition to Theseus code is `nano_core/src/ uefi.rs`.

### DIFF
--- a/kernel/nano_core/src/uefi.rs
+++ b/kernel/nano_core/src/uefi.rs
@@ -68,7 +68,7 @@ pub extern "C" fn _error() {
             // See <https://doc.rust-lang.org/nightly/rust-by-example/unsafe/asm.html#labels>
             "2:",
             "hlt",
-            "jmp 2b",
+            "jmp 2b", // jump backwards ("b") to label "2:".
             options(noreturn)
         )
     }
--- a/kernel/nano_core/src/uefi.rs
+++ b/kernel/nano_core/src/uefi.rs
@@ -1,0 +1,67 @@
+use crate::{early_setup, nano_core, try_exit};
+use boot_info::uefi::STACK_SIZE;
+use bootloader_api::{config::Mapping, BootloaderConfig};
+use core::arch::asm;
+use memory::VirtualAddress;
+
+#[used]
+#[link_section = ".bootloader-config"]
+pub static __BOOTLOADER_CONFIG: [u8; BootloaderConfig::SERIALIZED_LEN] = {
+    let mut config = BootloaderConfig::new_default();
+    config.mappings.physical_memory = Some(Mapping::Dynamic);
+    config.mappings.page_table_recursive =
+        Some(Mapping::FixedAddress(0o177777_776_000_000_000_0000));
+    config.kernel_stack_size = STACK_SIZE as u64;
+    config.serialize()
+};
+
+#[naked]
+#[no_mangle]
+#[link_section = ".init.text"]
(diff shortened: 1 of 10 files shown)


kernel/nano_core/src/uefi.rs
```diff
@@ -0,0 +1,67 @@
...
+    config.mappings.page_table_recursive =
+        Some(Mapping::FixedAddress(0o177777_776_000_000_000_0000));
+    config.kernel_stack_size = STACK_SIZE as u64;
+    config.serialize()
+};
+
+#[naked]
+#[no_mangle]
+#[link_section = ".init.text"]
+pub unsafe extern "C" fn _start(boot_info: &'static mut bootloader_api::BootInfo) {
```
kevinaboos: any particular reason why this is static mut? 
Also, how is it a rust-level reference and not a raw pointer if this is the first time we're entering Rust code? I understand they are interchangeable but it still conveys a misleading calling context.

kernel/nano_core/src/uefi.rs
```diff
@@ -0,0 +1,67 @@
...
+        // page.
+        "sub rsp, 4096",
+        "call rust_entry",
+        "jmp KEXIT",
+        options(noreturn),
+    );
+}
+
+#[no_mangle]
+unsafe extern "C" fn rust_entry(
```
kevinaboos: please separate this into an unsafe function (which will deref the boot into pointer) and a safe function (called by the unsafe one) that calls the `nano_core` fn

kernel/nano_core/src/uefi.rs
```diff
@@ -0,0 +1,67 @@
...
+    try_exit!(nano_core(
+        boot_info as &'static bootloader_api::BootInfo,
+        kernel_stack_start,
+    ));
+}
+
+#[naked]
+#[no_mangle]
+#[link_section = ".init.text"]
+pub unsafe extern "C" fn _error() {
```
kevinaboos: technically this function is not unsafe, only the interior asm block is.

kevinaboos (changes requested):
looks good, left a few small comments

kernel/nano_core/src/uefi.rs
tsoutsman: > any particular reason why this is static mut?

Changed

> Also, how is it a rust-level reference and not a raw pointer if this is the first time we're entering Rust code?

The bootloader runs rust code and directly calls `_start` with a Rust reference.

From <https://github.com/rust-osdev/bootloader>:
> Define an entry point function with the signature `fn kernel_main(boot_info: &'static mut bootloader_api::BootInfo) -> !`

kernel/nano_core/src/uefi.rs
tsoutsman: As I mentioned above, we don't need to use raw pointers. The function is still unsafe, though as `double_fault_stack` has to be pointing to the end of a mapped page.

I've added the `unsafe_op_in_unsafe_fn` lint to the `nano_core` crate, which means that the unsafe operations in an unsafe function are clearly marked. Is the lint good enough?

kernel/nano_core/src/uefi.rs
kevinaboos: ok i see, didn't realize we made the transition from Rust -> asm -> Rust again. Works for me.

kevinaboos (approved):

PR #765, opened 2022-12-23
Encapsulate ps2 functions

hecatia-elegua:
The only thing still unclear is how to pass the structs to their interrupt handlers.

i.e. https://github.com/theseus-os/Theseus/blob/639e30cc1e0c92e8dcaa33838d508705edb74a25/kernel/keyboard/src/lib.rs#L53 doesn't work (of course), but I don't know how to solve it. I tried changing the argument to `keyboard::init` to `&'static PS2Keyboard`, which wants the controller to be static as well, but I can't pass the controller out of `ps2_controller::init()` as static.
I'm again probably missing something small.


tsoutsman:
> The only thing still unclear is how to pass the structs to their interrupt handlers.

You are right; the only way to pass the struct to the interrupt handler is through a static.

> I tried changing the argument to keyboard::init to &'static PS2Keyboard, which wants the controller to be static as well, but I can't pass the controller out of ps2_controller::init() as static.

If you look at the definition for [`PS2Keyboard`](https://github.com/theseus-os/Theseus/blob/639e30cc1e0c92e8dcaa33838d508705edb74a25/kernel/ps2/src/lib.rs#L399-L401):
```rust
pub struct PS2Keyboard<'c> {
    controller: &'c PS2Controller,
}
```
You'll see it's generic over a lifetime. This means that it can only live for as long as `'c`. If you were to expand the lifetime elisions in the signature of `keyboard::init`, it would look like this:
```rust
pub fn init<'a>(keyboard: PS2Keyboard<'a>, keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
    //
}
```
Now that means our `Once<PS2Keyboard<'a>>` can only live for `'a`. But our `Once<PS2Keyboard<'a>>` is stored in a static, so it must live for `'static`, and hence the `T` returned by `Once::call_once` must be `'static`.  Now, the error should make sense:
```
error: lifetime may not live long enough
  --> kernel/keyboard/src/lib.rs:53:27
   |
31 | pub fn init(keyboard: PS2Keyboard, keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
   |             -------- has type `PS2Keyboard<'a>`
...
53 |     KEYBOARD.call_once(|| keyboard);
   |                           ^^^^^^^^ returning this value requires that `'a` must outlive `'static`

error: could not compile `keyboard` due to previous error
```
Rust expects the closure to return something with a static lifetime, but it returns something with a potentially shorter lifetime.

### Solution

Looking at the types reveals the solution:
`Once<PS2Keyboard<'a>>: 'static` -> `PS2Keyboard<'a>: 'static` -> `'a: 'static`

`keyboard::init`'s signature should look like so:
```rust
pub fn init(keyboard: PS2Keyboard<'static>, keyboard_queue_producer: Queue<Event>) -> Result<(), &'static str> {
    //
}
```

All this means the `PS2Keyboard` needs a static reference to the `PS2Controller`, which means the `PS2Controller` will also need to be stored in a static. After all, `PS2Keyboard` is just a wrapper around a `PS2Controller`, so you are indirectly passing the `PS2Controller` to the interrupt handler.

As a side note, it's probably best to name the lifetimes `'a` rather than `'c`.

hecatia-elegua:
@tsoutsman Thank you a lot!
Above I said the wrong thing. I tried `PS2Keyboard<'static>`, not `&'static PS2Keyboard`, oops. Maybe I did both at some point.
Still, that explained a bit more of what I was doing and helped my confidence.

Side note: I named them `'c` to point to it being the _controller's_ lifetime. Have seen some people put whole words in lifetimes, some people use a relevant first letter... I guess it's the same as with generics.

Now this works and all PS/2 crates save their respective struct in a static `Once<...>`. Would you want to review?

kernel/ps2_controller/src/lib.rs
```diff
@@ -23,84 +24,65 @@ pub fn init() -> Result<(), &'static str> {
...
-    reset_mouse()?;
+    controller.keyboard_handle().reset()?;
+    controller.mouse_handle().reset()?;
 
-    // for debugging purposes
-    debug!("Final PS/2 {:?}", read_config());
-    Ok(())
+    debug!("Final PS/2 {:?}", controller.read_config());
+    
+    static CONTROLLER: Once<PS2Controller> = Once::new();
```
tsoutsman: Could you please move the static into the crate's root (like in `keyboard`)?

kernel/keyboard/src/lib.rs
```diff
@@ -20,17 +20,18 @@ const PS2_KEYBOARD_IRQ: u8 = interrupts::IRQ_BASE_OFFSET + 0x1;
...
 static mut KBD_MODIFIERS: Lazy<KeyboardModifiers> = Lazy::new(KeyboardModifiers::new);
 
+static KEYBOARD: Once<PS2Keyboard> = Once::new();
 static KEYBOARD_PRODUCER: Once<Queue<Event>> = Once::new();
 
 /// Initialize the PS/2 keyboard driver and register its interrupt handler.
 /// 
 /// ## Arguments
 /// * `keyboard_queue_producer`: the queue onto which the keyboard interrupt handler
 ///    will push new keyboard events when a key action occurs.
```
tsoutsman: Could you please update the docs?

kernel/keyboard/src/lib.rs
```diff
@@ -20,17 +20,18 @@ const PS2_KEYBOARD_IRQ: u8 = interrupts::IRQ_BASE_OFFSET + 0x1;
 // TODO: avoid unsafe static mut
 static mut KBD_MODIFIERS: Lazy<KeyboardModifiers> = Lazy::new(KeyboardModifiers::new);
 
+static KEYBOARD: Once<PS2Keyboard> = Once::new();
 static KEYBOARD_PRODUCER: Once<Queue<Event>> = Once::new();
```
tsoutsman: Could you please define a new struct `Keyboard` like so:
```rust
struct Keyboard {
    inner: PS2Keyboard,
    producer: Queue<Event>,
}
```
And then have a singular static? Generally speaking, the more statics, the worse because it can introduce race conditions, for example, if we were to add the ability to change keyboards while the operating system is running.

kernel/mouse/src/lib.rs
```diff
@@ -9,30 +9,30 @@ use mpmc::Queue;
...
 use mouse_data::{MouseButtons, MouseEvent, MouseMovementRelative};
-use ps2::{mouse_id, set_mouse_id, read_mouse_packet, MouseId, MousePacket, status_register};
+use ps2::{PS2Mouse, MouseId, MousePacket};
 
 /// The first PS/2 port for the mouse is connected directly to IRQ 0xC.
 /// Because we perform the typical PIC remapping, the remapped IRQ vector number is 0x2C.
 const PS2_MOUSE_IRQ: u8 = interrupts::IRQ_BASE_OFFSET + 0xC;
 
+static MOUSE: Once<PS2Mouse> = Once::new();
 static MOUSE_PRODUCER: Once<Queue<Event>> = Once::new();
```
tsoutsman: Similar thing to `keyboard`. Could you consolidate the statics, please?

kernel/ps2/src/lib.rs
```diff
@@ -105,18 +106,380 @@ pub enum HostToControllerCommand {
...
+impl PS2Controller {
+    /// # Safety
+    /// 
+    /// Should only be called exactly once.
+    pub const unsafe fn new() -> Self {
+        Self {
+            data_port: Mutex::new(Port::new(PS2_DATA_PORT)),
+            command_and_status_port: Mutex::new(Port::new(PS2_COMMAND_AND_STATUS_PORT))
+        }
+    }
```
tsoutsman, line 127: We can have a public static like so:
```rust
static PS2_CONTROLLER: PS2Controller = PS2Controller {
    data_port: Mutex::new(0x60),
    command_and_status_port: Mutex::new(0x64),
};
```
Then we can remove `PS2Controller::new` so that there are no concerns about it being called multiple times.

kernel/ps2/src/lib.rs
```diff
@@ -105,18 +106,380 @@ pub enum HostToControllerCommand {
...
+
+    /// Get a handle to the keyboard.
+    /// 
+    /// The keyboard uses only [PS2_DATA_PORT].
```
tsoutsman: This is a private const, so it won't show up in the docs.

```suggestion
    /// The keyboard only uses the data port.
```

kernel/ps2/src/lib.rs
```diff
@@ -105,18 +106,380 @@ pub enum HostToControllerCommand {
...
+    pub fn keyboard_handle(&self) -> PS2Keyboard {
+        PS2Keyboard::new(self)
+    }
+
+    /// Get a handle to the keyboard.
+    /// 
+    /// The mouse uses both [PS2_DATA_PORT] and [PS2_COMMAND_AND_STATUS_PORT].
```
tsoutsman: This is a private const, so it won't show up in the docs.

```suggestion
    /// The mouse uses both the data, and command and status ports.
```

kernel/mouse/src/lib.rs
```diff
@@ -9,30 +9,30 @@ use mpmc::Queue;
...
-static MOUSE_ID: Once<MouseId> = Once::new();
 
 /// Initialize the PS/2 mouse driver and register its interrupt handler.
 /// 
 /// ## Arguments
 /// * `mouse_queue_producer`: the queue onto which the mouse interrupt handler
 ///    will push new mouse events when a mouse action occurs.
```
tsoutsman, line 30: Could you please update the docs?

tsoutsman (changes requested):
Looks good! Just a couple of nitpicks.

kernel/ps2_controller/src/lib.rs
hecatia-elegua: I would like to put it into this crate's root, but then your other suggestion in ps2 would either mean having two statics (one in ps2 and one here), or removing the one in ps2 but still using `PS2Controller::new`.

tsoutsman: You're right. Since we're performing initialisation, we shouldn't have a default static. Can you just implement this suggestion then?

hecatia-elegua:
Done. Thank you for the review @tsoutsman!

tsoutsman (approved):

tsoutsman:
Looks good. Thank you.

PR #766, opened 2022-12-23
Add task cancellation test

tsoutsman:
It doesn't currently pass, but should in the near future.

### DIFF
--- a/applications/test_task_cancel/src/lib.rs
+++ b/applications/test_task_cancel/src/lib.rs
@@ -5,12 +5,12 @@
 extern crate alloc;
 
 use alloc::{string::String, sync::Arc, vec::Vec};
-use core::sync::atomic::{AtomicUsize, Ordering};
+use core::sync::atomic::{AtomicBool, Ordering::Relaxed};
 use spin::Mutex;
 
 pub fn main(_: Vec<String>) -> isize {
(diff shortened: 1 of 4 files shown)


applications/test_task_cancel/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        unwind_row_generator();
+    }
+}
+
+#[inline(never)]
+fn unwind_row_generator() {
+    static __COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+    // Prevents unwind_row_generator from being optimised away.
+    __COUNTER.fetch_add(1, Ordering::Relaxed);
```
kevinaboos: perhaps `black_box()` would work here too?

kevinaboos (approved):
Great idea to add this test! I left one small comment but am still approving it; will merge in after discussion.

applications/test_task_cancel/src/lib.rs
tsoutsman: https://doc.rust-lang.org/beta/core/hint/fn.black_box.html
> Programs cannot rely on black_box for correctness in any way.

tsoutsman: Either way, we don't need it, as the `FALSE` static prevents the function from being optimised away.

PR #767, opened 2022-12-23
Synchronize memory when spawning and joining tasks

tsoutsman:
This is necessary for the following to always work:
```rust
static COUNTER: AtomicUsize = AtomicUsize::new(0);
COUNTER.fetch_add(1, Ordering::Relaxed);
spawn(|| {
    let value = COUNTER.load(1, Ordering::Relaxed);
    assert_eq!(value, 1);
    COUNTER.store(2, Ordering::Relaxed);
}).join();
assert_eq!(COUNTER.load(Ordering::Relaxed), 2);
```

As a side note, if the spawned closure used `fetch_add`, the assert
inside the closure would always succeed even without this PR, as read-
modify-write are [always guaranteed to read the last value][1].

The fence in the task cleanup is only necessary if the task is being
joined on but shrug.

TBH, I'm not sure about the fence orderings, so I've made them all
`SeqCst`.

A bunch of related links:
<https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_12>:
> The following functions synchronize memory with respect to other
> threads:
> pthread_create()
> pthread_join()

<https://en.cppreference.com/w/cpp/thread/thread/thread>
> The completion of the invocation of the constructor synchronizes-with
> (as defined in std::memory_order) the beginning of the invocation of
> the copy of f on the new thread of execution.

<https://en.cppreference.com/w/cpp/thread/thread/join>
> The completion of the thread identified by *this synchronizes with the
> corresponding successful return from join().

<https://www.modernescpp.com/index.php/relaxed-semantic>
> Thread creation is one synchronisation point. The other
> synchronisation point is the t.join() call in line 24. 

<https://comp.programming.threads.narkive.com/qS8xuYea/memory-barrier-implicit-in-pthread-create>
> According to the unwritten rules, all memory actions visible before
> pthread_create are visible to the created thread. And all memory
> actions visible before pthread_exit are visible to the thread that
> executed the pthread_join.

<https://stackoverflow.com/questions/9224542/is-a-memory-barrier-required-if-a-second-thread-waits-for-termination-of-the-fir>
> the API used to wait for thread termination would need to use memory
> barriers for its own purposes

<https://stackoverflow.com/questions/22347856/pthread-create3-and-memory-synchronization-guarantee-in-smp-architectures>
>> is there implied memory barrier before the thread starts running the
>> thread function so that it unfailingly sees the memory modifications
>> synchronized by pthread_create()?
>
> Yes.

[1]: https://eel.is/c++draft/atomics#order-10

### DIFF
--- a/kernel/spawn/src/lib.rs
+++ b/kernel/spawn/src/lib.rs
@@ -16,7 +16,7 @@
 #[macro_use] extern crate alloc;
 #[macro_use] extern crate log;
 
-use core::{marker::PhantomData, mem, ops::Deref};
+use core::{marker::PhantomData, mem, ops::Deref, sync::atomic::{fence, Ordering}};
 use alloc::{
     boxed::Box,
     string::{String, ToString},
@@ -347,6 +347,8 @@ impl<F, A, R> TaskBuilder<F, A, R>
 
     /// Finishes this `TaskBuilder` and spawns the new task as described by its builder functions.
     ///
+    /// Synchronizes memory with respect to the spawned task.
+    ///
     /// This merely creates the new task and makes it `Runnable`.
     /// It does not switch to it immediately; that will happen on the next scheduler invocation.
     #[inline(never)]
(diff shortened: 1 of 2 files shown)


tsoutsman:
@kevinaboos should be ready for review.

kevinaboos (approved):

PR #771, opened 2022-12-27
Add `test_runner`

tsoutsman:
The test runner is pretty simple. It runs all the applications that begin with `test_`. An alternative discovery method would be for it to parse `theseus_features/Cargo.toml` at compile time to generate a string of test paths.

The problem is the tests aren't standardised, so we can't just run them all:
- [ ] `test_backtrace`: fails to load unless backtrace is included
- [ ] `test_libc`: fails to load unless libc is included
- [ ] `test_wasmtime`: fails to load unless wasmtime is included
- [ ] `test_serial_echo`: never returns
- [ ] `test_channel`: requires a test to be selected using command line flags
- [ ] `test_ixgbe`: fails unless an IXGBE NIC is attached
- [ ] `test_mlx5`: fails unless an MLX5 NIC is attached
- [ ] `test_panic`: panics which `test_runner` interprets as a fail
- [ ] `test_realtime`: fails if the `realtime_scheduler` cfg isn't enabled

For `test_backtrace`, `test_libc`, `test_wasmtime`, and `test_realtime`, I think it only makes sense to compile and include them in the ISO if their respective features are enabled.

I don't think the hardware tests should fail if the appropriate device isn't attached.


kevinaboos:
I think this PR may jump the gun a bit. When I hear "test runner", I thought of something more like a testing harness that integrates a mechanism for attaching some form of QEMU I/O to the `cargo test`/`#[test]` attribute functionality. Once we have that, then it'd be much easier to standardize the tests, at which point this changset won't be needed.

PR #773, opened 2022-12-27
Specify recursive copy when copying dependencies

tsoutsman:
The build sometimes fails with:
```
cp: -r not specified; omitting directory './target/release/deps/rmetaMwvJeM'
make: *** [Makefile:208: build] Error 1
```

This PR fixes that. I don't think the change has any impact on the operating system.


kevinaboos:
yea, i've encountered this error occasionally, i think it comes from a partial or interrupted build. We shouldn't need to actually include any directories in `target/release/deps` but it also doesn't hurt.

kevinaboos (approved):

PR #774, opened 2022-12-27
Factor waker out of `dreadnought` and add `task_join`

tsoutsman:
The waker model can be used to implement blocking calls (e.g. `join`), and so it shouldn't be defined in `dreadnought`.

I hit a circular dependency problem similar to #766. `task_join::join`, relies on `waker`, which in turn relies on `scheduler` and `mutex_sleep`, both of which depend on `task`.


kernel/waker/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+}
+
+/// A waker that unblocks the given task when awoken.
+#[derive(Debug)]
+struct Waker {
+    /// Whether the waker has been activated.
+    ///
+    /// This field ensures `execute` detects if the waker was activated prior to
```
NathanRoyer: What is this `execute` mentioned in this comment?

tsoutsman: Nice catch. Meant to say `Blocker::block`.

kernel/waker/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+            activated: activated.clone(),
+            task: get_my_current_task().expect("failed to get current task"),
+        })),
+        Blocker { inner: activated },
+    )
+}
+
+/// A blocker that blocks the current task until the associated waker is woken.
+pub struct Blocker {
+    inner: Arc<Mutex<bool>>,
```
kevinaboos: should probably add a comment here explaining why we use a `Mutex<bool>` instead of an `AtomicBool`

kevinaboos (commented):
I like this, but it is a bit messy and likely confusing to newcomers why there needs to be a separate `task_join` crate.

Ideally, the action of blocking the current task while it is waiting to `join` another task would take place within `JoinableTaskRef::join()` itself. That would make the action of setting the current task's `waker` internal to the call to `join()`, which makes sense IMO from an API perspective. Is there any other use case where you would want to `set_waker` for a task but _not_ also call `join`?


I wonder how monolithic the `task` crate would have to become if we wanted to achieve this; there is likely some tension between that goal and our other goal of keeping crates small and plentiful.

kevinaboos:
Once #879 is merged in, we can use that here for simplicity.

kevinaboos (approved):

Issue #710, opened 2022-12-04
Port hardware clocks to new clocksource API


tsoutsman:
We should port the hardware clock sources to the new API added in #615.

- [ ] APIC
- [x] HPET
- [ ] PIT
- [x] TSC

kevinaboos:
yep, agreed. Been on my to-do list, just haven't gotten around to it. This one is solely about clock sources; next up would be a hardware timer abstraction.

amab8901:
I'll make an attempt on this

amab8901:
what does the porting entail? Can you elaborate?

tsoutsman:
> I'll make an attempt on this

Great!

> what does the porting entail? Can you elaborate?

We need to implement the [ClockSource](https://github.com/theseus-os/Theseus/blob/20809f37d0ce9d15f2b3a17e9a63d832177fb0c7/kernel/time/src/lib.rs#L205) trait for the clocks.

The [TSC](https://github.com/theseus-os/Theseus/blob/theseus_main/kernel/tsc/src/lib.rs) should be the easiest to start with. We'll want to implement the trait for a zero-sized type like so:
```rust
pub struct Tsc;

impl ClockSource for Tsc {
    // ...
}
```

The TSC is a monotonic clock source, so `now` should return an `Instant` with `counter` set to the value of the TSC register. One thing we missed in the initial PR was a `new` function for `Instant`, so we'll want to add that as well. The `unit_to_duration` and `duration_to_unit` functions should use the TSC's frequency to convert between the value of `counter` and a concrete `Duration`.

amab8901:
I noticed that `time::ClockType` is sealed. Do you want me to unseal it and write the implementation in each clock's crate? Or do you want me to write all the implementations of all the clocks in the `time` crate?

tsoutsman:
The naming may be confusing. `time::ClockType` differentiates categories of clock sources. It is implemented for `Monotonic` and `WallTime`. The TSC is not a clock type (i.e. it shouldn't implement `ClockType`) but rather a clock source (i.e. it should implement `ClockSource`). You shouldn't need to unseal `ClockType` to write the implementations in the clock crates.

```rust
pub struct Tsc;

impl time::ClockSource for Tsc {
    type ClockType = time::Monotonic;

    // ...
}
```

BTW it's probably best to create a new PR for each clock source port, e.g. a PR modifying only `tsc`.

amab8901:
[...]

PR #777, opened 2022-12-27
Port clock source to TSC

amab8901:
Addresses TSC checkbox in https://github.com/theseus-os/Theseus/issues/710

### DIFF
--- a/kernel/tsc/src/lib.rs
+++ b/kernel/tsc/src/lib.rs
@@ -2,9 +2,21 @@
 
 #[macro_use] extern crate log;
 extern crate pit_clock_basic;
+extern crate time;
 
 use core::sync::atomic::{AtomicUsize, Ordering};
 
+pub struct Tsc;
+
+impl time::ClockSource for Tsc {
+    type ClockType = time::Monotonic;
+
+    /// The current time according to the clock.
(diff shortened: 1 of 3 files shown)


kernel/tsc/src/lib.rs
```diff
@@ -2,9 +2,21 @@
...
 
+pub struct Tsc;
+
+impl time::ClockSource for Tsc {
+    type ClockType = time::Monotonic;
+
+    /// The current time according to the clock.
+    /// Monotonic clocks return an [`Instant`]
+    fn now() -> <Self::ClockType as time::ClockType>::Unit {
+        time::Instant::new(1)
```
kevinaboos, line 17: This returns a constant value of `1`. You need to actually read the current value of the timer's ticks; the functions for that are in this file.

kevinaboos (changes requested):
This PR doesn't actually use any functions from the TSC timer, so that clearly cannot be right.

In addition to fixing the one comment I left, you'd want to add a `tsc::init()` function or something that registers the TSC as a clock source too, otherwise it'll never be used. See how we do it for the existing types that implement `ClockSource`, you can use that as a guide.

PR #784, opened 2023-01-02
Safer logger on aarch64

### LINKED ISSUE #772: Undefined behaviour in aarch64 `logger`
## 1
```rust
pub static mut LOGGER: Logger = unsafe { MaybeUninit::uninit().assume_init() };
```
From the [`assume_init` docs](https://docs.rs/maybe-uninit/2.0.0/maybe_uninit/union.MaybeUninit.html#method.assume_init):
> It is up to the caller to guarantee that the MaybeUninit<T> really is in an initialized state. Calling this when the content is not yet fully initialized causes immediate undefined behavior.

We should either `assume_init` in `log` or wrap the logger in an `Option`.

## 2
```rust
let mutable = unsafe { (self as *const Self).cast_mut().as_mut().unwrap() };
```
From the [Nomicon](https://doc.rust-lang.org/nomicon/transmutes.html):
> - Transmuting an & to &mut is Undefined Behavior. While certain usages may appear safe, note that the Rust optimizer is free to assume that a shared reference won't change through its lifetime and thus such transmutation will run afoul of those assumptions. So:
>     - Transmuting an & to &mut is always Undefined Behavior.
>     - No you can't do it.
>     - No you're not special.

We should use a mutex or unsafe cell.

NathanRoyer:
Fixes #772; new code makes use of `irq_safety` and its `MutexIrqSafe` struct.
Also includes minimal documentation for this code.

It doesn't use `MaybeUninit` anymore; this has been replaced by wrapping the logger in an option, initially `None`, as suggested by @tsoutsman.

### DIFF
--- a/aarch64/Cargo.lock
+++ b/aarch64/Cargo.lock
@@ -170,6 +170,7 @@ dependencies = [
 name = "logger"
 version = "0.1.0"
 dependencies = [
+ "irq_safety",
  "log",
  "pl011_qemu",
 ]
@@ -300,7 +301,6 @@ dependencies = [
 name = "page_table_entry"
 version = "0.1.0"
 dependencies = [
- "bit_field 0.7.0",
  "frame_allocator",
  "kernel_config",
  "memory_structs",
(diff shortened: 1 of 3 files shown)


kevinaboos (approved):
Nice, looks good. When we combine `/aarch64` stuff into the main repo, this will be replaced anyway by the existing `DUMMY_LOGGER` technique, so no need to worry about it too much.

PR #785, opened 2023-01-03
Remove duplicate reserved memory regions in UEFI

tsoutsman:
This is the first step in supporting Theseus loaded at an arbitrary address. The UEFI memory map already reserves the loaded kernel file, boot info, and bootloader modules, so reserving them again is unnecessary.

The changes to `frame_allocator` are necessary for the frame allocator to correctly initialise when a reserved memory region overlaps with more than one reserved memory region that came before it. Reshuffling the order that the bootloader info and modules memory regions are added to the reserved memory regions revealed the problem.

The PR also removes `bios` as the default feature for `nano_core` as it makes it impossible to compile for UEFI, as discussed on Discord.


kernel/boot_info/src/multiboot2.rs
```diff
@@ -79,6 +79,77 @@ impl<'a> crate::Module for &'a multiboot2::ModuleTag {
...
+    .ok_or("kernel physical start address was invalid")?;
+
+    let virtual_end = VirtualAddress::new(
+        boot_info
+            .elf_sections()?
+            .into_iter()
+            .filter(|s| s.is_allocated())
+            .map(|s| s.end_address())
+            .max()
+            .ok_or("couldn't find kernel end address")? as usize,
```
kevinaboos: while we're at it, let's refactor this a bit to avoid iterating over the `elf_sections()` list twice unnecessarily.

kernel/boot_info/src/multiboot2.rs
```diff
@@ -169,6 +190,22 @@ impl crate::BootInformation for multiboot2::BootInformation {
...
+        Ok([
+            kernel_memory_region(self)?,
+            bootloader_info_memory_region(self)?,
+            modules_memory_region(self)?,
+        ]
+        .into_iter())
+    }
+
+    fn kernel_end(&self) -> Result<PhysicalAddress, &'static str> {
+        let reserved_region = kernel_memory_region(self)?;
```
kevinaboos, line 197: we're running this fairly expensive `kernel_memory_region` function twice; can we cache it or otherwise refactor to run it just once?

kevinaboos (changes requested):
Top-level comments:
1. Please move the changes to `frame_allocator` into a different PR; they're not otherwise relevant to UEFI.
2. I know it's inconvenient, but until we have a solution to the rust-analyzer issues surroudning default features, I'm reluctant to remove the `default = [ "bios" ]` featureset in `nano_core/Cargo.toml`. So let's undo that, and then keep the changes local to your machine for now. Sorry.

kernel/boot_info/src/multiboot2.rs
tsoutsman: I don't think so because UEFI only calls it once. `kernel_memory_region` iterates over a slice of approximately ten elements, so either way, I don't think it's worth the effort.

tsoutsman:
Depends on #789.

kevinaboos (approved):

PR #786, opened 2023-01-03
Add access to current page table in `PageTable::with`

### LINKED ISSUE #752: Allow access to current page table in `page_table::with`
As #747 :airplane: is merged in, I'd like to rework `page_table::with` to allow access to the current page table.

The PR stores the stack/boot info mappings in an array and then iterates over the stored mappings in the `page_table::with` function. We do this because, for the duration of the function, the recursive index points to the new page table and hence the current page table is inaccessible. However, it isn't technically necessary, as the mappings are still in the physical memory.

I [made an attempt](https://github.com/tsoutsman/Theseus/commit/1cb3d5084f319edfec3ed2c1dd90affd8af5a29c), but quickly realised it's trickier than it seems. My idea was to create a `PageTable<Active>` and `PageTable<Inactive>`. An inactive page table would store its temporary mapping, whereas an active page table would use the recursive address. However, this doesn't work because an inactive page table must also map the L3, L2, and L1 tables when accessing them.

An easier middle ground between the current hack and full access to the current page table would be to only map the L1 tables containing the stack/boot info mappings and iterate over them when modifying the new page table.

tsoutsman:
I'm not sure if adding another index like this is ok. Does the page allocator code need to be changed?

Closes #752.


kernel/kernel_config/src/memory.rs
```diff
@@ -36,18 +36,16 @@ pub const MAX_VIRTUAL_ADDRESS: usize = usize::MAX;
...
 /// Value: 510. The 510th entry is used for the recursive P4 mapping.
 pub const RECURSIVE_P4_INDEX: usize = ENTRIES_PER_PAGE_TABLE - 2;
 /// Value: 509. The 509th entry is used for the kernel heap
 pub const KERNEL_HEAP_P4_INDEX: usize = ENTRIES_PER_PAGE_TABLE - 3;
-/// Value: 508. The 508th entry is used for all kernel stacks
-pub const KERNEL_STACK_P4_INDEX: usize = ENTRIES_PER_PAGE_TABLE - 4;
-/// Value: 507. The 507th entry is used for all userspace stacks
-pub const USER_STACK_P4_INDEX: usize = ENTRIES_PER_PAGE_TABLE - 5;
+// Value: 508. The 508th entry is used as a temporary recursive entry when mapping a new page table.
+pub const TEMPORARY_RECURSIVE_P4_INDEX: usize = ENTRIES_PER_PAGE_TABLE - 6;
```
kevinaboos: small oversight i think?
```suggestion
pub const TEMPORARY_RECURSIVE_P4_INDEX: usize = ENTRIES_PER_PAGE_TABLE - 4;
```

kernel/memory/src/paging/mod.rs
```diff
@@ -125,7 +129,7 @@ impl PageTable {
         other_table: &mut PageTable,
         f: F,
     ) -> Result<(), &'static str>
-        where F: FnOnce(&mut Mapper) -> Result<(), &'static str>
+        where F: FnOnce(&mut Mapper, &Mapper) -> Result<(), &'static str>
```
kevinaboos, line 131: update the docs for this function to describe what the second parameter to the `F` type is, the immutable `Mapper` reference.

kernel/memory/src/paging/table.rs
```diff
@@ -29,6 +29,13 @@ pub const P4: *mut Table<Level4> = 0o177777_776_776_776_776_0000 as *mut _;
                                          // ^p4 ^p3 ^p2 ^p1 ^offset  
                                          // ^ 0o776 means that we're always looking at the 510th entry recursively
 
+/// By default this is equivalent to [`P4`] unless the temporary recursive entry is modified.
+///
+/// NOTE: this must be kept in sync with the recursive index in `kernel_config/memory.rs`.
+///
+/// All four table indexes need to be set to 0o772 so that `Table::next_table_address` works properly.
+pub const TEMP_P4: *mut Table<Level4> = 0o177777_772_772_772_772_0000 as *mut _;
```
kevinaboos: since we updated the P4 index used for this, we also need to fix this accordingly.

kevinaboos (changes requested):
Please update the module comments at the top of `kernel_config/src/memory.rs`, both the warning up top and the set of P4 mappings.

I briefly mentioned this in our meeting, but I also think we should change the term/name "temp" to something more descriptive that explains what that "temporary" P4 entry is used for.

kernel/memory/src/paging/mapper.rs
```diff
@@ -82,14 +89,13 @@ impl Mapper {
     /// and also shows their `PteFlags`.
     /// 
     /// The page table details are written to the the given `writer`.
-    pub fn dump_pte<W: Write>(&self, writer: &mut W, virtual_address: VirtualAddress) -> fmt::Result {
+    pub fn dump_pte(&self,virtual_address: VirtualAddress) {
```
kevinaboos: is there a hard reason for this change? It directly printed to the log ages ago, but I thought adding the `Write`r param made it more flexible.

kernel/memory/src/paging/mod.rs
```diff
@@ -104,6 +104,10 @@ impl PageTable {
                 frame.as_allocated_frame(),
                 PteFlagsArch::new().valid(true).writable(true),
             );
+            table[INACTIVE_PAGE_TABLE_RECURSIVE_P4_INDEX].set_entry(
+                frame.as_allocated_frame(),
+                PteFlagsArch::new().valid(true).writable(true),
+            );
```
kevinaboos: i realize doing this here is a bit faster, but i wonder if we should leave this P4 entry zeroed-out (invalid) just to ensure that no one can access these addresses at any point (except for, of course, in `PageTable:;with()`.

thoughts?

kevinaboos (changes requested):
Looks pretty good; left a few small comments.

Let's bikeshed the name "INACTIVE_..." on discord a bit.

kernel/memory/src/paging/mapper.rs
tsoutsman: AFAICT previously you needed to add a dependency on the `logger` crate and call it like this:
```rust
dump_pte(&mut logger::LOGGER.lock().unwrap().writers[0], virtual_address);
```
and I don't know where else you could write the output feasibly.

kevinaboos: well, the main use case would be writing it to a String, at which point you could do anything with it.

That does raise the related question of how we can get access to the current logger as a stream, which is apparently pretty shit rn.

tsoutsman: I can change it back, considering it's irrelevant to the PR, but it does seem strange that it's the only place in Theseus that uses a `Write` trait bound rather than `log`. E.g. `dump_frame_allocator_state` just prints to `log`.

kevinaboos: yea, i agree it is strange, especially considering it's a one-off.
It's fine, leave the change in ig, since this is a larger issue about being able to treat the logger stream as a `fmt::Write`r.

tsoutsman:
Sorry pushed to wrong branch gimme a sec.

tsoutsman:
Should be good now.

kevinaboos (approved):
Thanks, this is a big improvement. Perhaps later we can revamp the main mapping routines to avoid wasting 2/512ths of the virtual address space, but for now that's obviously nbd.

Sorry it took so damned long to merge it, i ran into a surprise issue with docs builds and had to triage those CI failures first.

PR #789, opened 2023-01-06
Fix frame allocator reserved list deduplication

tsoutsman:
The previous algorithm would fail if a reserved memory region overlaped with more than one reserved memory region came before it.


kevinaboos (approved):
Thanks!

PR #790, opened 2023-01-06
Separate `.init` and `.text` memory bounds

tsoutsman:
Since they are two different program segments, there are no guarantees that the bootloader will allocate them contiguously in physical memory.

### DIFF
--- a/kernel/memory/src/paging/mod.rs
+++ b/kernel/memory/src/paging/mod.rs
@@ -274,6 +274,8 @@ pub fn init(
         debug!("{:X?}", aggregated_section_memory_bounds);
         let mut index = 0;
 
+        let (init_start_virt,    init_start_phys)    = aggregated_section_memory_bounds.init.start;
+        let (init_end_virt,      init_end_phys)      = aggregated_section_memory_bounds.init.end;
         let (text_start_virt,    text_start_phys)    = aggregated_section_memory_bounds.text.start;
         let (text_end_virt,      text_end_phys)      = aggregated_section_memory_bounds.text.end;
         let (rodata_start_virt,  rodata_start_phys)  = aggregated_section_memory_bounds.rodata.start;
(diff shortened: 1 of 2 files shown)


kernel/memory/src/paging/mod.rs
```diff
@@ -274,24 +274,40 @@ pub fn init(
...
+
         let text_pages = page_allocator::allocate_pages_by_bytes_at(text_start_virt, text_end_virt.value() - text_start_virt.value())?;
         let text_frames = frame_allocator::allocate_frames_by_bytes_at(text_start_phys, text_end_phys.value() - text_start_phys.value())?;
         let text_pages_identity = page_allocator::allocate_pages_by_bytes_at(text_start_virt - KERNEL_OFFSET, text_end_virt.value() - text_start_virt.value())?;
         identity_mapped_pages[index] = Some(NoDrop::new( unsafe {
             Mapper::map_to_non_exclusive(mapper, text_pages_identity, &text_frames, text_flags)?
         }));
-        text_mapped_pages = Some(NoDrop::new(mapper.map_allocated_pages_to(text_pages, text_frames, text_flags)?));
+        init_mapped_pages.merge(mapper.map_allocated_pages_to(text_pages, text_frames, text_flags)?).map_err(|(error, _)| error)?;
+        text_mapped_pages = Some(NoDrop::new(init_mapped_pages));
```
kevinaboos, line 310: Obviously there's nothing wrong with doing this, but looking ahead to UEFI, will this cause any problems? If we're separating the bounds of `.init` and `.text`, we might as well treat them as separate sections.

I guess the real question is what is the true benefit of merging init and text back together here?

kevinaboos (changes requested):
👍 left one small question

kernel/memory/src/paging/mod.rs
tsoutsman: > Obviously there's nothing wrong with doing this, but looking ahead to UEFI, will this cause any problems?

No. At this point, it doesn't change anything. Only the identity mappings matter.

> I guess the real question is what is the true benefit of merging init and text back together here?

So we can store the `.init` section in the nano core's `LoadedCrate` text sections.

kevinaboos: thanks!

kevinaboos (approved):
LGTM thanks!

PR #794, opened 2023-01-09
Clarify error handling in `random` crate

### LINKED ISSUE #793: Flesh out `Error` type in `random`
#783 made a change to `random` that added a basic error type to placate clippy. All we need to do is rename the error type and variants to be more appropriate.

tsoutsman:
It doesn't make sense for errors from `rdseed_seed` and `rdrand_seed` to be in the public API, since they are private functions. The only error that can be returned to the user is from the initialisation of the CSPRNG, which we don't control.

Closes #793.

### DIFF
--- a/kernel/random/src/lib.rs
+++ b/kernel/random/src/lib.rs
@@ -28,6 +28,8 @@ use rand_chacha::{
 };
 use spin::mutex::Mutex;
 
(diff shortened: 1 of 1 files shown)


kevinaboos (approved):
Makes sense, using `Option` and the existing rand_core Error is a good fix.

PR #795, opened 2023-01-09
Fix clippy lints

amab8901:
Addresses https://github.com/theseus-os/Theseus/issues/526. Commit messages specify which lints are being addressed.

### DIFF
--- a/kernel/libterm/src/lib.rs
+++ b/kernel/libterm/src/lib.rs
@@ -193,7 +193,7 @@ impl Terminal {
             if new_line_indices.len() == 0 {
                 // indicates that the text is just one continuous string with no newlines and will therefore fill the buffer completely
                 end_idx += buffer_height * buffer_width;
-                if end_idx <= self.scrollback_buffer.len() -1 {
+                if end_idx == self.scrollback_buffer.len() {
                     return Ok(end_idx); 
                 } else {
                     return Err(ScrollError::OffEndBound);
@@ -234,7 +234,7 @@ impl Terminal {
                 end_idx += last_line_chars;
             }
 
-            if end_idx <= self.scrollback_buffer.len() -1 {
+            if end_idx == self.scrollback_buffer.len() {
                 return Ok(end_idx); 
             } else {
                 return Err(ScrollError::OffEndBound);
--- a/kernel/libterm/src/lib.rs
+++ b/kernel/libterm/src/lib.rs
@@ -193,7 +193,7 @@ impl Terminal {
             if new_line_indices.len() == 0 {
                 // indicates that the text is just one continuous string with no newlines and will therefore fill the buffer completely
                 end_idx += buffer_height * buffer_width;
-                if end_idx == self.scrollback_buffer.len() {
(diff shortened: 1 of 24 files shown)


kernel/libterm/src/lib.rs
```diff
@@ -193,7 +193,7 @@ impl Terminal {
             if new_line_indices.len() == 0 {
                 // indicates that the text is just one continuous string with no newlines and will therefore fill the buffer completely
                 end_idx += buffer_height * buffer_width;
-                if end_idx <= self.scrollback_buffer.len() -1 {
+                if end_idx == self.scrollback_buffer.len() {
```
kevinaboos: this changes the logic, right?

kernel/libterm/src/lib.rs
```diff
@@ -234,7 +234,7 @@ impl Terminal {
                 end_idx += last_line_chars;
             }
 
-            if end_idx <= self.scrollback_buffer.len() -1 {
+            if end_idx == self.scrollback_buffer.len() {
```
kevinaboos: same here, this changes the logic too

kernel/mlx_ethernet/src/command_queue.rs
```diff
@@ -398,8 +398,8 @@ pub enum QueryHcaCapMaxOpMod {
 /// Possible values of the opcode modifer when the opcode is [`CommandOpcode::QueryHcaCap`] and we want to retrieve current values of capabilities.
 #[derive(Copy, Clone)]
 pub enum QueryHcaCapCurrentOpMod {
-    GeneralDeviceCapabilities       = (0x0 << 1) | 0x1,
-    EthernetOffloadCapabilities     = (0x1 << 1) | 0x1
+    GeneralDeviceCapabilities       =              0x1,
```
kevinaboos: i think this is important to leave here in that it's reflecting the logic of how the const values of these hardware flags are calculated. So let's leave this one as is, and allow the `identity_op` just for this.

kernel/pit_clock/src/lib.rs
```diff
@@ -21,7 +21,7 @@ use pit_clock_basic::*;
 
 /// The PIT Channel 0 is connected directly to IRQ 0.
 /// Because we perform the typical PIC remapping, the remapped IRQ vector number is 0x20.
-const PIT_CHANNEL_0_IRQ: u8 = interrupts::IRQ_BASE_OFFSET + 0x0;
+const PIT_CHANNEL_0_IRQ: u8 = interrupts::IRQ_BASE_OFFSET;
```
kevinaboos: this is another case where the logic is encoding some information, so let's undo this change and allow this identity op lint.

kevinaboos (commented):
Looks pretty good! I left a few small comments, but they should be quick to address.

kernel/libterm/src/lib.rs
amab8901: oh! I think it should have been < rather than == :facepalm:

kevinaboos (approved):
Thanks, looks great!

PR #798, opened 2023-01-10
Fix clippy lints

amab8901:

### DIFF
--- a/kernel/compositor/src/lib.rs
+++ b/kernel/compositor/src/lib.rs
@@ -190,7 +190,7 @@ impl CompositableRegion for Rectangle {
                 Some(index) => index,
                 None => {continue;}
             };
-            dest_fb.composite_buffer(&(src_buffer[src_start_index..src_end_index]), dest_start_index as usize);
+            dest_fb.composite_buffer( &(src_buffer[src_start_index..src_end_index]), dest_start_index );
         }
 
         Ok(())
--- a/kernel/compositor/src/lib.rs
+++ b/kernel/compositor/src/lib.rs
@@ -190,7 +190,7 @@ impl CompositableRegion for Rectangle {
                 Some(index) => index,
                 None => {continue;}
             };
-            dest_fb.composite_buffer( &(src_buffer[src_start_index..src_end_index]), dest_start_index );
+            dest_fb.composite_buffer(&src_buffer[src_start_index..src_end_index], dest_start_index);
         }
 
         Ok(())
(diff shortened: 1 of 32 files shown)


kernel/unwind/src/lsda.rs
```diff
@@ -284,11 +284,11 @@ impl<R: Reader> FallibleIterator for CallSiteTableIterator<R> {
 fn read_encoded_pointer<R: gimli::Reader>(reader: &mut R, encoding: DwEhPe) -> gimli::Result<u64> {
     match encoding {
         DW_EH_PE_omit     => Err(gimli::Error::CannotParseOmitPointerEncoding),
-        DW_EH_PE_absptr   => reader.read_u64().map(|v| v as u64),
-        DW_EH_PE_uleb128  => reader.read_uleb128().map(|v| v as u64),
+        DW_EH_PE_absptr   => reader.read_u64().map(|v| v ), // v:u64
```
kevinaboos: all of these `.map(|v| v)` are what's causing the current clippy CI pass to fail, since they are now useless maps. So you can remove them.

kernel/compositor/src/lib.rs
```diff
@@ -190,7 +190,7 @@ impl CompositableRegion for Rectangle {
                 Some(index) => index,
                 None => {continue;}
             };
-            dest_fb.composite_buffer(&(src_buffer[src_start_index..src_end_index]), dest_start_index as usize);
+            dest_fb.composite_buffer( &(src_buffer[src_start_index..src_end_index]), dest_start_index );
```
kevinaboos: remove unnecessary parentheses, fix spacing
```suggestion
            dest_fb.composite_buffer(&src_buffer[src_start_index..src_end_index], dest_start_index);
```

kevinaboos (changes requested):
Logically everything looks good, but formatting-wise it's a bit sloppy. Can you go through all the proposed changes and ensure the spacing is correct? There's a lot of errant whitespace at the end of variable names or other tokens, between punctuation, etc.

I also see a few places where you allowed `clippy:type_complexity`. Is that actually what we want to do, or just a blanket fix to tell clippy to shut up? 😄 If it's the latter, maybe we can go through some of those examples to figure out what clippy would recommend instead (or do it in a future PR).

amab8901:
> Logically everything looks good, but formatting-wise it's a bit sloppy. Can you go through all the proposed changes and ensure the spacing is correct? There's a lot of errant whitespace at the end of variable names or other tokens, between punctuation, etc.

sure

> I also see a few places where you allowed `clippy:type_complexity`. Is that actually what we want to do, or just a blanket fix to tell clippy to shut up? smile If it's the latter, maybe we can go through some of those examples to figure out what clippy would recommend instead (or do it in a future PR).

Type complexity is a blanket implementation to tell Clippy to shut up. Because as I understand it, the idea behind `type_complexity` lint is that for the average Rust program (a library), there's going to be end-users interacting with methods and functions and stuff, and you want good user experience (including learning experience) to the user of the library. But in our case, it's like low-level stuff that the end-user probably won't interact with. It's an end-product rather than a library. It's just something that will run in the background. And if you ever need to interact with this code, then it's probably because you're a system developer who's trying to add a feature or fix some bug or refactor etc, in which case it may be helpful to see all the wrappers directly so that you will easily know how to handle all those layers.

kevinaboos (approved):
Thanks again! Appreciate the contribution.

PR #805, opened 2023-01-11
Fix Clippy lints

amab8901:


kevinaboos (approved):
Thanks again! Looks great.

PR #806, opened 2023-01-12
Fix clippy lints

amab8901:

### DIFF
--- a/kernel/pmu_x86/src/lib.rs
+++ b/kernel/pmu_x86/src/lib.rs
@@ -696,7 +696,7 @@ pub fn start_samples(event_type: EventType, event_per_sample: u32, task_id: Opti
     // and is therefore by definition in the range `u32::MIN..=u32::MAX`.
     // We'll check anyways, just in case `event_per_sample`'s type is changed.
     #[allow(clippy::absurd_extreme_comparisons)]
-    if !(core::u32::MIN..core::u32::MAX).contains(&event_per_sample) {
+    if !(core::u32::MIN..=core::u32::MAX).contains(&event_per_sample) {
         return Err("Number of events per sample invalid: must be within unsigned 32 bit");
     }
 
--- a/kernel/pmu_x86/src/lib.rs
+++ b/kernel/pmu_x86/src/lib.rs
@@ -696,7 +696,7 @@ pub fn start_samples(event_type: EventType, event_per_sample: u32, task_id: Opti
     // and is therefore by definition in the range `u32::MIN..=u32::MAX`.
     // We'll check anyways, just in case `event_per_sample`'s type is changed.
     #[allow(clippy::absurd_extreme_comparisons)]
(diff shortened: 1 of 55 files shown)


kernel/pmu_x86/src/lib.rs
```diff
@@ -696,7 +696,7 @@ pub fn start_samples(event_type: EventType, event_per_sample: u32, task_id: Opti
     // and is therefore by definition in the range `u32::MIN..=u32::MAX`.
     // We'll check anyways, just in case `event_per_sample`'s type is changed.
     #[allow(clippy::absurd_extreme_comparisons)]
-    if event_per_sample > core::u32::MAX || event_per_sample < core::u32::MIN {
+    if !(core::u32::MIN..=core::u32::MAX).contains(&event_per_sample) {
```
kevinaboos: might be better to change this to something like `u32::try_from(event_per_sample)` rather than this complex conditional

kernel/text_terminal/src/lib.rs
```diff
@@ -2329,9 +2329,10 @@ impl WideDisplayedUnit {
...
     #[inline(always)]
     fn is_continuance(&self) -> bool {
-        match self {
-            &Self::TabFill | &Self::MultiFill => true,
-            _ => false,
-        }
+        // match self {
+        //     &Self::TabFill | &Self::MultiFill => true,
+        //     _ => false,
+        // }
```
kevinaboos: this change is fine by me, so you can remove the commented-out old stuff.

kernel/unwind/src/lib.rs
```diff
@@ -98,9 +98,9 @@ pub struct UnwindingContext {
     /// A reference to the current task that is being unwound.
     current_task: TaskRef,
 }
-impl Into<(StackFrameIter, KillReason, TaskRef)> for UnwindingContext {
-    fn into(self) -> (StackFrameIter, KillReason, TaskRef) {
-        (self.stack_frame_iter, self.cause, self.current_task)
+impl From<UnwindingContext> for (StackFrameIter, KillReason, TaskRef){
```
kevinaboos:
```suggestion
impl From<UnwindingContext> for (StackFrameIter, KillReason, TaskRef) {
```

kevinaboos (changes requested):
Nearly perfect, thanks!

One request -- instead of adding empty SAFETY blocks and TODOs for the everywhere, just add one top-level `allow` command at the top of each file that requires this. The rationale behind that request is that adding a bunch of empty SAFETY/TODO blocks actually makes it _harder_ to discover unsafe blocks that aren't clearly explained. 
Adding an allow statement once per file reduces the number of unnecessary changes and also makes it easier to find where changes are still needed. Later, we'll go through all the places where each clippy lint is allowed and assign the correct person to address it.
Since going through every unsafe block and empty loop is a significant undertaking, it's best left for a future changeset rather than falsely placated now.

PR #809, opened 2023-01-12
Deduplicate run queue implementations

tsoutsman:
I renamed `runqueue` to `run_queue` because the struct was called `RunQueue` implying that it is two words.


kernel/ap_start/src/lib.rs
```diff
@@ -90,7 +90,6 @@ pub fn kstart_ap(
     // Now that the Local APIC has been initialized for this CPU, we can initialize the
     // task management subsystem and create the idle task for this CPU.
```
kevinaboos:
```suggestion
    // task management subsystem, which creates the idle task and
    // scheduler runqueues for this CPU.
```

kernel/scheduler/src/lib.rs
```diff
@@ -1,82 +1,158 @@
...
+        }
     };
 
-    let (did_switch, recovered_preemption_guard) = task::task_switch(
-        next_task,
-        cpu_id,
-        preemption_guard,
-    ); 
+    let next_task = run_queue.next();
+    drop(run_queue);
```
kevinaboos: i have experienced weird behavior before where `drop`ping a lock guard does not actually drop it until the end of the lexical scope. Therefore, I'd prefer to surround the guard with curly brackets (from which you can "return" `next_task`) to ensure that happens.

kernel/spawn/src/lib.rs
```diff
@@ -982,31 +987,29 @@ fn remove_current_task_from_runqueue(current_task: &ExitableTaskRef) {
...
-pub fn create_idle_task() -> Result<JoinableTaskRef, &'static str> {
-    let apic_id = cpu::current_cpu();
+pub fn create_idle_task(apic_id: u8) -> Result<TaskRef, &'static str> {
     debug!("Spawning a new idle task on core {}", apic_id);
 
     new_task_builder(idle_task_entry, apic_id)
         .name(format!("idle_task_core_{}", apic_id))
         .idle(apic_id)
         .spawn_restartable(None)
+        .map(|task| task.clone())
```
kevinaboos: this is correct, but quite tricky to understand. Most readers would likely misinterpret it as a "dumb unnecessary clone". 

It would be better to clearly show what is happening (or at least add a comment) explaining that we are dropping the `JoinableTaskRef` on purpose and obtaining a `TaskRef` since we do not want to join in later.

Even better would be to introduce a function like
```rust
impl JoinableTaskRef {
    /// Drops this `JoinableTaskRef`, making the task an orphan by ensuring
    /// that it can never be joined.
    ///
    /// This guarantees that it will be auto-reaped upon exit.
    pub fn orphan(self) -> TaskRef {
        ...
    }
}
```
and then call that `orphan()` function here.

kevinaboos: perhaps `into_orphan()` would be more Rusty

kevinaboos (changes requested):
damn, i had typed a nice long comment but my browser crashed and i lost it 😢 

anyway, the crux of it was that most of this is great, but there are some things to change:
* there should be one main `scheduler` crate that has only the "mechanisms" for scheduling: initialization, a public `schedule()` function, etc.
* All scheduler "policy" crates must be their own separate crate (i typed a bunch of reasons, but ... sigh)
* That main `scheduler` mechanism crate then invokes whichever scheduler policy crate to obtain the next task
* Must the runqueue be merged into the scheduler itself? It'd be nice it we could decouple them; I'm open to discussing the pros & cons of this.
* `runqueue_spillful` stuff can be removed entirely (or in a separate PR)
* The original design had a significant benefit in that there was no "a priori" defined interface to which a new scheduler policy implementation had to adhere. I've written about this in a prior paper, I can find it or explain it if you want more detail.
In contrast, the new merged/single runqueue has to have a lot of dummy implementations that do nothing (e.g., the round robin one returning an error when periodicity or priority is set), which defines an interface that limits flexibility to support new schedulers. 
  * I bet we can move those errors to compile time using trait objects (e.g., impl HasPriority for PriorityScheduler but not for RRScheduler), but this is something that can admittedly be done later

tsoutsman:
> Must the runqueue be merged into the scheduler itself? It'd be nice it we could decouple them; I'm open to discussing the pros & cons of this.

In the current configuration, `runqueue` exports internal implementation details. E.g. the fact that the run queue internally uses `PriorityTaskRef` or `RoundRobinTaskRef` or `RealtimeTaskRef`. The scheduler implementations only make sense in the context of their specific run queue, so I don't see the point of separating out the run queue implementation.

I'm pretty sure that currently, `scheduler_round_robin` could be compiled with `runqueue_realtime`, and the build system wouldn't complain. Obviously, that's a convoluted example, but I think that's indicative of the problem as that shouldn't be possible in the first place. The round-robin policy is a scheduler policy, and as such, I think it should be contained in a single crate. 

> In contrast, the new merged/single runqueue has to have a lot of dummy implementations that do nothing (e.g., the round robin one returning an error when periodicity or priority is set), which defines an interface that limits flexibility to support new schedulers.

I don't see how it's any different to [this](https://github.com/theseus-os/Theseus/blob/a5bba8aebc5c2ea62ba06d1f7bc0e075090ee5d8/kernel/scheduler/src/lib.rs#L52-L82). I wanted to minimise the number of `cfg()`s used, but I can switch it back to the old method.

kernel/spawn/src/lib.rs
tsoutsman: Added in `d6324142`. I think `orphan` is better because it's not just converting the `JoinableTaskRef` but also modifying the internal state.

tsoutsman:
It's impossible to have the schedulers in separate crates without also introducing generics on the run queue. Specifically the `queue: imp::Queue` field.

Issue #539, opened 2022-06-17
Adjust performance counter bitwidth assumptions to reflect possible sizes beyond `u32`


kevinaboos:
cc @Ramla-I 

> looking back over the sampling code, there are some issues. Foremost is the assumption that the maximum bit width of the counter is 32, when we should actually be taking it from the CPUID info.
> If the bitwidth is greater than 32, then our calculation for setting the start value of a register is incorrect, and an interrupt will occur after more than events_per_sample events occur.
> Also, in `start_samples()` we reserve a register for use, but the function can return with an error without freeing that register. So that register can never be used again.

Also, see this: https://github.com/theseus-os/Theseus/pull/538#discussion_r900225445

PR #812, opened 2023-01-13
Fix clippy lints

amab8901:


kernel/mutex_preemption/src/mutex_preempt.rs
```diff
@@ -138,13 +138,13 @@ impl<'a, T: ?Sized> Deref for MutexPreemptGuard<'a, T> {
     type Target = T;
 
     fn deref(&self) -> &T { 
-        & *(self.guard) 
+        &(self.guard)
```
kevinaboos: remove parentheses here, and everywhere else with a similar pattern

kernel/nano_core/build.rs
```diff
@@ -76,7 +82,7 @@ fn emit_built_rs_file() {
 
             custom_cfgs_str.push_str(&key);
             if !v.is_empty() {
-                custom_cfgs_str.push_str(&format!("=\"{}\"", v));
+                write!(custom_cfgs_str, "=\"{}\"", v).expect("Failed to add String value to custom_cfgs_str variable");
```
kevinaboos: how is this better? seems worse to me. What lint is suggesting this, so i can read about its rationale?

kernel/pmu_x86/src/lib.rs
```diff
@@ -701,7 +704,7 @@ pub fn start_samples(event_type: EventType, event_per_sample: u32, task_id: Opti
     // This check can never trigger since `event_per_sample` is a `u32`
     // and is therefore by definition in the range `u32::MIN..=u32::MAX`.
     // We'll check anyways, just in case `event_per_sample`'s type is changed.
-    if u32::try_from(event_per_sample).is_ok() {
+    if event_per_sample.type_id() == TypeId::of::<u32>() {
```
kevinaboos: TypeIDs aren't generally supposed to be used for this purpose. Let's keep it the way it was, since we do want a u32 out of it, and simply add an `allow` attribute above it.

kevinaboos (changes requested):

kernel/nano_core/build.rs
amab8901: https://rust-lang.github.io/rust-clippy/master/index.html#format_push_string

kernel/pmu_x86/src/lib.rs
amab8901: Why aren't TypeID:s supposed to be used for this purpose? The way it was triggered a lint (I don't remember which one), and the way I have it now doesn't trigger any lints

amab8901: the purpose of this condition is to check whether `event_per_sample` remained in the type `u32`, or throw error if the type has changed to something other than `u32`

kernel/nano_core/build.rs
amab8901: The way I put it now helps avoid unnecessary heap allocation, and also adds an error mechanism to spot a bug early in case there's a bug, instead of ignoring it. I added the wrong link at first. Now it should be correct.

kevinaboos: oh interesting, sounds good then. One suggestion:
```suggestion
                write!(custom_cfgs_str, "=\"{}\"", v).expect("Failed to write custom_cfgs_str");
```

kernel/pmu_x86/src/lib.rs
kevinaboos: Right,  but that's more of a concession for the fact that it needs to be changed anyway (see #539), so this is a temp fix. More specifically, the proper condition is not that it _must_ be a `u32`, but that the value of `event_per_sample` should fit into a `u32`. The old code is more correct.

applications/kill/src/lib.rs
```diff
@@ -35,79 +35,9 @@ pub fn main(args: Vec<String>) -> isize {
     return -1;
 }
 
-/*
```
kevinaboos: please restore this old code -- it will be re-used in an upcoming commit.

kernel/mutex_preemption/src/mutex_preempt.rs
```diff
@@ -138,13 +138,13 @@ impl<'a, T: ?Sized> Deref for MutexPreemptGuard<'a, T> {
...
     fn deref(&self) -> &T { 
-        & *(self.guard) 
+        &self.guard 
     }
 }
 
 impl<'a, T: ?Sized> DerefMut for MutexPreemptGuard<'a, T> {
     fn deref_mut(&mut self) -> &mut T { 
-        &mut *(self.guard)
+        &mut (self.guard)
```
kevinaboos: remove unneeded parentheses

kernel/pmu_x86/src/lib.rs
```diff
@@ -701,7 +699,8 @@ pub fn start_samples(event_type: EventType, event_per_sample: u32, task_id: Opti
     // This check can never trigger since `event_per_sample` is a `u32`
     // and is therefore by definition in the range `u32::MIN..=u32::MAX`.
     // We'll check anyways, just in case `event_per_sample`'s type is changed.
-    if u32::try_from(event_per_sample).is_ok() {
+    
+    if (u32::MIN..=u32::MAX).contains(&event_per_sample) {
```
kevinaboos: kindly keep the iteration of `u32::try_from()`. That is the _clearest_ and most correct semantic meaning of what we want to achieve here, and even when we change the type of `event_per_sample`, that code will continue to work. This change would not.

kevinaboos: Actually, the condition needs to be:
```suggestion
    if u32::try_from(event_per_sample).is_err() {
```
emphasis on the `is_err()`

kevinaboos (changes requested):
left a few more comments. There are still some unneeded parentheses, mostly around patterns like `&mut (self.guard)`

kevinaboos (approved):

PR #813, opened 2023-01-13
Rework sync primitives

tsoutsman:
`WaitQueue` now uses `mpmc_queue::Queue` internally.

All sync primitives now have a generic that implements `sync::Flavour`, one of:
- `Spin`
- `DisablePreemption`
- `DisableIrq`
- `Block`

Which modifies how the primitive behaves.

`DeadlockPrevention` is a more limited trait implemented for:
- `Spin`
- `DisablePreemption`
- `DisableIrq`

Which can be used for wait queues.

After documentation, the PR will be ready for review.

Future PRs:
- Implement generic `RwLock`
- Implement Condvar (should be easy since it can work for any mutex)
- Switch to using the new interface


libs/sync/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// Initial value for the lock data.
+    const INIT: Self::LockData;
+
+    /// Additional data stored on the lock.
+    type LockData;
+
+    type DeadlockPrevention: DeadlockPrevention;
+
+    /// Acquires the given mutex.
+    fn mutex_lock<'a, T>(
```
kevinaboos: nit: i think it's a bit clearer to have this be a verb phrase like `lock_mutex`

libs/sync/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+        Self: Sized;
+
+    /// Performs any necessary actions after unlocking the mutex.
+    fn post_unlock(mutex: &Self::LockData)
+    where
+        Self: Sized;
+}
+
+/// A deadlock prevention method.
+pub trait DeadlockPrevention {
```
kevinaboos: I generally like this name, but unfortunately it's a bit too narrow in scope to be correct -- we implement it for things that have no bearing on deadlock, e.g., blocking a task.

It's more like a pre/post lock action catch-all trait. We can bikeshed the name a bit, either here or on discord.

libs/sync_irq/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+/// A deadlock prevention method that disables interrupt requests.
+#[derive(Copy, Clone, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct DisableIrq {}
+
+impl !Send for DisableIrq {}
+
+impl sync::DeadlockPrevention for DisableIrq {
+    #[inline]
+    fn enter() {
+        // FIXME: Recursive disabling doesn't work.
```
kevinaboos: yeah, i think this is a symptom of this trait design, where `DeadlockPrevention`'s methods are a bit too simple. See my comment there.

libs/sync/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// Performs any necessary actions prior to locking.
+    fn enter();
+
+    /// Performs any necessary actions after unlocking.
+    fn exit();
```
kevinaboos: You've probably already considered this, but these trait methods are too limited because they have no stateful context. They need to offer a bit more complexity to be viable in other use cases, i.e., they need to be able to share state across the `enter` and `exit` calls. There are quite a few ways to do this, e.g., returning an object that is passed into exit.

Feel free to choose any method iface you want, but consider the most typical needs therein:
* maintaining a "guard" type (e.g., `HeldInterrupts`, `PreemptionGuard`) that is created in `enter()` and dropped in `exit()`
* performing an action that may require access to other mutex states (e.g., blocking a task)

kernel/sync_preemption/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+/// A deadlock prevention method that disables preemption.
+#[derive(Copy, Clone, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct DisablePreemption {}
+
+impl !Send for DisablePreemption {}
+
+impl sync::DeadlockPrevention for DisablePreemption {
+    #[inline]
+    fn enter() {
+        preemption::disable_preemption()
```
kevinaboos: as i mentioned in the comments on the `DeadlockPrevention` trait, this isn't correct, and unfortunately it cannot be correct due to a lacking interface. The issue here is that preemption can't just be arbitrarily enabled and disabled; to do it safely we need to maintain a count and ensure that it's only enabled or disabled when that count changes to or from 0.

That's the reason I introduced a `PreemptionGuard` type, and also why there is no other direct way to manually enable/disable preemption -- the guard type semantic is useless if there is also a broken way to circumvent it.

This is similar to what you said in your "FIXME" comment in the `DisableIrq` implementation; we should modify the `DeadlockPrevention` interface to allow guard types and other stateful contexts to be propagated across `enter` and `exit` calls.

kernel/wait_queue/src/lib.rs
```diff
@@ -1,239 +1,65 @@
...
-    Interrupted,
-    Timeout,
-    SpuriousWakeup,
-    CantBlockCurrentTask,
+use mpmc_queue::Queue;
+use sync::DeadlockPrevention;
+use sync_spin::Spin;
+use task::{get_my_current_task, TaskRef};
+
+pub struct WaitQueue<P = Spin>
```
kevinaboos, line 25: if a waitqueue only has the semantics of blocking a task, then why do we need to parameterize it on any form of `DeadlockPrevention`? It seems like this struct should be the one to make the choice as to what we're doing here, which AFAICT is just blocking a task.

kevinaboos (changes requested):
Thanks! I think the biggest outstanding design issue is the too-simple interface of `DeadlockPrevention` (and that its name is a bit of a misnomer). See my comments there.

A few top-level comments/requests:
1. Can you separate out everything that isn't a core sync abstractions and implementation from this PR? That is, open a new PR with the non-required/secondary changes to the channels crates, the queue crates, (and potentially the waitqueue and wait-related crates). That way we can consider those later, independently of the addition of these new sync primitives.
    * If/when we make changes to other crates (mpmc_queue, dfqueue, async_channel, rendezvous, etc), I'd like to have the sync primitive interface settled first such that we don't cause too much churn. The reason there is that other crates like our port of `wasmtime` implicitly depend on this, by means of using `mutex_sleep` and `waitqueue`.
2. Should we use our new waker functionality to realize a better waitqueue or general abstraction of a blocking wait condition/event? 
3. Isn't `sync_block` the same as `mutex_sleep`?
    * (I'm in favor of removing `mutex_sleep` later, just wanted to double-check that you intended them to be the same, functionality-wise).
4. The `VecDeque::new()` fn is now `const` in a recent version of Rust, so we can use that if needed to make `Waitqueue::new()` const.

tsoutsman:
I'll respond to all the comments here because I think they all touch on the same points.

I don't think I documented the API well enough.

The `DeadlockPrevention` trait is implemented only for:
1. `DisableIrq`
2. `DisablePreemption`
3. `Spin` (No deadlock prevention)

All of these types are related to preventing deadlock in some way. Notably, `sync_block::Block` does **not** implement `DeadlockPrevention`. It implements `Flavour`, which is the trait used by the `Mutex`.

However, note that in `sync`, there is an implementation of `Flavour` for all types that implement `DeadlockPrevention`.

There are two reasons for the `DeadlockPrevention` trait:
1. It removes the boilerplate for implementing `Flavour` for `DisableIrq`, `DisablePreemption` and `Spin` as their implementations would all follow the same pattern (seen in the implementation of `Flavour` for types that implement `DeadlockPrevention`)
2. `Queue`, and by extension `WaitQueue`, can be used with different forms of deadlock prevention. It doesn't make sense for `Queue` to accept a `Flavour` because `sync_block::Block` depends on `Queue`, and so that would lead to a recursive implementation. However, it means that `Queue` can be used in interrupt contexts or places where preemption must be disabled because the user can choose the deadlock prevention used for the inner mutex.

Now to answer your questions:
1. > Can you separate out everything that isn't a core sync abstractions and implementation from this PR?

Yes, but it's very hard to settle on an API without actually implementing it. I only saw the use in a `DeadlockPrevention` trait after reimplementing `WaitQueue`.

2. > Should we use our new waker functionality to realize a better waitqueue or general abstraction of a blocking wait condition/event?

Probably? The wait queue could store a list of wakers rather than `TaskRef`s. Nothing major that would change the API.

3. > Isn't sync_block the same as mutex_sleep?

Yes

4. > The VecDeque::new() fn is now const in a recent version of Rust, so we can use that if needed to make Waitqueue::new() const.

:+1: I'll make a PR to bump the Rust version

tsoutsman:
Note: Reached agreement in meeting.

Depends on theseus-os/irq_safety#7 for correctness.

tsoutsman:
It wouldn't be possible to store guards with `lock_api`, so I essentially just copied out the small bit of `spin` that we need to reimplement a spin mutex.

It's only 100 lines, but not a fan TBH. The only thing stopping us from using `spin` directly is the lack of a public `try_lock_weak` method. I'll open an upstream PR adding that method.

tsoutsman:
Should be good now. Fixed a potential deadlock in the current wait queue implementation along the way :smile:.

tsoutsman:
@kevinaboos should be ready for a review now.

kernel/rendezvous/src/lib.rs
```diff
@@ -126,12 +132,16 @@ impl<T> fmt::Debug for ExchangeState<T> {
...
-    let channel = Arc::new(Channel::<T> {
+/// Create a new channel that requires the sender and receiver to rendezvous
+/// when exchanging a message.
+///
+/// The rendezvous channel uses a wait queue internally and hence exposes a
+/// deadlock prevention type parameter. By default it is set to [`Spin`]. See
+/// [`WaitQueue`]'s documentation for more information on when to change this
+/// type parameter.
+#[allow(invalid_type_param_default)]
+pub fn new_channel<T: Send, P: DeadlockPrevention = Spin>() -> (Sender<T, P>, Receiver<T, P>) {
```
kevinaboos: see my msg on discord about how we can improve this to make it more ergonomic for the caller

kernel/mutex_sleep/src/mutex.rs
```diff
@@ -3,14 +3,15 @@ use core::ops::{Deref, DerefMut};
...
+use sync_spin::Spin;
 
 /// A mutual exclusion wrapper that puts a `Task` to sleep while waiting for the lock to become available. 
 /// 
 /// A sleeping `Task` has a "blocked" runstate, meaning that it will not be scheduled in. 
 /// Once the lock becomes available, `Task`s that are sleeping while waiting for the lock
 /// will be notified (woken up) so they can attempt to acquire the lock again.
 pub struct MutexSleep<T: ?Sized> {
-    queue: WaitQueue,
+    queue: WaitQueue<Spin>,
```
kevinaboos: any place we use `WaitQueue<Spin>` like this can be changed to just `WaitQueue`, right? given that it's a default parameter, we shouldn't need to specify it.

libs/dfqueue
kevinaboos, line 1: I think this was committed by accident -- we recently made some changes to the dfqueue module so you likely need to pull those down.

kernel/mutex_sleep/src/mutex.rs
kevinaboos: as in my other comment, i don't think it's needed to specify `<Spin>` here since it's the default (?)

kevinaboos (changes requested):
Generally it looks good. I left some questions about the ergonomics of channel creation, which is my only real concern that needs addressing before merging.

tsoutsman:
The type params on `Sender` and `Receiever` are reversed (`T` then `P`) because default type params have to go last. But users will rarely have to specify that parameter.

kevinaboos (approved):

PR #814, opened 2023-01-16
Use `time::Duration` in `sleep`

tsoutsman:


kevinaboos (approved):
Looks good, thanks!

PR #816, opened 2023-01-16
Update ACPI crates to edition 2021 and clippy fix

hecatia-elegua:
more like a kitchen sink PR:
- update edition (i.e. adjust imports)
- remove a few static_assertion dependencies which I had already written as `const _: ()` a few weeks ago
- make clippy happy, where the only place of interest is in acpi lib.rs, where instead of using `if let` like clippy wanted I'm instead printing that the case is unhandled
- AcpiTables has an uncomplete sentence in docs, what should we put there?


kernel/acpi/acpi_table/src/lib.rs
```diff
@@ -39,8 +36,8 @@ pub struct TableLocation {
 /// All ACPI tables are covered by a single large MappedPages object, 
 /// which is necessary because they may span multiple pages/frames,
 /// and generally should not be multiply aliased/accessed due to potential race conditions.
-/// As more ACPI tables are discovered, the single MappedPages object is 
-pub struct AcpiTables {
+/// As more ACPI tables are discovered, the single MappedPages object is
+pub struct AcpiTables { //^FIXME: is what?
```
hecatia-elegua: ^

kernel/acpi/src/lib.rs
```diff
@@ -131,7 +117,7 @@ pub fn init(rsdp_address: Option<PhysicalAddress>, page_table: &mut PageTable) -
                             debug!("                  path: {:?}", dev_scope.path());
                         }
                     }
-                    _ => { }
+                    _ => { error!("other DMAR entries are currently unhandled") }
```
hecatia-elegua: ^

kernel/acpi/dmar/src/device_scope.rs
```diff
@@ -18,8 +18,8 @@ pub(crate) struct DeviceScope {
     // It would look something like:
     // path: [u16],
 }
-const_assert_eq!(core::mem::size_of::<DeviceScope>(), 6);
-const_assert_eq!(core::mem::align_of::<DeviceScope>(), 1);
+const _: () = assert!(core::mem::size_of::<DeviceScope>() == 6);
+const _: () = assert!(core::mem::align_of::<DeviceScope>() == 1);
```
tsoutsman, line 22:
```suggestion
const _: () = assert_eq!(core::mem::size_of::<DeviceScope>(), 6);
const _: () = assert_eq!(core::mem::align_of::<DeviceScope>(), 1);
```

kernel/acpi/acpi_table/src/lib.rs
tsoutsman:
```suggestion
///
/// As more ACPI tables are discovered, the single MappedPages object is
/// extended to cover them.
pub struct AcpiTables {
```

kernel/acpi/src/lib.rs
tsoutsman: Is this logging statement necessary? I don't think it's an error to encounter an unknown DMAR entry.

tsoutsman (changes requested):
This is great, thank you. Updating Rust editions across the board is on the long list of TODOs.

One small request, could you change the `assert` to `assert_eq` (I commented an example)? This way, if they aren't equal, Rust will tell us the different values rather than just blindly throwing an error.

kernel/acpi/acpi_table/src/lib.rs
kevinaboos: oops! lol, thanks for catching that. What @tsoutsman wrote is sufficient.

kernel/acpi/src/lib.rs
kevinaboos: agreed, we don't use them yet so let's keep it as is.

kevinaboos:
Thanks, looks good to me! Will merge once the existing comments are addressed.

That `const` assert trick is neat. If you feel so inclined, would you mind submitting another PR that replaces the rest of the places where `static_assertions` is used?

kernel/acpi/dmar/src/device_scope.rs
hecatia-elegua: I would love to, but currently only `assert!` works in const-context, `assert_eq!` for whatever reason doesn't just call `assert!`. Maybe it has to do with declarative macros and `assert!` being a compiler built-in.
We can revert the change, though since one can hover over e.g. `DeviceScope` here and see size and align with rust-analyzer (in vscode at least), it should not matter that much

kernel/acpi/src/lib.rs
hecatia-elegua: warn would have made more sense, but I guess adding an `allow` here does the same thing until we need the other entries

kevinaboos: you could lower it to a `debug` or `trace` if you like, but we really don't need to be logging this since it doesn't provide any useful info atm.

hecatia-elegua:
Done

tsoutsman (approved):

kernel/acpi/src/lib.rs
```diff
@@ -106,6 +92,7 @@ pub fn init(rsdp_address: Option<PhysicalAddress>, page_table: &mut PageTable) -
             );
 
             for table in dmar_table.iter() {
+                #[allow(clippy::single_match)]
```
kevinaboos: might as well change it to the `if let` pattern (similar to what you did in the other file)

kevinaboos (changes requested):
Looks good -- just one minor request. Thanks!

kernel/acpi/src/lib.rs
hecatia-elegua: > in acpi lib.rs, where instead of using if let like clippy wanted I'm instead printing that the case is unhandled

I left it like this because I thought a future dev will handle the other variants

kevinaboos:
A recent commit fixed all remaining clippy issues. I think this PR is ready to go once you merge in the latest changes from` theseus_main`.

kernel/acpi/src/lib.rs
kevinaboos: Resolved in the main branch already.

kevinaboos (approved):
Looks good now. Thanks very much!

PR #817, opened 2023-01-16
Combine aarch64 code into main repo workspace

NathanRoyer:
This is a draft PR showing my progress regarding the merge of the aarch64 workspace and the main one.

To build:
```bash
make ARCH=aarch64 FEATURES= CROSS=aarch64-linux-gnu- run
```

List of changes:
- renamed main `logger` → `logger_x86_64` and aarch64 logger `logger_aarch64`
- changed `theseus-os/pl011_qemu` so that the PL011 address isn't a fixed constant anymore; changed `logger_aarch64` to reflect that
- arch-gated some dependencies of nano_core: dependencies either temporarily incompatible like `captain` or forever incompatible such as `vga_buffer`
- made context_switch_regular cross-platform
- modified panic_entry so that it builds for aarch64  (arch-gated some dependencies and handling code - replaced latter with infinite loop)
- introduced an aarch64-specific linker script which is identical to the x86-one, except for the kernel offset
- tiny edits to kernel_config::memory so that addresses aren't x86-canonical on aarch64 + simpler maths
- slightly modified the Makefile so that an aarch64 build is possible (see current build command above); `uefi` boot_spec by default on aarch64
- `memory`: some flag changes, arch-gated vga-specific code, a `tlb_flush_all()` added for each page switch
- EFI firmware is downloaded from [/retrage/edk2-nightly](/retrage/edk2-nightly) from the Makefile via wget
- new json target file: cfg/aarch64-unknown-theseus.json
- removed the aarch64 workspace


tools/uefi_builder/aarch64/Cargo.toml
```diff
@@ -10,7 +10,8 @@ uefi_builder_common = { path = "../common" }
 
 [dependencies.uefi-bootloader]
 artifact = "bin"
-git = "https://github.com/theseus-os/uefi-bootloader"
+git = "https://github.com/NathanRoyer/uefi-bootloader"
+branch = "aarch64-fixes"
 target = "aarch64-unknown-uefi"
```
NathanRoyer: This is temporary, I'll leave Klim some time to review the relevant PR.

tools/uefi_builder/common/src/fs.rs
```diff
@@ -159,7 +159,7 @@ pub fn create_fat_filesystem(
         .truncate(true)
         .open(out_fat_path)
         .unwrap();
-    let fat_size_padded_and_rounded = ((needed_size + 1024 * 64 - 1) / MB + 1) * MB;
+    let fat_size_padded_and_rounded = (needed_size / MB + 10) * MB;
     fat_file.set_len(fat_size_padded_and_rounded).unwrap();
```
NathanRoyer: Hey @tsoutsman, this is a crash workaround; I think you wrote the original size computation, could you tell me the reasoning behind it?

kevinaboos: Modified in #826, tho the fix is still pretty hacky.

tools/uefi_builder/aarch64/Cargo.toml
kevinaboos: Can be removed now :+1:

Makefile
```diff
@@ -171,18 +180,19 @@ full: iso
 iso: $(iso)
 
 ### This target builds an .iso OS image from all of the compiled crates.
-$(iso) $(efi_firmware): clean-old-build build extra_files copy_kernel $(bootloader)
-ifeq ($(boot_spec), uefi)
+$(iso): clean-old-build build extra_files copy_kernel $(iso)-$(boot_spec)
+
+$(iso)-bios: $(bootloader)
+
+$(iso)-uefi: $(efi_firmware)
```
kevinaboos: what's the purpose of these additional targets? They'd be very difficult, if not impossible, to invoke from the command line. It'd be best to avoid splitting the main build-related targets if possible.

Makefile
```diff
@@ -379,6 +389,12 @@ limine:
 	@$(MAKE) -C $(LIMINE_DIR)
 	@$(LIMINE_DIR)/limine-deploy $(iso)
 
+bootloader-none:
+# On UEFI, no bootloader is required
```
kevinaboos: what's this for?

Makefile
```diff
@@ -379,6 +389,12 @@ limine:
...
+
+$(efi_firmware):
```
kevinaboos, line 403: Add a small comment documenting what this does

Makefile
```diff
@@ -879,12 +899,12 @@ QEMU_FLAGS += $(QEMU_EXTRA)
 
 ### Old Run: runs the most recent build without rebuilding
 orun:
-	qemu-system-x86_64 $(QEMU_FLAGS)
+	qemu-system-$(ARCH) $(QEMU_FLAGS)
```
kevinaboos: this is good but I think it'd be better to define a single `QEMU` variable somewhere, which would be `qemu-system-$(ARCH)`.

cfg/aarch64-unknown-theseus.json
```diff
@@ -0,0 +1,16 @@
+{
```
kevinaboos, line 1: let's make this as close to `x86_64-unknown-theseus.json` as possible.

For things you're unsure about, check the target spec of the aarch64-unknown-none built-in rust target

kernel/memory/src/paging/mapper.rs
```diff
@@ -203,8 +206,16 @@ impl Mapper {
...
+        #[cfg(target_arch = "x86_64")]
+        let actual_flags = flags
+            .valid(true)
+            .exclusive(Frames::OWNED);
+
+        #[cfg(target_arch = "aarch64")]
         let actual_flags = flags
             .valid(true)
+            .accessed(true)
+            .page_descriptor(true)
```
kevinaboos: isn't this not necessary? Setting the `accessed` and `page_descriptor` bitflags is already handled in `pte_flags`, right?

kernel/memory/src/paging/mapper.rs
```diff
@@ -275,9 +286,17 @@ impl Mapper {
...
-            .valid(true);
+            .valid(true)
+            .exclusive(true);
+
+        #[cfg(target_arch = "aarch64")]
+        let actual_flags = flags
+            .accessed(true)
+            .page_descriptor(true)
+            .valid(true)
+            .exclusive(true);
```
kevinaboos: same question as above

kernel/memory/src/paging/mod.rs
```diff
@@ -183,13 +190,20 @@ impl PageTable {
...
             Cr3::write(
                 PhysFrame::containing_address(PhysAddr::new_truncate(new_table.p4_table.start_address().value() as u64)),
                 Cr3Flags::empty(),
             )
         };
+
+        #[cfg(target_arch = "aarch64")]
+        set_as_active_page_table_root(new_table.physical_address());
+
+        tlb_flush_all();
```
kevinaboos: on x86, setting CR3 automatically flushes the TLB, so you only need to do this on aarch64. You can also include that as a comment here.

kernel/memory_aarch64/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+/// if a full-address-space identity mapping
+/// was active. When it's enabled, the TTB0_EL1
+/// register is expected to point to a valid
+/// page table (using its physical address).
+pub fn enable_mmu() {
+    SCTLR_EL1.modify(SCTLR_EL1::M::Enable);
+    unsafe { barrier::isb(barrier::SY) };
+}
+
+/// Configures paging for Theseus.
```
kevinaboos, line 98: It'd be great to move (or, where appropriate, duplicate) these internal comments to doc comments atop this function. That way people browsing the docs could see them.

kevinaboos: A good example to model your comments on is the `## Assumed System Configuration` section in `PteFlagsAarch64`.

kernel/memory_aarch64/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+
+            // Allow the MMU to update the DIRTY flag.
+            + TCR_EL1::HD::Enable
+        );
+
+        barrier::isb(barrier::SY);
+    }
+}
+
+/// Installs a page table in the TTBR CPU register
```
kevinaboos: "Install" is a bit of a loaded term. 
```suggestion
/// Sets the given `page_table` as active by updating the TTBR0 register.
```

kernel/memory_aarch64/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+
+
+/// Finds the addresses in memory of the main kernel sections, as specified by the given boot information. 
+/// 
+/// Returns the following tuple, if successful:
+///  * The combined size and address bounds of key sections, e.g., .text, .rodata, .data.
+///    Each of the these section bounds is aggregated to cover the bounds and sizes of *all* sections 
+///    that share the same page table mapping flags and can thus be logically combined.
+///  * The list of all individual sections found. 
+pub fn find_section_memory_bounds<F>(boot_info: &impl BootInformation, translate: F) -> Result<(AggregatedSectionMemoryBounds, [Option<SectionMemoryBounds>; 32]), &'static str>
```
kevinaboos, line 255: i'm a bit confused -- this is being moved into `boot_info`, right?

kernel/nano_core/Cargo.toml
```diff
@@ -68,7 +42,7 @@ optional = true
 cfg-if = "1.0.0"
 
 [features]
-default = [ "bios" ]
+# default = [ "bios" ]
```
kevinaboos: please undo. See my changes in #826, which should make this work by default when you specify `boot_spec=uefi`

kernel/nano_core/linker_higher_half-aarch64.ld
```diff
@@ -0,0 +1,103 @@
+ENTRY(_start)
+OUTPUT_FORMAT(elf64-littleaarch64)
+
+/* the last 2GiB of mappable memory */
+KERNEL_OFFSET = 0xFFFF80000000;
```
kevinaboos:
```suggestion
/*
 * The upper-most 2GiB of accessible virtual addresses on aarch64,
 * because Theseus uses 48-bit virtual addresses with an ASID of 0.
 */
KERNEL_OFFSET = 0x0000_FFFF_8000_0000;
```

kernel/nano_core/src/uefi.rs
```diff
@@ -43,7 +44,41 @@ pub extern "C" fn _start(boot_info: &'static BootInformation) {
...
+            //
+            // The guard page and double fault stack are both one page in size;
+            // the kernel stack is `KERNEL_STACK_SIZE_IN_PAGES` pages.
+            //
+            // Stacks grow downwards on aarch64, meaning that the stack pointer will grow
+            // towards the guard page. That's why we start it at the top (the highest vaddr).
+
+            // Before invoking `rust_entry`, we need to set up:
+            // 1. First arg  (in x0): a reference to the boot info (just pass it through).
+            // 2. Second arg (in x1): the top vaddr of the double fault handler stack.
```
kevinaboos: dedup this comment section and move it out of the asm!() block.

kernel/pte_flags/src/pte_flags_aarch64.rs
```diff
@@ -220,6 +220,7 @@ impl PteFlagsAarch64 {
         Self::from_bits_truncate(
             Self::NORMAL_MEMORY.bits
             | Self::OUTER_SHAREABLE.bits
+            | Self::PAGE_DESCRIPTOR.bits
```
kevinaboos: Is this right? I thought we only wanted to set this for upper-level page table entries?

If it is correct, please update the function docs accordingly.

kevinaboos (changes requested):
Looks pretty good! Very exciting too!

There are a lot of changes related to the `logger`. In my mind, this should be arch-independent -- can we not use the typical pattern of an arch-agnostic `logger` crate that exposes a unified API for the underlying `logger_aarch64` and `logger_x86_64` crates? Then we wouldn't need many changes. 
Actually, I'm not sure why `logger` itself needs to have arch-specific instances. Seems like it only depends on `serial_port_basic`, which I assume would have an equivalent implementation on aarch64.

Nit: for consistency, let's use "aarch64" instead of "AArch64".

Makefile
NathanRoyer: These are the result of our discussion on discord:

> might be good to add new targets called something like qemu-$(ARCH) so you can use that exisintg $(ARCH) make variable to differentiate between the needed EFI file

to which I replied, "good idea, that's even better".

(In this messaged you were mentioning qemu & ARCH but the discussion was about efi firmware download which is required when boot_spec = uefi; this snippet is my interpretation of what you meant, was I wrong?)

Makefile
NathanRoyer: the `$(bootloader)` target is a dependency of the `$(iso)-$(boot_spec)` target when boot_spec = uefi; this is now never reached but was before I introduced this `$(iso)-$(boot_spec)`, as previously the `$(bootloader)` target was a dep of `$(iso)`; I'll remove that

kernel/memory/src/paging/mapper.rs
NathanRoyer: No, it's not. If you use PteFlagsAarch64::new, then these two flags are set in the value you get, but if you use .into, they're not added in PteFlagsAarch64::from(PteFlags) so you have to add them manually.

As a reminder, I was initially setting them in page_table_entry::Entry::set_entry as you first suggested [here](https://github.com/theseus-os/Theseus/pull/699#discussion_r1030829572); later it was decided it's better to have them set explicitely, which is why we thought it was right to set them here, in mapper.

If you're unhappy with that, we can also add a `prepare_for_entry` or `validate` method to each PteFlagsArch, which would add `VALID` on x86 and `(VALID + ACCESSED + PAGE_DESCRIPTOR)` on aarch64.

kernel/memory_aarch64/src/lib.rs
NathanRoyer: Well no, as stated in your "PR #817 todo:" message on discord:

> move find_section_memory_bounds() changes into a new separate PR

There were two changes:

 - find_section_memory_bounds was made compatible with aarch64
 - find_section_memory_bounds was moved from memory_$(ARCH) to boot_info

As this PR is about aarch64 compatiblity, I kept only the first change and reverted the second one for a future PR... isn't that right?

kernel/nano_core/linker_higher_half-aarch64.ld
NathanRoyer: underscores aren't allowed but apart from that, np

kernel/pte_flags/src/pte_flags_aarch64.rs
NathanRoyer: No, it's required in the last of the four levels (even if it has no effect at this point). It's optional in the first 3. Qemu ensures it's set in the last level. I'll update the docs.

Makefile
NathanRoyer: Ok, I added QEMU_BIN

kernel/memory_aarch64/src/lib.rs
kevinaboos: ohhhh, I think we had a slight misunderstanding. I was suggesting we do it in the inverse order, but that's totally fine. We can move `find_section_memory_bounds` into a new crate after this PR is merged, that's fine by me.

cfg/aarch64-unknown-theseus.json
```diff
@@ -0,0 +1,20 @@
...
+  "llvm-target": "aarch64-unknown-none-elf",
+  "max-atomic-width": 128,
+  "merge-functions": "disabled",
+  "os": "theseus",
+  "panic-strategy": "unwind",
+  "relocation-model": "static",
+  "relro-level": "full",
+  "supported-sanitizers": [
+    "kcfi"
+  ],
```
kevinaboos, line 18: do we support this? I'm not aware of any such support in Theseus, unless it's built into Rust or LLVM.

cfg/aarch64-unknown-theseus.json
```diff
@@ -0,0 +1,20 @@
+{
+  "arch": "aarch64",
+  "data-layout": "e-m:e-i8:8:32-i16:16:32-i64:64-i128:128-n32:64-S128",
+  "disable-redzone": true,
+  "features": "+strict-align,+neon,+fp-armv8",
```
kevinaboos: this appears to enable SIMD / hardware floating point instructions. Do we want this? The default `x86_64-unknown-theseus` target explicitly disables hardware floating point in favor of soft floats, which simplifies things.

Typically if hardware floating point is enabled, we also need to handle those registers in the context creation and context switching procedures. So it's probably best to disable it in favor of software floating point, at least initially.

kernel/logger_aarch64/lib.rs
```diff
@@ -45,12 +49,30 @@ static mut LOGGER: Option<QemuVirtUartLogger> = None;
...
+    let mmio_flags = PteFlags::DEVICE_MEMORY
+                   | PteFlags::NOT_EXECUTABLE
+                   | PteFlags::WRITABLE;
+
+    let pages = allocate_pages(1).unwrap();
+
+    let qemu_uart_frame = PhysicalAddress::new_canonical(0x0900_0000);
+    let frames = allocate_frames_at(qemu_uart_frame, 1).unwrap();
+
+    let mapped_pages = page_table.map_allocated_pages_to(pages, frames, mmio_flags).unwrap();
```
kevinaboos: instead of `.unwrap()`, return an error with `.ok_or()`

kernel/logger_aarch64/lib.rs
```diff
@@ -33,7 +34,10 @@ impl Log for QemuVirtUartLogger {
 }
 
 // Global logger Singleton
-static mut LOGGER: Option<QemuVirtUartLogger> = None;
+static mut LOGGER: Option<UartLogger> = None;
```
kevinaboos: avoid unsafety by not using `static mut`. You're already using an irq-safe Mutex for to wrap the `PL011` crate's instance, so perhaps just move it outside (up one layer) so we can directly use it as a static.

kernel/logger_aarch64/lib.rs
```diff
@@ -33,7 +34,10 @@ impl Log for QemuVirtUartLogger {
...
+
+// Mapped Pages for the UART MMIO
+static mut UART_MMIO: MappedPages = MappedPages::empty();
```
kevinaboos: If this just needs to be init'd once, use `spin::Once` or `once_cell`.

If you need to reobtain or modify it, use `spin::Mutex`.

kernel/memory/src/paging/mapper.rs
kevinaboos: My goal here is to simply reduce the number of places in the code where we have to remember to set those two bits. 

How about we just change the implementation of `From<PteFlags> for PteFlagsAarch64` to set those bits, would that be correct? 

I notice that we already do that in  `PteFlagsAarch64::adjust_for_higher_level_pte()`. In your other comment you state that `PAGE_DESCRIPTOR` actually only needs to be set for the lowest-level (P1) PTE, so I think the above suggestion of changing the `From` implementation would work correctly? Try that out.

kernel/logger_aarch64/lib.rs
```diff
@@ -1,18 +1,19 @@
...
 
-type QemuVirtUart = PL011<UART1>;
+use memory::{PhysicalAddress, MappedPages, PteFlags, get_kernel_mmi_ref, allocate_pages, allocate_frames_at};
 
 /// This wraps a UART channel handle.
-pub struct QemuVirtUartLogger {
-    pub(crate) uart: MutexIrqSafe<QemuVirtUart>,
+pub struct UartLogger {
+    pub(crate) uart: MutexIrqSafe<PL011>,
 }
```
kevinaboos: as with my other comment, move the `MutexIrqSafe` outside of the `UartLogger` so we don't need the unsafe `static mut`.

kernel/pte_flags/src/pte_flags_aarch64.rs
kevinaboos: Ok I see, thanks for clarifying. We can't really rely on QEMU to do anything for us or even to perform equally to real hardware. Although perhaps you meant that "QEMU checks (and faults) if it's not set in the last level".

See my other comment about the flags.

kernel/memory_x86_64/src/lib.rs
```diff
@@ -110,19 +110,19 @@ where
...
+        // | (1)  | .text             | end of executable pages       |
+        // | (2)  | .rodata           | start of read-only pages      |
+        // | (3)  | .eh_frame         | part of read-only pages       |
+        // | (4)  | .gcc_except_table | part/end of read-only pages   |
+        // | (5)  | .tdata            | part/end of read-only pages   |
+        // | (6)  | .tbss             | part/end of read-only pages   |
+        // | (7)  | .data             | start of read-write pages     | 
+        // | (8)  | .bss              | end of read-write pages       |
+        // | (9)  | .page_table       | separate .data-like section   |
+        // | (10) | .stack            | separate .data-like section   |
```
kevinaboos: I noticed you updated these shndxs -- those are supposed to match what is shown in the ELF file. I don't think `0` is a valid one, so kindly update those to match the output of `readelf`... unless I'm mis-remembering

tools/uefi_builder/aarch64/Cargo.toml
```diff
@@ -10,7 +10,8 @@ uefi_builder_common = { path = "../common" }
 
 [dependencies.uefi-bootloader]
 artifact = "bin"
-git = "https://github.com/theseus-os/uefi-bootloader"
+git = "https://github.com/NathanRoyer/uefi-bootloader"
+branch = "aarch64-fixes"
```
kevinaboos: do we still need this? I thought https://github.com/theseus-os/uefi-bootloader/pull/2 fixed this.

kevinaboos (changes requested):
left a few more comments

kernel/memory/src/paging/mapper.rs
NathanRoyer: > I notice that we already do that in `PteFlagsAarch64::adjust_for_higher_level_pte()`. In your other comment you state that `PAGE_DESCRIPTOR` actually only needs to be set for the lowest-level (P1) PTE, so I think the above suggestion of changing the `From` implementation would work correctly? Try that out.

it's required by qemu and the architecture that P1 has the flag; but we always need the flag set because all our tables contain page descriptors.

kernel/pte_flags/src/pte_flags_aarch64.rs
NathanRoyer: > QEMU checks (and faults) if it's not set in the last level

that's correct

cfg/aarch64-unknown-theseus.json
NathanRoyer: I don't know what this does, it comes from the built-in aarch64-unknown-none target.

cfg/aarch64-unknown-theseus.json
NathanRoyer: Ok, I'll change it to soft floats.

kernel/memory_x86_64/src/lib.rs
NathanRoyer: I changed that so that they match the output of objdump (zero-based); I'll revert that

tools/uefi_builder/aarch64/Cargo.toml
NathanRoyer: we don't, sorry

kernel/memory/src/paging/mapper.rs
NathanRoyer: > How about we just change the implementation of `From<PteFlags> for PteFlagsAarch64` to set those bits, would that be correct?

Sure

kernel/logger_aarch64/lib.rs
NathanRoyer: I can't get this one out because the log method only has an immutable self; I'm going to use a Once in the global variable

cfg/aarch64-unknown-theseus.json
NathanRoyer: Hmmm rustc reports that the soft-float mode isn't available for aarch64

cfg/aarch64-unknown-theseus.json
kevinaboos: let's test with and without the sanitizer. If generated code is the same, we can include it, if not, let's exclude it

cfg/aarch64-unknown-theseus.json
kevinaboos: solved: there is a different target that specifies `abi: "softfloat"` called `aarch64-unknown-none-softfloat`

cfg/aarch64-unknown-theseus.json
NathanRoyer: The generated code is the same with and without the sanitizer.

kevinaboos (approved):

PR #818, opened 2023-01-17
Fix lints

amab8901:

### DIFF
--- a/kernel/wait_queue/src/lib.rs
+++ b/kernel/wait_queue/src/lib.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::new_without_default)]
 #![no_std]
 
 extern crate alloc;
--- a/kernel/wait_queue/src/lib.rs
+++ b/kernel/wait_queue/src/lib.rs
@@ -236,3 +236,9 @@ impl WaitQueue {
         }
     }
 }
+
+impl Default for WaitQueue {
+    fn default() -> Self {
+        Self::new()
+    }
+}
--- a/kernel/wait_queue/src/lib.rs
+++ b/kernel/wait_queue/src/lib.rs
@@ -236,9 +236,3 @@ impl WaitQueue {
         }
     }
 }
-
-impl Default for WaitQueue {
-    fn default() -> Self {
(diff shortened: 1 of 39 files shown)


kernel/wait_queue/src/lib.rs
```diff
@@ -236,3 +236,9 @@ impl WaitQueue {
         }
     }
 }
+
+impl Default for WaitQueue {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
kevinaboos: why is this needed? If it's not, please remove.

kevinaboos (changes requested):
Thanks! Looks good overall. 

Kindly undo the changes the `kernel/debug_info` and `waitqueue`, which seem accidental and unrelated to this PR.

kernel/wait_queue/src/lib.rs
amab8901: Clippy recommended it

amab8901:
> Thanks! Looks good overall.
> 
> Kindly undo the changes the `kernel/debug_info` and `waitqueue`, which seem accidental and unrelated to this PR.

What about these lints?
https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default
https://rust-lang.github.io/rust-clippy/master/index.html#needless_doctest_main

kernel/wait_queue/src/lib.rs
kevinaboos: is it actually used somewhere? I don't want to add or derive a bunch of unnecessary trait impls until they are actually needed.

kevinaboos: If this is the only place where the lint triggers, I suppose it's fine to add it. But if this is a frequent pattern, I'd rather just disable the entire lint. Code isn't free, this style lint is rather silly imo.

kevinaboos:
Thanks for responding, see my comments on discord -- you could try changing `no_run` to `ignore` and/or changing `fn main` to `fn foo`

kevinaboos (approved):

PR #819, opened 2023-01-18
Fix lints

amab8901:

### DIFF
--- a/kernel/exceptions_full/src/lib.rs
+++ b/kernel/exceptions_full/src/lib.rs
@@ -223,8 +223,8 @@ fn kill_and_halt(
             }
             kill_result
         });
-        if let Err(task_error) = res {
-            println_both!("BUG: kill_and_halt(): Couldn't get current task in order to kill it. {}", task_error);
+        if res.is_err() {
+            println_both!("BUG: kill_and_halt(): Couldn't get current task in order to kill it.");
         }
     }
 
--- a/kernel/exceptions_full/src/lib.rs
+++ b/kernel/exceptions_full/src/lib.rs
@@ -224,7 +224,7 @@ fn kill_and_halt(
             kill_result
         });
         if let Err(task_error) = res {
-            println_both!("BUG and {}: kill_and_halt(): Couldn't get current task in order to kill it.", task_error);
+            println_both!("BUG: kill_and_halt(): Couldn't get current task in order to kill it. {}", task_error);
         }
     }
 
--- a/kernel/exceptions_full/src/lib.rs
+++ b/kernel/exceptions_full/src/lib.rs
@@ -223,8 +223,8 @@ fn kill_and_halt(
             }
             kill_result
         });
-        if let Err(()) = res {
-            println_both!("BUG: kill_and_halt(): Couldn't get current task in order to kill it.");
+        if let Err(task_error) = res {
+            println_both!("BUG and {}: kill_and_halt(): Couldn't get current task in order to kill it.", task_error);
         }
     }
 
--- a/kernel/external_unwind_info/src/lib.rs
+++ b/kernel/external_unwind_info/src/lib.rs
@@ -95,11 +95,11 @@ pub unsafe fn register_unwind_info(
 /// for the given `text_section_base_address`.
 pub unsafe fn deregister_unwind_info(
     text_section_base_address: *mut u8
-) -> Result<(), ()> {
+) -> Result<(), ExternalUnwindInfoError> {
     EXTERNAL_UNWIND_INFO.lock()
         .remove(&VirtualAddress::new_canonical(text_section_base_address as usize))
         .map(|_| ())
-        .ok_or(())
+        .ok_or(ExternalUnwindInfoError::NotFound)
 }
 
 
@@ -120,4 +120,5 @@ pub fn get_unwind_info(
 
 pub enum ExternalUnwindInfoError {
     AlreadyExists,
+    NotFound,
 }
--- a/kernel/external_unwind_info/src/lib.rs
+++ b/kernel/external_unwind_info/src/lib.rs
@@ -80,26 +80,26 @@ pub unsafe fn register_unwind_info(
         unwind_info:  uw_start   .. uw_end,
     };
 
-    uw.try_insert(text_start, uw_info).map_err(|_e| {
-        error!("External unwind info for {text_start:#X} was already registered");
-        ExternalUnwindInfoError::AlreadyExists
-    })?;
-
-    Ok(())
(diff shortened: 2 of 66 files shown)


kernel/signal_handler/src/lib.rs
```diff
@@ -94,11 +94,13 @@ pub enum Signal {
     /// Bad arithmetic operation, e.g., divide by zero.
     /// Analogous to SIGFPE.
     ArithmeticError                 = 3,
+    /// Failed to register signal handlar
+    RegistrationError               = 4,
```
kevinaboos: this is an incorrect solution. You'll need to create a new enum type or struct that represents an Signal registration error, not add a new signal type.

kernel/external_unwind_info/src/lib.rs
```diff
@@ -95,11 +95,11 @@ pub unsafe fn register_unwind_info(
 /// for the given `text_section_base_address`.
 pub unsafe fn deregister_unwind_info(
     text_section_base_address: *mut u8
-) -> Result<(), ()> {
+) -> Result<(), ExternalUnwindInfoError> {
```
kevinaboos, line 98: please update this function's doc accordingly, e.g.,
```rust
/// Returns [`ExternalUnwindInfoError::NotFound` if no unwinding information
/// was registered for the given `text_section_base_address`.
```

kernel/mod_mgmt/src/lib.rs
```diff
@@ -1655,16 +1655,16 @@ impl CrateNamespace {
...
+        const UNLIKELY_PREFIX:         &str = "unlikely."; // the full section prefix is ".text.unlikely."
+        const RODATA_PREFIX:           &str = ".rodata.";
+        const DATA_PREFIX:             &str = ".data.";
+        const BSS_PREFIX:              &str = ".bss.";
+        const TLS_DATA_PREFIX:         &str = ".tdata.";
+        const TLS_BSS_PREFIX:          &str = ".tbss.";
         // const RELRO_PREFIX:            &'static str = "rel.ro.";
-        const GCC_EXCEPT_TABLE_PREFIX: &'static str = ".gcc_except_table.";
-        const EH_FRAME_NAME:           &'static str = ".eh_frame";
+        const GCC_EXCEPT_TABLE_PREFIX: &str = ".gcc_except_table.";
```
kevinaboos, line 1666: fix the above `RELRO_PREFIX` too even though it is commented out.

kernel/page_attribute_table/src/lib.rs
```diff
@@ -60,6 +60,9 @@ pub static FIXED_PAT: PageAttributeTable = PageAttributeTable::from_bytes([
     MemoryCachingType::UncachedMinus  as u8, // 7: 0b111
 ]);
 
+pub enum PageAttributeTableError {
+    Unsupported,
+}
```
kevinaboos: might as well make this an empty struct or enum called `PatUnsupported`, since there's only one variant

kernel/rtc/src/lib.rs
```diff
@@ -133,15 +133,20 @@ pub fn read_rtc() -> RtcTime {
     }
 }
 
-pub fn get_rtc_ticks() -> Result<usize, ()> {
+/// Returns rtc ticks if exists, otherwise returns RtcError::NotFound
+pub fn get_rtc_ticks() -> Result<usize, RtcError> {
```
kevinaboos: this should return an `Option`, that was an oversight.

kernel/serial_port/src/lib.rs
```diff
@@ -66,6 +66,9 @@ static COM2_SERIAL_PORT: Once<Arc<MutexIrqSafe<SerialPort>>> = Once::new();
 static COM3_SERIAL_PORT: Once<Arc<MutexIrqSafe<SerialPort>>> = Once::new();
 static COM4_SERIAL_PORT: Once<Arc<MutexIrqSafe<SerialPort>>> = Once::new();
 
+pub enum SerialPortError {
+    AlreadyExists,
+}
```
kevinaboos: this error/enum variant name is misleading. It sounds like it'd be used to indicate that the serial port already exists, but instead it is actually used to indicate that a serial port already had a data sender. It should be named accordingly.

kernel/task/src/lib.rs
```diff
@@ -77,6 +77,26 @@ use x86_64::registers::model_specific::FsBase;
...
+            TaskError::AlreadyExists => {
+                write!(f, "AlreadyExists")
+            },
+            TaskError::NotFound => {
+                write!(f, "NotFound")
+            },
+        }
+    }
+}
+
```
kevinaboos: 1. This should be moved into the `tls_current_task` mod.
2. This should be named something related to accessing/getting the current task, since that's what it's about. It doesn't have anything to do with the rest of the task subsystem at large.

kernel/rtc/src/lib.rs
```diff
@@ -185,10 +190,10 @@ fn log2(value: usize) -> usize {
 
 /// sets the period of the RTC interrupt. 
 /// `rate` must be a power of 2, between 2 and 8192 inclusive.
-pub fn set_rtc_frequency(rate: usize) -> Result<(), ()> {
+pub fn set_rtc_frequency(rate: usize) -> Result<(), RtcError> {
```
kevinaboos: Let's use a more specific error type (empty enum or struct) that is strictly about an invalid frequency rate value. `InvalidInput` is vague.

kernel/exceptions_full/src/lib.rs
```diff
@@ -223,8 +223,8 @@ fn kill_and_halt(
             }
             kill_result
         });
-        if let Err(()) = res {
-            println_both!("BUG: kill_and_halt(): Couldn't get current task in order to kill it.");
+        if let Err(task_error) = res {
+            println_both!("BUG and {}: kill_and_halt(): Couldn't get current task in order to kill it.", task_error);
```
kevinaboos:
```suggestion
            println_both!("BUG: kill_and_halt(): Couldn't get current task in order to kill it. {}", task_error);
```

kevinaboos (changes requested):
Thanks! Looks pretty good, I only left a few small comments mostly about the ways you handled errors to address the `result_unit_err` lint.

kevinaboos:
Notes for #526:
* `ptr-arg`
* `ptr-eq`
* `redundant-closure`
* `redundant-field-names`
* `redundant-pattern-matching`
* `redundant-static-lifetimes`
* `result-unit-err`

kevinaboos (approved):

PR #821, opened 2023-01-18
Only store runnable tasks in run queue

tsoutsman:
Results in approximately a 40% speedup for the round robin run queue.
This is also a precursor to the planned rework of idle threads as it
makes checking whether a core has any runnable tasks much quicker.

Depends on https://github.com/theseus-os/Theseus/pull/820.


kevinaboos:
I see you closed this, bu I'd still like to make this change. Do you plan to re-do it as a different PR or is there another reason for closing this?

PR #822, opened 2023-01-19
Spread preemption counters across multiple cache lines

tsoutsman:
`preemption_eval` goes from approximately 8s to 6.7s.

It's not a particularly important performance improvement considering that modifying the preemption counter is already fast, but I had this idea in the middle of the night and decided to implement it.

The optimisations don't use any unsafe code.


tsoutsman:
Running a similar benchmark on my x64 Linux laptop yields a 10x improvement, so I think the relatively small improvement is due to the fact that QEMU is a functional emulator, not a cycle-accurate one.

kernel/preemption/src/lib.rs
```diff
@@ -4,14 +4,16 @@
...
 #![feature(negative_impls)]
 
 extern crate alloc;
 
 use core::sync::atomic::{AtomicU8, Ordering};
+use cache_line::CacheLineOptimisedArray;
 
 /// Theseus uses a `u8` to hold each CPU core's ID (apic ID),
-/// so the maximum number of cores it supports is `u8::MAX` (256).
+/// so the maximum number of cores it supports is `u8::MAX` (255).
```
kevinaboos: 255 is the max value, but there are 256 possible values, [0:255] inclusively.

libs/cache_line/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+
+const LINE_SIZE: usize = 64;
+
+/// A cache line optimised array.
+///
+/// It spreads elements across as many cache lines as possible. For example if
+/// only the first four elements of a 256 byte array are being accessed, each
+/// element will have its own cache line.
+///
+/// Indexing is faster if the number of cache lines that the array spreads is a
```
kevinaboos: nit: "spreads" is odd, do you mean "spans"?

libs/cache_line/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+/// power of two.
+#[derive(Debug)]
+// Alignment must be the same as `LINE_SIZE`.
+#[repr(align(64))]
+pub struct CacheLineOptimisedArray<T, const N: usize> {
```
kevinaboos: naming is hard. I feel like "Cache line-optimized array" is a bit vague, specifically "optimized". Can we replace optimized with a different, more specific word that indicates what the array is actually doing to its elements?

possible ideas to get the creative juices flowing: "aligned", "sized", "distributed", "segregated", "placed", "separated". We can bikeshed it a bit if you want

libs/cache_line/Cargo.lock
```diff
@@ -0,0 +1,7 @@
+# This file is automatically @generated by Cargo.
```
kevinaboos: no need to commit this, we don't typically include lockfiles for library crates.

libs/cache_line/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "cache_line"
+version = "0.1.0"
+edition = "2021"
```
kevinaboos, line 6: add a small description and yourself as the author

libs/cache_line/src/lib.rs
```diff
@@ -0,0 +1,151 @@
+#![feature(const_trait_impl)]
```
kevinaboos, line 8: Add a small crate-level doc comment here, can mostly just link to the main array type.

libs/cache_line/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+#![no_std]
+
+use core::{mem, ops};
+
+const LINE_SIZE: usize = 64;
+
+/// A cache line optimised array.
```
kevinaboos: Also we should clearly state that we assume 64-byte cache lines in both the crate-level and here in this type's docs, or alternatively provide a way for the user to set a cache line size as a type parameter

libs/cache_line/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+#[repr(align(64))]
+pub struct CacheLineOptimisedArray<T, const N: usize> {
+    inner: [T; N],
+}
+
+impl<T, const N: usize> const From<[T; N]> for CacheLineOptimisedArray<T, N> {
+    fn from(value: [T; N]) -> Self {
+        Self { inner: value }
+    }
+}
```
kevinaboos: hmm, perhaps i'm misunderstanding this approach, but don't we actually want each element in the array to occupy a separate cache line? That would imply that each `T` needs to be 64-byte aligned in addition to the array itself. This also means that we wouldn't be able to directly construct the array from an existing `[T; N]` value unless the alignment of T was already 64 and the size of T was <= 64.

It seems like the current impl is more focused on putting the multiple elements in a single cache line, which is less beneficial than one element per cache line, at least for the preemption example where we want to avoid cache line thrashing and false sharing.

kevinaboos (changes requested):
Looks good, I left some questions/comments.

One major change request -- currently the only way to access the array itself is by indexing operators, which inherently panic. Can you add a way to access the inner array such that a user can leverage the rest of the methods provided by the core array/slice types?

libs/cache_line/src/lib.rs
tsoutsman: No. This implementation spreads the elements across different cache lines as much as possible without increasing the memory footprint of the array. So, for example, the array of preemption counters will still only take up 256 bytes, but in the majority of cases when we are only using 4 CPUs, each CPU preemption counter will have its own cache line. If you add more CPUs, they will be equally spread out across the cache lines, so CPUs 0 and 4 will be in cache line 1, CPUs 1 and 5 in line 2, CPUs 2 and 6 in line 3, and CPUs 3 and 7 in line 4.

kevinaboos: Right, but at least for things that should be CPU-local, we don't want data for one CPU to share a cache line with that of another CPU. It's possible that other categories of data may be okay with this sharded approach, but I can't see how it's beneficial for this specific case.

Basically, I would expect this to look more like `[CachePadded<T>; N]`, since that's what would actually reduce the bad false sharing (across CPUs) that we want to avoid.

As you know, we're now about to use CPU locals for this instead, but I think it would be good to go through other types that are frequently shared across CPUs and use the `crossbeam_utils::CachedPadded` type to wrap them. There are other places where we want to align data structures up to a cache line size, but the items shared across CPUs are the lowest hanging fruit therein.

PR #823, opened 2023-01-22
Replace old graphics api with Porthole

ouz-a:
_Longer comment will be here_

Adds on top of what's built on #722, replaces entire graphics api with **Porthole**

# Tasks:
- [X] Open the PR
- [ ] Clean the code
- [ ] Create documentation
- [X] Replace book's chapter on graphics with new one


kernel/porthole/src/window.rs
```diff
@@ -0,0 +1,382 @@
...
+    pub fn should_resize_framebuffer(&mut self) -> Result<(), &'static str> {
+        if self.resized() {
+            self.resize_framebuffer()?;
+            self.resized = false;
+        }
+        Ok(())
+    }
+
+    /// Draws the rectangular shape representing the `Window`
+    pub fn draw_rectangle(&mut self, color: Color) -> Result<(), &'static str> {
```
kevinaboos: this function name is a bit vague -- it appears to be a "fill" command in that it fills the entire window with a single color. Let's name it as such.

kernel/porthole/src/window.rs
```diff
@@ -0,0 +1,382 @@
...
+        for pixel in self.frame_buffer.buffer.iter_mut() {
+            *pixel = color;
+        }
+        self.draw_title_border();
+        Ok(())
+    }
+
+    /// Resizes framebuffer after to Window's width and height
+    fn resize_framebuffer(&mut self) -> Result<(), &'static str> {
+        self.frame_buffer = VirtualFrameBuffer::new(self.rect.width, self.rect.height).or(Err(
```
kevinaboos: you almost certainly don't want to call `.or()` here. I think you want `.map_err()` instead.

kernel/porthole/src/window.rs
```diff
@@ -0,0 +1,382 @@
...
+            "Unable to resize framebuffer to current width and height",
+        ))?;
+        Ok(())
+    }
+
+    /// Returns visible part of self's `rect` with relative bounds applied
+    /// e.g if visible rect is `Rect{width: 358, height: 400, x: 0, y: 0}`
+    /// this will return `Rect{width: 358, height: 400, x: 42, y: 0}`
```
kevinaboos: in this example, where does `x: 42` come from? I think you're missing some additional context, e.g., the window's current location within the screen (?)

kernel/porthole/src/framebuffer.rs
```diff
@@ -0,0 +1,229 @@
...
+    rect: Rect,
+    /// Amount of pixels in a line of `Framebuffer`
+    stride: usize,
+    /// Where we start the row
+    row_index_beg: usize,
+    /// Where we end the row
+    row_index_end: usize,
+    /// The index of the row we are getting
+    current_column: usize,
+    _marker: PhantomData<&'a mut T>,
```
kevinaboos: not necessary, can remove.

kernel/porthole/src/lib.rs
```diff
@@ -0,0 +1,552 @@
...
+
+                            _ => {
+                                break;
+                            }
+                        }
+                    }
+                    if x != 0 || y != 0 {
+                        window_manager
+                            .lock()
+                            .update_mouse_position(ScreenPos::new(x as i32, -(y as i32)));
```
kevinaboos: this is misleading -- this doesn't seem to be a `ScreenPos` (which is an absolute location within the screen's coordinate system. Instead, it appears to be a relative offset that describes the difference in position (within the screen's coordinate system) that the current mouse position should be augmented by.

kevinaboos: Either that, or `ScreenPos` itself is a misnomer (?)

kevinaboos (changes requested):
forgot to submit this review, sorry. Some things may have already been addressed.

kernel/porthole/src/window.rs
ouz-a: I added detailed comment explaining what and why of it.

kernel/porthole/src/lib.rs
ouz-a: This was misleading, changed it

PR #824, opened 2023-01-25
Remove static_assertions from most places

hecatia-elegua:
* it is still needed for assert_not_impl_any

I used this search & replace:
`const_assert_eq!\(core::mem::size_of::<([A-Za-z0-9_]*)>\(\),\s`
`const _: () = assert!(core::mem::size_of::<$1>() == `
and changed the others by hand.
We can do something similar if assert_eq ever becomes const.


kevinaboos (approved):
Thanks, looks perfect!

PR #825, opened 2023-01-25
Add pci I/O bar access and use bitfields

hecatia-elegua:
* edition update/clippy as well

I added some TODOs because I'm not done with pci, will do more PRs. I actually stated a question there too: Do we / does the device author always know what kind of address a `determine_mem_base` call returns?
I can look that up in a pci spec, just asking if anybody knows.

### DIFF
--- a/kernel/pci/src/lib.rs
+++ b/kernel/pci/src/lib.rs
@@ -2,20 +2,17 @@
 
 #![allow(dead_code)]
 
-#[macro_use] extern crate log;
 extern crate alloc;
-extern crate spin;
-extern crate port_io;
-extern crate memory;
-extern crate bit_field;
 
 use core::fmt;
 use core::ops::{Deref, DerefMut};
 use alloc::vec::Vec;
+use log::{trace, debug};
+use modular_bitfield::{bitfield, BitfieldSpecifier};
+use modular_bitfield::specifiers::{B1, B30, B28};
 use port_io::Port;
 use spin::{Once, Mutex};
 use memory::PhysicalAddress;
-use bit_field::BitField;
 
 // The below constants define the PCI configuration space. 
 // More info here: <http://wiki.osdev.org/PCI#PCI_Device_Structure>
@@ -52,10 +49,6 @@ pub const PCI_MAX_LATENCY:           u16 = 0x3F;
 pub const MSI_CAPABILITY:           u16 = 0x05;
 pub const MSIX_CAPABILITY:          u16 = 0x11;
 
-/// If a BAR's bits [2:1] equal this value, that BAR describes a 64-bit address.
-/// If not, that BAR describes a 32-bit address.
-const BAR_ADDRESS_IS_64_BIT: u32 = 2;
-
 /// The maximum number of PCI buses.
 const MAX_NUM_PCI_BUSES: u16 = 256;
 /// The maximum number of PCI slots on one PCI bus.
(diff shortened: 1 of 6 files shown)


kernel/pci/src/lib.rs
```diff
@@ -502,8 +503,51 @@ impl DerefMut for PciDevice {
...
+    Reserved = 0x1
+}
+
+/// A base address can either be in I/O space or memory space
+pub enum BaseAddress {
+    IO(PhysicalAddress),
+    Memory(PhysicalAddress)
+}
 
+//TODO: incorporate this somewhere else
```
kevinaboos, line 550: this type can now be deleted since you've replaced it with the two enums

kernel/pci/src/lib.rs
```diff
@@ -420,7 +421,7 @@ impl PciDevice {
 
         self.pci_write(bar_offset, 0xFFFF_FFFF);          // Step 1
         let mut mem_size = self.pci_read_32(bar_offset);  // Step 2
-        mem_size.set_bits(0..4, 0);                       // Step 3
+        mem_size &= 0xFFFFFFF0;                           // Step 3
```
kevinaboos, line 424: is this better? I don't necessarily think it's any clearer

kevinaboos (changes requested):
Overall, i like this PR, but am concerned that the representation of the PCI BAR is just way too complex with the multiple enums and bitfields.

Given the fact that the register layout is so dynamic (e.g., address field changes width and type based on bit 0), I think a design like this would be much simpler (pseudocode):
```rust
/// A PCI Base Address Register (BAR) ... finish docs
#[repr(transparent)]
pub struct BaseAddressRegister(u32);
impl BaseAddressRegister {
    
    /// TODO: here, add const defs for the various magic number

    /// Docs here
    pub fn get(&self) -> PciBar {
        match self.0 & 0x1 {
            0 => {
                let typ = self.0.get_bits(1..=2);
                if typ != 0x0 {
                    panic!("PCI BAR of type {typ:#X} is is unsupported")
                }
                PciBar::MemorySpace {
                    base_addr: PhysicalAddress::new_canonical((self.0 >> 4) as usize),
                    prefetchable: self.get_bit(3),
                }
            }
            1 => PciBar::IoSpace(self.0 >> 2)
        }
    }
}

/// Docs here
pub enum PciBar {
    /// Docs here
    MemorySpace {
        base_addr: PhysicalAddress,
        /// If `true`, reading from this PCI memory can be cached,
        /// and should thus be mapped as write-through, not uncacheable.
        prefetchable: bool,
    }
    /// Docs here
    IoSpace {
        addr: u32,
    }
}
```

and then the `PciDevice` struct would use the `BaseAddressRegister` type in its `bars` field.

kevinaboos:
To answer your question:
> Do we / does the device author always know what kind of address a determine_mem_base call returns?

Yes, but we should verify that by actually checking the PCI BAR content. Your approach in the network device drivers of returning an error if it's not a memory space BAR is correct.

kernel/pci/src/lib.rs
hecatia-elegua: this is temporary since I wanted to remove the bitfield crate and this was the only place left

hecatia-elegua:
Thanks for the comments, I'll think about this some more when I'm stuck on my main tasks.
I do think that your version is more readable, though.

PR #827, opened 2023-01-26
PCI: Use u8 as in the spec, add port docs

hecatia-elegua:
* also adds generic read/write methods for one of the ports

Please merge after https://github.com/theseus-os/Theseus/pull/825.
I'm unsure if this is the right way to, like, do a follow-up PR?
I guess if the content is unrelated, I can start a completely different branch..
would be wild if one could specify that this PR depends on the previous one being done (which is very natural, similar to the git branch concept)

### DIFF
--- a/kernel/pci/src/lib.rs
+++ b/kernel/pci/src/lib.rs
@@ -19,56 +19,62 @@ use bit_field::BitField;
 
 // The below constants define the PCI configuration space. 
 // More info here: <http://wiki.osdev.org/PCI#PCI_Device_Structure>
-pub const PCI_VENDOR_ID:             u16 = 0x0;
-pub const PCI_DEVICE_ID:             u16 = 0x2;
-pub const PCI_COMMAND:               u16 = 0x4;
-pub const PCI_STATUS:                u16 = 0x6;
-pub const PCI_REVISION_ID:           u16 = 0x8;
-pub const PCI_PROG_IF:               u16 = 0x9;
-pub const PCI_SUBCLASS:              u16 = 0xA;
-pub const PCI_CLASS:                 u16 = 0xB;
-pub const PCI_CACHE_LINE_SIZE:       u16 = 0xC;
-pub const PCI_LATENCY_TIMER:         u16 = 0xD;
-pub const PCI_HEADER_TYPE:           u16 = 0xE;
-pub const PCI_BIST:                  u16 = 0xF;
-pub const PCI_BAR0:                  u16 = 0x10;
-pub const PCI_BAR1:                  u16 = 0x14;
-pub const PCI_BAR2:                  u16 = 0x18;
-pub const PCI_BAR3:                  u16 = 0x1C;
-pub const PCI_BAR4:                  u16 = 0x20;
-pub const PCI_BAR5:                  u16 = 0x24;
-pub const PCI_CARDBUS_CIS:           u16 = 0x28;
-pub const PCI_SUBSYSTEM_VENDOR_ID:   u16 = 0x2C;
-pub const PCI_SUBSYSTEM_ID:          u16 = 0x2E;
-pub const PCI_EXPANSION_ROM_BASE:    u16 = 0x30;
-pub const PCI_CAPABILITIES:          u16 = 0x34;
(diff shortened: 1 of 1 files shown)


kernel/pci/src/lib.rs
```diff
@@ -2,73 +2,70 @@
...
+pub const MSIX_CAPABILITY:          u8 = 0x11;
 
-/// If a BAR's bits [2:1] equal this value, that BAR describes a 64-bit address.
-/// If not, that BAR describes a 32-bit address.
-const BAR_ADDRESS_IS_64_BIT: u32 = 2;
-
-/// The maximum number of PCI buses.
-const MAX_NUM_PCI_BUSES: u16 = 256;
+/// The maximum number of PCI buses, minus 1 (to fit into u8).
+const MAX_NUM_PCI_BUSES_MINUS_1: u8 = 255;
```
kevinaboos: rename this `MAX_PCI_BUS_NUMBER` to make it clearer

kernel/pci/src/lib.rs
```diff
@@ -2,73 +2,70 @@
...
-const PCI_CONFIG_ADDRESS_OFFSET_MASK: u16 = 0xFC; 
+const PCI_CONFIG_ADDRESS_OFFSET_MASK: u8 = 0xFC; 
 const CONFIG_ADDRESS: u16 = 0xCF8;
 const CONFIG_DATA: u16 = 0xCFC;
 
+/// specifies the configuration address to be accessed
 static PCI_CONFIG_ADDRESS_PORT: Mutex<Port<u32>> = Mutex::new(Port::new(CONFIG_ADDRESS));
+/// accesses to this:
+/// - will actually generate the configuration access
+/// - will transfer the data to or from this register
```
kevinaboos: These doc comments are nice, but please properly format them such that they look good when viewed after theh Markdown transformation.

kernel/pci/src/lib.rs
```diff
@@ -215,44 +212,52 @@ impl PciLocation {
...
 
+    fn data_port_write(value: u32) {
+        unsafe {
+            PCI_CONFIG_DATA_PORT.lock().write(value);
+        }
+    }
+
+    fn data_port_read() -> u32 {
+        PCI_CONFIG_DATA_PORT.lock().read()
+    }
```
kevinaboos: I don't think these are much of an improvement, they just hide what's going on. Probably best to delete since the existing code was minimal anyway.

kevinaboos (changes requested):
Thanks, this looks pretty good! I'm not sure why we were using `u16`s instead of `u8`s in the first place.

One request -- I'm not sure how I feel about #825, I may not merge that in yet. But the changes from commit https://github.com/theseus-os/Theseus/pull/827/commits/815484fcbb9297f12cd01f1d62230a0dfa222165 are generally good, and also independent from #825. So let's keep this PR strictly about `u16` --> `u8` such that we can address #825 separately.

Basically, please remove all changes related to #825, e.g., the `determine_mem_base` and `BaseAddress` thing.

kernel/pci/src/lib.rs
hecatia-elegua: I can read it better without a lot of `unsafe {}` everywhere and it is similar to the ps2 code.
Oh and I should make this a little clearer: I have some more local changes which also adjust the address port reads/writes (then we have even less unsafe blocks)

hecatia-elegua:
About #825: I guess you mean that the API for getting the base address is bad? I have more changes to come so it will not stay that way and I can read up on pci BARs to find out if returning an enum is sensible

kernel/pci/src/lib.rs
hecatia-elegua: I've made it even more clear by saving the range instead

hecatia-elegua:
Sorry for the second force-push, now it should be good though. The docs are better too now.

kernel/pci/src/lib.rs
kevinaboos: I'll accept it this time because they're private functions, but in general, hiding unsafe blocks without actually assuring the safety of an outside invocation is bad practice

kevinaboos (approved):
Thanks, everything looks good now.

PR #831, opened 2023-02-02
Set up an early dummy TLS data image

kevinaboos:
This fixes a rare issue where a page fault occurs upon a very early panic or error, only when the `mirror_log_to_vga` cfg option is enabled.

This stemmed from the TLS variables that hold the current task and current task ID being accessed (for fault recovery purposes) upon a panic, but those TLS data sections aren't generally available until the tasking subsystem is initialized.

Now, we initialize the TLS data image (using `FsBase` on x86_64`) as soon as we possibly can, which is right after the `nano_core` is parsed and first initialized.
We use a dummy/empty TLS data image for this, which ensures that the TLS register points to a real initialized TLS data image, avoiding weird problems like an unexpected page fault.

Note that the prior behavior wasn't a real problem or had any safety implications, it's just caused an annoying issue where you couldn't get a clean error message when an early error happened on real hardware.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2006,6 +2006,7 @@ dependencies = [
  "lz4_flex",
  "memfs",
  "memory",
+ "no_drop",
  "path",
  "qp-trie",
  "rangemap",
@@ -2014,6 +2015,7 @@ dependencies = [
  "serde",
  "spin 0.9.4",
  "vfs_node",
+ "x86_64",
  "xmas-elf",
 ]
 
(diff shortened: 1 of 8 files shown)


kevinaboos:
Note: still need to drop the initial `tls_data` to avoid wasting heap memory.

kevinaboos:
Done with TLS area cleanup. 

there is one remaining issue: loadable mode doesn't work.
To fix this, we must re-generate the TLS data image AFTER loading the
captain and all of its constituent crates, because in loadable mode, the
CURRENT_TASK / CURRENT_TASK_ID TLS variables aren't in the static kernel
image, they're dynamic TLS sections. We then need to reload `FsBase`
with that new TLS data image.

So we need to create a separate crate that does this, something like
`early_tls_init`, which we can then call from the nano_core when
appropriate, instead of only doing it within `parse_nano_core()`.

Also, we should perform the same `early_tls_init` procedure on each CPU
via `kstart_ap()`.

Also, we probably shouldn't drop the initial TLS data image until `spawn::init()`
has been run on every CPU. We could require passing that initial TLS data image
into `spawn::init()` as proof that it has been held until that point.

PR #833, opened 2023-02-03
Frame allocator improvement: allocate a contiguous chunk that's not found in allocator

Ramla-I:
1. Added two more variants to `AllocationError`:
	- `AddressNotFree` is split into two variants: `AddressNotFree` and `AddressNotFound`.
	- `ContiguousChunkNotFound` for when the first part of a requested range is found, but not the rest

2. `allocate_from_chosen_chunk()` no longer returns a result since it can never return an `Err`.

3. `find_specific_chunk()` can return `Ok`, `AllocationError::ContiguousChunkNotFound` or `AllocationError::AddressNotFound` (not `AddressNotFree` because at the end of this function, we can't say this for certain).

4. `allocate_frames_deferred()` now deals with the different possible` AllocationError`s. 
    - If, after calling `find_specific_chunk`, we get a `ContiguousChunkNotFound` error, then directly try to allocate the requested range. 
    - If `AddressNotFound` is returned, first try to search general list. If not found in general list then try to allocate original requested chunk.


kevinaboos (commented):
Thanks! Looks good, I made some changes to reduce indentation and make it easier to read. I also left some more detailed comments on discord which likely merit further discussion; once those are resolved I'd be happy to merge this in.

Ramla-I:
looks good to me.

kevinaboos (approved):

PR #834, opened 2023-02-03
Use longer timeout for PS/2 keyboard reset command

hecatia-elegua:
This is either a fix for VirtualBox only or also fixes it on real hardware.
When the other real-hw bugs are fixed, I can try, otherwise someone else might.


kevinaboos (approved):
While this isn't sufficient to fix the PS/2 issues on real hardware, there's no problem with merging it in now. We'll continue to investigate those issues.

PR #838, opened 2023-02-14
Fix the APIC timer enable/disable procedure

### LINKED ISSUE #679: APIC timer calibration is inaccurate on KVM --> timeslice period is too short
The APIC timer calibration is wrong or somehow inaccurate when running on KVM (`make run host=yes`), which results in the timeslice period being VERY short, and effectively just  context switching on all CPUs incessantly. Thus, it looks as if the system makes no progress or is hung, but in reality it's running correctly.

By choosing a fixed, very large APIC LVT timer calibration value (e.g., `cfg(apic_timer_fixed)` setting a value of 0x10000000), the system works perfectly well... 

_Originally posted by @kevinaboos in https://github.com/theseus-os/Theseus/issues/676#issuecomment-1295695879_

kevinaboos:
* The problem: the APIC's timer initial count register must be re-written to when re-enabling that timer after it had been previously disabled.
* We store the current timer count when disabling the APIC timer so that we can write that value back to the initial count register the next time we re-enable the APIC timer.

* Also do a bit of refactoring.

* Makes KVM work again, closes #679

### DIFF
--- a/kernel/apic/src/lib.rs
+++ b/kernel/apic/src/lib.rs
@@ -396,9 +396,9 @@ pub struct LocalApic {
     processor_id: u8,
     /// Whether this Local APIC is the BootStrap Processor (the first CPU to boot up).
     is_bootstrap_cpu: bool,
-    /// The current count value from when this lapic's timer was last disabled.
-    /// This is written to the timer's initial count register when the timer is re-enabled.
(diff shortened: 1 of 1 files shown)


kevinaboos:
Update: doesn't actually work. Problem still TBD

PR #843, opened 2023-02-19
Add idle states for Broadwell CPUs

tsoutsman:
The states are from Linux: https://elixir.bootlin.com/linux/latest/source/drivers/idle/intel_idle.c#L1396.

In a future PR, the predefined idle states will be combined with the idle states provided by ACPI (which QEMU doesn't seem to provide) to enter a sleep state in the idle loop.


kernel/cpu/src/arch/x86_64/intel.rs
```diff
@@ -0,0 +1,81 @@
...
+
+pub(crate) enum Model {
+    Broadwell,
+}
+
+impl Model {
+    pub(crate) fn current() -> Option<Self> {
+        let eax = unsafe { core::arch::x86_64::__cpuid(1).eax };
+        let model = (eax >> 4) & 0xf;
+        let extended_model = (eax >> 16) & 0xf;
```
kevinaboos: Does the raw_cpuid crate cover this? If so, we should use it; if not, we should add it there.

kevinaboos (changes requested):
Thanks! Looks good, but I vote we put this content in a separate crate, since `cpu` is strictly about multicore CPU identification and mgmt. We can re-re-name `cpu` or move its functionality into another sub-crate that can be re-exported from `cpu`, but in any case let's try to keep crates small according to separation of concerns.

kevinaboos (approved):
Great, thanks for the changes!

PR #845, opened 2023-02-22
Set FsBase/TPIDR_EL1 in mod_mgmt::TldDataImage::set_as_current_tls_base

NathanRoyer:
First step of porting `task`. Second step is in `preemption` and depends on #727.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2027,6 +2027,7 @@ dependencies = [
  "bincode",
  "bootloader_modules",
  "const_format",
+ "cortex-a",
  "cow_arc",
  "cpio_reader",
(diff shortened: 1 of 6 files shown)


kevinaboos:
Thanks, this looks pretty good, but I actually have some changes already in progress to the TLS area components, related to this to-do item: https://github.com/theseus-os/Theseus/pull/831#issuecomment-1413151840

I still may be able to merge this in, I'll have to see if it's incompatible with my changes.

kevinaboos (approved):
Works for me, I can rebase my prior changes on top of this with no problems. Thanks!

PR #850, opened 2023-02-23
Fix erroneous import in tls_initializer

NathanRoyer:
Apparently I didn't include this important line in #845; the build obviously fails without it for aarch64... sorry about that.

(Initially, the code wrote to TPIDR_EL1 as we're in EL1, but then I remembered that the rust compiler uses TPIDR_EL0 as it has no clue we're in EL1. That's fine, but I forgot to include this line when I made the modification.)


kevinaboos (approved):
Ah thanks, yeah i also missed that with my TLS modifications yesterday. oops!

PR #851, opened 2023-02-23
Move interrupts implementation into x86_64 subdirectory

NathanRoyer:
This leaves the `lib.rs` path available, allowing for another PR to add it as well as an aarch64 implementation.

Hopefully this will allow Github to keep the history of the x86_64 implementation.


kevinaboos (approved):
Thanks, appreciate it.

PR #855, opened 2023-02-24
Only update waker if not waking up same task

kavan-mevada:


tsoutsman:
Thanks, in the current state, I think the change adds unnecessary overhead, but if the check was moved into `Task::set_waker` so that the `Task::inner` was locked only once, we could potentially avoid a clone.

i.e.

```rust
pub fn set_waker(&self, waker: &Waker) {
    let inner = self.inner.lock();.
    if inner.waker.map_or(true, |waker| !waker.will_wake(waker)) {
        inner.waker = Some(waker.clone());
    }
}
```

tsoutsman (changes requested):
Thanks, in the current state, I think the change adds unnecessary overhead, but if the check was moved into `Task::set_waker` so that the `Task::inner` was locked only once, we could potentially avoid a clone.

i.e.

```rust
pub fn set_waker(&self, waker: &Waker) {
    let inner = self.inner.lock();.
    if inner.waker.map_or(true, |waker| !waker.will_wake(waker)) {
        inner.waker = Some(waker.clone());
    }
}
```

PR #857, opened 2023-02-28
Partially port `cpu` to aarch64

NathanRoyer:
This PR implements CpuId for aarch64. Some other `cpu` functions were temporarily implemented as dummies: I think the multicore ARM switch should happen in another PR.

Three CpuId flavors are implemented:
- CpuId: an opaque type wrapping a u32, the compact representation
- MpidrValue: a wrapper for the value which is read from the MPIDR_EL1 register. The inner u64 can be obtained through an unsafe method; this should be used by PSCI code when we want to bring another core up.
- OptionalCpuId, which is used by `task_struct`:
```rust
/// An equivalent to Option<CpuId>, which internally encodes None as
/// `u32::MAX`, which is an invalid CpuId (bits [4:7] of affinity level
/// 0 must always be cleared). This guarantees that it compiles down to
/// lock-free native atomic instructions when using it inside of an atomic
/// type like [`AtomicCell`], as u32 is atomic when running on ARMv8.
```

The PR also contains two unrelated & unimportant modifications which are:
- ~~the removal of a whitespace character at the end of a line in `sleep`'s manifest~~
- the switch to more compact arch gates in `interrupt`

### DIFF
--- a/kernel/cpu/src/aarch64.rs
+++ b/kernel/cpu/src/aarch64.rs
@@ -63,12 +63,7 @@ impl CpuId {
 
 impl MpidrValue {
     /// Obtain the inner raw u64 that was read from the MPIDR_EL1 register
-    ///
-    /// This is not dangerous, the unsafe modifier is simply there to
-    /// indicate that this value should only be used when interacting with
-    /// hardware/peripherals, in driver code for instance; it should not be
-    /// used to identify/reference a CPU core.
-    pub unsafe fn get(self) -> u64 {
+    pub fn get(self) -> u64 {
         self.0
     }
 }
--- a/kernel/cpu/src/aarch64.rs
+++ b/kernel/cpu/src/aarch64.rs
@@ -1,0 +1,110 @@
+use cortex_a::registers::MPIDR_EL1;
+use tock_registers::interfaces::Readable;
+
+/// A unique identifier for a CPU core.
+#[derive(Copy, Clone, Debug)]
+#[repr(transparent)]
+pub struct CpuId(u32);
+
+/// An equivalent top Option<CpuId>, which internally encodes None as
+/// `u32::MAX`, which is an invalid CpuId (bits [4:7] of affinity level
+/// 0 must always be cleared). This guarantees that it compiles down to
+/// lock-free native atomic instructions when using it inside of an atomic
+/// type like [`AtomicCell`], as u32 is atomic when running on ARMv8.
+#[derive(Clone, Copy)]
+pub struct OptionalCpuId(u32);
+
+/// A unique identifier for a CPU core.
+#[derive(Copy, Clone, Debug)]
+#[repr(transparent)]
+pub struct MpidrValue(u64);
+
+/// Returns the number of CPUs (SMP cores) that exist and
+/// are currently initialized on this system.
+pub fn cpu_count() -> u32 {
+    // The ARM port doesn't start secondary cores for the moment.
+    1
+}
+
+/// Returns the ID of the bootstrap CPU (if known), which
+/// is the first CPU to run after system power-on.
+pub fn bootstrap_cpu() -> CpuId {
+    // The ARM port doesn't start secondary cores for the moment,
+    // so the current CPU can only be the "bootstrap" CPU.
+    current_cpu()
+}
+
+/// Returns true if the currently executing CPU is the bootstrap
+/// CPU, i.e., the first procesor to run after system power-on.
+pub fn is_bootstrap_cpu() -> bool {
+    // The ARM port doesn't start secondary cores for the moment,
+    // so the current CPU can only be the "bootstrap" CPU.
+    true
+}
+
+/// Returns the ID of the currently executing CPU.
+pub fn current_cpu() -> CpuId {
+    MpidrValue(MPIDR_EL1.get() as u64).into()
+}
+
+impl CpuId {
+    /// Reads an affinity level from this CpuId
+    ///
+    /// Valid affinity levels are 0, 1, 2, 3
+    pub fn affinity(self, level: u8) -> u8 {
+        assert!(level < 4, "Valid affinity levels are 0, 1, 2, 3");
+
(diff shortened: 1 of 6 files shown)


kernel/cpu/src/aarch64.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// Valid affinity levels are 0, 1, 2, 3
+    pub fn affinity(self, level: u8) -> u8 {
+        assert!(level < 4, "Valid affinity levels are 0, 1, 2, 3");
+
+        (self.0 >> (level * 8)) as u8
+    }
+}
+
+impl MpidrValue {
+    pub unsafe fn get(self) -> u64 {
```
kevinaboos: what's unsafe about this?

kernel/cpu/src/aarch64.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl From<OptionalCpuId> for Option<CpuId> {
+    fn from(val: OptionalCpuId) -> Self {
+        match val.0 {
+            u32::MAX => None,
+            v => Some(CpuId(v)),
+        }
+    }
+}
```
kevinaboos, line 108: So, the manual way of doing this won't result in the same performance as the compiler-internal form of NPO (null pointer optimization, which is what this technique is called). Unfortunately the only way that NPO is exposed outside of rustc itself is via the NonZero* or NonNull types.

If we want to support a value of 0 directly, we can pretty easily get around that with some simple arithmetic. See this SO comment for an example of how: https://stackoverflow.com/questions/30414068/can-i-use-the-null-pointer-optimization-for-my-own-non-pointer-types#comment48914781_30414145

kevinaboos: Really, we should make decisions based on performance. The 3 alternatives are:
* my old alignment-based trick
* your new manual option with u32::MAX as the impossible sentinel value
* an existing NonZeroU32 type with a bit of addition math to adjust the real value.

kernel/cpu/src/lib.rs
```diff
@@ -6,11 +6,8 @@
...
 
-#[cfg(target_arch = "x86_64")]
-pub use apic::{
-    CpuId,
-    cpu_count,
-    bootstrap_cpu,
-    is_bootstrap_cpu,
-    current_cpu,
-};
+#[cfg_attr(target_arch = "x86_64", path = "x86_64.rs")]
```
kevinaboos, line 13: the docs on Line 5 of this crate need an update.

kernel/sleep/Cargo.toml
```diff
@@ -14,7 +14,7 @@ version = "1.2.0"
 path = "../task"
 
 [dependencies.irq_safety]
-git = "https://github.com/theseus-os/irq_safety" 
+git = "https://github.com/theseus-os/irq_safety"
```
kevinaboos: undo changes to this unrelated file

kevinaboos (changes requested):

kernel/cpu/src/aarch64.rs
NathanRoyer: I wanted to indicate that should not be used often; I forgot to document that by the way

kernel/cpu/src/aarch64.rs
NathanRoyer: Oh, indeed we can store the cpu id "plus one" so that zero would be available for use by None. This is completely the same as storing None as u32::MAX except it adds an addition/subtraction in the conversion. I think my solution is simple and the doc explains it clearly. Perhaps I should add a showcase/example in the doc?

kernel/cpu/src/aarch64.rs
NathanRoyer: New doc:
```rust
    /// Obtain the inner raw u64 that was read from the MPIDR_EL1 register
    ///
    /// This is not dangerous, the unsafe modifier is simply there to
    /// indicate that this value should only be used when interacting with
    /// hardware/peripherals, in driver code for instance; it should not be
    /// used to identify/reference a CPU core.
```

kevinaboos: ok i see, then that's not the proper usage of unsafe. Getting an inner raw value is canonically safe in all Rust libraries, e.g., obtaining a pointer to something. _Using_ that raw value to re-obtain a higher-typed value is what's unsafe, e.g., dereferencing a pointer into a reference, or in this case, converting a raw integer into a `MpidrValue`.

We don't and shouldn't offer a public-facing way to convert a raw integer into a `MpidrValue`, so no unsafety is needed.

Rather than using unsafe, the way to indicate that a raw value should not be used to identify a CPU core is to _prevent_ it from happening by accepting only a higher-typed value that originated from a getter function, not a raw value.

kernel/cpu/src/aarch64.rs
kevinaboos: I was suggesting that you could try the 3 options and see which is more performant; alternatively we can implement the 3 options and gate them behind cfgs for future testing. Your solution is indeed simple but it has branches, which is usually what kills perf, not addition.

No need to worry about it now I suppose, but we can experiment soon.

kernel/cpu/src/aarch64.rs
NathanRoyer: > Rather than using unsafe, the way to indicate that a raw value should not be used to identify a CPU core is to _prevent_ it from happening by accepting only a higher-typed value that originated from a getter function, not a raw value.

I agree, however driver crates will need to access that value...

I will remove the unsafe modifier.

kevinaboos: > I agree, however driver crates will need to access that value...

Right, and they can do some completely safely when obtaining that value from a bonafide instance of `MpidrValue` or something that was constructed from it, e.g., `CpuId`. This is the underlying principle of how type-carried invariants work, we're not really reinventing the wheel here.

> I will remove the unsafe modifier.

Great, thanks. Resolved.

kevinaboos (approved):

PR #859, opened 2023-03-01
Add a generic waker/blocker pair that doesn't depend on `task`.

kevinaboos:
* Use it for `JoinableTaskStruct::join()` to make `join()` a blocking call that puts the waiting task to sleep.

This is a funky alternative to #774, except that it doesn't include the changes to `dreadnought`. 

We can implement a regular waker that uses the `waker_generic`, such that it's more ergonomic for all other "normal" users of `waker` to invoke one.


kevinaboos:
@tsoutsman grep my new code for `@tsoutsman` to see where I wrote some notes/thoughts to you. We can discuss on discord or in person.

Kindly don't merge until we discuss

kernel/waker_generic/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+        B: Fn() -> R,
+    {
+        // Care must be taken not to introduce race conditions.
+        // After creating the waker, the wake condition must be checked to ensure
+        // that it did not get woken prior to registering the waker;
+        // otherwise, the waker will never be woken, and this function will loop forever.
+        loop {
+            let woken = self.woken.lock();
+            if *woken {
+                // *woken = false;  // To @tsoutsman: why do this? there's no point, right?
```
tsoutsman: Wakers can be reused. If `woken` isn't reset, subsequent calls to `block` will immediately return.

kernel/waker_generic/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+//! and blocking the [Blocker].
+
+#![no_std]
+
+extern crate alloc;
+
+use alloc::sync::Arc;
+// @tsoutman: as I said before, I think we definitely do need a preemption-safe mutex here.
+//            It's not technically required, but it'd prevent task switch thrashing
+//            between the task that is being blocked and the task that is waking the waker.
```
tsoutsman: We can benchmark this, but I don't think we should use preemption-safe mutexes when they aren't strictly required. Disabling the APIC takes time and also prevents tasks with a higher priority from running.

Disabling preemption is also useless in the likely scenario that the waker task is running on another core.

kernel/task/src/lib.rs
```diff
@@ -409,14 +411,31 @@ impl JoinableTaskRef {
...
+                ScheduleOnDrop { }
+            };
+            blocker.block(block_action);
+        }
+        
+        // @tsoutsman: double-check my reasoning in the comments below.
+        //
+        // Note: previously, we waited for this task to actually stop running;
+        //       however, I don't think there's any need to actually do that.
+        // while self.is_running() { }
```
tsoutsman:
```rust
        // Note: previously, we waited for this task to actually stop running;
        //       however, I don't think there's any need to actually do that.
```
Is this the reasoning?

I don't think the fence will synchronise temporally with the fence in `task_wrapper`. We will be woken before `task_cleanup_final` has run but the exit value has already been stored so IDK if we really care.

tsoutsman (commented):
Sidenote, I think `generic_waker` is a better name for the crate. It's not like we will have multiple `waker_*` crates. Alternatively, just `waker`.

kernel/task/src/lib.rs
kevinaboos: Yeah, basically that's the logic. Since we're only waiting for it to exit, we actually only care about the user-provided entry function being complete, _not_ our `task_wrapper` & friends being complete.

So yea, i think this is okay -- the various `task_cleanup*` functions can continue executing even after that task has been `join`ed by another task.

kernel/waker_generic/src/lib.rs
kevinaboos: Fair enough. The one concern i had is demonstrated by this sequence below
```rust
let mut woken = self.woken.lock();
if *woken {
    *woken = false;
    return;
} else {
    let r = block_action(); // blocks the current thread
    // <-- preemptive task switch occurs here, before `woken` lock can be dropped
    drop(woken);
    drop(r);
}
```
if a preemptive task switch occurs where indicated by my comment, the lock on `woken` will not be dropped. 
Say the "waker" task then runs and attempts to wake the associated waker, it will not be able to obtain the lock on its inner `woken` field and will just have to spin endlessly until the "blocker" task gets to run again and release its lock.

Disabling preemption here may indeed be too expensive, but it would at least prevent that scenario.

Perhaps there's another better solution -- just dropping the lock on `woken` right before invoking the `block_action()`. I think that wouldn't cause potential interleaving issues with the waker, so long as we keep the `wake()` function as is... but I haven't thought deeply about it. Let me know what you think.

kernel/waker_generic/src/lib.rs
kevinaboos: alright, i will restore that statement. 

In the current code i didn't see any examples of wakers being reused. What would the interface to that look like?
I only ask because i wouldn't want to be overly restrictive with the interface we expose to access a waker, e.g., requiring the waker to be consumed.

On a related note -- should we consider making waker `wake()` itself when it gets dropped? I might be overthinking it, but if another caller sets a task's registered waker instance, the existing waker instance (if any) will be overwritten and dropped. If we assume that wakers can never be spuriously woken, then this wouldn't be a good idea. But it's not ideal to have a waker get registered with a task and then it not actually ever be woken up. 
Of course we can just remove that interface for now.... just wondering if you'd need it in `dreadnought`.

kernel/task/src/lib.rs
kevinaboos: > I don't think the fence will synchronise temporally with the fence in `task_wrapper`. We will be woken before `task_cleanup_final` has run but the exit value has already been stored so IDK if we really care.

Good point. We can leave the fence in `task_cleanup_final()` there as a backup in case wakers aren't used or the task isn't joined. But I will also put one in `TaskRef::internal_exit()` right before the waker is woken up.

Side note, i think we have too many useless fences, lol. But I'm not gonna worry about that.

kevinaboos:
Thanks for the feedback! I pushed a new commit.

> Sidenote, I think `generic_waker` is a better name for the crate. It's not like we will have multiple `waker_*` crates. Alternatively, just `waker`.

Sure, I can rename it after this PR. I do plan to implement another `waker` crate that is more "normal", basically one that is similar to your `waker` in #774 (in that it directly depends on `task`), but I'll implement it in terms of the generic waker from this PR.

kernel/waker_generic/src/lib.rs
tsoutsman: E.g. An async I/O driver that wakes the waker every time new data is available.

> I might be overthinking it, but if another caller sets a task's registered waker instance, the existing waker instance (if any) will be overwritten and dropped

Currently, it can only be done through a `JoinableTaskRef` of which only one exists.

> If we assume that wakers can never be spuriously woken

In the context of async, wakers can be woken spuriously, but it leads to the executor unnecessarily polling the future again.

> just wondering if you'd need it in `dreadnought`

I don't think so. If a future doesn't wake the waker, then it shouldn't be woken.

As a side note, we aren't using a `generic_waker::Waker`, we are using a `core::task::Waker` (a glorified vtable) so the drop handler would actually be called in `core::task::Waker::from` in `new_waker`.

kernel/task/src/lib.rs
```diff
@@ -221,6 +221,16 @@ impl TaskRef {
             *self.0.exit_value_mailbox.lock() = Some(val);
             self.0.task.runstate().store(RunState::Exited);
 
+            // Synchronize with the acquire fence in `JoinableTaskRef::join()`,
+            // as we have just stored the exit value that `join()` will load.
+            fence(Ordering::Release);
```
tsoutsman, line 227: We don't need a fence here because we lock (and unlock the mutex via `drop`), which has all the necessary fences to sync the data stored in the mutex.

kernel/waker_generic/src/lib.rs
kevinaboos: Agree on all counts. Thanks!

kernel/task/src/lib.rs
kevinaboos: yes, for now we do use a mutex, so the fence is covered. I should've written a note about this, but I was looking ahead to the future where i might do something without a mutex. But thinking deeper, we'd probably always use one.

kernel/waker_generic/src/lib.rs
tsoutsman: > I think that wouldn't cause potential interleaving issues with the waker, so long as we keep the wake() function as is... but I haven't thought deeply about it.

I think it would in the following order:
1. Blocker drops the lock and is preempted
2. Waker wakes the blocker (unblocking an already unblocked task)
3. Blocker is rescheduled and blocks itself, but won't ever be unblocked.

hm, yea. then we probably do need to disable preemption for correctness.

kernel/task/src/lib.rs
tsoutsman: See new review comment.

kernel/waker_generic/src/lib.rs
kevinaboos: so, to be clear, the problematic interleaving would only occur if we did drop the lock on `woken` _before_ invoking `block_action()`. It wouldn't occur in the current code.

So we only need to disable preemption if we want to be able to reorder the locking and unlocking of `woken`.

EDIT: this is wrong, we do need to always disable preemption.

kevinaboos: One alternative idea is to add a new function that cheaply & quickly disables preemption function (without messing with interrupt chip timers). So it'd be appropriate for this scenario and all other scenarios where we want to atomically block a task and then do something else without the scheduler interfering.

tsoutsman: Yea I was thinking about that as well. I'm not sure how long disabling the APIC takes, but I think it would be faster to soft disable preemption (continue ticks) if it's just for one scheduler invocation.

As a side note, I think Theseus preempts much more often than other operating systems so that's something we could adjust.

tsoutsman (approved):

kernel/waker_generic/src/lib.rs
kevinaboos: So after more thought, we can't reorder the blocking operation and lock release, even if we use preemptive-safe locks. That's fine.

> Yea I was thinking about that as well. I'm not sure how long disabling the APIC takes, but I think it would be faster to soft disable preemption (continue ticks) if it's just for one scheduler invocation.

I don't think it's expensive, but it does require acquiring yet another lock. I can experiment with that disabling preemption during the blocking procedure in a future PR, since it doesn't affect correctness, only potential performance. 

> As a side note, I think Theseus preempts much more often than other operating systems so that's something we could adjust.

Do we? I think our simple preemptive scheduling timer interrupt occurs every 6ms or so, which I thought was on the lower side of things -- most commercial OSes vary their timeslice periods from 0.5 to 8ms, but in any case I'm happy to change it if we determine it'd be valuable.

PR #861, opened 2023-03-02
Use trusted `CpuId` type (almost) everywhere

kevinaboos:
* `CpuId` represents the unique ID of a CPU that is guaranteed to exist on the system, unlike a raw `u8`.

* Almost all code now uses this type, except runqueue stuff because those are liable to change soon.

* x86_64-specific code now uses full 32-bit CPU ID values based on the new `ApicId` type, which is another opaque type that can be converted into `CpuId`.


applications/scheduler_eval/src/lib.rs
```diff
@@ -45,7 +45,7 @@ pub fn main(args: Vec<String>) -> isize {
                 // the benchmark doesn't incorporate work stealing, but the only reason we're having
                 // this problem in the first place is because work stealing isn't implemented so...
                 // TODO: Remove this when work stealing is implemented.
-                .pin_on_core(3)
+                .pin_on_cpu( unsafe { core::mem::transmute(3) })
                 .block()
```
NathanRoyer: Didn't know about this code, do we check that the CpuId exists? how will the spawning fail if it doesn't?

Also there's a space between the parenthesis and `unsafe`

applications/bm/src/lib.rs
```diff
@@ -405,7 +408,7 @@ fn do_ctx() -> Result<(), &'static str> {
 /// This is measured by creating two tasks and pinning them to the same core.
 /// The tasks yield to each other repetitively.
 /// Overhead is measured by doing the above operation with two tasks that just return.
-fn do_ctx_inner(th: usize, nr: usize, child_core: u8) -> Result<u64, &'static str> {
+fn do_ctx_inner(th: usize, nr: usize, child_core: cpu::CpuId) -> Result<u64, &'static str> {
     let start_hpet: u64;
 	let end_hpet: u64;
```
NathanRoyer: `cpu::` looks unnecessary

applications/test_downtime/src/lib.rs
```diff
@@ -417,7 +395,7 @@ pub fn main(args: Vec<String>) -> isize {
         
         let taskref1  = new_task_builder(fault_graphics_task, arg_val)
             .name(String::from("fault_graphics_task"))
-            .pin_on_core(2)
+            .pin_on_cpu( unsafe { core::mem::transmute(2_u32) })
             .spawn_restartable(None)
```
NathanRoyer: Same as above, seems dangerous + there is a whitespace between `(` & `unsafe`

applications/test_scheduler/src/lib.rs
```diff
@@ -5,15 +5,20 @@ extern crate alloc;
...
 
 use alloc::string::String;
 use alloc::vec::Vec;
+use cpu::CpuId;
 
 
 pub fn main(_args: Vec<String>) -> isize {
+    // temp hack to get a fixed CpuId(1), which is unsafe.
+    let cpu_1: CpuId = unsafe { core::mem::transmute(1_u32) };
+
```
NathanRoyer: Same as above, seems dangerous

kernel/ap_start/src/lib.rs
```diff
@@ -32,26 +24,26 @@ pub static AP_READY_FLAG: AtomicBool = AtomicBool::new(false);
 
 /// Temporary storage for transferring allocated `Stack`s from 
 /// the main bootstrap processor (BSP) to the AP processor being booted in `kstart_ap()` below.
-static AP_STACKS: MutexIrqSafe<BTreeMap<u8, NoDrop<Stack>>> = MutexIrqSafe::new(BTreeMap::new());
+static AP_STACKS: MutexIrqSafe<BTreeMap<u32, NoDrop<Stack>>> = MutexIrqSafe::new(BTreeMap::new());
```
NathanRoyer, line 28: Can't we use CpuId directly here?

kernel/apic/src/lib.rs
```diff
@@ -38,6 +36,20 @@ const _APIC_TIMER_MODE_TSC_DEADLINE:   u32 = 0b10 << 17;
 pub const LOCAL_APIC_LVT_IRQ:          u8  = 0x22;
 
 
+/// A unique identifier for a Local APIC, i.e., a CPU on x86_64.
+#[derive(
+    Clone, Copy, Debug, Display, PartialEq, Eq, PartialOrd, Ord,
+    Hash, Binary, Octal, LowerHex, UpperHex,
+)]
```
NathanRoyer, line 43: Why would we need Ord to be implemented? Betrays the "opaque" aspect a bit imo

kernel/cpu/src/aarch64.rs
```diff
@@ -50,20 +33,25 @@ pub fn current_cpu() -> CpuId {
...
+)]
+#[repr(transparent)]
+pub struct MpidrValue(u64);
+
+impl MpidrValue {
+    /// Reads an affinity level from this `MpidrValue`.
     ///
-    /// Valid affinity levels are 0, 1, 2, 3
+    /// Panics if an affinity value Valid affinity levels are 0, 1, 2, 3. Panics i
     pub fn affinity(self, level: u8) -> u8 {
```
NathanRoyer: I think you mistyped this comment

kernel/cpu/src/aarch64.rs
```diff
@@ -50,20 +33,25 @@ pub fn current_cpu() -> CpuId {
...
         assert!(level < 4, "Valid affinity levels are 0, 1, 2, 3");
-
         (self.0 >> (level * 8)) as u8
```
NathanRoyer: This code is wrong, affinity level 3 is not at bits [24:31] in an MpidrValue, it's in bits [32:39]. This code is valid for a compact CpuId u32.

kernel/cpu/src/aarch64.rs
```diff
@@ -1,27 +1,10 @@
...
-/// type like [`AtomicCell`], as u32 is atomic when running on ARMv8.
-#[derive(Clone, Copy, PartialEq, Eq)]
-#[repr(transparent)]
-pub struct OptionalCpuId(u32);
-
-/// A unique identifier for a CPU core.
-#[derive(Copy, Clone, Debug)]
-#[repr(transparent)]
-pub struct MpidrValue(u64);
-
```
NathanRoyer, line 24: I personally think people should first learn about the types when they open a file, which is why I was placing these definitions here. Was there a motivation for moving them lower?

Also, CpuId isn't imported from `super` so using it shouldn't compile?

kernel/cpu/src/lib.rs
```diff
@@ -15,3 +15,31 @@
...
+        self.0
+    }
+
+    /// A temporary function (will be removed later) that converts the given `CpuId`
+    /// into a `u8`, panicking if its inner `u32` value does not fit into a `u8`.
+    pub fn into_u8(self) -> u8 {
+        self.0
+            .try_into()
+            .unwrap_or_else(|_| panic!("couldn't convert CpuId {self} into a u8"))
+    }
```
NathanRoyer, line 44: This must be defined per-arch, as I'm expecting this to fail even for some quad core ARMv8 SoCs, as I said on discord.

kernel/ioapic/src/lib.rs
```diff
@@ -140,18 +135,18 @@ impl IoApic {
...
     ///    For example, 0x20 is the PIT timer, 0x21 is the PS2 keyboard, etc.
-    pub fn set_irq(&mut self, ioapic_irq: u8, lapic_id: u8, irq_vector: u8) {
+    pub fn set_irq(&mut self, ioapic_irq: u8, apic_id: ApicId, irq_vector: u8) {
         let low_index: u32 = 0x10 + (ioapic_irq as u32) * 2;
         let high_index: u32 = 0x10 + (ioapic_irq as u32) * 2 + 1;
 
         let mut high = self.read_reg(high_index);
         high &= !0xff000000;
-        high |= (lapic_id as u32) << 24;
+        high |= apic_id.value() << 24;
```
NathanRoyer, line 168: I suggest checking that apic_id < 256

kernel/libtest/src/lib.rs
```diff
@@ -58,18 +60,16 @@ pub fn check_myrq() -> bool {
...
+pub fn pick_free_core() -> Result<CpuId, &'static str> {
 	// a free core will only have 1 task, the idle task, running on it.
 	const NUM_TASKS_ON_FREE_CORE: usize = 1;
 
-	// try with current core -1
-	let child_core: u8 = CPU_ID!() as u8 - 1;
-	if nr_tasks_in_rq(child_core) == Some(NUM_TASKS_ON_FREE_CORE) {return Ok(child_core);}
-
-	// if failed, iterate through all cores
+	// if failed, iterate through all CPUs
```
NathanRoyer: comment could be outdated (nothing is done above so nothing can fail)

kernel/preemption/src/lib.rs
```diff
@@ -9,26 +9,30 @@
...
 
-/// Theseus uses a `u8` to hold each CPU core's ID (apic ID),
-/// so the maximum number of cores it supports is `u8::MAX` (256).
-const MAX_CPU_CORES: usize = u8::MAX as usize;
+/// Theseus previously used a `u8` to hold each CPU core's ID
+/// so the maximum number of cores it supports is 256.
+///
+/// In the future, we will use CPU-local storage for each CPU's preemption count,
+/// so we won't need to use a system-wide primitive array here.
+const MAX_CPU_CORES: usize = 2_usize.pow(u8::BITS);
```
NathanRoyer: Why use pow? `u8::MAX` was simpler to understand

kernel/preemption/src/lib.rs
```diff
@@ -9,26 +9,30 @@
...
 static PREEMPTION_COUNT: [AtomicU8; MAX_CPU_CORES] = [ATOMIC_U8_ZERO; MAX_CPU_CORES];
 
 
 /// Prevents preemption (preemptive task switching) from occurring
 /// until the returned guard object is dropped.
 pub fn hold_preemption() -> PreemptionGuard {
-    let cpu_id = apic::current_cpu();
-    let prev_val = PREEMPTION_COUNT[cpu_id as usize].fetch_add(1, Ordering::Relaxed);
+    let cpu_id = cpu::current_cpu();
+    let prev_val = PREEMPTION_COUNT[cpu_id.value() as usize].fetch_add(1, Ordering::Relaxed);
```
NathanRoyer, line 35: Would you agree if we used the `u8` representation to index into this? that way it would work as-is on most ARMv8 CPUs and panic on other ones. When we have Cpu Local Storage we'd change this.

NathanRoyer (changes requested):
Looks generally good

applications/scheduler_eval/src/lib.rs
kevinaboos: yea, this was a temp fix since it's only ever used in evaluation/testing code.
I've come up with a better way, though, basically `impl TryFrom<u32> for CpuId`

applications/bm/src/lib.rs
kevinaboos: yep, thanks.

kernel/ap_start/src/lib.rs
kevinaboos: unfortunately not, because the origin of this value is an ACPI system table, which is a raw value that must be used in order to actually bring up this particular APIC/CPU. So we couldn't possibly create a `CpuId` or `ApicId` at this point because that APIC/CPU wouldn't yet exist.

kernel/apic/src/lib.rs
kevinaboos: the type isn't actually opaque since you can access the inner value. It's opaque in the sense that you cannot instantiate one yourself because the inner value is hidden. But this raises a good point; I will clarify the docs.

There's no harm in implementing `Ord`, in fact it's necessary to allow it to be used in an ordered/sorted type like `BTreeMap`.

kevinaboos: Actually the docs don't use the word opaque, so they're already fine.

kernel/cpu/src/aarch64.rs
kevinaboos: yep, fixed. Thanks!

kernel/cpu/src/aarch64.rs
kevinaboos: it's more canonical to group a type's definition with its implementations, otherwise they can be hard to find. The only place you can't do that is when the impls must be in different modules, e.g., due to different target archs.

> Also, CpuId isn't imported from `super` so using it shouldn't compile?

Huh, that is indeed weird. It builds on my system, let me see what the problem is.

kevinaboos: The `cpu` crate is not included in an aarch64 build, that's why. I'll let you deal with that in a future PR.

kernel/ioapic/src/lib.rs
kevinaboos: ah, thanks for catching this. Checking that apic_id < 256 isn't actually the right solution, but rather I think it's necessary to use a bitshift of 24 for only xapic, and no bitshift for x2apic. I'll double-check.

kevinaboos: Turns out that you were indeed correct, that's a limitation of IOAPIC redirection tables. man, sometimes i hate x86. 
Anyway, i've fixed it now via an explicit check.

kernel/cpu/src/aarch64.rs
kevinaboos: ah, thanks for catching my mistake, that was accidental.

kevinaboos: I have undone this change now, but personally I believe that the `affinity` method should _only_ be implemented on `MpidrValue` since it is specific to aarch64 only, whereas `CpuId` should be kept arch-agnostic.

I"ll let you make that change in your next PR.

In this way, `CpuId` is analogous to `PteFlags`, `ApicId` is like `PteFlagsX86_64`, and `MpidrValue` is like `PteFlagsAarch64`.

kernel/cpu/src/lib.rs
kevinaboos: that's exactly the point -- it's a method used for temporary compatibility that's **explicitly designed to fail** such that we know when a CPU ID is being treated as a `u8` invalidly. 

Future changes to both x86 and aarch64 code would be made which would then make this method unnecessary.

kernel/libtest/src/lib.rs
kevinaboos: yep, fixed.

kernel/preemption/src/lib.rs
kevinaboos: `u8:MAX` was incorrect. It should be 256, not 255. I can just use a const instead.

kernel/preemption/src/lib.rs
kevinaboos: no, that'd require a bunch of extra type wrappers to implement the `Index` trait for `u8`, as it typically is only implemented for `usize`. 
There's no point messing with this, as it'll change soon anyway, and it already will panic as expected for ID values that are above 255.

kevinaboos: If/when you actually encounter CPU ID values larger than `u8::MAX` on aarch64, then we can switch to a map structure like a `BTreeMap`.

PR #870, opened 2023-03-06
Debug state for PS/2 real hardware testing with logs

hecatia-elegua:
- just prints out debug info to vga
- should loop on error so you can see logs

Please try this as is and type some stuff, then update `keyboard.keyboard_scancode_set(ScancodeSet::Set1)?;` to use Set2 and maybe Set 3 as well. Some of these might print bogus scancode errors, I'm interested in seeing "ps2_keyboard_handler" logs.

Then also try putting an infinite loop in device_manager after the ps2 setup (I mean after all the ps2 setup is done, not just the controller), so I can see the debug logs of the setup.

(Of course this is just for testing, the real PR will come later)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -186,6 +186,7 @@ full: iso
 iso: $(iso)
 
 ### This target builds an .iso OS image from all of the compiled crates.
+$(iso) : export override THESEUS_CONFIG += mirror_log_to_vga
 $(iso): clean-old-build build extra_files copy_kernel $(iso)-$(boot_spec)
 
 ## This target is invoked by the '$(iso)' target when boot_spec = 'bios'.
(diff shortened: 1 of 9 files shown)


kevinaboos:
thanks! I have left comments and log files on discord.

In short, it works on my real hw and in QEMU normally! However, it doesn't work in QEMU with KVM enabled (`make run host=yes`), as that results in the virtual hw being different. So something's still not fully correct.

hecatia-elegua:
I'd love to have automatic test runs on different emulators for making changes to something like this.

kevinaboos:
yeah, that'd be great. We could do a qemu test easily, but everything else would be difficult, even kvm. Not sure about other emulators like VirtualBox or VMware.

kevinaboos:
Shall we close this now that #884 has been merged in? Or did you want to keep it around for future usage?

hecatia-elegua:
I would like to keep it around, especially for others to test real hardware as well.
Might update it. Even better (and I will try this sometime) would be some feature/build option just for debug, or some more flexibility with respect to logging. Something like the log crate but for choosing in which scope debug/trace is enabled.

PR #876, opened 2023-03-07
Added aarch64 scheduling code

NathanRoyer:
Two unrelated changes to `interrupts` & `scheduler`
- implements preemptive scheduling via timer interrupts on aarch64
- a `time::ClockSource` was implemented for aarch64 in `interrupts` using the physical system timer

### DIFF
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -249,8 +249,11 @@ pub fn eoi(irq_num: InterruptNumber) {
     gic.end_of_interrupt(irq_num);
 }
 
+// A ClockSource for the time crate, implemented using
+// the System Counter of the Generic Arm Timer. The
+// period of this timer is computed in `init` above.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
-pub struct PhysicalSystemCounter;
+struct PhysicalSystemCounter;
 
 impl ClockSource for PhysicalSystemCounter {
     type ClockType = Monotonic;
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -63,10 +63,16 @@ pub struct ExceptionContext {
     esr_el1: EsrEL1,
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub enum EoiBehaviour {
+    CallerMustSignalEoi,
+    HandlerHasSignaledEoi,
+}
+
 /// Return value:
 /// - true if you sent an End Of Interrupt signal in the handler
 /// - false if you want the caller to do it for you after you return
-type HandlerFunc = extern "C" fn(&ExceptionContext) -> bool;
+type HandlerFunc = extern "C" fn(&ExceptionContext) -> EoiBehaviour;
 
 // called for all exceptions other than interrupts
 fn default_exception_handler(exc: &ExceptionContext, origin: &'static str) {
@@ -75,7 +81,7 @@ fn default_exception_handler(exc: &ExceptionContext, origin: &'static str) {
 }
 
 // called for all unhandled interrupt requests
-extern "C" fn default_irq_handler(exc: &ExceptionContext) -> bool {
+extern "C" fn default_irq_handler(exc: &ExceptionContext) -> EoiBehaviour {
     log::error!("Unhandled IRQ:\r\n{:?}\r\n[looping forever now]", exc);
     loop {}
 }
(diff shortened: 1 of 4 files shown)


kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -90,6 +96,15 @@ pub fn init() -> Result<(), &'static str> {
         static __exception_vector_start: extern "C" fn();
     }
 
+    let counter_freq_hz = CNTFRQ_EL0.get() as f64;
+    let fs_in_one_sec = 1_000_000_000_000_000.0;
+
+    // https://doc.rust-lang.org/reference/expressions/operator-expr.html
+    // "Casting from a float to an integer will round the float towards zero"
+    let period_femtoseconds = (fs_in_one_sec / counter_freq_hz) as u64;
```
kevinaboos: we should avoid using floating point types in general (they're slow and require special hw or emulation by the compiler), especially in the core kernel subsystems.

Here, you're converting an unsigned integer into a float and then converting it back into an unsinged integer, so you might as well just treat it as an unsigned integer for all operations.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -117,23 +132,29 @@ pub fn init() -> Result<(), &'static str> {
...
-    if let Err(existing_handler) = register_interrupt(CPU_LOCAL_TIMER_IRQ, timer_handler) {
-        if timer_handler as *const HandlerFunc != existing_handler {
-            return Err("An interrupt handler has already been setup for the timer IRQ number");
+/// This function registers an interrupt handler for the CPU-local
+/// timer, enables the routing of this interrupt in the GIC, and
+/// turns the timer on, when `enable` is true.
+///
+/// When `enable` is false, the handler is deregistered, routing is
+/// disabled in the GIC, and the timer is turned off.
+pub fn enable_timer_interrupts(enable: bool, timer_tick_handler: HandlerFunc) -> Result<(), &'static str> {
```
kevinaboos: this function does too many things. I think it's supposed to be a counterpart to `LocalApic::enable_lvt_timer()`, but it includes both initialization, registration, and enabling.

It should be possible to initialize and register an interrupt handler in one function, and then enable/disable it in a different function. This is required for `preemption`, among other crates.

kernel/scheduler/src/lib.rs
```diff
@@ -53,15 +53,32 @@ pub fn init() -> Result<(), &'static str> {
         })
     }
 
-    #[cfg(not(target_arch = "x86_64"))] {
+    #[cfg(target_arch = "aarch64")] {
+        interrupts::enable_timer_interrupts(true, aarch64_timer_handler)
+    }
+
+    #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))] {
```
kevinaboos: we can remove this whole log statement if you now support preemptive task switching on aarch64. That was just meant to be a placeholder for you.

kernel/scheduler/src/lib.rs
```diff
@@ -53,15 +53,32 @@ pub fn init() -> Result<(), &'static str> {
...
+}
+
 /// The handler for each CPU's local timer interrupt, used for preemptive task switching.
 #[cfg(target_arch = "x86_64")]
 extern "x86-interrupt" fn lapic_timer_handler(_stack_frame: x86_64::structures::idt::InterruptStackFrame) {
+    cpu_local_timer_tick_handler()
+}
+
+// Cross platform scheduling code
+fn cpu_local_timer_tick_handler() {
```
kevinaboos, line 78: this is nice, i like this design!

One request -- we should probably change the below "LAPIC_TIMER_TICKS` and the associated logging statement to something non-LAPIC-specific, just as you did with cpu_local_timer_handler

kevinaboos (changes requested):
looks good, just a few minor requests.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: You're totally right! It used to be two parts in my code: `(1/<a lot>) * <even more>`, which would have been impossible with integers, but now that it's `<a lot>/<a bit less>` integers are fine :ok_hand: thank you

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: > I think it's supposed to be a counterpart to `LocalApic::enable_lvt_timer()`

Ah, that wasn't my intention; I'll split it in two.

kernel/scheduler/src/lib.rs
NathanRoyer: > One request -- we should probably change the below "LAPIC_TIMER_TICKS` and the associated logging statement to something non-LAPIC-specific, just as you did with cpu_local_timer_handler

Np, done.

By the way, when I want to disable debugging statements like you did here, I usually just put an if statement rather than a config:
```rust
    if false { // stripped away on optimized builds
        use core::sync::atomic::{AtomicUsize, Ordering};
        static CPU_LOCAL_TIMER_TICKS: AtomicUsize = AtomicUsize::new(0);
        let _ticks = CPU_LOCAL_TIMER_TICKS.fetch_add(1, Ordering::Relaxed);
        log::info!("(CPU {}) CPU-LOCAL TIMER HANDLER! TICKS = {}", cpu::current_cpu(), _ticks);
    }
```
which has three benefits:
- it's also stripped away on release builds
- the content is still checked for errors
- last one is less important: to re-enable it, you just have to double-click false and type true, which is shorter than selecting the cfg stuff.

I didn't replace your way of doing it but I thought I'd mention it.

kevinaboos: oh that's a good way of doing it -- thanks for pointing that out! I wish rust supported syntax like `#[cfg(false)]`, but your point of it being checked during compilation is a key improvement.

You can definitely go ahead and change that here, i'll use that pattern elsewhere when i encounter the need for it.

kevinaboos: Another thing we should do (these can be done in a future PR) is to promote the `EoiBehavior` type to be used across architectures. I've frequently considered doing something like that on x86, so it'd be welcome there too, especially once we offer an arch-agnostic handler function type that is wrapped in an arch-specific entry point.

kevinaboos (approved):

PR #877, opened 2023-03-07
Jump to captain from nano_core, fix cpuid on aarch64

NathanRoyer:
- `nano_core` depends on `captain` and jumps to it just like on x86_64
- `captain::init` calls are almost entirely arch-gated:
```
- TSC stuff
- device manager initialization
- special stacks
- exceptions_full initialization
- multicore bringup
- tlb_shootdown
- multiple_heaps
- window_manager
- console initialization
- first_app start
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3669,6 +3669,7 @@ dependencies = [
  "stack",
  "static_assertions",
  "task_struct",
+ "waker_generic",
 ]
 
 [[package]]
@@ -4360,6 +4361,23 @@ dependencies = [
  "log",
  "scheduler",
  "task",
+]
+
+[[package]]
+name = "waker"
+version = "0.1.0"
+dependencies = [
+ "spin 0.9.4",
+ "task",
+ "waker_generic",
+]
+
+[[package]]
+name = "waker_generic"
+version = "0.1.0"
+dependencies = [
+ "preemption",
+ "spin 0.9.4",
 ]
 
 [[package]]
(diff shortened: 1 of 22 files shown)


kevinaboos (changes requested):
In general everything looks fine, functionality-wise.

I understand some of these cfg statements are temporary, but there are just so many that it makes the code hard to read. I also cannot tell which ones are permanent arch-specific cfgs, and which ones are temporary... adding comments to the temporary ones would help this situation.

I think having lots of cfg statements (temporary or not) is unavoidable and thus okay in the captain, but i'd like to avoid that in other crates.
In general I'd prefer to cfg-gate **entire crates** in or out, rather than putting a cfg on each specific call or use of that crate's functions/types. Specifically in this PR, `catch_unwind` and `fault_log` can probably just be completely cfg'd out until you port them to aarch64, and likely `app_io` too. You can make the decision for these and others.

One thing I will do shortly is to wrap all of the `ap_realmode*` stuff in an x86-specific multicore bringup type, which should help reduce some of the complexity of that part of the `nano_core`-`captain` interface.

kevinaboos:
See #878. If you merge in those changes, it should make some things a bit cleaner.

NathanRoyer:
> Specifically in this PR, `catch_unwind` and `fault_log` can probably just be completely cfg'd out until you port them to aarch64, and likely `app_io` too. You can make the decision for these and others.

I agree :+1: I merged in #878 & applied your suggestions.

kevinaboos (approved):
looks great now. I made a few tiny changes, nothing you need to review though.

PR #879, opened 2023-03-08
Add a standard `waker` that directly uses task (un)blocking

kevinaboos:


tsoutsman (approved):
`TaskRef::join` has an identical waker implementation. Is the eventual goal to move this `waker` implementation into `task` so that `join` can use this implementation?

kevinaboos:
> `TaskRef::join` has an identical waker implementation. Is the eventual goal to move this `waker` implementation into `task` so that `join` can use this implementation?

Yeah, they're similar though not wholly identical, so no they wouldn't be combined... at least not in the current form.

If we changed the `waker_generic::Blocker` struct to hold its `block_action` function as a field instead of accepting it as a parameter in `block()`, then the two impls would be more similar. I weighed this design choice in my mind when implementing #859, but didn't see much of an advantage to that design at that time.
But if that's a pattern that would make things more ergonomic for waker usage elsewhere, e.g., in dreadnought, then i could change the API of `waker` and _potentially_ dedup them.

PR #881, opened 2023-03-13
Small scheduling improvements

NathanRoyer:
- Added preemption rate control for aarch64
- Changed the style of a debug-gate (cfg() → constant boolean)
- Removed the uefi gate in captain for aarch64

### DIFF
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -8,12 +8,13 @@ use tock_registers::interfaces::Writeable;
 use tock_registers::interfaces::Readable;
 use tock_registers::registers::InMemoryRegister;
 
+use kernel_config::time::CONFIG_TIMESLICE_PERIOD_MICROSECONDS;
 use gic::{qemu_virt_addrs, ArmGic, InterruptNumber, Version as GicVersion};
 use irq_safety::{RwLockIrqSafe, MutexIrqSafe};
 use memory::get_kernel_mmi_ref;
 use log::{info, error};
 
-use time::{Duration, Monotonic, ClockSource, Instant, Period, register_clock_source};
+use time::{Monotonic, ClockSource, Instant, Period, register_clock_source};
 
 // This assembly file contains trampolines to `extern "C"` functions defined below.
 global_asm!(include_str!("table.s"));
@@ -162,11 +163,11 @@ pub fn init_timer(timer_tick_handler: HandlerFunc) -> Result<(), &'static str> {
 }
 
 /// Disables the timer, schedules its next tick, and re-enables it
-pub fn schedule_next_timer_tick(delay: Duration) {
+pub fn schedule_next_timer_tick() {
     enable_timer(false);
     let period_femtoseconds = read_timer_period_femtoseconds();
-    let period_nanoseconds = period_femtoseconds / 1_000_000;
-    let ticks = delay.as_nanos() / (period_nanoseconds as u128);
+    let period_microseconds = period_femtoseconds / 1_000_000_000;
+    let ticks = CONFIG_TIMESLICE_PERIOD_MICROSECONDS / (period_microseconds as u32);
     CNTP_TVAL_EL0.set(ticks.try_into().unwrap());
     enable_timer(true);
 }
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -23,9 +23,6 @@ global_asm!(include_str!("table.s"));
 // The global Generic Interrupt Controller singleton
 static GIC: MutexIrqSafe<Option<ArmGic>> = MutexIrqSafe::new(None);
 
-// The number of femtoseconds between each internal timer tick
-static TICK_PERIOD_FEMTOSECS: Once<u64> = Once::new();
-
 /// The IRQ number reserved for CPU-local timer interrupts,
 /// which Theseus currently uses for preemptive task switching.
 //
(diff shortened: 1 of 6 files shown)


kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -157,6 +161,16 @@ pub fn init_timer(timer_tick_handler: HandlerFunc) -> Result<(), &'static str> {
     Ok(())
 }
 
+/// Disables the timer, schedules its next tick, and re-enables it
+pub fn schedule_next_timer_tick(delay: Duration) {
+    enable_timer(false);
+    let period_femtoseconds = read_timer_period_femtoseconds();
+    let period_nanoseconds = period_femtoseconds / 1_000_000;
+    let ticks = delay.as_nanos() / (period_nanoseconds as u128);
```
kevinaboos: this is a pretty expensive calc to be re-doing on every interrupt. Can we calculate this value once and then re-use it?

kevinaboos (changes requested):

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: Yep; I just put it in a spin::Once cache.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -86,6 +91,15 @@ extern "C" fn default_irq_handler(exc: &ExceptionContext) -> EoiBehaviour {
...
 }
 
+#[inline(always)]
+fn read_timer_period_femtoseconds() -> u64 {
+    let counter_freq_hz = CNTFRQ_EL0.get();
+    let fs_in_one_sec = 1_000_000_000_000_000;
+    let period_femtoseconds = fs_in_one_sec / counter_freq_hz;
+    TICK_PERIOD_FEMTOSECS.call_once(|| period_femtoseconds);
+    period_femtoseconds
+}
```
kevinaboos: this is an unusual way to store a cached value. Typically you would do something like this:
```rust
fn read_timer_period_femtoseconds() -> u64 { 
    static TICK_PERIOD_FEMTOSECS: Once<u64> = Once::new();
    *TICK_PERIOD_FEMTOSECS.call_once(|| {
        let counter_freq_hz = CNTFRQ_EL0.get();
        let fs_in_one_sec = 1_000_000_000_000_000;
        let period_femtoseconds = fs_in_one_sec / counter_freq_hz;
        period_femtoseconds
    })
}
```
This has several advantages:
* the `static` variable is scoped within this function, ensuring nobody else can mess with it
* This function doesn't have to be called in an earlier init() function
* This function cannot fail -- if it hasn't been initialized, it initializes itself
* This also avoids the need to call `Once::try()`, which itself avoids the need to handle errors
* Guarantees the cached value will always be used, thanks to the semantics of `call_once`

kevinaboos (changes requested):
Please simplify the way the `Once` variable is lazily initialized, per my suggestion

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: Ah yes sorry about that; my reasoning was that `read_timer_period_femtoseconds` was called only once but you're right.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -157,6 +168,16 @@ pub fn init_timer(timer_tick_handler: HandlerFunc) -> Result<(), &'static str> {
     Ok(())
 }
 
+/// Disables the timer, schedules its next tick, and re-enables it
+pub fn schedule_next_timer_tick() {
+    enable_timer(false);
+    let timeslice_femtosecs = (CONFIG_TIMESLICE_PERIOD_MICROSECONDS as u64) * 1_000_000_000;
+    let tick_period_femtosecs = read_timer_period_femtoseconds();
+    let ticks = timeslice_femtosecs / tick_period_femtosecs;
```
kevinaboos: Sorry, but you're still doing a lot of unnecessary operations here. All of these can be done once, cached, and then reused.

The only reason I'm being so strict about this is that it has to happen on _every single_ timer interrupt, so it needs to be as cheap/fast as possible.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -21,6 +23,9 @@ global_asm!(include_str!("table.s"));
 // The global Generic Interrupt Controller singleton
 static GIC: MutexIrqSafe<Option<ArmGic>> = MutexIrqSafe::new(None);
 
+// The number of femtoseconds between each internal timer tick
+static TICK_PERIOD_FEMTOSECS: Once<u64> = Once::new();
+
```
kevinaboos: I still think it's a good idea to move this into the body of `read_timer_period_femtoseconds()`. It's clearer in that it both explicitly demonstrates and mandates that only that one function can and should access that, i.e., the principle of least privilege.

kevinaboos (changes requested):
not quite there yet, i think you missed a couple of my points

kernel/interrupts/src/aarch64/mod.rs
kevinaboos: For more, see how we do it in `LocalApic`. I basically calculate the value once and then store it in each `LocalApic` struct.

NathanRoyer: Ok, I pub all computations in these two functions:

```rust
fn read_timer_period_femtoseconds() -> u64 {
    let counter_freq_hz = CNTFRQ_EL0.get();
    let fs_in_one_sec = 1_000_000_000_000_000;
    fs_in_one_sec / counter_freq_hz
}

fn get_timeslice_ticks() -> u64 {
    // The number of femtoseconds between each internal timer tick
    static TIMESLICE_TICKS: Once<u64> = Once::new();

    *TIMESLICE_TICKS.call_once(|| {
        let timeslice_femtosecs = (CONFIG_TIMESLICE_PERIOD_MICROSECONDS as u64) * 1_000_000_000;
        let tick_period_femtosecs = read_timer_period_femtoseconds();
        timeslice_femtosecs / tick_period_femtosecs
    })
}
```

- `read_timer_period_femtoseconds` is called only twice per run; no need to cache it.
- `get_timeslice_ticks` is called for each timer interrupt; it makes sense to cache it.

`schedule_next_timer_tick` is now:
```rust
/// Disables the timer, schedules its next tick, and re-enables it
pub fn schedule_next_timer_tick() {
    enable_timer(false);
    CNTP_TVAL_EL0.set(get_timeslice_ticks());
    enable_timer(true);
}
```

Is that good?

kevinaboos: yep, perfect!

kevinaboos (approved):

PR #884, opened 2023-03-16
Handle PS/2 better on real hardware

hecatia-elegua:
This is the second, actual PR (opposed to #870, which is just for debugging).

We still need to test on more real hardware, so this is a draft and still has some debug statements, which might be useful if early framebuffer logging is possible soon.
When everything works I can deduplicate all those loops as well.

### DIFF
--- a/kernel/keyboard/src/lib.rs
+++ b/kernel/keyboard/src/lib.rs
@@ -47,7 +47,11 @@ pub fn init(keyboard: PS2Keyboard<'static>, keyboard_queue_producer: Queue<Event
     }
 
     // TODO: figure out what we should do, for now using set 1
-    keyboard.keyboard_scancode_set(ScancodeSet::Set1)?;
+    // some datapoints:
+    // - laptop 1 allows setting Scancode Set 1 and also returns Set 1 on GetScancode, but doesn't interrupt
+    // - laptop 1 allows setting Scancode Set 2 and works
+    // -> this means we can't check by hardware, we can only try setting the default, Set 2 and then try GetScancode
+    keyboard.set_keyboard_scancode_set(ScancodeSet::Set1)?;
 
     // Register the interrupt handler
     interrupts::register_interrupt(PS2_KEYBOARD_IRQ, ps2_keyboard_handler).map_err(|e| {
@@ -63,6 +67,7 @@ pub fn init(keyboard: PS2Keyboard<'static>, keyboard_queue_producer: Queue<Event
 
 /// The interrupt handler for a PS/2-connected keyboard, registered at IRQ 0x21.
 extern "x86-interrupt" fn ps2_keyboard_handler(_stack_frame: InterruptStackFrame) {
+    debug!("ps2_keyboard_handler");
     // Some of the scancodes are "extended", which means they generate two different interrupts,
     // the first handling the E0 byte, the second handling their second byte.
     static EXTENDED_SCANCODE: AtomicBool = AtomicBool::new(false);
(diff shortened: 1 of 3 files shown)


kernel/keyboard/src/lib.rs
```diff
@@ -47,7 +47,11 @@ pub fn init(keyboard: PS2Keyboard<'static>, keyboard_queue_producer: Queue<Event
     }
 
     // TODO: figure out what we should do, for now using set 1
-    keyboard.keyboard_scancode_set(ScancodeSet::Set1)?;
+    // some datapoints:
+    // - laptop 1 allows setting Scancode Set 1 and also returns Set 1 on GetScancode, but doesn't interrupt
+    // - laptop 1 allows setting Scancode Set 2 and works
```
kevinaboos: you have "laptop 1" written twice; i presume one of them is one laptop, and one is another? 

If you want to leave this comment in, it'd be best to provide as many details about the real hw as possible, e.g., by saying something like "A thinkpad X220 allows..."

kernel/ps2/src/lib.rs
```diff
@@ -266,6 +276,15 @@ impl PS2Controller {
         self.data_port.lock().read()
     }
 
+    /// For debugging purposes
+    #[doc(hidden)]
```
kevinaboos: it's a private fn, so this doesn't have any effect.

kernel/ps2/src/lib.rs
```diff
@@ -464,38 +514,98 @@ impl<'c> PS2Keyboard<'c> {
...
     ///
     /// Note:
     /// On the identify command, [KeyboardType::AncientATKeyboard] usually returns no bytes at all, but `PS2Controller::read_data` always returns a byte.
-    /// This means `PS2Controller::read_data` would return 0x00 here, even though the value is already reserved for the device type "Standard PS/2 mouse".
+    /// This means `PS2Controller::read_data` would presumably return 0x00 here, even though the value is already reserved for the device type "Standard PS/2 mouse".
     /// As we only care about detecting keyboard types here, it should work.
     pub fn keyboard_detect(&self) -> Result<KeyboardType, &'static str> {
+        // return Ok(KeyboardType::MF2Keyboard);
+        //I get random startup blackscreens when multicore is on (no logging)
+        //when doing command_to_keyboard, but it could also be caused by completely different things
```
kevinaboos: if you merge in the latest changes from `theseus_main`, you should be able to disable the code routine that changes graphics mode when bringing up multiple CPUs. You can do that by commenting out this line: https://github.com/theseus-os/Theseus/blob/0f4f28ff86cf0be028ae5cdb6c36cbf2b2340fa7/kernel/multicore_bringup/src/lib.rs#L338

That way, the system will keep using the same framebuffer until the graphics/WM subsystem is initialized.

kevinaboos (changes requested):
Thanks so much! I'm happy to confirm that this PR makes the keyboard work on my real hw machine, as well as on QEMU and QEMU+KVM! Nice work.

The mouse still does not work on real hw, though it does work on QEMU and QEMU+KVM. Currently the only symptom I'm able to observe on real hw is that as soon as I move or click the mouse, the keyboard also stops working. This might be due to the myriad `loop`s you've inserted throughout the code, or just some other form of interference between the kbd and mouse w.r.t. shared ps2 data ports. We can discuss more here or on discord if you have other ideas to try. 

When you get a chance, can you clean up this PR to reduce the amount of verbose logging (or submit a different PR if you want to keep this as is)? Also be sure to remove an `loop`s inserted. After that, I'd be happy to merge it in, since it does at least allow the keyboard to work, which is must more important than the mouse.

Feel free to check out the latest changes to `theseus_main`, which enable a graphical framebuffer for early printing. It should work for you to get more info during early init more easily.

hecatia-elegua:
Should be good for merging now.

kevinaboos (approved):
Great, thanks a lot!

PR #886, opened 2023-03-17
Replace busy-waiting with halting

amab8901:
If Theseus enters any of the tight loops (`loop {}`), then it's going to waste 100% of CPU without doing any meaningful work. This PR should replace this busy-waiting with halting so that the CPU doesn't do anything until it receives the next interrupt.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1043,6 +1043,7 @@ dependencies = [
  "locked_idt",
  "memory",
  "mod_mgmt",
+ "pause 0.1.0",
  "spin 0.9.4",
  "tss",
  "x86_64",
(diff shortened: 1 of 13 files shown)


kevinaboos (changes requested):
Not a bad idea, but this is x86-specific. Also, I don't think `hlt` is what we want in most of these cases, as that stops the CPU until an interrupt occurs. We are working on power mgmt behind the scenes to put the CPU into a low-power state, but that's a while off right now.

For now, the best we can do is to insert a `spin_loop` hint in these empty loops; I'd be okay with that change.

kevinaboos:
Superseded by #895

PR #887, opened 2023-03-17
`nano_core`: minor clean up and typo fixes

hecatia-elegua:
I found some stuff while reading and studying :)

* Simplifies rust_entry
* Fixes some typos
* `unwrap_or_else` might be pedantic


kernel/nano_core/src/bios.rs
```diff
@@ -1,25 +1,25 @@
...
+    VirtualAddress::new(boot_info_vaddr)
+        .ok_or("BUG: multiboot2 info virtual address is invalid")?;
+    let boot_info = unsafe { multiboot2::load(boot_info_vaddr) }.ok()
+        .ok_or("BUG: failed to load multiboot 2 info")?;
+    let kernel_stack_start = VirtualAddress::new(double_fault_stack_top - boot_info.stack_size()?)
+        .ok_or("BUG: kernel_stack_start virtual address is invalid")?;
+    let double_fault_stack_top = VirtualAddress::new(double_fault_stack_top)
+        .ok_or("BUG: double_fault_stack_top virtual address is invalid")?;
+
+    nano_core(boot_info, double_fault_stack_top, kernel_stack_start)
```
kevinaboos: Hmm, i guess my question is... why? This isn't really any better than what we had, and you've discarded info from the `multiboot2::load` error condition.

kernel/nano_core/build.rs
```diff
@@ -33,20 +28,24 @@ cfg_if::cfg_if! {
...
+    "TARGET_ARCH",
+    "TARGET_ENDIAN",
+    "TARGET_ENV",
+    "TARGET_FEATURE",
+    "TARGET_HAS_ATOMIC",
+    "TARGET_HAS_ATOMIC_EQUAL_ALIGNMENT",
+    "TARGET_HAS_ATOMIC_LOAD_STORE",
+    "TARGET_OS",
+    "TARGET_POINTER_WIDTH",
+    "TARGET_VENDOR",
```
kevinaboos: i find this to be less clear / worse, since the reader cannot quickly observe and understand that we're searching for default/built-in `CARGO_CFG_*` values. Please revert.

kernel/nano_core/build.rs
```diff
@@ -79,30 +78,33 @@ fn emit_built_rs_file() {
...
-
-            num_custom_cfgs += 1;
         }
     }
 
     // Append all of the custom cfg values to the built.rs file as an array.
     write!(
         &mut built_file,
-        "#[allow(dead_code)]\npub const CUSTOM_CFG: [(&str, &str); {num_custom_cfgs}] = [{custom_cfgs}];\n",
+        "#[allow(dead_code)]\n
```
kevinaboos: you need to append another `\` at the end of the line (after the `\n`) in order to keep the spacing alignment correct.

kernel/nano_core/build.rs
```diff
@@ -79,30 +78,33 @@ fn emit_built_rs_file() {
...
+        pub const CUSTOM_CFG: [(&str, &str); {num_custom_cfgs}] = [{custom_cfgs}];\n",
     ).unwrap();
 
     // Append all of the custom cfg values to the built.rs file as a single string.
     write!(
         &mut built_file,
-        "#[allow(dead_code)]\npub const CUSTOM_CFG_STR: &str = r#\"{custom_cfgs_str}\"#;\n",
+        "#[allow(dead_code)]\n
+        pub const CUSTOM_CFG_STR: &str = r#\"{custom_cfgs_str}\"#;\n",
```
kevinaboos: same here

kevinaboos (changes requested):
Thanks! most of the changes are good, but some don't seem like improvements to me. See my comments.

kernel/nano_core/src/bios.rs
hecatia-elegua: I tried to get that error condition by propagating format_args but it looked worse, then I thought "when is this going to happen" and decided that if it was to happen, it would still be easy enough to debug without the info. If the multiboot Error  just implemented Error or to_string or sth and or we also had error types here, it would be easier.

Besides that, I think this is easier to read and removes one more macro. The entry point just handles control flow setup -> shutdown and preferrably this would be the only shutdown-point in theseus.

kernel/nano_core/build.rs
hecatia-elegua: It's worse for grepping I guess. I did that for dedup and code flow reasons below. Will revert.

kernel/nano_core/build.rs
hecatia-elegua: I somehow thought normal rust strings don't care about new lines and only raw strings do care, oops

kernel/nano_core/src/bios.rs
kevinaboos: fair enough

PR #888, opened 2023-03-17
Update `uefi-bootloader` version

tsoutsman:
Adds `virt` field to `FrameBufferInfo`.


kevinaboos (commented):
Thanks! I'll need to test this with my latest changes before merging, but it should work. Looks good so far.

kevinaboos (approved):
Looks good. It works but I need to make some follow-up improvements to our new early_printer to better leverage the new UEFI-mapped framebuffer.

PR #890, opened 2023-03-17
Multicore step 1 - Implement `handle_ap_cores` for aarch64

NathanRoyer:
This PR implements the discovery and setup of secondary/AP cores for aarch64.

This depends on #889.


kernel/captain/src/lib.rs
```diff
@@ -132,13 +132,10 @@ pub fn init(
     
     // boot up the other cores (APs)
     // arch-gate: no multicore support on aarch64 at the moment
```
kevinaboos: outdated comment

kernel/cpu/src/aarch64.rs
```diff
@@ -43,6 +43,14 @@ impl CpuId {
         assert!(level < 4, "Valid affinity levels are 0, 1, 2, 3");
         (self.0 >> (level * 8)) as u8
     }
+
+    pub fn new(aff3: u8, aff2: u8, aff1: u8, aff0: u8) -> Self {
```
kevinaboos: `CpuId` is supposed to be a trusted type that nobody can arbitrarily instantiate, so a publicly-available constructor like this breaks that type guarantee.

Is this absolutely necessary, or can we get around it somehow? If required, let's write our or discuss the actual concrete requirements so we can brainstorm an alternate solution.

kernel/nano_core/src/lib.rs
```diff
@@ -120,11 +120,16 @@ where
...
+    #[cfg(target_arch = "x86_64")]
+    let framebuffer_info;
+
+    #[cfg(target_arch = "x86_64")] {
+        // If the bootloader already mapped the framebuffer for us, the we can use it now
+        // before initializing the memory mgmt subsystem.
+        framebuffer_info = boot_info.framebuffer_info();
+        if let Some(ref fb_info) = framebuffer_info && fb_info.is_mapped() {
+            early_printer::init(fb_info)?;
+        }
```
kevinaboos: i think this is a bit cleaner and easier to read
```suggestion
    // If the bootloader already mapped the framebuffer for us, the we can use it now
    // before initializing the memory mgmt subsystem.
    #[cfg(target_arch = "x86_64")]
    let framebuffer_info = boot_info.framebuffer_info();
    #[cfg(target_arch = "x86_64")]
    if let Some(ref fb_info) = framebuffer_info && fb_info.is_mapped() {
        early_printer::init(fb_info)?;
```

kevinaboos: Also, let's refactor this PR a bit -- move the changes to nano_core into a separate PR, since it has nothing to do with multicore support.

kernel/multicore_bringup/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+//!
+//! These functions are intended to be invoked from the BSP
+//! (the Bootstrap Processor, the main CPU in x86 terms)
+//! in order to bring up secondary CPUs (APs in x86 terms).
+
+#![no_std]
+#![cfg_attr(target_arch = "x86_64", feature(let_chains))]
+#![cfg_attr(target_arch = "aarch64", feature(naked_functions))]
+
+#[cfg_attr(target_arch = "x86_64", macro_use)] extern crate log;
```
kevinaboos: you should be able to remove this and add the following into the x86_64 module:
```rust
use log::*;
```

kernel/multicore_bringup/src/aarch64.rs
```diff
@@ -0,0 +1,214 @@
...
+        let entry_point_virt_addr = VirtualAddress::new_canonical(ap_entry_point as usize);
+        entry_point_phys_addr = page_table.translate(entry_point_virt_addr).unwrap();
+
+        // get the physical address of the MmuConfig
+        let mmu_config_virt_addr = VirtualAddress::new_canonical(&mmu_config as *const _ as usize);
+        ap_data.ap_mmu_config.write(page_table.translate(mmu_config_virt_addr).unwrap());
+
+        // Note: this could be variable on aarch64
+        // we'd need a way to find any identity mappable page+frame range
+        const AP_STARTUP: usize = 0x10000;
```
kevinaboos, line 69: identity mapping shouldn't be too hard. What are the requirements; does it need to be at a certain address in low memory? For ex, on x86, it needs to be within the bounds of 16 bits, hence why this is so complicated.

kernel/multicore_bringup/src/aarch64.rs
```diff
@@ -0,0 +1,214 @@
...
+            flags,
+        )?;
+
+        // get physical address of the ApTrampolineData structure
+        ap_data_phys_addr = page_table.translate(ap_data.ap_data_virt_addr.read()).unwrap();
+
+        // copy the entry point code
+        let dst = ap_startup_mapped_pages.as_slice_mut(0, PAGE_SIZE).unwrap();
+        let src = unsafe { (ap_entry_point as *const [u8; PAGE_SIZE]).as_ref() }.unwrap();
+        dst.copy_from_slice(src);
```
kevinaboos, line 92: same question for this -- what are the requirements here? This whole complex concept of copying code into a dedicated area is just to ensure that everything can be accessed in x86 real mode (16-bit addressing, no paging, etc)

kernel/multicore_bringup/src/aarch64.rs
```diff
@@ -0,0 +1,214 @@
...
+        &mut kernel_mmi_ref.lock().page_table,
+    ).ok_or("could not allocate AP stack!")?));
+
+    log::error!("Bruteforcing 256 CpuIds that are likely to be valid");
+    for aff1 in 0..16 {
+        for aff0 in 0..16 {
+            let cpu_id = CpuId::new(0, 0, aff1, aff0);
+
+            // Write the per-core fields of ApTrampolineData
+            let stack = ap_stack.take().unwrap();
```
kevinaboos: i love the idea of reusing stacks for CPUs that don't actually exist. if you reorganize this loop a bit, you can easily avoid some complexity and also the leaking of the last stack.

kernel/ap_start/src/lib.rs
```diff
@@ -27,10 +31,13 @@ pub static AP_READY_FLAG: AtomicBool = AtomicBool::new(false);
 static AP_STACKS: MutexIrqSafe<BTreeMap<u32, NoDrop<Stack>>> = MutexIrqSafe::new(BTreeMap::new());
 
 /// Insert a new stack that was allocated for the AP with the given `cpu_id`.
-pub fn insert_ap_stack(cpu_id: u32 , stack: Stack) {
-    AP_STACKS.lock().insert(cpu_id, NoDrop::new(stack));
+pub fn insert_ap_stack(cpu_id: u32, stack: NoDrop<Stack>) {
+    AP_STACKS.lock().insert(cpu_id, stack);
```
kevinaboos: is there any reason why we force the caller to deal with `NoDrop` now rather than doing it within this function?

kevinaboos (changes requested):
is the term "AP" used in aarch64 parlance? I was under the impression that was an x86-specific term, so we probably shouldn't use that in aarch64 code. Then, we can use the term CPU or secondary CPU in arch-generic code. But of course we can do that in a follow-up PR.

I think we probably need to meet up and discuss the requirements of aarch64 multicore bringup, since a lot of the complexity in the existing implementation comes from the fact that APs boot up in 16-bit real mode on x86, meaning we have to do a lot of weird shit to allow that code to run initially. For example, do we need the `ApTrampolineData` struct and its semantics on aarch64? Basically I'm wondering how closely the aarch64 implementation of this should follow the x86 version -- i had expected the aarch64 version to be quite a bit simpler.

kernel/multicore_bringup/src/aarch64.rs
NathanRoyer: > identity mapping shouldn't be too hard. What are the requirements; does it need to be at a certain address in low memory? For ex, on x86, it needs to be within the bounds of 16 bits, hence why this is so complicated.

Nope; we only need an ID-mapping, whatever the address.

kernel/multicore_bringup/src/aarch64.rs
NathanRoyer: Secondary AArch64 cores are waiting for startup with their MMU disabled. When we ask them to start and jump to a physical address, the startup code will configure the MMU and then enable it; for that reason, the code must be identity mapped so that the instruction pointer is valid whether the MMU is enabled or disabled. Most of theseus' code - including the `ap_entry_point` - is not identity mapped; this is why we identity-map one page of RWX memory and copy the startup code to it; we can then make secondary CPU cores jump to this code area.

kernel/cpu/src/aarch64.rs
NathanRoyer: I moved this (as well as `affinity`) to `MpidrValue`:
```rust
impl MpidrValue {
    // ...

    /// Reads an affinity `level` from this `MpidrValue`.
    ///
    /// Panics if the given affinity level is not 0, 1, 2, or 3.
    pub fn affinity(self, level: u8) -> u8 {
        let shift = match level {
            0 => 0,
            1 => 8,
            2 => 16,
            3 => 32,
            _ => panic!("Valid affinity levels are 0, 1, 2, 3"),
        };
        (self.0 >> shift) as u8
    }

    /// Create an `MpidrValue` from its four affinity numbers
    pub fn new(aff3: u8, aff2: u8, aff1: u8, aff0: u8) -> Self {
        let aff3 = (aff3 as u64) << 32;
        let aff2 = (aff2 as u64) << 16;
        let aff1 = (aff1 as u64) <<  8;
        let aff0 = (aff0 as u64) <<  0;
        Self(aff3 | aff2 | aff1 | aff0)
    }
}
```

kernel/multicore_bringup/src/aarch64.rs
NathanRoyer: Good point. I did what you advised.

kernel/ap_start/src/lib.rs
NathanRoyer: Not anymore :ok_hand:

PR #893, opened 2023-03-19
Add preserves_flags

amab8901:
> preserves_flags: The asm! block does not modify the flags register (defined in the rules below). This allows the compiler to avoid recomputing the condition flags after the asm! block.

(https://doc.rust-lang.org/nightly/reference/inline-assembly.html)

the flags registers aren't modified here, so it's safe to use this option. (https://www.felixcloutier.com/x86/pause) <-- In this link, it doesn't mention anything about flags and it says `This instruction does not change the architectural state of the processor`

### DIFF
--- a/kernel/pause/src/lib.rs
+++ b/kernel/pause/src/lib.rs
@@ -15,5 +15,5 @@
 pub fn spin_loop_hint() {
     // core::hint::spin_loop();
     #[cfg(target_arch = "x86_64")]
-    unsafe { core::arch::asm!("pause", options(nomem, nostack)); }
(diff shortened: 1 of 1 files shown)


kevinaboos (changes requested):
Ah, thanks. While this is correct, you can go ahead and completely remove this crate, as it's old and no longer used now that the `core::hint::spin_loop()` generates the correct code on non-SSE x86 platforms.

You can also remove any usage of this crate too.

kevinaboos:
Closed in favor of #894

PR #894, opened 2023-03-20
remove pause crate

amab8901:
this applies the recommendation in https://github.com/theseus-os/Theseus/pull/893


kevinaboos (approved):
thanks, looks good!

Issue #808, opened 2023-01-12
Add documentation/comments to infinite loops


amab8901:
After https://github.com/theseus-os/Theseus/pull/806 has been merged, go through every empty loop (marked by `#[allow(clippy::empty_loop)]`) and add explanation for why the empty loop is necessary and criteria for replacing it with something else.

Each empty loop should have its own independent explanation case-by-case (rather than having a single explanation that covers multiple empty loops, or copy+pasting boilerplate). Because that extra scrutiny might reveal an (otherwise hard to spot) opportunity for improvement to future system developers. 

@kevinaboos

kevinaboos:
Thanks for filing this!

PR #895, opened 2023-03-20
Add spin-loop hint

amab8901:
replaces https://github.com/theseus-os/Theseus/pull/886


kevinaboos:
Thanks! Does this address #808? Also i recall you adding TODO comments to a few places about addressing empty loops; would you mind removing those as part of this PR?

amab8901:
> Does this address https://github.com/theseus-os/Theseus/issues/808?

apparently it does. Thanks for reminding me about it. 

> Also i recall you adding TODO comments to a few places about addressing empty loops; would you mind removing those as part of this PR?

I fixed it now

kevinaboos (approved):
Thanks, looks all good now!

PR #900, opened 2023-03-21
Make clippy deny

amab8901:
Clippy CI has thus far allowed warnings to pass through. But now it will require Clippy lints to be applied before it passes the commit.

Also, we don't need to use `make clippy` syntax any more since we can use `cargo clippy` instead


.github/workflows/check-clippy.yaml
```diff
@@ -17,4 +17,4 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         working-directory: .
         run: |
-          make clippy
+          cargo clippy -- -D warnings
```
kevinaboos: we still need the make target to allow devs to run clippy locally. Also, using make ensures we run clippy with the correct target configuration and include all crates in the build.

Instead of changing this, you can change the `clippy` target in the `Makefile`.

Also, should we use `-D all` or something more all-encompassing instead of just `-D warnings`?

kernel/boot_info/src/lib.rs
```diff
@@ -95,8 +95,8 @@ pub enum Address {
 impl fmt::Debug for Address {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Address::Physical(paddr) => write!(f, "p{:#X}", paddr),
-            Address::Virtual(vaddr)  => write!(f, "v{:#X}", vaddr),
+            Address::Physical(paddr) => write!(f, "p{paddr:#X}"),
+            Address::Virtual(vaddr)  => write!(f, "v{vaddr:#X}"),
```
kevinaboos: please merge in the latest changes from main, this was removed yesterday or the day before

kernel/tls_initializer/src/lib.rs
```diff
@@ -84,6 +84,10 @@ const TLS_SELF_POINTER_SIZE: usize = size_of::<usize>();
 #[cfg(target_arch = "aarch64")]
 const TLS_SELF_POINTER_SIZE: usize = 0;
 
+pub enum TlsInitializerError {
+    InvalidData,
+}
+
```
kevinaboos: this is good, but we need more variants for the possible error conditions. See my comments below.

kernel/tls_initializer/src/lib.rs
```diff
@@ -156,7 +160,7 @@ impl TlsInitializer {
         #[cfg(target_arch = "aarch64")]
         let virt_addr_value = original_offset;
 
-        tls_section.virt_addr = VirtualAddress::new(virt_addr_value).ok_or(())?;
+        tls_section.virt_addr = VirtualAddress::new(virt_addr_value).ok_or(TlsInitializerError::InvalidData)?;
```
kevinaboos: this should use an error variant related to the TLS section having an invalid virtual address.

kernel/tls_initializer/src/lib.rs
```diff
@@ -146,7 +150,7 @@ impl TlsInitializer {
         if self.static_section_offsets.contains_key(&range.start) || 
             self.static_section_offsets.contains_key(&(range.end - 1))
         {
-            return Err(());
+            return Err(TlsInitializerError::InvalidData);
```
kevinaboos: this should use an error variant related to the TLS section's bounds overlapping with an existing section.

kernel/tls_initializer/src/lib.rs
```diff
@@ -200,7 +204,7 @@ impl TlsInitializer {
             }
         }
 
-        let start = start_index.ok_or(())?;
+        let start = start_index.ok_or(TlsInitializerError::InvalidData)?;
```
kevinaboos: this should use an error variant related to not being able to find space to insert the TLS section into the TLS data image.

kernel/tls_initializer/src/lib.rs
```diff
@@ -209,7 +213,7 @@ impl TlsInitializer {
         #[cfg(target_arch = "aarch64")]
         let virt_addr_value = start + self.end_of_static_sections - max(16, 8 /* TODO FIXME: pass in the TLS segment's alignment */);
 
-        section.virt_addr = VirtualAddress::new(virt_addr_value).ok_or(())?;
+        section.virt_addr = VirtualAddress::new(virt_addr_value).ok_or(TlsInitializerError::InvalidData)?;
```
kevinaboos: this should use an error variant related to the TLS section having an invalid virtual address (same as above)

kernel/tls_initializer/src/lib.rs
kevinaboos: also, please add comments to each future variant describing what they mean, according to my review comments below.

kevinaboos (changes requested):
Thanks! Generally it looks good, I left a few minor comments.

.github/workflows/check-clippy.yaml
amab8901: I think we can start with merging `-D warnings` and then add `-D all` later. I think this kind of gradual transition will be more doable because then I won't need to worry about new warning lints being added from other contributors while trying to fix the `-D all` lints

kevinaboos: that's fine with me. We do still need to make this change in the `clippy` makefile target, not here.

kernel/tls_initializer/src/lib.rs
```diff
@@ -84,6 +84,22 @@ const TLS_SELF_POINTER_SIZE: usize = size_of::<usize>();
...
+
+/// OverlapWithExistingSection:
+    ///   An error if inserting the given section at the given offset
+    ///   would overlap with an existing section. 
+    ///   An error occurring here would indicate a link-time bug 
+    ///   or a bug in the symbol parsing code that invokes this function.
+/// TLSHasInvalidVirtualAddress
+    /// TLS section has an invalid virtual address.
+/// NoRemainingSpace
+    /// Can't find find enough space to insert the TLS section into the TLS data image.
```
kevinaboos: these doc comments should go above each variant, as such:
```rust
/// Errors that may occur when registering new TLS sections.
pub enum TlsInitializerError {
    /// insert docs here
    OverlapWithExistingSection,
    /// insert docs here
    TLSHasInvalidVirtualAddress,
    /// insert docs here
    NoRemainingSpace,
}

File: kernel/spawn/src/lib.rs [hunk 9]
```diff
@@ -984,7 +984,7 @@ where
 
     scheduler::schedule();
     error!("BUG: task_cleanup_final(): task was rescheduled after being dead!");
-    loop { core::hint::spin_loop() }
+    loop { core::hint::spin_loop()   }
```
kevinaboos: remove excess whitespace

kevinaboos (changes requested):
The changes look good! I left a few small comments.

One thing -- I don't see your changes to the `libs/dfqueue` submodule here. It looks like you didn't actually commit those to this branch. You'll need to do that in order to make sure that the main Theseus repo actually depends on the newest commit of dfqueue: https://github.com/theseus-os/DFQueue/commit/896cc00a826007a8aa99cc6ccd96525d865882bc

kevinaboos (approved):

PR #903, opened 2023-03-23
Clean up memory crates

hecatia-elegua:
- remove unused or stabilized features
- sort and unify cargo deps
- update editions
- change one glob import to named imports


kevinaboos (approved):
thanks!

PR #904, opened 2023-03-23
Add a pre-push hook

hecatia-elegua:
- add a pre-push hook running clippy (can be expanded to run docs (view-docs?))
- add a make target for contributors to use our hooks (this should not be done automatically due to safety) and one to un-use them (if build fails but changes are ok this is useful)
- add a sentence to README about this

This currently won't do much besides ensuring that the build succeeds.
After #900 it will help.
What I dislike a bit about this is that the usual vscode git ui doesn't really tell you "hey, push fails because of a hook" but just tells you to "maybe pull in changes" or sth like that. Will take a look if there is some way to improve this later.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+set -e
+
+echo '+make clippy'
+make clippy
--- a/Makefile
+++ b/Makefile
@@ -683,8 +683,13 @@ clean-doc:
 	@cargo clean --doc
 	@rm -rf $(RUSTDOC_OUT) $(BOOK_OUT)
 
+### Sets this repo's git hooks directory
 hooks:
 	@git config --local core.hooksPath .githooks/
+
+### Unsets this repo's git hooks directory
+unhook:
+	@git config --local --unset core.hooksPath
 
 ### The primary documentation for this makefile itself.
 help: 
--- a/Makefile
+++ b/Makefile
@@ -682,7 +682,9 @@ view-book: book
 clean-doc:
 	@cargo clean --doc
 	@rm -rf $(RUSTDOC_OUT) $(BOOK_OUT)
-	
+
+hooks:
+	@git config --local core.hooksPath .githooks/
 
 ### The primary documentation for this makefile itself.
 help: 
--- a/README.md
+++ b/README.md
@@ -277,7 +277,8 @@ For VS Code, recommended plugins are:
  * **rust-analyzer**, by matklad
  * **Better TOML**, by bungcip
  * **x86 and x86_64 Assembly**, by 13xforever
- 
+
+If you want to contribute, please run `make hooks` once. This way, `clippy` runs automatically on `git push` so you can't forget.
 
 ## Acknowledgements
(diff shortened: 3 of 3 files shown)


kevinaboos:
Hmm, I appreciate the idea, but philosophically I disagree with this requirement. We frequently push things that aren't ready and shouldn't be validated. Also, I don't want to suffer a delay on each push action on the order of the time required to build Theseus and run clippy. 

Checks like this should only be done as a prerequisite for a _merge_, not a push, imo. Also, if we make a slow procedure like this opt-in, it should probably be opt-in on each action (i.e., each push), not opt-in until you opt out again.

If you find this to be helpful, I do encourage you to use it, but I don't feel that it's something we should apply to all contributors/users. Feel free to share more thoughts.

hecatia-elegua:
I completely forgot that many workflows involve pushing to a PR frequently.
I also didn't consider some worse hardware, since clippy takes like 5sec on my mid-range hardware. Though it also has to do with rust-analyzer and maybe cargo caching all the build stuff.

One of the reasons I had was to reduce checking-time in CI, I have some other ideas around that.
Hmm..
Let's say a full CI run takes 4min, that's 500-750 CI runs depending on the Github Pricing Plan. That's kinda far-off too, I guess.

The other reason was to just not have to remember the things you should do before pushing/merging. So always opt-in would only work with some kind of dialog.
Another idea would be to just have a prompt like "Really push? Have you done _?", which would work for command line but might be even worse for vscode git ui. I will try that shortly.

hecatia-elegua:
Well, it works nicely for command line but I see no integration path into vscode git extensions.

kevinaboos:
> One of the reasons I had was to reduce checking-time in CI, I have some other ideas around that.

I totally agree, we need a smarter CI pass that doesn't always re-run everything blindly. That I am in favor of.

All in all, this PR isn't really going to work because the hook is far too heavyweight. I don't think pre-push hooks are intended for this sort of thing like a full-system build check. It's also not something I want to run unconditionally on every push. On my desktop it takes 23 seconds to run `make clippy`, and on my laptop it's over 2 minutes.

In general, the point I made earlier stands: the checks should be a prerequisite for _merging_ into the main branch, as that is the final action for which we want to ensure correctness. We only care about validating what exists in `theseus_main`, not every tiny feature branch. The act of pushing something up a repo is _not_ something we need to nor want to validate; for example, if i wanted to push up some buggy or failing code in order to share with other folks (a _very_ common occurrence), this would effectively prevent me from doing that, or at least require me to jump through extra hoops. That would even apply to people's own personal forks of Theseus, which would be even stranger. There are also many cases where we want to push code up in a hurry just to save it or back it up, e.g., right before leaving your workplace or taking off on a plane -- I definitely don't want any delays there.

I don't see any real benefit in forcing devs to overcome another hidden git hurdle just to push code up to a remote, so I'm going to close this. Thanks for the effort, though, hopefully we can improve our CI passes instead of (ab)using git hooks.

hecatia-elegua:
I do understand, though I didn't even argue for that anymore.
The thing I tried after your first comment was about just asking "hey, have you thought about testing/clippy/...?", which isn't heavyweight, though it doesn't work so I stopped.

Besides that, maybe ` "rust-analyzer.check.command": "clippy",` is faster because doesn't pick up the whole project. I was just wondering why we have such different times, since I really don't have the greatest hardware.

PR #908, opened 2023-03-28
-D clippy::all

amab8901:
doesn't really make any semantic difference but I think the value lies in clearly communicating to the code user that all lints have been addressed


kevinaboos (approved):
Thanks!

PR #909, opened 2023-03-28
Make spin_loop() the default in loop

amab8901:
`loop{}` is basically the most expensive operation in existence. I think it would be a good idea to default to having `spin_loop()` in a `loop {}` to avoid eating 95% of CPU when you actually need just 1% of CPU

### DIFF
--- a/applications/cat/src/lib.rs
+++ b/applications/cat/src/lib.rs
@@ -9,7 +9,7 @@ extern crate path;
 extern crate fs_node;
 extern crate core2;
 
-use core::str;
+use core::{str, hint::spin_loop};
 use alloc::{
     string::{String, ToString},
     vec::Vec,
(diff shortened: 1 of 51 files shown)


amab8901:
actually I'm not sure if this is a good idea anymore. I talked to some other Rust developers and they said a `sleep` function would be more suitable in this kind of situation, rather than `spin_loop`, and that we might have to design that `sleep` function from scratch

kevinaboos:
Ah, so the issue about CPU usage doesn't have anything to do with spin_loop() here, and this isn't really the proper usage of the spin_loop hint, since most of these loops aren't spinning on a lock.

The issue is that we aren't dropping the CPU down to a low-power state when it's idle, so QEMU keeps running the CPU at full speed. We're working on this separately, but idle CPU power state management is a complex and hard problem, and each CPU model/generation has different power states and interfaces via ACPI/AML. Either way, inserting spin_loop won't actually make any difference.

PR #910, opened 2023-03-28
Fixes and improvements in GIC driver

NathanRoyer:
Small changes to the `gic` crate:
- Differentiate `SpiDestination` & `IpiTargetCpu`, which were previously `TargetCpu`
- Use CpuId & MpidrValue when appropriate
- Fixed TargetList so that it doesn't allow routing SPIs & IPIs to arbitrary CpuId values
- Implemented `TryFrom<u64>` for `MpidrValue` to identify existing CPUs based on GIC register contents

### DIFF
--- a/kernel/arm_boards/src/mpidr.rs
+++ b/kernel/arm_boards/src/mpidr.rs
@@ -23,3 +23,16 @@ impl DefinedMpidrValue {
         Self(aff3 | aff2 | aff1 | aff0)
     }
 }
+
+/// Tries to find this MPIDR value among those known for this board.
+///
+/// Returns None if no CPU has this MPIDR value.
+pub fn find_mpidr(mpidr_value: u64) -> Option<DefinedMpidrValue> {
+    for def_mpidr in crate::BOARD_CONFIG.cpu_ids {
+        if def_mpidr.value() == mpidr_value {
+            return Some(def_mpidr)
+        }
+    }
+
+    None
+}
--- a/kernel/arm_boards/src/mpidr.rs
+++ b/kernel/arm_boards/src/mpidr.rs
@@ -23,16 +23,3 @@ impl DefinedMpidrValue {
         Self(aff3 | aff2 | aff1 | aff0)
     }
 }
-
-/// Tries to find this MPIDR value among those known for this board.
-///
-/// Returns None if no CPU has this MPIDR value.
-pub fn find_mpidr(mpidr_value: u64) -> Option<DefinedMpidrValue> {
-    for def_mpidr in crate::BOARD_CONFIG.cpu_ids {
-        if def_mpidr.value() == mpidr_value {
-            return Some(def_mpidr)
-        }
-    }
-
-    None
-}
--- a/kernel/cpu/src/aarch64.rs
+++ b/kernel/cpu/src/aarch64.rs
@@ -6,7 +6,7 @@ use derive_more::{Display, Binary, Octal, LowerHex, UpperHex};
 use irq_safety::RwLockIrqSafe;
 use core::fmt;
 use alloc::vec::Vec;
-use arm_boards::mpidr::DefinedMpidrValue;
+use arm_boards::{mpidr::DefinedMpidrValue, BOARD_CONFIG};
 
 use super::CpuId;
 
@@ -125,6 +125,22 @@ impl From<MpidrValue> for CpuId {
         let aff_0_1_2 =  (mpidr.0 & 0x0000ffffff) as u32;
 
         Self(aff_3 | aff_0_1_2)
+    }
+}
+
+impl TryFrom<u64> for MpidrValue {
+    type Error = &'static str;
+
+    /// Tries to find this MPIDR value in those defined by
+    /// `arm_boards::cpu_ids`. Fails if No CPU has this MPIDR value.
+    fn try_from(mpidr_value: u64) -> Result<Self, Self::Error> {
+        for def_mpidr in BOARD_CONFIG.cpu_ids {
+            if def_mpidr.value() == mpidr_value {
+                return Ok(def_mpidr.into())
+            }
+        }
+
+        Err("No CPU has this MPIDR value")
     }
 }
 
--- a/kernel/cpu/src/aarch64.rs
+++ b/kernel/cpu/src/aarch64.rs
@@ -91,8 +91,6 @@ impl MpidrValue {
     }
 
     /// Reads an affinity `level` from this `MpidrValue`.
-    ///
-    /// Panics if the given affinity level is not 0, 1, 2, or 3.
     pub fn affinity(self, level: AffinityShift) -> u64 {
         (self.0 >> (level as u64)) & (u8::MAX as u64)
     }
--- a/kernel/cpu/src/aarch64.rs
+++ b/kernel/cpu/src/aarch64.rs
@@ -80,7 +80,7 @@ impl MpidrValue {
     /// Reads an affinity `level` from this `MpidrValue`.
     ///
     /// Panics if the given affinity level is not 0, 1, 2, or 3.
-    pub fn affinity(self, level: u8) -> u8 {
+    pub fn affinity(self, level: u8) -> u64 {
         let shift = match level {
             0 => 0,
             1 => 8,
@@ -88,7 +88,7 @@ impl MpidrValue {
             3 => 32,
             _ => panic!("Valid affinity levels are 0, 1, 2, 3"),
         };
-        (self.0 >> shift) as u8
+        (self.0 >> shift) & 0xff
     }
 }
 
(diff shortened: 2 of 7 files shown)


kernel/gic/src/gic/mod.rs
```diff
@@ -46,38 +48,70 @@ bitflags! {
...
     /// more important task
     AnyCpuAvailable,
     GICv2TargetList(TargetList),
 }
 
-const U32BYTES: usize = core::mem::size_of::<u32>();
+/// Target of an inter-processor interrupt
+pub enum IpiTargetCpu {
+    /// That interrupt must be handled by
+    /// a specific PE in the system.
```
kevinaboos:
```suggestion
    /// The interrupt will be delivered to a specific CPU.
```

kernel/gic/src/gic/mod.rs
```diff
@@ -46,38 +48,70 @@ bitflags! {
...
+    Specific(CpuId),
+    /// All PEs will receive this interrupt
+    /// except the sender
```
kevinaboos:
```suggestion
    /// The interrupt will be delivered to all CPUs except the sender.
```

kernel/cpu/src/aarch64.rs
```diff
@@ -52,15 +52,16 @@ impl MpidrValue {
     /// Reads an affinity `level` from this `MpidrValue`.
     ///
     /// Panics if the given affinity level is not 0, 1, 2, or 3.
-    pub fn affinity(self, level: u8) -> u8 {
+    pub fn affinity(self, level: u8) -> u64 {
```
kevinaboos: you can avoid all of these panics if you create a new enum and use that as the argument instead of a raw `u8`.

ex:
```rust
pub enum MpidrAffinity {
    Zero = 0,
    One  = 8,
    Two  = 16,
    Three = 32,
}
```
note that if you need the values of each variant to be the integers 0, 1, 2, 3, then you can do that too.

kernel/gic/src/gic/mod.rs
```diff
@@ -46,38 +48,70 @@ bitflags! {
...
+    AllOtherCpus,
+    GICv2TargetList(TargetList),
```
kevinaboos, line 99: add doc here, something like this?
```rust
/// The interrupt will be delivered to all CPUs specified by the included target list
```

kevinaboos (changes requested):
Everything looks generally fine, I just made a few suggestions for minor improvements.

kernel/gic/src/gic/mod.rs
```diff
@@ -50,24 +50,22 @@ bitflags! {
...
 pub enum SpiDestination {
-    /// That interrupt must be handled by
-    /// a specific PE in the system.
+    /// The interrupt must be delivered to a specific CPU.
     Specific(CpuId),
-    /// That interrupt can be handled by
-    /// any PE that is not busy with another,
-    /// more important task
+    /// That interrupt can be handled by any PE that is not busy with another, more
+    /// important task
```
kevinaboos, line 86: hmm, this makes it sound like if one of the other PEs is busy with another more important task, the interrupt won't actually be handled/delivered at all. Is that correct, i.e., is it guaranteed to be delivered to just a single PE? Will that chosne PE eventually handle that interrupt once it has completed it's current "higher priority" task?

Let's not use the phrase "more important", that's quite vague; use "higher priority" or whatever the most correct and specific term is.

It'd also be nice to explain a bit about how the PE is actually chosen. Is it simply chosen by hardware? And is it guaranteed to be handled by just _one_ PE? I only ask because x86 has different behavior; typically you can have a single interrupt get delivered to multiple CPUs simultaneously, which makes it really hard to handle correctly.

kevinaboos (changes requested):
Looks solid, just left one comment clarifying `SpiDestination` behavior

kernel/gic/src/gic/mod.rs
hecatia-elegua: If I'm allowed to chime in:
This variant (AnyCpuAvailable, also named Mode 1) routes the interrupts to any connected PE that is participating in distribution of the interrupt's group. It's chosen by the Distributor (hardware) and can vary each time.
Haven't read about priority here, I do assume that's explained somewhere else.

GICv2 did send it to multiple targets, with the first to respond getting it (see below).

NathanRoyer: From the GICv3 architecture specification:

image

- First condition = if the redistributor of that PE was initialized
- E1NWF stands for "Enable 1 of N Wakeup Functionality"; it's a bit in the control reg of the distributor (one per system)
- 3rd condition = whether the redistributor of that PE is configure to accept the interrupt group of that interrupt
- (4th condition, mentioned elsewhere = if the CPU interface is configured to accept the interrupt group of that interrupt)

In Theseus currently:
- 1st condition is asserted for the BSP only
- 2nd condition is not asserted (although we never put any core to sleep atm)
- 3rd condition is not asserted
- 4th condition: configured to accept group 1 interrupts only

I used to think that the redistributor is only relevant to LPIs; that was wrong.

Regarding which of the selectable PE is actually selected, it is implementation defined, so the running priority level is actually not necessarily involved in the choice:

image

I'll try to implement assertions for these conditions, even though we have no SPI to test this with currently.

NathanRoyer: We can resolve that conversation now that these assertions were implemented by #915 :ok_hand:

kernel/gic/src/gic/dist_interface.rs
```diff
@@ -138,82 +138,70 @@ impl super::ArmGic {
...
-    pub fn get_spi_target(&self, int: InterruptNumber) -> TargetCpu {
+    /// Returns the destination of an SPI if it's valid, i.e. if it
+    /// points to existing CPU(s).
+    ///
+    /// Note: If the destination is a `GICv2TargetList`, the validity
+    /// of that destination is not checked.
+    pub fn get_spi_target(&self, int: InterruptNumber) -> Option<SpiDestination> {
         assert!(int >= 32, "interrupts number below 32 (SGIs & PPIs) don't have a target CPU");
-        if !self.affinity_routing() {
+        Some(if !self.affinity_routing() {
```
kevinaboos: wrapping a long `if` statement in `Some()` is pretty hard to read. Move the `Some` wrapper elsewhere.

kernel/gic/src/gic/mod.rs
```diff
@@ -23,38 +23,100 @@ pub type InterruptNumber = u32;
...
-        const CPU_4 = 1 << 4;
-        const CPU_5 = 1 << 5;
-        const CPU_6 = 1 << 6;
-        const CPU_7 = 1 << 7;
-        const ALL_CPUS = u8::MAX;
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct TargetList(u8);
+
+impl TargetList {
+    pub fn new(list: &[CpuId]) -> Result<Self, &'static str> {
```
kevinaboos: to be most flexible, this should accept an iterator of `CpuId`s rather than a slice.

kernel/gic/src/gic/mod.rs
```diff
@@ -23,38 +23,100 @@ pub type InterruptNumber = u32;
...
+        Ok(Self(this))
+    }
+
+    /// Tries to create a `TargetList` from `arm_boards::BOARD_CONFIG.cpu_ids`
+    pub fn new_all_cpus() -> Result<Self, &'static str> {
+        let cpu_ids: [CpuId; NUM_CPUS] = core::array::from_fn(|i| {
+            CpuId::from(BOARD_CONFIG.cpu_ids[i])
+        });
+
+        Self::new(&cpu_ids).map_err(|_| "Some CPUs in the system cannot be stored in a TargetList")
```
kevinaboos: if `Self::new()` accepts an iterator, you won't need to try to create a fixed array here.

kernel/arm_boards/src/mpidr.rs
```diff
@@ -23,3 +23,16 @@ impl DefinedMpidrValue {
         Self(aff3 | aff2 | aff1 | aff0)
     }
 }
+
+/// Tries to find this MPIDR value among those known for this board.
+///
+/// Returns None if no CPU has this MPIDR value.
+pub fn find_mpidr(mpidr_value: u64) -> Option<DefinedMpidrValue> {
```
kevinaboos: I was a bit confused by this method name until I saw how it was used. Regardless of naming, this logic doesn't actually have anything to do with board-specific configuration, it's only related to our requirement that a `CpuId` object represents a CPU that is guaranteed to exist. Therefore, this logic should be moved into the `cpu` crate.

Instead of logic like this, you should do something like `impl TryFrom<u64> for CpuId` (or perhaps `impl TryFrom<u64> for MpidrValue`, but I think implementing it for `CpuId` makes more sense given what the types represent). We do this on x86 and it makes sense this way, and is also much clearer semantically.

kevinaboos (changes requested):
left a few comments about iterator usage and type conversion implementations

kevinaboos (approved):

PR #911, opened 2023-03-28
Make secondary cores usable on aarch64

NathanRoyer:
Little PR implementing the storage of online CpuIds, enabling further boot of secondary CPUs.
Another change is that the CPU-local timer is also configured for these cores now.

### DIFF
--- a/kernel/ap_start/src/lib.rs
+++ b/kernel/ap_start/src/lib.rs
@@ -50,7 +50,6 @@ pub fn kstart_ap(
     nmi_lint: u8,
     nmi_flags: u16,
 ) -> ! {
-    // As a cheap precaution
     irq_safety::disable_interrupts();
 
     info!("Booting AP: proc: {}, CPU: {}, stack: {:#X} to {:#X}, nmi_lint: {}, nmi_flags: {:#X}",
--- a/kernel/ap_start/src/lib.rs
+++ b/kernel/ap_start/src/lib.rs
@@ -50,13 +50,12 @@ pub fn kstart_ap(
     nmi_lint: u8,
     nmi_flags: u16,
 ) -> ! {
+    // As a cheap precaution
+    irq_safety::disable_interrupts();
+
     info!("Booting AP: proc: {}, CPU: {}, stack: {:#X} to {:#X}, nmi_lint: {}, nmi_flags: {:#X}",
         processor_id, cpu_id, _stack_start, _stack_end, nmi_lint, nmi_flags
     );
-
-    // As a cheap precaution
-    #[cfg(target_arch = "aarch64")]
-    irq_safety::disable_interrupts();
 
     // set a flag telling the BSP that this AP has entered Rust code
     AP_READY_FLAG.store(true, Ordering::SeqCst);
--- a/kernel/ap_start/src/lib.rs
+++ b/kernel/ap_start/src/lib.rs
@@ -54,6 +54,10 @@ pub fn kstart_ap(
         processor_id, cpu_id, _stack_start, _stack_end, nmi_lint, nmi_flags
     );
 
+    // As a cheap precaution
+    #[cfg(target_arch = "aarch64")]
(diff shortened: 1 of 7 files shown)


kernel/ap_start/src/lib.rs
```diff
@@ -54,6 +54,10 @@ pub fn kstart_ap(
         processor_id, cpu_id, _stack_start, _stack_end, nmi_lint, nmi_flags
     );
 
+    // As a cheap precaution
+    #[cfg(target_arch = "aarch64")]
+    irq_safety::disable_interrupts();
```
kevinaboos: actually this is a great idea to have in general. Can you move it to the top of this function and make it always active for all architectures?

kernel/cpu/src/aarch64.rs
```diff
@@ -2,32 +2,61 @@
...
+use derive_more::{Display, Binary, Octal, LowerHex, UpperHex};
+use irq_safety::RwLockIrqSafe;
 use core::fmt;
+use alloc::vec::Vec;
+
 use super::CpuId;
-use derive_more::{Display, Binary, Octal, LowerHex, UpperHex};
+
+// The vector of CpuIds for known and online CPU cores
+static ONLINE_CORES: RwLockIrqSafe<Vec<CpuId>> = RwLockIrqSafe::new(Vec::new());
```
kevinaboos: nit: for consistency, we're migrating terms to `cpu` instead of `core`
```suggestion
static ONLINE_CPUS: RwLockIrqSafe<Vec<CpuId>> = RwLockIrqSafe::new(Vec::new());
```

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -119,8 +119,16 @@ fn set_vbar_el1() {
 }
 
 /// Sets `VBAR_EL1` to the start of the exception vector
+/// and enables timer interrupts
 pub fn init_ap() {
     set_vbar_el1();
+
+    // Enable the CPU-local timer
+    let mut gic = GIC.lock();
+    let gic = gic.as_mut().expect("GIC is uninitialized");
```
kevinaboos: this is a bug, right? we should clarify a bit
```suggestion
    let gic = gic.as_mut().expect("BUG: init_ap(): GIC was uninitialized");
```

kernel/cpu/src/aarch64.rs
kevinaboos: do we need a `RwLock` for this? If it's not absolutely required, we should use `Mutex` because `RwLock` is way slower

kevinaboos (changes requested):
I like this changeset; eventually we should abstract `LocalApic` and the aarch64 `cpu` module such that both use an interface like `register_cpu()`.

kernel/cpu/src/aarch64.rs
NathanRoyer: Using Mutex would prevent us from making `cpu_count` cheap

kevinaboos: for a cheap `cpu_count()` impl, we should just store that in an atomicusize, like we already do on x86. Also, you may be surprised how much slower `RwLock::read()` is than `Mutex::lock()` ... so if you're concerned about making this cheap, `RwLock` is the opposite of that.

Regardless of perf, my question is primarily about semantics -- do we actually _need_ multiple readers to be able to immutably access the vector of online `CpuId`s? If so, why, and in what circumstances? (I'm not saying there aren't any, I just want to explore this)

NathanRoyer: > Regardless of perf, my question is primarily about semantics -- do we actually _need_ multiple readers to be able to immutably access the vector of online `CpuId`s? If so, why, and in what circumstances? (I'm not saying there aren't any, I just want to explore this)

The only use case I have in mind really is for `cpu_count`, so an atomicusize solves the issue.

kevinaboos (approved):

PR #913, opened 2023-03-28
Enable traditional TLB-shootdowns on aarch64

NathanRoyer:
This enables the first TLB shootdown strategy: sending an IPI to other cores and waiting for them to patch their TLBs.

:warning: depends on #910

### DIFF
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -124,16 +124,8 @@ fn set_vbar_el1() {
 }
 
 /// Sets `VBAR_EL1` to the start of the exception vector
-/// and enables timer interrupts
 pub fn init_ap() {
     set_vbar_el1();
-
-    // Enable the CPU-local timer
-    let mut gic = GIC.lock();
-    let gic = gic.as_mut().expect("GIC is uninitialized");
-    gic.set_interrupt_state(CPU_LOCAL_TIMER_IRQ, true);
-
-    enable_timer(true);
 }
 
 /// Please call this (only once) before using this crate.
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -31,7 +31,9 @@ pub const CPU_LOCAL_TIMER_IRQ: InterruptNumber = 30;
 
 /// The IRQ/IPI number for TLB Shootdowns
 ///
-/// This is arbitrarily defined (range: 0..16)
+/// Note: This is arbitrarily defined in the range 0..16,
+/// which is reserved for IPIs (SGIs - for software generated
+/// interrupts - in GIC terminology)
 pub const TLB_SHOOTDOWN_IPI: InterruptNumber = 2;
 
 const MAX_IRQ_NUM: usize = 256;
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -33,7 +33,7 @@ pub const CPU_LOCAL_TIMER_IRQ: InterruptNumber = 30;
 ///
 /// Note: This is arbitrarily defined in the range 0..16,
 /// which is reserved for IPIs (SGIs - for software generated
-/// interrupts - in GIC terminology)
+/// interrupts - in GIC terminology).
 pub const TLB_SHOOTDOWN_IPI: InterruptNumber = 2;
 
 const MAX_IRQ_NUM: usize = 256;
@@ -171,7 +171,7 @@ pub fn init() -> Result<(), &'static str> {
 /// This function registers an interrupt handler for the CPU-local
 /// timer and handles GIC configuration for the timer interrupt.
 pub fn init_timer(timer_tick_handler: HandlerFunc) -> Result<(), &'static str> {
-    // register/deregister the handler for the timer IRQ.
+    // register the handler for the timer IRQ.
     if let Err(existing_handler) = register_interrupt(CPU_LOCAL_TIMER_IRQ, timer_tick_handler) {
         if timer_tick_handler as *const HandlerFunc != existing_handler {
             return Err("A different interrupt handler has already been setup for the timer IRQ number");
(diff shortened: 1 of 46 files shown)


kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -183,6 +188,30 @@ pub fn init_timer(timer_tick_handler: HandlerFunc) -> Result<(), &'static str> {
     Ok(())
 }
 
+/// This function registers an interrupt handler for an inter-processor interrupt
+/// and handles GIC configuration for that interrupt.
+pub fn init_ipi(handler: HandlerFunc, irq_num: InterruptNumber) -> Result<(), &'static str> {
```
kevinaboos: `init_ipi` is a bit vague, maybe something like `register_ipi_handler`?

kernel/tlb_shootdown/src/lib.rs
```diff
@@ -50,26 +39,32 @@ pub fn handle_tlb_shootdown_ipi(pages_to_invalidate: PageRange) {
...
-pub fn send_tlb_shootdown_ipi(my_lapic: &mut LocalApic, pages_to_invalidate: PageRange) {        
+fn broadcast_tlb_shootdown(pages_to_invalidate: PageRange) {        
     // skip sending IPIs if there are no other cores running
     let cpu_count = cpu_count();
     if cpu_count <= 1 {
         return;
     }
 
-    // log::trace!("send_tlb_shootdown_ipi(): from AP {}, cpu_count: {}, {:?}", my_lapic.apic_id(), cpu_count, pages_to_invalidate);
+    // log::trace!("send_tlb_shootdown_ipi(): cpu_count: {}, {:?}", cpu_count, pages_to_invalidate);
```
kevinaboos: let's keep the "from CPU" info too
```suggestion
    // log::trace!("send_tlb_shootdown_ipi(): from CPU {:?}, cpu_count: {}, {:?}", cpu::current_cpu(), cpu_count, pages_to_invalidate);
```

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -29,6 +29,11 @@ static GIC: MutexIrqSafe<Option<ArmGic>> = MutexIrqSafe::new(None);
 // aarch64 manuals define the default timer IRQ number to be 30.
 pub const CPU_LOCAL_TIMER_IRQ: InterruptNumber = 30;
 
+/// The IRQ/IPI number for TLB Shootdowns
+///
+/// This is arbitrarily defined (range: 0..16)
+pub const TLB_SHOOTDOWN_IPI: InterruptNumber = 2;
```
kevinaboos: does this need to be fixed/hardcoded? on x86, we have a function that will basically "allocate"/pick a free IRQ number for a new handler (see `register_msi_interrupt` for the basic logic) and then return the IRQ number that it chose. So, you can do this in the function that you currently call `init_ipi`, and then store that IRQ number using `call_once()` after you redefine this as something like:
```rust
pub const TLB_SHOOTDOWN_IPI: Once<InterruptNumber> = Once::new();
```

I think doing something like that here is best, as won't we have devices that may need a specific IRQ number later? (for ex, on x86, ATA disks, keyboard, mouse, etc all have fixed IRQ numbers that cannot be chosen by software).

FWIW, i think this is currently hardcoded on x86 too, but we don't actually use a custom IPI, we use the NMI handler.

kevinaboos (changes requested):
Looks good, i'm happy to merge this after #910. I left a few small suggestions.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: This function not only registers the handler but also enables the corresponding interrupt in the GIC, so I went for `setup_ipi_handler`.

kernel/tlb_shootdown/src/lib.rs
NathanRoyer: Yep. I also changed this to the `if false` pattern.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: I'm not sure that peripheral IRQs can land in this range:
- 0..16 is for Software Generated Interrupts = IPIs
- 16..32 is for Local Peripheral Interrupts = the cpu-local timer, for instance
- 32..1024 is for Shared Peripheral Interrupts = PCIe devices for instance

(GIC spec)

So... in this SGI-only range, there shouldn't be any unpredictable user of these numbers 0..16.
But that doesn't prevent us from doing what you said; With this "IRQ number layout" in mind, do you still think it's desirable?

kernel/interrupts/src/aarch64/mod.rs
kevinaboos: great, sounds good

kernel/tlb_shootdown/src/lib.rs
kevinaboos: perfect!

kernel/interrupts/src/aarch64/mod.rs
kevinaboos: ok, thanks for clarifying. I think it's fine to hardcode it for now. Let's add a note in the comment there that mentions those 3 types of ranges, and how this is an SGI, so that's why it needs to be within `0..16`

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -183,6 +188,30 @@ pub fn init_timer(timer_tick_handler: HandlerFunc) -> Result<(), &'static str> {
...
 
+/// This function registers an interrupt handler for an inter-processor interrupt
+/// and handles GIC configuration for that interrupt.
+pub fn setup_ipi_handler(handler: HandlerFunc, irq_num: InterruptNumber) -> Result<(), &'static str> {
+    assert!(irq_num < 16, "Inter-processor interrupts must have a number in the range 0..16");
+
+    // register/deregister the handler for the timer IRQ.
+    if let Err(existing_handler) = register_interrupt(irq_num, handler) {
+        if handler as *const HandlerFunc != existing_handler {
+            return Err("A different interrupt handler has already been setup for the timer IRQ number");
```
kevinaboos: copy/paste error in this message, this isn't the timer IRQ

kernel/tlb_shootdown/src/lib.rs
```diff
@@ -85,8 +82,15 @@ pub fn send_tlb_shootdown_ipi(my_lapic: &mut LocalApic, pages_to_invalidate: Pag
     *TLB_SHOOTDOWN_IPI_PAGES.write() = Some(pages_to_invalidate);
     TLB_SHOOTDOWN_IPI_COUNT.store(cpu_count - 1, Ordering::SeqCst); // -1 to exclude this core 
 
-    // let's try to use NMI instead, since it will interrupt everyone forcibly and result in the fastest handling
-    my_lapic.send_nmi_ipi(LapicIpiDestination::AllButMe); // send IPI to all other cores but this one
+    #[cfg(target_arch = "x86_64")] {
+        let my_lapic = apic::get_my_apic().expect("If there are more than one CPU ready, this one should be registered");
```
kevinaboos: a bit clearer, plus "BUG"
```suggestion
        let my_lapic = apic::get_my_apic()
            .expect("BUG: broadcast_tlb_shootdown(): couldn't get LocalApic");
```

kevinaboos (changes requested):
a few minor nits

kevinaboos (approved):
Looks great now! Thanks

PR #915, opened 2023-03-31
Introduction of `arm_boards` crate & GIC driver fixes

NathanRoyer:
This PR brings the following changes:
- `arm_boards` crate: Per-board definitions for AArch64 builds - currently the number of CPUs, their IDs, and the interrupt controller configuration including one GIC redistributor base address per core (motivation for this PR)
- the `gic` crate was modified to use this
- the redistributor initialization code now enables the dispatch of "1 of N"-distributed SPIs (motivation for this PR)
- the redistributors of all cores are now initialized by the bootstrap processor as part of `interrupts::init`
- the distributor initialization code makes sure that asleep CPU cores can be selected for "1 of N"-distributed SPIs
- `multicore_bringup` uses the list of `CpuId`s from `arm_boards` (no longer bruteforcing them)
- some conversions in `cpu` were made const
- Bump `vte` dependency in text-terminal so that `gic` can use version `0.7.2` of `arrayvec` (used during initialization)

Two changes were directly imported from #910:
- Fix a bug in `get_spi_target` & `set_spi_target` where an atomic u32 R/W was used to manipulate a u64 MMIO register
- Introduce `Offset32` & `Offset64` types to distinguish 32b & 64b registers

All in all, this fixes the problems discovered via discussion on #910 (some conditions for "1 of N" distribution of SPIs were not asserted).


kernel/gic/src/gic/mod.rs
```diff
@@ -208,25 +235,36 @@ impl ArmGic {
...
-                    mapped.into_borrowed_mut(0).map_err(|(_, e)| e)?
-                };
-
-                let redist_sgippi = {
-                    let pages = allocate_pages(1).ok_or("couldn't allocate pages for the extended redistributor interface")?;
-                    let frames = allocate_frames_at(redist + REDIST_SGIPPI_OFFSET, 1)?;
-                    let mapped = page_table.map_allocated_pages_to(pages, frames, mmio_flags)?;
-                    mapped.into_borrowed_mut(0).map_err(|(_, e)| e)?
-                };
+                let mut redistributors: ArrayVec<ArmGicV3RedistPages, { arm_boards::CPUS }> = ArrayVec::new();
```
hecatia-elegua: https://github.com/rust-lang/rust/pull/75644 if we don't want ArrayVec here, maybe this works.
```rust
let redistributors: [ArmGicV3RedistPages; arm_boards::CPUS] = core::array::try_from_fn(|i| {
    let phys_addr = redist[i];
    /* ... */
    Ok(ArmGicV3RedistPages { redistributor, redist_sgippi })
})?;
```

kernel/gic/src/gic/redist_interface.rs
```diff
@@ -44,18 +57,17 @@ const TIMEOUT_ITERATIONS: usize = 0xffff;
...
+    registers.write_volatile(offset::WAKER, reg);
 
     // then poll ChildrenAsleep until it's cleared
 
     let children_asleep = || {
-        registers.read_volatile(offset::RD_WAKER) & RD_WAKER_CHLIDREN_ASLEEP > 0
+        registers.read_volatile(offset::WAKER) & WAKER_CHLIDREN_ASLEEP > 0
     };
 
     let mut counter = 0;
```
hecatia-elegua, line 71: If this closure based timeout loop is supposed to not be optimized away, I think it still is: https://godbolt.org/z/9WK56qdqr
wait no I'm wrong, though now I wonder why it works

hecatia-elegua (commented):
Solid. Only two suggestions :)

kernel/gic/src/gic/mod.rs
NathanRoyer: Ah yes thank you very much, that's what I was looking for :ok_hand:

kernel/text_terminal/Cargo.toml
```diff
@@ -8,7 +8,7 @@ description = "A text-based terminal emulator that supports ANSI, VT100, xterm,
 bitflags = "1.1.0"
 core2 = { version = "0.4.0", default-features = false, features = ["alloc", "nightly"] }
 unicode-width = "0.1.8"
-vte = "0.10.1"
+vte = "0.11.0"
```
kevinaboos: undo unrelated change

kernel/gic/src/gic/redist_interface.rs
kevinaboos: huh? 

I'm not sure what this comment means. @NathanRoyer?

kernel/gic/src/gic/mod.rs
kevinaboos: nice suggestion, thanks @hecatia-elegua !

kernel/gic/src/gic/mod.rs
```diff
@@ -132,6 +148,13 @@ impl GicRegisters {
 
 const_assert_eq!(core::mem::size_of::<GicRegisters>(), 0x1000);
 
+fn current_redist_index() -> usize {
```
kevinaboos: nit: give this a name (and ideally, some doc comments too) that indicate it is the redist index for the current _cpu_, not just any other type of "current" concept

kernel/arm_boards/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+authors = ["Nathan Royer <nathan.royer.pro@gmail.com>"]
+name = "arm_boards"
+description = "Per-board AArch64 configuration"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+memory_structs = { path = "../memory_structs" }
+cpu = { path = "../cpu" }
```
kevinaboos: see my other comment -- we should invert this dependency such that `cpu` depends on this crate instead of this crate depending on `cpu`.

kernel/arm_boards/src/board.rs
```diff
@@ -0,0 +1,8 @@
+//! Default board config file.
```
kevinaboos: Based on the name `board.rs`, I was expecting this file to contain a struct definition that includes all the fields that each board will need to specify, and then each board file will instantiate an instance of that struct. In fact, I think you should make that change now; otherwise it'll be difficult to standardize the content of each board config file in a maintainable way.

This module could contain a default board file too, or that could be moved to a module called `default.rs`.

kernel/arm_boards/src/qemu_virt.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use super::{InterruptControllerConfig::{self, GicV3}, GicV3InterruptControllerConfig};
+use cpu::{CpuId, MpidrValue};
+use memory_structs::PhysicalAddress;
+
+// local utility function to generate the CPU id from
+// the affinity level 0 number of the cpu core
+const fn cpu_id(aff0: u8) -> CpuId {
+    CpuId::from(MpidrValue::new(0, 0, 0, aff0))
+}
```
kevinaboos: I think I mentioned this a while back on a previous PR, but the ultimate goal of `CpuId` is to be a safe/private type that cannot be invalidly/arbitrarily instantiated.

Now that we have board config files, we can make this happen again. The way to do that is:
* Define a new public type in this crate's `lib.rs`, with a private inner value that allows another crate to access it.
    ```rust
    // You can choose a better name for this struct if you want. Or make it u64 if needed.
    pub struct DefinedCpuId(u32);
    impl DefinedCpuId {
        pub fn value(&self) -> u32 { self.0 }
    }
    ```
* And then in the `cpu` crate, you add a `From<DefinedCpuId> for CpuId` implementation, or alternatively for `MpidrValue`, whichever one works best.
* You should also use this in the `BoardConfig` struct; see my other comment.

kernel/arm_boards/src/qemu_virt.rs
```diff
@@ -0,0 +1,36 @@
+//! AArch64 Board Config for the `virt` machine of Qemu
```
kevinaboos: All of these board-specific details should be encapsulated in a struct defined in the `board.rs` module (or in `lib.rs`, e.g.,
```rust
pub struct BoardConfig<const NUM_CPUS: usize> {
    cpu_list: [DefinedCpuId; NUM_CPUS],
    interrupt_controller: InterruptControllerConfig,
    ...
}
```
Then you instantiate it here in this file (pseudo-code-y):

```rust
pub const NUM_CPUS: usize = 4;
pub static ARM_BOARD: BoardConfig = BoardConfig::<NUM_CPUS> {
    cpu_list: [
        DefinedCpuId(0),
        DefinedCpuId(1),
        DefinedCpuId(2),
        DefinedCpuId(3),
    ],
    ...
}
```

You've already done this for the interrupt controller config, so nice job there -- just extend it to everything.

kevinaboos (changes requested):

kernel/text_terminal/Cargo.toml
NathanRoyer: You may have not seen the top comment, which explains why this was required; anyway, thanks to hecatia-elegua's suggestion, I wont need arrayvec so this can indeed be reverted.

kernel/gic/src/gic/redist_interface.rs
hecatia-elegua: Afaik the rust compiler can optimize timeout loops away unless there is some operation inside the loop which can't be reasoned about by the compiler, i.e. volatile accesses or an `asm!` block. I guess in this case it optimizes the `timeout` closure (on line 74, not shown in this comment window) in a way that the volatile read happens inside of it.

NathanRoyer: Yeah, if rustc optimized a `read_volatile` away it would be a bug imo. Anyway, this actually compiles to a volatile read at the moment.

kernel/arm_boards/src/board.rs
NathanRoyer: I think the makefile should explicitely select a board in the future so that they don't try to run a kernel which is inadequate for their board (it would crash easily), which is why I put a compile-time error in `board.rs`, so I'll rename `board.rs` to `unselected.rs` and put the error there.

Note that right now, the qemu-virt feature is enabled by default as we currently have no other option.

kernel/arm_boards/src/qemu_virt.rs
NathanRoyer: I have another solution, which would avoir us duplicating the Mpidr parsing/packing code: we could move the definition of MpidrValue to arm_boards, and put the constructor as `pub(crate)`.

kernel/arm_boards/src/board.rs
kevinaboos: agreed, i like that choice.

kevinaboos (approved):

kernel/gic/src/gic/redist_interface.rs
kevinaboos: Ah, I see, github hid the relevant code -- @hecatia-elegua was talking about the timeout checking loop being optimized away, not the volatile register read. They're right, I checked the actual asm and it wasn't in the loop body. I've changed it to fix that.

hecatia-elegua: Wait I thought it was 🤣 
well whatever, if it works for you

PR #917, opened 2023-04-02
Implement basic per-CPU / CPU-local storage

kevinaboos:
* Not yet used anywhere
* Only supported on x86_64

Reminder: remove temp tests and log statements before merging


kernel/ap_start/src/lib.rs
```diff
@@ -102,9 +102,10 @@ pub fn kstart_ap(
         nmi_lint,
         nmi_flags,
     ).unwrap();
-
+
```
tsoutsman: Whitespace.

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    size:   usize,
+    align:  usize,
+}
+// NOTE: These fields must be kept in sync with `cpu_local::FixedCpuLocal`.
+impl FixedCpuLocal {
+    const SELF_PTR_OFFSET: usize = 0;
+    pub const CPU_ID:                       Self = Self { offset: 8,  size: 4, align: 4 };
+    pub const PREEMPTION_COUNT:             Self = Self { offset: 12, size: 1, align: 1 };
+    pub const TASK_SWITCH_PREEMPTION_GUARD: Self = Self { offset: 16, size: 8, align: 4 };
+    pub const DROP_AFTER_TASK_SWITCH:       Self = Self { offset: 24, size: 8, align: 8 };
```
tsoutsman: Could we somehow test these offsets in `per_cpu` using `const_assert`?

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+struct CpuLocalDataImage(MappedPages);
+impl CpuLocalDataImage {
+    /// This function does 3 main things:
+    /// 1. Allocates a new CPU-local data image for this CPU.
+    /// 2. Sets the self pointer value such that it can be properly accessed.
+    /// 3. Sets this CPU's base register (e.g., GsBase on x86_64) to the address
+    ///    of this new data image, making it "currently active" and accessible.
+    fn new() -> Result<CpuLocalDataImage, &'static str> {
+        // 1. Allocate a single page to store each CPU's local data.
+        let mut mp = memory::create_mapping(1, PteFlags::new().writable(true).valid(true))?;
```
tsoutsman: Minor thing, but asserting that we fit into a single page i.e.
```rust
const_assert_eq!(DROP_AFTER_TASK_SWITCH.offset + DROP_AFTER_TASK_SWITCH.size, memory::PAGE_SIZE)
```

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+        func(local_ref_mut)
+    }
+
+    /// Returns the value of the self pointer, which points to this CPU's `PerCpuData`.
+    fn self_ptr(&self) -> usize {
+        let self_ptr: usize;
+        unsafe {
+            #[cfg(target_arch = "x86_64")]
+            asm!(
+                concat!("mov {}, gs:[0]"), // the SELF_PTR_OFFSET is 0
```
tsoutsman: Is there a reason for `concat!`?

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    ///    Must be one of the public constants defined in [`FixedCpuLocal`].
+    ///
+    /// # Safety
+    /// This is unsafe because we currently do not have a way to guarantee
+    /// that the caller-provided type `T` is the same as the type intended for use
+    /// by the given `FixedCpuLocal`.
+    ///
+    /// Thus, the caller must guarantee that the type `T` is correct for the
+    /// given `FixedCpuLocal`.
+    pub const unsafe fn new_fixed(
```
tsoutsman: When do we use this function?

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    let entry = regions.entry(cpu_id);
+    let data_image = match entry {
+        Entry::Vacant(v) => v.insert(CpuLocalDataImage::new()?),
+        Entry::Occupied(_) => return Err("BUG: cannot init CPU-local data more than once"),
+    };
+    let self_ptr = data_image.0.start_address().value();
+
+    // Run the given initializer function on the per-CPU data.
+    let new_data_image = CpuLocal::<{FixedCpuLocal::SELF_PTR_OFFSET}, P>(PhantomData);
+    new_data_image.with_mut(|per_cpu_data_mut| {
```
tsoutsman: Converting the address to a `&mut P` (in `with_mut`) even if we don't use `per_cpu_data_mut` is undefined behaviour as the memory is uninitialised.

kernel/per_cpu/src/lib.rs
```diff
@@ -0,0 +1,108 @@
...
+/// by other crates using the functions in the [`cpu_local`] crate.
+#[allow(dead_code)] // These fields are accessed via `cpu_local` functions.
+#[repr(C)]
+pub struct PerCpuData {
+    /// A pointer to the start of this struct in memory, similar to a TLS self pointer.
+    /// This has a different initial value for each CPU's data image, of course.
+    ///
+    /// We use this to allow writes to this entire structure (for initialization),
+    /// and also to allow faster access to large fields herein () accelerate accesses to large items
+    self_ptr: usize,
```
tsoutsman: Could you elaborate on how this field allows faster access to large fields because this field is causing the problems with uninitialised memory in `cpu_local`?

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+            todo!("CPU Locals are not yet supported on non-x86_64 platforms");
+        };
+        self_ptr 
+    }
+}
+
+impl<const OFFSET: usize, T: Copy> CpuLocal<OFFSET, T> {
+    /// Returns a copy of this `CpuLocal`'s inner value of type `T`.
+    ///
+    /// This is a convenience functiononly available for types where `T: Copy`.
```
tsoutsman:
```suggestion
    /// This is a convenience function only available for types where `T: Copy`.
```

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    pub const PREEMPTION_COUNT:             Self = Self { offset: 12, size: 1, align: 1 };
+    pub const TASK_SWITCH_PREEMPTION_GUARD: Self = Self { offset: 16, size: 8, align: 4 };
+    pub const DROP_AFTER_TASK_SWITCH:       Self = Self { offset: 24, size: 8, align: 8 };
+}
+
+/// A reference to a CPU-local variable.
+///
+/// Note that this struct doesn't contain an instance of the type `T`,
+/// and dropping it has no effect.
+pub struct CpuLocal<const OFFSET: usize, T>(PhantomData<*mut T>);
```
tsoutsman: If we use the enum representation above, I *think* this could be defined as:
```rust
unsafe trait Field: Sized {
    // Idk if this would work.
    const SIZE_CHECK: () = const_assert_eq!(core::mem::size_of::<Self>(), Self::field().size());
    const ALIGN_CHECK: () = const_assert_eq!(core::mem::align_of::<Self>(), Self::field().align());

    const fn field() -> CpuLocalField;
}

pub struct CpuLocal<T: Field>(PhantomData<*mut T>);
```

the offset would be calculated using `T::field().offset()`.

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+use alloc::collections::{BTreeMap, btree_map::Entry};
+use memory::{MappedPages, PteFlags};
+use preemption::{hold_preemption, PreemptionGuard};
+use spin::Mutex;
+
+#[cfg(target_arch = "x86_64")]
+use x86_64::{registers::model_specific::GsBase, VirtAddr};
+
+/// Info use to obtain a reference to a CPU-local variable; see [`CpuLocal::new()`].
+pub struct FixedCpuLocal {
```
tsoutsman: I think the name and implementation are quite confusing. Maybe something like this would be better:
```rust
pub enum CpuLocalField {
    SelfPointer,
    CpuId,
    PreemptionCount,
    TaskSwitchPreemptionGuard,
    DropAfterTaskSwitch,
}

impl CpuLocalField {
    const fn offset(&self) -> usize {
        match self {
            SelfPointer => 0,
            CpuId => 8,
            PreemptionCount => 12,
            TaskSwitchPreemptionGuard => 16,
            DropAfterTaskSwitch => 24,
    }

    // const fn size

    // const fn align
}

Comment (tsoutsman) [reply kernel/cpu_local/src/lib.rs]: See comment below for more details.

**Review by tsoutsman — COMMENTED** [review]
I've proposed a slightly different API that would decrease the amount of unsafe *syncing* we have to do between `PerCpuData` and `FixedCpuLocal`.

Just to make sure we're on the same page `gs` is a register that isn't used in regular code and so we can use it for a different purpose right?

File: kernel/cpu_local/src/lib.rs (hunk 2 above)
Comment (kevinaboos) [reply kernel/cpu_local/src/lib.rs]: i was trying to figure out a way to do that, but I don't think it's possible given the cyclic dependency issues it would cause. If you have another idea though, I'm open to it and would love to have these be more thoroughly tested and checked for errors, especially because this design is brittle towards change.

File: kernel/cpu_local/src/lib.rs (hunk 4 above)
Comment (kevinaboos) [reply kernel/cpu_local/src/lib.rs]: leftover relic of old changes, thanks for catching.

File: kernel/cpu_local/src/lib.rs (hunk 5 above)
Comment (kevinaboos) [reply kernel/cpu_local/src/lib.rs]: we don't use it yet, but it would be used like so:
```rust
// in `kernel/preemption`:
static PREEMPTION_COUNT: CpuLocal<..., PreemptionCount> = unsafe { CpuLocal::new_fixed(FixedCpuLocal::PREEMPTION_COUNT) };

// in hold_preemption, increase the count:
let prev_val = PREEMPTION_COUNT.with(|count| count.fetch_add(1, ...));

// in PreemptionGuard::drop(), decrease the count:
let prev_val = PREEMPTION_COUNT.with(|count| count.fetch_sub(1, ...));
```

kernel/cpu_local/src/lib.rs
kevinaboos: ya, basically i was looking at clever ways to avoid `core::ptr::write()` here, but i don't think there is a good way. You're right.

kernel/per_cpu/src/lib.rs
kevinaboos: basically it avoids having to load the entire field before modifying it. I forgot to link to this discussion in the comments, but I meant to add this in: https://github.com/rust-osdev/x86_64/pull/257#issuecomment-849514649

kernel/cpu_local/src/lib.rs
kevinaboos: Ah yes, good call. An even better solution would be to pass in the size of the data structure from `per_cpu`, which ensures we'll have enough memory.

kernel/cpu_local/src/lib.rs
kevinaboos: yea i like the enum approach and actually had that as my first design attempt. couldn't figure out how to do the const asserts easily, but I'll try it with your trait idea above!

kevinaboos:
> Just to make sure we're on the same page `gs` is a register that isn't used in regular code and so we can use it for a different purpose right?

Correct, the GS segment register is unused and is frequently used for CPU-local storage on x86. It's similar to how the FS segment register is used for thread-local storage.

kernel/cpu_local/src/lib.rs
kevinaboos: ah ok, so now I remember why i moved away from an enum+trait based design -- you can't have a `const fn` in traits yet.

kevinaboos: if the rust PR for "enum variants as types" ever gets merged in, then we could improve this, but for now I hacked around that using enum variants that contain empty structs for each field.
EDIT: nvm, this didn't work.

kevinaboos: Actually, we need the user of `CpuLocal` to be able to specify the type `T`, and at least with the design I had in mind, that user wouldn't be able to impl our private trait for their type `T` ... unless we defined and impl'd that trait inside `per_cpu`, which would introduce another cyclic dep problem. hmm

kernel/cpu_local/src/lib.rs
kevinaboos: nvm, i misread what you suggested. Done now.

kevinaboos:
@tsoutsman I attempted a few different ways to use traits in the `CpuLocal` type, with the goal being to make the RAII declaration of a CPU-local variable as simple as possible. It's pretty hard to make the type signature simple, but I was able to get it down to a single generic type parameter (which itself has one type parameter). I didn't push it to this branch because I don't love it, but take a look here: https://github.com/theseus-os/Theseus/commit/3b24b1daddb215a4e352027a49115b8396579bbf.  At the very least, it removes/hides the `const OFFSET` parameter from `CpuLocal`, which I do think was a bit of a flawed/leaky design. This design lets the caller do something like:
```rust
static PREEMPTION_COUNT: CpuLocal<CpuLocalPreemptionCount<PreemptionCount>> = unsafe { CpuLocal::new() };
```

--------------

The only other option I was able to get to work was to parameterize the `CpuLocalVariable` trait with a type, like so:
```rust
pub trait CpuLocalVariable<T: Sized> { ... }
```
which works, but you then need to have both `T` and `V` parameters for `CpuLocal`:
```rust
pub struct CpuLocal<T: Sized, V: CpuLocalVariable<T>> {
    _typ: PhantomData<T>,
    _var: PhantomData<V<T>>,
}

impl<T: Sized, V: CpuLocalVariable<T>> CpuLocal<T, V> {
    pub const unsafe fn new(_variable_marker: V) -> Self {
        Self { _typ: PhantomData, _var: PhantomData }
    }
```
This is clearer but ergonomically worse, as you have to declare/instantiate it like so:
```rust
static PREEMPTION_COUNT: CpuLocal<PreemptionCount, CpuLocalVariable<PreemptionCount>> = unsafe { CpuLocal::new(CpuLocalPreemptionCount) };
```

-------------
If you have any ideas on how to further simplify this, please do let me know.

tsoutsman:
Apologies force push was because I pushed to the wrong remote repo.

Here's a quick enum implementation I made:
https://github.com/theseus-os/Theseus/commit/b0ce230b862b6eb22a96cc2fc716214e58b2dc99

It removes size and alignment from `cpu_local`, removes `OFFSET` const parameter, prevents `CpuLocal` from being instantiated with arbitrary offsets and types. The one bad thing is that `per_cpu` now needs to define a bunch of wrapper types so that it can implement `Field` for them.

tsoutsman:
right but that introduces a circular dependency hmm

we could have the Field impls in the crated where the type is defined e.g. cpu for CpuId

kevinaboos:
Thanks, that enum impl is similar to what I tried, but the addition of the unsafe trait with its associated const does help. I was trying to avoid an unsafe trait -- i wonder if there's a clever way to prevent other crates from implementing the `Field` trait.

> The one bad thing is that `per_cpu` now needs to define a bunch of wrapper types so that it can implement `Field` for them.

This is fine, but we should impl Deref+DerefMut on those wrappers so the crates that use `cpu_local` don't have to be aware of said newtype wrappers.

> we could have the Field impls in the crates where the type is defined, e.g. cpu for CpuId

Yeah, i don't really have any other ideas for workarounds here, so I think this is fine. But we should expose it via a macro like `define_cpu_local_type!(<type>)` such that it's not easy/recommended to directly impl `Field`. This ties into trying to prevent someone from implementing `Field`, but I do realize that macros wouldn't prevent that.

I suppose you could also define a proc_macro that would implement hidden traits for a cpu-local type, but that might be too complex right now. My future idea for TLS-style CPU locals incorporates proc_macros for custom attributes, but we can discuss that later.

---------

Finally, you left this comment, which I agree with, but yes that's one of the many reasons why `CpuLocal::new*` was unsafe -- the type provided had to match the offset. 
> // NOTE: The fact that this previously compiled in `cpu_local` is
>        // indicative of an unsafe API, as the offsets (and types) are
>        // completely arbitrary. There's nothing stopping us from trying to
>        // access CpuLocal::<16, String> which would be undefined behaviour.

kevinaboos:
> Here's a quick enum implementation I made: [b0ce230](https://github.com/theseus-os/Theseus/commit/b0ce230b862b6eb22a96cc2fc716214e58b2dc99)
> 
> It removes size and alignment from `cpu_local`, removes `OFFSET` const parameter, prevents `CpuLocal` from being instantiated with arbitrary offsets and types. The one bad thing is that `per_cpu` now needs to define a bunch of wrapper types so that it can implement `Field` for them.

Thanks, I have merged/included that commit into this PR branch, and am going from there.

> we could have the Field impls in the crated where the type is defined e.g. cpu for CpuId

Working on this now. I've also addressed some naming issues.

--------

Will re-request a review from you when done.

kernel/cpu_local/src/lib.rs
kevinaboos: fixed in latest changeset

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+///
+/// This macro implements the [`CpuLocalField`]
+#[macro_export]
+macro_rules! define_per_cpu_field {
+    ($TypeWrapper:ident, $InnerType:ty, $PerCpuField:expr) => {
+        #[repr(transparent)]
+        struct $TypeWrapper($InnerType);
+        unsafe impl $crate::CpuLocalField for $TypeWrapper {
+            const FIELD: $crate::PerCpuField = $PerCpuField;
+        }
```
tsoutsman: I don't like the fact that we have a macro to implement an unsafe trait.

The trait isn't marked unsafe to satisfy the compiler; it's marked unsafe to warn the user that they must uphold certain invariants. The following definition would cause undefined behaviour but would never use the unsafe keyword:
```rust
define_per_cpu_field!(CpuId, cpu::CpuId, PerCpuField::PreemptionCount);
```

There isn't much point in marking it unsafe if we provide a macro that avoids the `unsafe` keyword completely.

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    // const SIZE_CHECK:  () = assert!(Self::FIELD.size() == size_of::<Self>());
+    // const ALIGN_CHECK: () = assert!(Self::FIELD.align() == align_of::<Self>());
+}
+
+/// A macro used to define a type wrapper for inclusion in the `PerCpuData` struct.
+///
+/// This macro implements the [`CpuLocalField`]
+#[macro_export]
+macro_rules! define_per_cpu_field {
```
tsoutsman: Can we move this macro into `per_cpu`? No other crate should be using this macro.

kernel/per_cpu/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+            test_string: TestString(alloc::string::String::from("test_string hello")),
+
+        }
+    }
+}
+
+define_per_cpu_field!(CpuId, cpu::CpuId, PerCpuField::CpuId);
+define_per_cpu_field!(PreemptionCount, preemption::PreemptionCount, PerCpuField::PreemptionCount);
+define_per_cpu_field!(TaskSwitchPreemptionGuard, Option<preemption::PreemptionGuard>, PerCpuField::TaskSwitchPreemptionGuard);
+define_per_cpu_field!(DropAfterTaskSwitch, Option<TaskRef>, PerCpuField::DropAfterTaskSwitch);
```
tsoutsman: AFAICT, this creates a circular dependency that `per_cpu` was trying to avoid in the first place.

`preemption` will not be able to access the CPU local `PreemptionCount` as that would lead to a circular dependency (`per_cpu` depends on `preemption` depends on `per_cpu`). The trait has to be implemented in the crate where the type is defined (e.g. in `cpu` or `preemption`).

tsoutsman (changes requested):

kernel/cpu_local/src/lib.rs
kevinaboos: That's fair, agreed.

kernel/cpu_local/src/lib.rs
kevinaboos: Agreed, will do.

kernel/per_cpu/src/lib.rs
kevinaboos: Yeah, that's true.

`preemption` is a special case, given that `cpu_local` itself must depend on `preemption` because its `CpuLocal::with()` accessors use `PreemptionGuard` and `hold_preemption()`. So in the next PR when we start using per-CPU preemption counts, I will have to merge `cpu_local` and `preemption`. That's a known issue, so no worries there.

I'll work on this.
```rust

```

kernel/cpu_local/src/lib.rs
kevinaboos: Actually I'll remove it entirely.

kevinaboos:
Made the changes, thanks for the suggestion. Though not included in this changeset, I confirmed it's possible to instantiate a CPU-local task type, i.e., this works:
```rust
static TASK_SWITCH_PREEMPTION_GUARD: cpu_local::CpuLocal<TaskSwitchPreemptionGuard> = cpu_local::CpuLocal::new(PerCpuField::TaskSwitchPreemptionGuard);
pub fn test_me() -> Option<PreemptionGuard> {
    TASK_SWITCH_PREEMPTION_GUARD.with_mut(|t| t.0.take())
}
```

see [my latest commit msg](https://github.com/theseus-os/Theseus/pull/917/commits/5d64794e646fdef4f8bb5f75b8e997ea0b19f321) for other details

kernel/cpu_local/src/lib.rs
```diff
@@ -0,0 +1,309 @@
...
+impl<T: CpuLocalField> CpuLocal<T> {
+    /// Creates a new reference to a predefined CPU-local variable.
+    ///
+    /// ## Arguments
+    /// * `field`: the field in the `per_cpu::PerCpuData` struct that
+    ///   you wish to access via the returned `CpuLocal` object.
+    ///
+    /// The type `T: CpuLocalField` must be specified with the turbofish operator:
+    /// ```rust,no_run
+    /// static CPU_ID: CpuLocal<CpuId> = CpuId::new(PerCpuField::CpuId);
```
tsoutsman:
```suggestion
    /// static CPU_ID: CpuLocal<CpuLocalCpuId> = CpuId::new(PerCpuField::CpuId);
````

tsoutsman (approved):
One small doc change.

PR #933, opened 2023-04-11
Support ARMv8 PL011 UARTs in `serial_port_basic`

NathanRoyer:
Changes:
- Store PL011 configuration in `arm_boards`
- Support reading from and writing to PL011 UARTs in `serial_port_basic`; interrupt management is not implemented.
- Remove `logger_aarch64`; Rename `logger_x86_64` as `logger`

I suggest merging #932 before reviewing this.


kevinaboos:
#932 has been merged in

NathanRoyer:
> #932 has been merged in

Thanks; I rebased on main

kevinaboos (approved):
I'll accept this for now because it technically works, but I'd like to see a series of future improvements to the serial port driver design in general.

1. You've identified a few small commonalities between the two drivers, which is good, but we can go so much further. We should effectively combine the two into a single `SerialPort` type, since they share like 95% of the code, especially stuff for I/O. Most of the methods are `SerialPort` are duplicated across the two archs. You could optionally refactor out the differences into an arch-specific `SerialPortInner` type, or something similar.
    * Another choice would be to create a serial port trait that both architecture `mod`ules implement, in which the "higher-level" functions (like `out_str()`) are provided trait methods, since they're the same for both archs and just rely on the required `out_byte()` method.
2. The identification of a serial port should probably move away from the "COM" term, which is PC-specific. AFAIK, that's not typically used on ARM devices (?) 
3. The actual PL011 driver should be redone to be safer; right now there is a lot of unmanaged unsafe code there, even the constructor. Safe access to MMIO registers is part of the reason behind our design of `MappedPages`.
4. Re-implementing the PL011 driver will also help to address the issue with the weird `Option`-filled design, which I presume stems from the need to construct a dummy instance for the "replace on drop" semantics we use with the `TriState` wrapper. (Of course, we can change that part too.)
5. Just because the existing x86/legacy implementation of something is infallible doesn't mean that we should avoid returning results in the corresponding new aarch64 version of something. It's preferable to return a `Result<SerialPort, ...>` from the `new()` constructor rather than panicking a lot just to match the existing function signature.

It'd be great if you could submit 3-4 separate PRs for these design points soon. Happy to help as needed.

PR #935, opened 2023-04-13
Use `spin` in `sync`

tsoutsman:
Recent upstream changes expose the `try_lock_weak` function on `spin`'s mutexes allowing us to use them in our implementation.


kevinaboos (approved):
Awesome! Thanks, looks great.

PR #939, opened 2023-04-19
GIC driver fixes & improvements

NathanRoyer:
- Implement `ArmGic::{get_interrupt_priority, set_interrupt_priority}`
- Implement Copy, Clone & Debug for SpiDestination & IpiTargetCpu
- Fix obvious execution flow issue in `ArmGic::set_spi_target` (method had never been tested)
- Fix GIC configuration for TLB Shootdown IPIs (initialize CPU interface for secondary cores + actually enable the interrupt for the BSP)

### DIFF
--- a/kernel/gic/src/gic/dist_interface.rs
+++ b/kernel/gic/src/gic/dist_interface.rs
@@ -184,7 +184,7 @@ impl super::ArmGic {
                 } else {
                     let mpidr = reg & 0xff00ffffff;
                     let cpu_id = MpidrValue::try_from(mpidr)?.into();
-                    return Ok(SpiDestination::Specific(cpu_id));
+                    Ok(SpiDestination::Specific(cpu_id))
                 }
             }
         }
--- a/kernel/gic/src/gic/dist_interface.rs
+++ b/kernel/gic/src/gic/dist_interface.rs
@@ -221,10 +221,10 @@ impl super::ArmGic {
             };
 
             v3.dist_extended.write_volatile_64(offset::P6IROUTER, value);
+        } else {
+            // If we're on gicv2 then affinity routing is off
+            // so we landed in the first block
+            unreachable!()
         }
-
-        // If we're on gicv2 then affinity routing is off
-        // so we landed in the first block
-        unreachable!()
     }
 }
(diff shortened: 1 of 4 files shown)


kernel/gic/src/gic/dist_interface.rs
```diff
@@ -208,10 +221,10 @@ impl super::ArmGic {
             };
 
             v3.dist_extended.write_volatile_64(offset::P6IROUTER, value);
+        } else {
+            // If we're on gicv2 then affinity routing is off
+            // so we landed in the first block
+            unreachable!()
```
kevinaboos: in general, `unreachable` code found in if-else statements is a code smell indicating a design flaw. Is there a better way to represent this to avoid the if-else issue, e.g., a match statement or something?

kevinaboos (changes requested):
Looks good, just one small request/question.

kernel/gic/src/gic/dist_interface.rs
NathanRoyer: Yep, sure:

```rust
    /// Sets the destination of an SPI.
    pub fn set_spi_target(&mut self, int: InterruptNumber, target: SpiDestination) {
        assert!(int >= 32, "interrupts number below 32 (SGIs & PPIs) don't have a target CPU");
        match self {
            Self::V2(_) | Self::V3(ArmGicV3 { affinity_routing: false, .. }) => {
                if let SpiDestination::Specific(cpu) = &target {
                    assert!(cpu.value() < 8, "affinity routing is disabled; cannot target a CPU with id >= 8");
                }

                let value = match target {
                    SpiDestination::Specific(cpu) => 1 << cpu.value(),
                    SpiDestination::AnyCpuAvailable => {
                        let list = TargetList::new_all_cpus()
                            .expect("This is invalid: CpuId > 8 AND affinity routing is disabled");
                        list.0 as u32
                    },
                    SpiDestination::GICv2TargetList(list) => list.0 as u32,
                };

                self.distributor_mut().write_array_volatile::<4>(offset::ITARGETSR, int, value);
            },
            Self::V3(ArmGicV3 { affinity_routing: true, dist_extended, .. }) => {
                let value = match target {
                    SpiDestination::Specific(cpu) => MpidrValue::from(cpu).value(),
                    // bit 31: Interrupt Routing Mode
                    // value of 1 to target any available cpu
                    SpiDestination::AnyCpuAvailable => P6IROUTER_ANY_AVAILABLE_PE,
                    SpiDestination::GICv2TargetList(_) => {
                        panic!("Cannot use SpiDestination::GICv2TargetList with affinity routing enabled");
                    },
                };

                dist_extended.write_volatile_64(offset::P6IROUTER, value);
            }
        }
    }
```

pushing this

kevinaboos: perfect, well done!

kevinaboos (approved):

PR #942, opened 2023-04-23
Add `sync::RwLock`

tsoutsman:
Depends on theseus-os/spin-rs#2.

### DIFF
--- a/libs/sync/src/lib.rs
+++ b/libs/sync/src/lib.rs
@@ -22,7 +22,7 @@ pub mod spin {
     };
 }
 
-pub trait MutexFlavour {
+pub trait MutexFlavor {
     const INIT: Self::LockData;
 
     type LockData;
@@ -46,7 +46,7 @@ pub trait MutexFlavour {
     fn post_unlock(data: &Self::LockData);
 }
 
-pub trait RwLockFlavour {
+pub trait RwLockFlavor {
     const INIT: Self::LockData;
 
     type LockData;
@@ -91,7 +91,7 @@ pub trait DeadlockPrevention {
     fn enter() -> Self::Guard;
 }
 
-impl<P> MutexFlavour for P
+impl<P> MutexFlavor for P
 where
     P: DeadlockPrevention,
 {
@@ -136,7 +136,7 @@ where
     fn post_unlock(_: &Self::LockData) {}
 }
 
-impl<P> RwLockFlavour for P
+impl<P> RwLockFlavor for P
 where
     P: DeadlockPrevention,
 {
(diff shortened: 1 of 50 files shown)


kevinaboos:
theseus-os/spin-rs#2 has been merged in

tsoutsman:
~~Should be ready.~~

Nevermind.

kevinaboos:
> ~Should be ready.~
> 
> Nevermind.

Ok lmk when you'd like another review.

tsoutsman:
@kevinaboos should be ready.

libs/sync/src/lib.rs
```diff
@@ -22,35 +24,59 @@ pub mod spin {
...
+        rw_lock: &'a spin::RwLock<T>,
+        data: &'a Self::RwLockData,
+    ) -> (spin::RwLockReadGuard<'a, T>, Self::Guard);
+
+    fn write_rw_lock<'a, T>(
+        rw_lock: &'a spin::RwLock<T>,
+        data: &'a Self::RwLockData,
+    ) -> (spin::RwLockWriteGuard<'a, T>, Self::Guard);
+
+    fn post_rw_lock_unlock(data: &Self::RwLockData, is_writer_or_last_reader: bool);
```
kevinaboos: missing docs on these functions

kevinaboos (changes requested):
Sorry for the delay, i left on vacation right before you readied this PR. 

Anyway, i like everything here, nice work!  One change request -- I think it'd be a good design choice to separate out the trait representing the flavor of actions for a Mutex from the trait representing the flavor of actions for an RwLock. There are a few reasons for this:
* I think it makes sense to separate them as they are typically implemented separately in both the std lib as well as crates like `spin`. 
* Also, some concrete impls of those flavor traits may not want to (or may not be able to) offer both a Mutex and an RwLock; separating them into two traits allows one to be implemented without the other, and is also a bit clearer imo.
* `RwLock<T>` typically has different Send/Sync bounds for its type `T` than `Mutex<T>` does. I realize this isn't part of the Flavor trait API, but it still speaks to the expectation of separation between the two.
* It would give the Flavor trait a more specific name: `MutexFlavor` and `RwLockFlavor`

kevinaboos (changes requested):
Sorry for the delay, i left on vacation right before you readied this PR. 

Anyway, i like everything here, nice work!  One change request -- I think it'd be a good design choice to separate out the trait representing the flavor of actions for a Mutex from the trait representing the flavor of actions for an RwLock. There are a few reasons for this:
* I think it makes sense to separate them as they are typically implemented separately in both the std lib as well as crates like `spin`. 
* Also, some concrete impls of those flavor traits may not want to (or may not be able to) offer both a Mutex and an RwLock; separating them into two traits allows one to be implemented without the other, and is also a bit clearer imo.
* `RwLock<T>` typically has different Send/Sync bounds for its type `T` than `Mutex<T>` does. I realize this isn't part of the Flavor trait API, but it still speaks to the expectation of separation between the two.
* It would give the Flavor trait a more specific name: `MutexFlavor` and `RwLockFlavor`

kevinaboos (approved):
awesome work!

PR #943, opened 2023-04-25
Macro-based cross platform interrupt handling

NathanRoyer:
- This creates a way for interrupt handler to be declared in a cross-platform way:

```rust
// Cross platform scheduling code
interrupt_handler!(timer_tick_handler, None, _stack_frame, {
    // ...
});
```

- Handlers in `scheduler` & `serial_ports` were modified to adopt this.
- Handlers in `exceptions_early`, `exceptions_full`, `keyboard`, `mouse` & `pit_clock` were left untouched as they are x86-specific.
- Handlers in `ata` & `e1000` were not modified because there's no way to test them on aarch64 atm.

There's a debate to have: Is it preferable to use macros (as this PR implements) or to declare handlers as a normal function/callback.

The macro-based approach has the benefit of allowing x86 handlers to be optimized using the "x86-interrupt" special ABI. 

Also, this removes the need to convert the existing x86_64 part of `interrupts`, because there's no need to declare raw interrupt handlers for every possible interrupt number, as would be the case with normal callbacks (or maybe there's another solution that I don't see?).

On the other hand, having all handlers declared as normal callbacks would make the code a bit easier to read and understand.

This is mergeable but I'm open to discussion :ok_hand:

### DIFF
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -2,6 +2,8 @@ use core::arch::global_asm;
 use core::ops::DerefMut;
 use core::fmt;
 
+use crate::EoiBehaviour;
+
 use cortex_a::registers::*;
 
 use tock_registers::interfaces::Writeable;
@@ -53,6 +55,14 @@ struct SpsrEL1(InMemoryRegister<u64, SPSR_EL1::Register>);
 #[repr(transparent)]
 struct EsrEL1(InMemoryRegister<u64, ESR_EL1::Register>);
 
+#[cfg(target_arch = "aarch64")]
+#[macro_export]
+macro_rules! interrupt_handler {
+    ($name:ident, $x86_64_interrupt_number:expr, $context:ident, $code:block) => {
+        extern "C" fn $name($context: &$crate::InterruptStackFrame) -> $crate::EoiBehaviour $code
+    }
+}
+
 /// The exception context as it is stored on the stack on exception entry.
 ///
 /// Warning: `table.s` assumes this exact layout. If you modify this,
@@ -75,14 +85,8 @@ pub struct ExceptionContext {
     esr_el1: EsrEL1,
 }
 
-#[derive(Debug, PartialEq, Eq)]
-#[repr(C)]
-pub enum EoiBehaviour {
-    CallerMustSignalEoi,
-    HandlerHasSignaledEoi,
-}
-
-type HandlerFunc = extern "C" fn(&ExceptionContext) -> EoiBehaviour;
+pub type HandlerFunc = extern "C" fn(&InterruptStackFrame) -> EoiBehaviour;
+pub type InterruptStackFrame = ExceptionContext;
 
 // called for all exceptions other than interrupts
 fn default_exception_handler(exc: &ExceptionContext, origin: &'static str) {
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -456,7 +456,7 @@ extern "C" fn current_elx_irq(exc: &mut ExceptionContext) {
         false => default_irq_handler,
     };
 
-    if handler(exc) == EoiBehaviour::CallerMustSignalEoi {
+    if handler(exc) == EoiBehaviour::HandlerDidNotSendEoi {
         // handler has returned, we can lock again
         let mut int_ctrl = INTERRUPT_CONTROLLER.lock();
         int_ctrl.as_mut().unwrap().end_of_interrupt(irq_num);
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -45,7 +45,7 @@ const MAX_IRQ_NUM: usize = 256;
 // it's an array of function pointers which are meant to handle IRQs.
 // Synchronous Exceptions (including syscalls) are not IRQs on aarch64;
 // this crate doesn't expose any way to handle them at the moment.
-static IRQ_HANDLERS: RwLockIrqSafe<[HandlerFunc; MAX_IRQ_NUM]> = RwLockIrqSafe::new([default_irq_handler; MAX_IRQ_NUM]);
+static IRQ_HANDLERS: RwLockIrqSafe<[InterruptHandler; MAX_IRQ_NUM]> = RwLockIrqSafe::new([default_irq_handler; MAX_IRQ_NUM]);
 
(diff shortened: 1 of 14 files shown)


kernel/interrupts/src/x86_64/mod.rs
```diff
@@ -32,6 +32,19 @@ static RESERVED_IRQ_LIST: [u8; 3] = [
 ];
 
 
+#[macro_export]
+macro_rules! interrupt_handler {
+    ($name:ident, $x86_64_interrupt_number:expr, $stack_frame:ident, $code:block) => {
+        extern "x86-interrupt" fn $name(sf: $crate::InterruptStackFrame) {
+            let $stack_frame = &sf;
+            if let CallerMustSignalEoi = $code {
```
kevinaboos: this should probably be fully qualified in order to avoid name clashes or prevent calling crates from having to import it specifically.
```suggestion
            if let $crate::EoiBehavior::CallerMustSignalEoi = $code {
```

kernel/interrupts/src/lib.rs
```diff
@@ -9,3 +9,10 @@
 mod arch;
 
 pub use arch::*;
+
+#[derive(Debug, PartialEq, Eq)]
+#[repr(C)]
+pub enum EoiBehaviour {
+    CallerMustSignalEoi,
+    HandlerHasSignaledEoi,
```
kevinaboos: add enum-level docs and docs for each variant.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -75,14 +85,8 @@ pub struct ExceptionContext {
...
 
-#[derive(Debug, PartialEq, Eq)]
-#[repr(C)]
-pub enum EoiBehaviour {
-    CallerMustSignalEoi,
-    HandlerHasSignaledEoi,
-}
-
-type HandlerFunc = extern "C" fn(&ExceptionContext) -> EoiBehaviour;
+pub type HandlerFunc = extern "C" fn(&InterruptStackFrame) -> EoiBehaviour;
```
kevinaboos: i vote to rename this `InterruptHandler` for clarity. Applies to both archs.

kernel/interrupts/src/lib.rs
kevinaboos: I think we could also improve the names of the variants, specifically the first one. It's a bit counter-intuitive to think of an interrupt handler as having a "caller", since they are typically described as entry points that the hw jumps to directly. I understand we are now wrapping it, but it's still not "called" in the traditional sense of a function.

Perhaps it should be something like:
```rust
pub enum EoiBehaviour {
    HandlerSentEoi,
    HandlerDidNotSendEoi,
}
```
which has no ambiguity.

kernel/interrupts/src/x86_64/mod.rs
```diff
@@ -32,6 +32,19 @@ static RESERVED_IRQ_LIST: [u8; 3] = [
 ];
 
 
+#[macro_export]
+macro_rules! interrupt_handler {
+    ($name:ident, $x86_64_interrupt_number:expr, $stack_frame:ident, $code:block) => {
+        extern "x86-interrupt" fn $name(sf: $crate::InterruptStackFrame) {
+            let $stack_frame = &sf;
```
kevinaboos, line 40: what's this line for?

kernel/interrupts/src/x86_64/mod.rs
```diff
@@ -32,6 +32,19 @@ static RESERVED_IRQ_LIST: [u8; 3] = [
 ];
 
 
+#[macro_export]
+macro_rules! interrupt_handler {
```
kevinaboos: add some docs with an explanation of each macro argument and also some example use cases. Just by looking at the macro definition itself, you wouldn't be able to tell that the `$x86_64_interrupt_number` needs to be an `Option<u8>`.

kernel/scheduler/src/lib.rs
```diff
@@ -46,36 +46,25 @@ pub fn init() -> Result<(), &'static str> {
...
+        interrupts::init_timer(timer_tick_handler)?;
         interrupts::enable_timer(true);
         Ok(())
     }
 }
 
-/// The handler for each CPU's local timer interrupt, used for preemptive task switching.
-#[cfg(target_arch = "aarch64")]
-extern "C" fn aarch64_timer_handler(_exc: &interrupts::ExceptionContext) -> interrupts::EoiBehaviour {
+// Cross platform scheduling code
```
kevinaboos:
```suggestion
// Architecture-independent timer interrupt handler for preemptive scheduling.
```

kevinaboos (changes requested):

kernel/interrupts/src/lib.rs
NathanRoyer: No problem. Here's the doc I added:
```rust
#[derive(Debug, PartialEq, Eq)]
#[repr(C)]
pub enum EoiBehaviour {
    /// The interrupt handler hasn't called the [`eoi`] function,
    /// in which case it will be called automatically once the
    /// handler returns.
    HandlerDidNotSendEoi,
    /// The interrupt handler has called the [`eoi`] function.
    HandlerSentEoi,
}
```

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: I agree with this name.

kernel/interrupts/src/x86_64/mod.rs
NathanRoyer: If you're talking about the assignment (line 39), it's only there to provide a reference to the stack frame because on aarch64 the stack frame is a reference, so I thought I'd mimic that, as no handler was exploiting the fact that it's a copy. Additionally, it cannot hurt to pass this by reference, as it's a pretty big structure, right?

kernel/interrupts/src/x86_64/mod.rs
NathanRoyer: Here's the doc I added (same for both archs):
image

(hi res screenshot)

kernel/interrupts/src/lib.rs
kevinaboos: great!

kernel/interrupts/src/x86_64/mod.rs
kevinaboos: oops yea, i was asking about 39. Never mind, i mis-read it initially, i see what it is now.

kevinaboos (approved):
nice work, thanks!

PR #948, opened 2023-05-02
Serial port interrupt handling on aarch64 & COM1 console

NathanRoyer:
Changes:
- Interrupts from the PL011 serial port are configured in the GIC and handled like on x86_64
- `device_manager` implements serial port configuration only on aarch64
- `deferred_interrupt_tasks` is now compatible with aarch64
- COM1 switches to a `hull` console (which doesn't have any app available) when we write to it. Log messages still appear while this console runs.

We can revert some changes of [4e8a63f](https://github.com/theseus-os/Theseus/pull/948/commits/4e8a63f68c204f00c7a635e123a5ec9cfc34105f) once relocations work on aarch64.

### DIFF
--- a/kernel/console/src/lib.rs
+++ b/kernel/console/src/lib.rs
@@ -109,17 +109,27 @@ fn shell_loop(
         .name(format!("{address:?}_to_tty"))
         .spawn()?;
 
-    let new_app_ns = mod_mgmt::create_application_namespace(None)?;
 
-    let (app_file, _ns) =
-        mod_mgmt::CrateNamespace::get_crate_object_file_starting_with(&new_app_ns, "hull-")
-            .expect("Couldn't find shell in default app namespace");
+    let task;
+    #[cfg(target_arch = "x86_64")] {
+        let new_app_ns = mod_mgmt::create_application_namespace(None)?;
 
-    let path = path::Path::new(app_file.lock().get_absolute_path());
-    let task = spawn::new_application_task_builder(path, Some(new_app_ns))?
-        .name(format!("{address:?}_shell"))
-        .block()
-        .spawn()?;
+        let (app_file, _ns) =
+            mod_mgmt::CrateNamespace::get_crate_object_file_starting_with(&new_app_ns, "hull-")
+                .expect("Couldn't find shell in default app namespace");
+
+        let path = path::Path::new(app_file.lock().get_absolute_path());
+        task = spawn::new_application_task_builder(path, Some(new_app_ns))?
+            .name(format!("{address:?}_shell"))
+            .block()
+            .spawn()?;
+    }
+    #[cfg(target_arch = "aarch64")] {
+        task = spawn::new_task_builder(hull::main, alloc::vec::Vec::new())
+            .name(alloc::format!("{address:?}_shell"))
+            .block()
+            .spawn()?;
+    }
 
     let id = task.id;
     let stream = Arc::new(tty.slave());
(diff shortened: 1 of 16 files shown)


kernel/console/src/lib.rs
```diff
@@ -109,17 +109,27 @@ fn shell_loop(
         .name(format!("{address:?}_to_tty"))
         .spawn()?;
 
-    let new_app_ns = mod_mgmt::create_application_namespace(None)?;
 
-    let (app_file, _ns) =
-        mod_mgmt::CrateNamespace::get_crate_object_file_starting_with(&new_app_ns, "hull-")
-            .expect("Couldn't find shell in default app namespace");
+    let task;
```
hecatia-elegua, line 113: Is this possible?
```rust
let task = {
    #[cfg(target_arch = "x86_64")] {
        //...
        .spawn()?
    }
    #[cfg(target_arch = "aarch64")] {
        //...
        .spawn()?
    }
}
```

kernel/serial_port_basic/src/aarch64.rs
```diff
@@ -77,19 +77,28 @@ impl SerialPort {
...
             port_address: serial_port_address,
-            inner: Some(PL011::new(addr as *mut _)),
+            inner: Some(pl011),
             _mapped_pages: Some(mapped_pages),
         }
     }
 
     /// Enable or disable interrupts on this serial port for various events.
     ///
     /// Note: currently unimplemented on `aarch64`.
```
hecatia-elegua: now it is halfway there

hecatia-elegua (commented):
Haven't seen anything suspicious. I'm guessing the `unwrap`s and `expect`s are needed.
I like the edition updates and so on.
Git really produced a great diff in device_manager :)


Some of the serial port changes are here: https://github.com/theseus-os/pl011/commit/464dbf2288a5d9e7445b0b1f404ddd41b1fd1c1e
I didn't look at them, though.

kernel/console/src/lib.rs
NathanRoyer: I think not. But we could use `cfg_if`, if you prefer.
Note that this is part of the temporary fix which will be removed once relocations work on aarch64.

hecatia-elegua: yeah sorry, doesn't matter then

kevinaboos:
> Git really produced a great diff in device_manager :)

jeez you're not kidding. That is horrendous. I'll try to fix it by breaking some indentation rules...

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -32,6 +34,12 @@ static INTERRUPT_CONTROLLER: MutexIrqSafe<Option<ArmGic>> = MutexIrqSafe::new(No
 // aarch64 manuals define the default timer IRQ number to be 30.
 pub const CPU_LOCAL_TIMER_IRQ: InterruptNumber = 30;
 
+/// The IRQ number reserved for the PL011 Single-Serial-Port Controller
+/// which Theseus currently uses for logging and UART console.
+//
+// Qemu attributes number 33 to this interrupt.
+pub const PL011_RX_SPI: InterruptNumber = 33;
```
kevinaboos, line 41: should this go in the device file imitation we have? (the board config crate)

kevinaboos (approved):
I'll approve this since the code itself is fine, but I left one inline comment and have another general comment here.

I think it's probably time to move the PL011 driver into our code base directly since it's diverged significantly from that of the original crate you forked. There are more changes I want to make to it too (and to our x86 serial port driver), so we sort of need it to be in-kernel.

Feel free to open a new issue to track that, or add it to the existing aarch64 one, or just do it in a PR without any corresponding issue.

PR #949, opened 2023-05-05
Update MacOS setup script

tsoutsman:
Changes:
  - Script now exclusively uses `brew`
  - Tools are installed to `/usr/local` so that the user doesn't need to change `PATH`.
  - Tools are copied into `/tmp/theseus_tools_src` to not clog up the home directory
  - Tools are only installed if they aren't already


kevinaboos (approved):
Excellent, thanks!

PR #950, opened 2023-05-07
Fix some typos and wrong links in the book

hecatia-elegua:
also changed Spawning -> Initializing in the book to match the actual code, spell checker said "Initing" isn't a word

### DIFF
--- a/book/cspell.json
+++ b/book/cspell.json
@@ -83,6 +83,7 @@
         "linkcheck",
         "manycore",
         "MMIO",
+        "OSDI",
         "Phalereus",
         "reinterpretive",
         "rodata",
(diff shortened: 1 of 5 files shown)


kevinaboos (approved):
Thanks!

I used this PR as an opportunity to improve other parts of the task subsystem's docs too.

hecatia-elegua:
Nice.
Just noting because I didn't even know this:
I noticed "Kevin Boos' " -> "Kevin Boos's", I thought the rule is "no 's if last letter is s".
After a quick search, I now learned "With regular plural nouns we use ' not 's." and "The rule depends on the house style manual you’re following." So s's is used in more formal situations, interesting.

kevinaboos:
It's a matter of preference, but in my experience something like `Boos'` is an older writing style that folks have moved away from. I also tend to not use that style because it is ambiguous with plural possessives, that is, do you mean a singular Boos-owned item, or multiple Boo-owned items?

lol

PR #951, opened 2023-05-07
Update docs.yaml

hecatia-elegua:
see https://github.blog/changelog/2023-05-04-github-actions-all-actions-will-run-on-node16-instead-of-node12/


kevinaboos (approved):
Thanks, appreciate this update!

PR #952, opened 2023-05-08
Fix real hardware graphics mode picking

hecatia-elegua:
full context from discord:
> After startup (after switching away from captain), I see 4 windows mirroring each other, side by side. And they're striped black horizontally.

> https://github.com/theseus-os/Theseus/blob/40805a7934710f16f90000f65497d3a61bb7b681/kernel/multicore_bringup/src/x86_64.rs#L59 I would keep this at the start of the struct, like the field definition in https://github.com/theseus-os/Theseus/blob/40805a7934710f16f90000f65497d3a61bb7b681/kernel/nano_core/src/asm/ap_realmode.asm#L66. I also tried doing this without changing https://github.com/theseus-os/Theseus/blob/40805a7934710f16f90000f65497d3a61bb7b681/kernel/nano_core/src/asm/ap_realmode.asm#L258 to be at the first byte, and for some reason that worked for my laptop? So I hypothesize mode or width is causing the visual glitches.

> mode is 292 / 0x124 / 0001_0010_0100

> GraphicInfo._attributes is 0xBA is 0b10111010, sooo, the best_mode.mode we select is not supported?
> see VBE core functions, ModeAttributes
> 0 means unsupported by hardware

There's just a small visual bug left at the end of captain, which I think is happening because we set the VGA card via BIOS interrupt immediately on the last AP's init in handle_ap_cores, so it changes graphic modes while still logging to the screen (and maybe BIOS or the window manager or sth doesn't like that).

### DIFF
--- a/kernel/nano_core/src/asm/ap_realmode.asm
+++ b/kernel/nano_core/src/asm/ap_realmode.asm
@@ -181,9 +181,10 @@ get_vbe_card_info:
     jne .next_mode          ; We failed to get info about this mode. Go back and try the next mode.
 
     ; Check whether the current mode is supported by the hardware.
+    ; If bit 0 is clear, the mode is unsupported, so continue on to the next mode.
     mov word ax, [VBEModeInfo.attributes]
     test word ax, 1
-    jz .next_mode ; if first bit is clear, mode is unsupported, so fall through.
+    jz .next_mode
 
     ; We only support modes with 32-bit pixel sizes
     cmp byte [VBEModeInfo.bitsperpixel], 32
--- a/kernel/nano_core/src/asm/ap_realmode.asm
+++ b/kernel/nano_core/src/asm/ap_realmode.asm
@@ -173,16 +173,22 @@ get_vbe_card_info:
     ; Here, we attempt to get the mode info for the mode we just iterated to
     push esi
     mov [current.mode], cx  ; Store the current mode in `current.mode`
-    mov ax, 0x4F01          ; 0x4F01 is the argument fo the BIOS 0x10 interrupt used to get the currrent mode information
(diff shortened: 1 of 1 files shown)


kevinaboos:
Thanks a lot for digging into this and solving it! Been quite busy lately, preoccupied with other stuff and some travel, so I apologize for the long delay. Hopefully I can test this on my real hw shortly and merge it by early next week.

kevinaboos (approved):
I tested this and can confirm it works on all h/w and emulator configurations I have available to me. Thanks so much!

Side note, it's really weird that some hardware exposes a particular VBE mode that isn't actually supported. I had figured the mode wouldn't be available at all if it wasn't supported.

kevinaboos:
> There's just a small visual bug left at the end of captain, which I think is happening because we set the VGA card via BIOS interrupt immediately on the last AP's init in handle_ap_cores, so it changes graphic modes while still logging to the screen (and maybe BIOS or the window manager or sth doesn't like that).

To clarify, you mean this visual bug still occurs even with the changes from this PR? If so, can you describe the nature of the "visual bug" in more detail, or perhaps capture an image of it? I understand if you can't, it may be hard to capture.

hecatia-elegua:
> weird that some hardware exposes a particular VBE mode that isn't actually supported

I wondered about that as well, my only idea is:
"VBE is made available through the video card's BIOS"
-> so, the laptop might have a video card which supports higher/different resolution, maybe even allow to connect an external display

> visual bug still occurs

A new one, I'm guessing on real hardware when the graphics mode picking changes something. That would mean your hardware just stays on the same e.g. resolution, while my hardware doesn't, so only I see a slight glitch.

PR #953, opened 2023-05-08
Read one mouse packet byte per interrupt

### LINKED ISSUE #832: Moving PS/2 mouse and typing at the same time
The only thing working in that interaction is mouse clicks, which get interweaved with keyboard key events like it should be.
This is _still the case_ if I just get _one byte_ per interrupt, even if the only code running in these interrupts gets the byte and prints it.

Maybe qemu is at fault, that's why I would like to get bochs/real hardware running first.

hecatia-elegua:
Fix #832
`Mutex` should be good here, right?
Can these handlers be called from multiple cores at the same time? Interleaved, or nested?

### DIFF
--- a/kernel/mouse/src/lib.rs
+++ b/kernel/mouse/src/lib.rs
@@ -4,7 +4,7 @@
 #![feature(abi_x86_interrupt)]
 
 use log::{error, warn};
-use spin::Once;
+use spin::{Mutex, Once};
 use mpmc::Queue;
 use event_types::Event;
 use x86_64::structures::idt::InterruptStackFrame;
@@ -15,11 +15,51 @@ use ps2::{PS2Mouse, MousePacket};
 /// Because we perform the typical PIC remapping, the remapped IRQ vector number is 0x2C.
 const PS2_MOUSE_IRQ: u8 = interrupts::IRQ_BASE_OFFSET + 0xC;
 
+const PS2_MAX_MOUSE_BYTES: usize = 4;
+
 static MOUSE: Once<MouseInterruptParams> = Once::new();
 
+/// Everything we need in [`ps2_mouse_handler`].
 struct MouseInterruptParams {
     mouse: PS2Mouse<'static>,
     queue: Queue<Event>,
+    packet_bytes: PacketBytes,
+}
+
+/// Somewhat like an array/vec mixture to allow pushing single bytes
+/// of a mouse packet per interrupt into an array.
+/// This can handle MouseId 0 (3 bytes) and 3, 4 (4 bytes).
+struct PacketBytes {
+    len: usize,
+    inner: Mutex<[u8; PS2_MAX_MOUSE_BYTES]>,
+    // where to push the next element
+    cursor: Mutex<usize>,
+}
+
+impl PacketBytes {
+    const fn new(len: usize) -> Self {
+        Self { len, inner: Mutex::new([0; PS2_MAX_MOUSE_BYTES]), cursor: Mutex::new(0) }
+    }
+    // TODO: we can use a u32 once we switch to the bilge crate
+    fn push(&self, value: u8) {
+        let mut cursor = self.cursor.lock();
+        if *cursor < PS2_MAX_MOUSE_BYTES {
+            self.inner.lock()[*cursor] = value;
+            *cursor += 1;
+        }
+    }
+
+    /// Return the packet bytes if they're filled
+    fn filled_bytes(&self) -> Option<[u8; 4]> {
+        let mut cursor = self.cursor.lock();
+
+        if *cursor == self.len {
+            *cursor = 0;
+            Some(*self.inner.lock())
+        } else {
+            None
+        }
+    }
 }
 
 /// Initialize the PS/2 mouse driver and register its interrupt handler.
@@ -41,9 +81,13 @@ pub fn init(mut mouse: PS2Mouse<'static>, mouse_queue_producer: Queue<Event>) ->
         "PS/2 mouse IRQ was already in use! Sharing IRQs is currently unsupported."
     })?;
 
+    // Initialize the mouse packet bytes, which will be filled by 3-4 interrupts,
+    // depending on the MouseId
+    let packet_bytes = PacketBytes::new(mouse.packet_size());
+
     // Final step: set the producer end of the mouse event queue.
     // Also add the mouse struct for access during interrupts.
-    MOUSE.call_once(|| MouseInterruptParams { mouse, queue: mouse_queue_producer });
+    MOUSE.call_once(|| MouseInterruptParams { mouse, queue: mouse_queue_producer, packet_bytes });
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


kernel/mouse/src/lib.rs
```diff
@@ -15,11 +15,51 @@ use ps2::{PS2Mouse, MousePacket};
...
     mouse: PS2Mouse<'static>,
     queue: Queue<Event>,
+    packet_bytes: PacketBytes,
+}
+
+/// Somewhat like an array/vec mixture to allow pushing single bytes
+/// of a mouse packet per interrupt into an array.
+/// This can handle MouseId 0 (3 bytes) and 3, 4 (4 bytes).
+struct PacketBytes {
+    len: usize,
```
kevinaboos, line 33: if `len` is always a statically-known value, we can use const generics here.

kevinaboos: Also `len` is a bit vague, give it a more precise name (and docs) that conveys it's an "expected byte count" or the "number of bytes in a complete mouse packet"

kernel/mouse/src/lib.rs
```diff
@@ -15,11 +15,51 @@ use ps2::{PS2Mouse, MousePacket};
...
+    inner: Mutex<[u8; PS2_MAX_MOUSE_BYTES]>,
+    // where to push the next element
+    cursor: Mutex<usize>,
```
kevinaboos, line 36: there's no real benefit here to using two mutexes. I'd condense this into one.

kevinaboos: If you want to get even fancier, you can pack all of this data into an atomic u64 since you only need 5 bytes. That would be the fastest option of all.

kernel/mouse/src/lib.rs
```diff
@@ -41,9 +81,13 @@ pub fn init(mut mouse: PS2Mouse<'static>, mouse_queue_producer: Queue<Event>) ->
         "PS/2 mouse IRQ was already in use! Sharing IRQs is currently unsupported."
     })?;
 
+    // Initialize the mouse packet bytes, which will be filled by 3-4 interrupts,
+    // depending on the MouseId
+    let packet_bytes = PacketBytes::new(mouse.packet_size());
```
kevinaboos, line 86: here's where you'd specify the const generic parameter, either 3 or 4 depending upon mouse ID.

kernel/mouse/src/lib.rs
```diff
@@ -56,18 +100,20 @@ pub fn init(mut mouse: PS2Mouse<'static>, mouse_queue_producer: Queue<Event>) ->
...
-            let mouse_packet = mouse.read_mouse_packet();
-            
-            if mouse_packet.always_one() != 1 {
-                // this could signal a hardware error or a mouse which doesn't conform to the rule
-                warn!("ps2_mouse_handler(): Discarding mouse data packet since its third bit should always be 1.");
-            } else if let Err(e) = handle_mouse_input(mouse_packet, queue) {
-                error!("ps2_mouse_handler(): {e:?}");
+            packet_bytes.push(mouse.read_packet_byte());
+
+            if let Some(bytes) = packet_bytes.filled_bytes() {
```
kevinaboos, line 108: these two instructions acquire and release a total of 4 locks. You can likely combine their functionality together such that a "push" operation returns an `Option<[u8; N]>` indicating whether the expected number of bytes have been received (and providing a copy of those 3 or 4 bytes).

Also, if you adopt a single-Mutex design (or better yet, the AtomicU64 design), that would reduce this overhead from 4 to 1 lock acquisitions.

kevinaboos (changes requested):
> Mutex should be good here, right?
> Can these handlers be called from multiple cores at the same time? Interleaved, or nested?

In our current configuration, a device interrupt is only delivered to one CPU core at once, and they generally aren't able to be nested (i.e., the next mouse interrupt cannot occur until the prior mouse interrupt has been completely handled). However, using just a `Mutex` for this generally isn't advised, considering that those aspects could change at any moment.

`MutexIrqSafe` is required when another execution context (like a normal function) is also accessing variables that are shared with and accessed by an interrupt handler. However, here, only the interrupt handler accesses this state, so we don't technically need one.

On a related note, see my inline comments about using an AtomicU64 to pack all the mouse packet bytes and the cursor into an efficient u64. This approach would be best.
If you don't want to do that and would prefer to stick with a `Mutex` instead, you should move that static mutex instance into the interrupt handler function scope itself, in order to ensure that nobody else can access it.

--------
One other question: is it valid/possible to read multiple bytes from the mouse for a single interrupt? For example, if a lot of mouse activity is occurring, must we wait for each successive interrupt to read all the bytes? I'm not saying you should always forcibly read either 3 or 4 bytes at once, since that could indeed cause perf issues or infinite loops, but rather that you could iteratively check whether more bytes are available and read them all until no more bytes are available, all within a single interrupt handler invocation.

I'm asking because I'm not sure if the PS/2 mouse behavior allows/supports this. 
In other devices, like the serial port, you do receive an interrupt when data is ready to be read in, but you can read more than just one byte per interrupt, which is faster than only reading a single byte per interrupt.

kernel/mouse/src/lib.rs
```diff
@@ -56,18 +100,20 @@ pub fn init(mut mouse: PS2Mouse<'static>, mouse_queue_producer: Queue<Event>) ->
 /// 
 /// In some cases (e.g. on device init), [the PS/2 controller can also send an interrupt](https://wiki.osdev.org/%228042%22_PS/2_Controller#Interrupts).
 extern "x86-interrupt" fn ps2_mouse_handler(_stack_frame: InterruptStackFrame) {
-    if let Some(MouseInterruptParams { mouse, queue }) = MOUSE.get() {
+    if let Some(MouseInterruptParams { mouse, queue, packet_bytes }) = MOUSE.get() {
+        // using `while` here didn't interact well with the window manager and increases handler runtime
```
hecatia-elegua, line 104: I've tried using a loop here and will try again, since some of the code changed.
This would handle multi-byte interrupts.

kernel/mouse/src/lib.rs
hecatia-elegua: Nice idea, didn't consider const generics.

kernel/mouse/src/lib.rs
hecatia-elegua: Again, thanks, exactly what I need to learn.

kevinaboos: yea np at all, this is just the kind of thing that only becomes apparent once you have another set of eyes on it. :smile:

kernel/mouse/src/lib.rs
kevinaboos: yeah, I recall the previous code there. I phrased my comment as a question because legitimately not sure if the PS/2 mouse device actually allows for multiple bytes to be read per interrupt (i.e., does it do "batching" of multiple bytes into a single interrupt?); once we answer that then we can determine what the best choice is here.

hecatia-elegua: Yes it does, I've seen some osdev wiki people use `while` here 👍

kevinaboos:
Hi @hecatia-elegua, I'm doing some PR triage and noticed this one was mostly ready, just needed a few minor fixes. 

If you have time, would you be interested in making those changes? If not, I totally understand; just let me know and I can take over and fix things up before merging.

hecatia-elegua:
I think back when I tried switching this to const generics + AtomicU64 it didn't just work easily and then I switched to other tasks. Feel free I guess

kevinaboos:
> I think back when I tried switching this to const generics + AtomicU64 it didn't just work easily and then I switched to other tasks. 

Oh interesting! What did you try, and do you have your attempted code anywhere where I could check it out? (even if it's messy)

> Feel free I guess

Ok thanks, will do.

hecatia-elegua:
can't find it, I think I just removed the folder when switching computers

PR #957, opened 2023-05-10
Move convenience mapping functions from `nic_initialization` to `memory`

NathanRoyer:
This replaces the following items from `nic_initialization`:
- `allocate_memory`
- `allocate_device_register_memory` (which was unused)
- `NIC_MAPPING_FLAGS`

With:
- `memory::MMIO_FLAGS`: Mapping flags that can be used to map MMIO registers.
- `memory::map_frame_range`: A convenience function that creates a new memory mapping by allocating specific frames from physical memory.
- `memory::map_mmio_range`: A convenience function that creates a new memory mapping by allocating specific frames from physical memory with [`MMIO_FLAGS`] as flags.
- `pci::pci_map_bar_mem` as a clean replacement for `allocate_device_register_memory`, even though it's also unused. Maps device memory specified by a Base Address Register.

`gic` & `serial_port_basic` were also modified so that they use `memory::map_mmio_range`.

### DIFF
--- a/kernel/e1000/src/lib.rs
+++ b/kernel/e1000/src/lib.rs
@@ -34,13 +34,13 @@ use regs::*;
 use spin::Once; 
 use alloc::{collections::VecDeque, format, sync::Arc, vec::Vec};
 use irq_safety::MutexIrqSafe;
-use memory::{PhysicalAddress, BorrowedMappedPages, BorrowedSliceMappedPages, Mutable};
+use memory::{PhysicalAddress, BorrowedMappedPages, BorrowedSliceMappedPages, Mutable, map_mmio_range};
 use pci::{PciDevice, PCI_INTERRUPT_LINE, PciConfigSpaceAccessMechanism};
 use kernel_config::memory::PAGE_SIZE;
 use interrupts::eoi;
 use x86_64::structures::idt::InterruptStackFrame;
 use network_interface_card:: NetworkInterfaceCard;
-use nic_initialization::{allocate_memory, init_rx_buf_pool, init_rx_queue, init_tx_queue};
+use nic_initialization::{init_rx_buf_pool, init_rx_queue, init_tx_queue};
 use intel_ethernet::descriptors::{LegacyRxDescriptor, LegacyTxDescriptor};
 use nic_buffers::{TransmitBuffer, ReceiveBuffer, ReceivedFrame};
 use nic_queues::{RxQueue, TxQueue, RxQueueRegisters, TxQueueRegisters};
@@ -289,7 +289,7 @@ impl E1000Nic {
     /// * `device`: reference to the nic device
     /// * `mem_base`: the physical address where the NIC's memory starts.
     fn map_e1000_regs(
-        _device: &PciDevice, 
+        _device: &PciDevice,
         mem_base: PhysicalAddress,
     ) -> Result<(
         BorrowedMappedPages<E1000Registers, Mutable>, 
@@ -303,10 +303,10 @@ impl E1000Nic {
         const TX_REGISTERS_SIZE_BYTES: usize = 4096;
         const MAC_REGISTERS_SIZE_BYTES: usize = 114_688;
 
-        let nic_regs_mapped_page     = allocate_memory(mem_base, GENERAL_REGISTERS_SIZE_BYTES)?;
-        let nic_rx_regs_mapped_page  = allocate_memory(mem_base + GENERAL_REGISTERS_SIZE_BYTES, RX_REGISTERS_SIZE_BYTES)?;
-        let nic_tx_regs_mapped_page  = allocate_memory(mem_base + GENERAL_REGISTERS_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES, TX_REGISTERS_SIZE_BYTES)?;
-        let nic_mac_regs_mapped_page = allocate_memory(mem_base + GENERAL_REGISTERS_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES + TX_REGISTERS_SIZE_BYTES, MAC_REGISTERS_SIZE_BYTES)?;
+        let nic_regs_mapped_page     = map_mmio_range(mem_base, GENERAL_REGISTERS_SIZE_BYTES)?;
+        let nic_rx_regs_mapped_page  = map_mmio_range(mem_base + GENERAL_REGISTERS_SIZE_BYTES, RX_REGISTERS_SIZE_BYTES)?;
+        let nic_tx_regs_mapped_page  = map_mmio_range(mem_base + GENERAL_REGISTERS_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES, TX_REGISTERS_SIZE_BYTES)?;
+        let nic_mac_regs_mapped_page = map_mmio_range(mem_base + GENERAL_REGISTERS_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES + TX_REGISTERS_SIZE_BYTES, MAC_REGISTERS_SIZE_BYTES)?;
 
         let regs     = nic_regs_mapped_page.into_borrowed_mut(0).map_err(|(_mp, err)| err)?;
         let rx_regs  = nic_rx_regs_mapped_page.into_borrowed_mut(0).map_err(|(_mp, err)| err)?;
--- a/kernel/e1000/src/lib.rs
+++ b/kernel/e1000/src/lib.rs
@@ -34,7 +34,7 @@ use regs::*;
 use spin::Once; 
 use alloc::{collections::VecDeque, format, sync::Arc, vec::Vec};
 use irq_safety::MutexIrqSafe;
-use memory::{PhysicalAddress, BorrowedMappedPages, BorrowedSliceMappedPages, Mutable, map_mmio_range};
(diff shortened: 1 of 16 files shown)


kernel/memory/src/lib.rs
```diff
@@ -99,6 +107,37 @@ pub fn create_contiguous_mapping<F: Into<PteFlagsArch>>(
...
+/// Currently, this function acquires the lock on the frame allocator and the kernel's `MemoryManagementInfo` instance.
+/// Thus, the caller should ensure that the locks on those two variables are not held when invoking this function.
+pub fn map_frame_range<F: Into<PteFlagsArch>>(
+    range: FrameRange,
+    flags: F,
+) -> Result<MappedPages, &'static str> {
+    let kernel_mmi_ref = get_kernel_mmi_ref().ok_or("map_range(): KERNEL_MMI was not yet initialized!")?;
+    let num_frames = range.size_in_frames();
+    let allocated_pages = allocate_pages(num_frames).ok_or("memory::map_range(): couldn't allocate contiguous pages!")?;
+    let allocated_frames = allocate_frames_by_bytes_at(range.start_address(), num_frames)
```
kevinaboos: this is incorrect, you're passing in the number of frames to a function that expects the number of bytes.
```suggestion
    let allocated_frames = allocate_frames_at(range.start_address(), num_frames)
```

kevinaboos (commented):
I don't mind convenience functions, but sometimes having too many reduces clarity or decreases readability, especially when it comes to new users reading documentation to determine which function is most appropriate for their use case. I'm getting a bit concerned that we'll have far too many convenience functions for mapping memory; in my opinion, we really only need two: one for mapping pages to specific frames, and one for mapping pages to any frames.
Debatably, there are more combinations of use cases that we could also support (either with const generics or `Option` parameters), such as your `map_frame_range` function that maps any pages to specific frames. I don't know if those really merit a dedicated function.
Either way, I don't think we need one for MMIO specifically, as it's a subcase of one of the aforementioned cases.

I'm sort-of okay with having an easy definition for MMIO flags, but @Ramla-I and I were talking a couple months ago and realized that mapping MMIO registers as non-cacheable (i.e., what `DEVICE_MEMORY` currently is) is actually *not* necessary and can surprisingly harm performance in a major way.
Thus, I'm reluctant to apply this changeset system-wide, especially to a bunch of advanced NIC drivers that are hard to test (I cannot personally do so due to lack of hardware access). I'd also like to hear @Ramla-I's thoughts on MMIO mapping flags, since they're often different from general device memory mappings.

kernel/memory/src/lib.rs
NathanRoyer: Good catch, thank you

NathanRoyer:
> I don't mind convenience functions, but sometimes having too many reduces clarity or decreases readability, especially when it comes to new users reading documentation to determine which function is most appropriate for their use case. I'm getting a bit concerned that we'll have far too many convenience functions for mapping memory.

Well, I didn't add any; this is just a refactor for functions I think we're going to need in a lot of drivers and that were already present (but not in the right place imo).

Ramla-I:
@kevinaboos the slowdown in the driver was because we had used the same flags for MMIO as for DMA memory, which is incorrect. The MMIO registers are not cacheable, but all other memory used for DMA is e.g. descriptor rings and packet buffers.

It may be useful to define two sets of flags: MMIO_FLAGS and DMA_FLAGS.
The current driver implementations in theseus_main have not been fixed; they still use MMIO flags for descriptor rings and packet buffers. I can make that change as part of a separate commit or it can be part of this one too.

kevinaboos:
> > I don't mind convenience functions, but sometimes having too many reduces clarity or decreases readability, especially when it comes to new users reading documentation to determine which function is most appropriate for their use case. I'm getting a bit concerned that we'll have far too many convenience functions for mapping memory.
> 
> Well, I didn't add any; this is just a refactor for functions I think we're going to need in a lot of drivers and that were already present (but not in the right place imo).

hmm ok, yea that's a fair way of looking at it. Apologies if I came off rude.

I think we ought to consider the matrix of all possible invocations of various memory mapping functions and come up with a good, easy interface that represents all cases.

kevinaboos:
> @kevinaboos the slowdown in the driver was because we had used the same flags for MMIO as for DMA memory, which is incorrect. The MMIO registers are not cacheable, but all other memory used for DMA is e.g. descriptor rings and packet buffers.
> 
> It may be useful to define two sets of flags: MMIO_FLAGS and DMA_FLAGS. The current driver implementations in theseus_main have not been fixed; they still use MMIO flags for descriptor rings and packet buffers. I can make that change as part of a separate commit or it can be part of this one too.

ah ok, glad i asked since I had remembered it backwards. So, to summarize:
* MMIO_FLAGS are not cacheable
* DMA_FLAGS are cacheable

NathanRoyer:
> > Well, I didn't add any; this is just a refactor for functions I think we're going to need in a lot of drivers and that were already present (but not in the right place imo).
> 
> hmm ok, yea that's a fair way of looking at it. Apologies if I came off rude.
> 
> I think we ought to consider the matrix of all possible invocations of various memory mapping functions and come up with a good, easy interface that represents all cases.

You were right, I actually added `map_mmio_range` which is only one line long. I'll remove it.

kernel/e1000/src/lib.rs
```diff
@@ -303,14 +303,21 @@ impl E1000Nic {
...
+
+        let nic_rx_regs_mapped_page = map_frame_range(offset, RX_REGISTERS_SIZE_BYTES, MMIO_FLAGS)?;
+        let rx_regs = nic_rx_regs_mapped_page.into_borrowed_mut(0).map_err(|(_mp, err)| err)?;
+        offset += RX_REGISTERS_SIZE_BYTES;
+
+        let nic_tx_regs_mapped_page = map_frame_range(offset, TX_REGISTERS_SIZE_BYTES, MMIO_FLAGS)?;
+        let tx_regs = nic_tx_regs_mapped_page.into_borrowed_mut(0).map_err(|(_mp, err)| err)?;
+        offset += TX_REGISTERS_SIZE_BYTES;
+
+        let nic_mac_regs_mapped_page = map_frame_range(offset, MAC_REGISTERS_SIZE_BYTES, MMIO_FLAGS)?;
```
kevinaboos: this is really nice! A good mix of convenience functions with a simple representation of NIC-specific logic.

:+1:

kernel/e1000/src/lib.rs
```diff
@@ -303,14 +303,21 @@ impl E1000Nic {
         const TX_REGISTERS_SIZE_BYTES: usize = 4096;
         const MAC_REGISTERS_SIZE_BYTES: usize = 114_688;
 
-        let nic_regs_mapped_page     = allocate_memory(mem_base, GENERAL_REGISTERS_SIZE_BYTES)?;
-        let nic_rx_regs_mapped_page  = allocate_memory(mem_base + GENERAL_REGISTERS_SIZE_BYTES, RX_REGISTERS_SIZE_BYTES)?;
-        let nic_tx_regs_mapped_page  = allocate_memory(mem_base + GENERAL_REGISTERS_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES, TX_REGISTERS_SIZE_BYTES)?;
-        let nic_mac_regs_mapped_page = allocate_memory(mem_base + GENERAL_REGISTERS_SIZE_BYTES + RX_REGISTERS_SIZE_BYTES + TX_REGISTERS_SIZE_BYTES, MAC_REGISTERS_SIZE_BYTES)?;
+        let mut offset = mem_base;
```
kevinaboos: nit: `offset` implies there is a base value that the offset gets added to. Here, what you call `offset` is actually a physical address, so let's call it that. That also makes it match better with the signature of `map_frame_range()`.

kevinaboos: Also i realize you didn't write this original code, but the same can be applied to the code in `ixgbe`.

kernel/memory/src/lib.rs
```diff
@@ -99,6 +114,25 @@ pub fn create_contiguous_mapping<F: Into<PteFlagsArch>>(
 }
 
 
+/// A convenience function that creates a new memory mapping by allocating specific frames from physical memory.
+/// Returns the new `MappedPages`.
```
kevinaboos: can simplify a bit. Either way, we ought to mention that the pages it allocates are random and not specifiable when using this function.
```suggestion
/// A convenience function that maps randomly-allocated pages to the given range of frames.
```

kevinaboos (changes requested):
looks pretty good. I left a few small comments.

I think in a future PR, I can make some more improvements to actually implement these convenience functions on the `Mapper` itself, such that you could use them when the lock on the kernel's `PageTable` is already acquired (i.e., when you already have a `&mut PageTable` or `&mut Mapper`).
Then, the standalone functions in `memory` can just acquire the kernel's page table and invoke the corresponding convenience method in `Mapper`.

kevinaboos:
Also there's a conflict with the aarhc64 serial port basic file.

kevinaboos:
> It may be useful to define two sets of flags: MMIO_FLAGS and DMA_FLAGS.

Good idea, Nathan has done it as part of this PR.

> The current driver implementations in theseus_main have not been fixed; they still use MMIO flags for descriptor rings and packet buffers. I can make that change as part of a separate commit or it can be part of this one too.

@Ramla-I Yeah, I don't think the driver's in the main branch do this. We defined the flags now, but I'll let you make those changes (to actually use the correct flags) as part of a separate PR.

kevinaboos (approved):

PR #958, opened 2023-05-10
`pci` improvements

NathanRoyer:
This reduces the number of exposed `pci` internals:
- All internal consts are now private.
- `pci_read_*` internal methods are now private as well.
- MSI-X mapping and configuration is now done via `PciDevice::pci_mem_map_msix`. Related structs (`MsixVectorEntry` & `MsixVectorTable`) were moved to `pci`, and there is now a `configure` method in `MsixVectorEntry`.
- The `PciDevice::pci_get_interrupt_line` getter was added.
- `nic_queues::{RxQueue, TxQueue}` now use `OptionalCpuId` instead of `Option<u8>`.

This depends on :
- commits from #957
- items added in #948 (`interrupts::InterruptNumber`)

Consequently, CI will fail at least until `interrupts::InterruptNumber` exists.
Quick fix for testing: Add
```rust
pub use core::primitive::u8 as InterruptNumber;
```
to `interrupts::x86_64`.

### DIFF
--- a/kernel/ixgbe/src/lib.rs
+++ b/kernel/ixgbe/src/lib.rs
@@ -1178,7 +1178,7 @@ impl IxgbeNic {
                 Some(cpu_id) => cpu_id,
                 None => cpu::bootstrap_cpu().ok_or("Couldn't read BSP CpuId")?,
             };
-            vector_table[i].configure(cpu_id, msi_int_num);
+            vector_table[i].init(cpu_id, msi_int_num);
         }
 
         Ok(interrupt_nums)
--- a/kernel/ixgbe/src/lib.rs
+++ b/kernel/ixgbe/src/lib.rs
@@ -986,12 +986,11 @@ impl IxgbeNic {
     /// Sets up DCA for the rx queues that have been enabled.
     /// You can optionally choose to have the descriptor, header and payload copied to the cache for each received packet
     fn enable_rx_dca(
-        rx_queues: &mut Vec<RxQueue<IxgbeRxQueueRegisters,AdvancedRxDescriptor>>
+        rx_queues: &mut Vec<RxQueue<IxgbeRxQueueRegisters, AdvancedRxDescriptor>>
     ) -> Result<(), &'static str> {
 
         for rxq in rx_queues {            
             // the cpu id will tell which cache the data will need to be written to
-            // TODO: choose a better default value
             let cpu_id = match rxq.cpu_id {
                 Some(cpu_id) => cpu_id,
                 None => cpu::bootstrap_cpu().ok_or("Couldn't read BSP CpuId")?,
@@ -1173,7 +1172,6 @@ impl IxgbeNic {
 
             // find core to redirect interrupt to
             // we assume that the number of msi vectors are equal to the number of rx queues
-            // TODO: choose a better default value
             let cpu_id = match rxq[i].cpu_id {
                 Some(cpu_id) => cpu_id,
                 None => cpu::bootstrap_cpu().ok_or("Couldn't read BSP CpuId")?,
--- a/kernel/ixgbe/src/lib.rs
+++ b/kernel/ixgbe/src/lib.rs
@@ -53,7 +53,7 @@ use alloc::{
     vec::Vec,
 };
 use irq_safety::MutexIrqSafe;
-use memory::{PhysicalAddress, MappedPages, Mutable, BorrowedSliceMappedPages, BorrowedMappedPages, map_mmio_range};
+use memory::{PhysicalAddress, MappedPages, Mutable, BorrowedSliceMappedPages, BorrowedMappedPages, map_frame_range, MMIO_FLAGS};
 use pci::{PciDevice, MSIX_CAPABILITY, PciConfigSpaceAccessMechanism, PciLocation};
 use bit_field::BitField;
 use interrupts::{register_msi_interrupt, InterruptHandler};
@@ -446,25 +446,25 @@ impl IxgbeNic {
 
         // Allocate memory for the registers, making sure each successive memory region begins where the previous region ended.
         let mut offset = mem_base;
-        let nic_regs1_mapped_page = map_mmio_range(offset, GENERAL_REGISTERS_1_SIZE_BYTES)?;
+        let nic_regs1_mapped_page = map_frame_range(offset, GENERAL_REGISTERS_1_SIZE_BYTES, MMIO_FLAGS)?;
 
         offset += GENERAL_REGISTERS_1_SIZE_BYTES;
-        let nic_rx_regs1_mapped_page = map_mmio_range(offset, RX_REGISTERS_SIZE_BYTES)?;
+        let nic_rx_regs1_mapped_page = map_frame_range(offset, RX_REGISTERS_SIZE_BYTES, MMIO_FLAGS)?;
 
         offset += RX_REGISTERS_SIZE_BYTES;
-        let nic_regs2_mapped_page = map_mmio_range(offset, GENERAL_REGISTERS_2_SIZE_BYTES)?;  
+        let nic_regs2_mapped_page = map_frame_range(offset, GENERAL_REGISTERS_2_SIZE_BYTES, MMIO_FLAGS)?;
 
         offset += GENERAL_REGISTERS_2_SIZE_BYTES;
-        let nic_tx_regs_mapped_page = map_mmio_range(offset, TX_REGISTERS_SIZE_BYTES)?;
+        let nic_tx_regs_mapped_page = map_frame_range(offset, TX_REGISTERS_SIZE_BYTES, MMIO_FLAGS)?;
 
         offset += TX_REGISTERS_SIZE_BYTES;
-        let nic_mac_regs_mapped_page = map_mmio_range(offset, MAC_REGISTERS_SIZE_BYTES)?;
+        let nic_mac_regs_mapped_page = map_frame_range(offset, MAC_REGISTERS_SIZE_BYTES, MMIO_FLAGS)?;
 
         offset += MAC_REGISTERS_SIZE_BYTES;
-        let nic_rx_regs2_mapped_page = map_mmio_range(offset, RX_REGISTERS_SIZE_BYTES)?;   
+        let nic_rx_regs2_mapped_page = map_frame_range(offset, RX_REGISTERS_SIZE_BYTES, MMIO_FLAGS)?;
 
         offset += RX_REGISTERS_SIZE_BYTES;
-        let nic_regs3_mapped_page = map_mmio_range(offset, GENERAL_REGISTERS_3_SIZE_BYTES)?;
+        let nic_regs3_mapped_page = map_frame_range(offset, GENERAL_REGISTERS_3_SIZE_BYTES, MMIO_FLAGS)?;
 
         // Map the memory as the register struct and tie the lifetime of the struct with its backing mapped pages
         let regs1    = nic_regs1_mapped_page.into_borrowed_mut(0).map_err(|(_mp, err)| err)?;
@@ -558,7 +558,7 @@ impl IxgbeNic {
 
         // debug!("msi-x vector table bar: {}, base_address: {:#X} and size: {} bytes", bar, mem_base, mem_size_in_bytes);
 
-        let msix_mapped_pages = map_mmio_range(mem_base, mem_size_in_bytes)?;
+        let msix_mapped_pages = map_frame_range(mem_base, mem_size_in_bytes, MMIO_FLAGS)?;
         let vector_table = BorrowedMappedPages::from_mut(msix_mapped_pages, 0)
             .map_err(|(_mp, err)| err)?;
 
--- a/kernel/ixgbe/src/lib.rs
+++ b/kernel/ixgbe/src/lib.rs
@@ -445,26 +445,26 @@ impl IxgbeNic {
         const GENERAL_REGISTERS_3_SIZE_BYTES:   usize = 18 * 4096;
 
         // Allocate memory for the registers, making sure each successive memory region begins where the previous region ended.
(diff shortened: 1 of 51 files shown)


kevinaboos:
@NathanRoyer #948 has been merged in.

kernel/pci/src/lib.rs
```diff
@@ -1,56 +1,55 @@
...
+const PCI_INTERRUPT_LINE:        u8 = 0x3C;
+const PCI_INTERRUPT_PIN:         u8 = 0x3D;
+const PCI_MIN_GRANT:             u8 = 0x3E;
+const PCI_MAX_LATENCY:           u8 = 0x3F;
 
 // PCI Capability IDs
-pub const MSI_CAPABILITY:           u8 = 0x05;
-pub const MSIX_CAPABILITY:          u8 = 0x11;
+const MSI_CAPABILITY:           u8 = 0x05;
+const MSIX_CAPABILITY:          u8 = 0x11;
```
kevinaboos: i think these still need to be public since we don't have accessor methods for all parts of the config space.

Similarly, the `pci_read_*` functions should also be public. Though it's probably best to define an enum with all of the offsets for the config space, and then require that enum in those `pci_read_*` (and write) functions instead of accepting a raw `offset: u8`.

kevinaboos: Probably a good idea to keep the `write` functions private though.

kernel/pci/src/lib.rs
```diff
@@ -500,14 +499,45 @@ impl PciDevice {
...
-        map_mmio_range(mem_base, self.determine_mem_size(bar_index) as usize)
+    /// * `bar_index`: index of the Base Address Register to use
+    pub fn pci_map_bar_mem(&self, bar_index: usize) -> Result<MappedPages, &'static str> {
+        let mem_base = self.determine_mem_base(bar_index)?;
+        let mem_size = self.determine_mem_size(bar_index);
+        map_mmio_range(mem_base, mem_size as usize)
+    }
+
+    /// Reads the `PCI_INTERRUPT_LINE` register
+    pub fn pci_get_interrupt_line(&self) -> InterruptNumber {
```
kevinaboos: does this always return an interrupt number? From the PCI docs on osdev wiki, i see this:
> For the x86 architecture, this register corresponds to the PIC IRQ numbers 0-15 (and not I/O APIC IRQ numbers). A value of 0xFF defines no connection.

So we'd want to possibly handle the 0xFF case specially, perhaps with an option.

Also, this doesn't include the interrupt pin info, which is probably a good thing to return along with the interrupt line. So this function could be something more like `pci_get_interrupt_info()` that returns an interrupt line and interrupt pin.

For more, see osdev wiki again:

> Specifies which interrupt pin the device uses. A value of 0x1 is INTA#, 0x2 is INTB#, 0x3 is INTC#, 0x4 is INTD#, and 0x0 means the device does not use an interrupt pin.

An enum for the interrupt pin would be the proper way to represent this.

kernel/pci/src/lib.rs
```diff
@@ -288,7 +287,7 @@ impl PciLocation {
     /// with each capability storing the pointer to the next capability right after its ID.
     /// The function returns a None value if capabilities are not valid for this device 
     /// or if the requested capability is not present. 
-    pub fn find_pci_capability(&self, pci_capability: u8) -> Option<u8> {
+    fn find_pci_capability(&self, pci_capability: u8) -> Option<u8> {
```
kevinaboos: this can also be public. Generally, harmless functions that just provide info in a read-only manner are okay to expose publicly.

kernel/nic_queues/src/lib.rs
```diff
@@ -21,6 +22,7 @@ use alloc::{
 use memory::{create_contiguous_mapping, BorrowedSliceMappedPages, Mutable, MMIO_FLAGS};
 use intel_ethernet::descriptors::{RxDescriptor, TxDescriptor};
 use nic_buffers::{ReceiveBuffer, ReceivedFrame, TransmitBuffer};
+use cpu::OptionalCpuId;
```
kevinaboos: The `OptionalCpuId` type is just a hack for use in task mgmt. You should just use a regular `Option<CpuId>` here

kernel/pci/src/lib.rs
```diff
@@ -530,4 +560,77 @@ impl DerefMut for PciDevice {
...
+
+impl MsixVectorEntry {
+    /// Sets interrupt destination & number for this entry
+    pub fn configure(&mut self, cpu_id: CpuId, int_num: InterruptNumber) {
+        // unmask the interrupt
+        self.vector_control.write(MSIX_UNMASK_INT);
+        let lower_addr = self.msg_lower_addr.read();
+
+        // set the core to which this interrupt will be sent
+        let dest_id = (cpu_id.into_u8() as u32) << MSIX_DEST_ID_SHIFT;
```
kevinaboos, line 640: `CpuId::into_u8()` is a temp hack function, you can just convert it directly into a u32 instead.

kernel/ixgbe/src/lib.rs
```diff
@@ -1015,7 +992,10 @@ impl IxgbeNic {
         for rxq in rx_queues {            
             // the cpu id will tell which cache the data will need to be written to
             // TODO: choose a better default value
-            let cpu_id = rxq.cpu_id.unwrap_or(0) as u32;
+            let cpu_id = match rxq.cpu_id.into() {
+                Some(cpu_id) => cpu_id,
+                None => cpu::bootstrap_cpu().expect("Couldn't read BSP CpuId"),
```
kevinaboos: return a result instead of panicking

kernel/ixgbe/src/lib.rs
```diff
@@ -1015,7 +992,10 @@ impl IxgbeNic {
...
+            }.into_u8() as u32;
```
kevinaboos, line 997: same comment, you can directly convert a CpuId into a u32, or just use the `.value()` function.

Or... wait. Does it need to be a CPU ID that fits within the bounds of a `u8`?

kernel/ixgbe/src/lib.rs
```diff
@@ -1194,15 +1174,11 @@ impl IxgbeNic {
...
-            vector_table.msi_vector[i].vector_control.write(MSIX_UNMASK_INT);
-            let lower_addr = vector_table.msi_vector[i].msg_lower_addr.read();
-            // set the core to which this interrupt will be sent
-            vector_table.msi_vector[i].msg_lower_addr.write((lower_addr & !MSIX_ADDRESS_BITS) | MSIX_INTERRUPT_REGION | (core_id << MSIX_DEST_ID_SHIFT)); 
-            //allocate an interrupt to msix vector            
-            vector_table.msi_vector[i].msg_data.write(msi_int_num as u32);
-            // debug!("Created MSI vector: control: {}, core: {}, int: {}", vector_table.msi_vector[i].vector_control.read(), core_id, interrupt_nums[i]);
+            let cpu_id = match rxq[i].cpu_id.into() {
+                Some(cpu_id) => cpu_id,
+                None => cpu::bootstrap_cpu().expect("Couldn't read BSP CpuId"),
```
kevinaboos: return result

kernel/pci/src/lib.rs
```diff
@@ -530,4 +560,77 @@ impl DerefMut for PciDevice {
...
+    msg_upper_addr:         Volatile<u32>,
+    /// The data portion of the msi vector which contains the interrupt number.
+    msg_data:               Volatile<u32>,
+    /// The control portion which contains the interrupt mask bit.
+    vector_control:         Volatile<u32>,
+}
+
+impl MsixVectorEntry {
+    /// Sets interrupt destination & number for this entry
+    pub fn configure(&mut self, cpu_id: CpuId, int_num: InterruptNumber) {
```
kevinaboos: more descriptive name? "configure" is a bit vague. 

Based on the comments within this function, something like "enable interrupt for cpu" or the like

kevinaboos (changes requested):
Looks good, though I only reviewed the changes from the latest 6 commits, and ignored the changes borrowed from #957.

Considering that we're unsure about #957, can you remove those changes from this PR and have it be standalone?

kernel/pci/src/lib.rs
NathanRoyer: I was thinking it would be preferable to have safe getters in this crate instead of allowing raw pci_reads; I grepped the source to find uses of them and replaced that with new getters (which is why I moved MSIX code to this crate). Should pci_read really be publicly visible?

kernel/pci/src/lib.rs
NathanRoyer: Sure, I'll change it.

kernel/pci/src/lib.rs
NathanRoyer: The day CpuId isn't 8-bits anymore, this could cause hard to diagnose bugs as this code expects the cpu id to only occupy 8 bits in this u32 (other bits are used for other purposes). This `into_u8` will loudly fail once this happens, which is better imo.

kernel/ixgbe/src/lib.rs
NathanRoyer: Yes

kernel/pci/src/lib.rs
NathanRoyer: I'd have named it so but there's always a value in there, so it could already be enabled. Should I still name it `enable_...`?

kevinaboos: "configure" doesn't really say anything about what it does. If it enables something, then yea, i'd vote enable. That implies there should be a corresponding way to disable it.
If it initializes something, you could use init_... too.

kernel/pci/src/lib.rs
kevinaboos: eh, it's a matter of opinion, but I think it's harmless?  Always nice to have for the sake of debugging, but I guess we don't technically need it.

kernel/pci/src/lib.rs
kevinaboos: oh god, why on earth is this a thing.... sigh. x86 blows, lol. I had hoped to remove `CpuId::into_u8()`, but now we can't? hmm.

So I'm struggling to actually find documentation on the bit-wise format of this field. Specifically I don't see where the CPU ID gets involved; perhaps @Ramla-I can chime in too since she originally implemented this?

kernel/ixgbe/src/lib.rs
kevinaboos: ah ok. we'll discuss on the other comment and wait for Ramla's response.

kernel/pci/src/lib.rs
NathanRoyer: So, I changed it, and I changed some logic in accordance to this becoming an option; `E1000Nic.interrupt_num` is an Option too, now:
```rust
    pub fn init_interrupts(
        &mut self,
        interface: Arc<net::NetworkInterface>,
        if let Some(interrupt_num) = self.interrupt_num {
            self.enable_interrupts();
            let deferred_task = deferred_interrupt_tasks::register_interrupt_handler(
                interrupt_num,
                e1000_handler,
                poll_interface,
                interface,
                Some(format!("e1000_deferred_task_irq_{:?}", self.interrupt_num)),
            )
            .map_err(|error| {
                error!("error registering e1000 handler: {:?}", error);
                "e1000 interrupt number was already in use! Sharing IRQs is currently unsupported."
            })?;
            self.deferred_task = Some(deferred_task);
        }

        Ok(())
    }
```

kernel/pci/src/lib.rs
NathanRoyer: Ok, but I'll convert `pci_capability` to an enum then.

kernel/pci/src/lib.rs
NathanRoyer: Simply `init` maybe? since this function simply encodes the cpu ID, the interrupt number and makes sure the interrupt is unmasked.

With that signature:
```rust
/// Sets interrupt destination & number for this entry and makes sure the
/// interrupt is unmasked (PCI Controller side)
MsixVectorEntry::init(CpuId, InterruptNumber);
```

I feel like it's sufficiently clear?

kevinaboos: sure. I'd need to know more about what the API actually should be for MSI vector entries before I can make an informed decision, but in any case let's just stick with `init` for now i guess.

kernel/pci/src/lib.rs
Ramla-I: The Intel Architecture manual ([Vol 3a, 10.11](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3a-part-1-manual.pdf)) has the information on the layout of the MSI fields.

The destination ID field refers to to the target processor.

kernel/ixgbe/src/lib.rs
Ramla-I: yes, that field is only 8 bits.

kernel/pci/src/lib.rs
kevinaboos: Thanks, I see now. It's not necessarily an 8-bit CPU ID, it's just an IOAPIC-defined destination ID. Since we currently only use physical destination mode to specify an interrupt's destination, it's the same thing as a CPU ID (an APIC ID on x86). If we ever switched to logical destination mode (which is very unlikely), it wouldn't be a straight CPU ID.

kernel/pci/src/lib.rs
kevinaboos: update: we decided to make this an enum in a future PR.

kernel/pci/src/lib.rs
kevinaboos: I changed this back to a non-`Option` design in order to move the failure condition earlier.

kevinaboos (approved):

PR #959, opened 2023-05-11
Add condvar implementation

tsoutsman:
Depends on https://github.com/theseus-os/Theseus/pull/942.

### DIFF
--- a/kernel/sync_block/src/condvar.rs
+++ b/kernel/sync_block/src/condvar.rs
@@ -1,5 +1,5 @@
 use crate::MutexGuard;
-use cpu_local_preemption::hold_preemption;
+use cpu_local_preemption::hold_preemption_no_timer_disable;
 use mpmc_queue::Queue;
 use sync::DeadlockPrevention;
 use sync_spin::Spin;
@@ -43,7 +43,7 @@ where
             .inner
             .push_if_fail(task.clone(), || {
                 drop(guard);
-                let preemption_guard = hold_preemption();
+                let preemption_guard = hold_preemption_no_timer_disable();
                 task.block().unwrap();
                 Result::<(), _>::Err(preemption_guard)
             })
@@ -57,7 +57,7 @@ where
                 if let Some(mutex_guard) = mutex.try_lock() {
                     Ok(mutex_guard)
                 } else {
-                    let preemption_guard = hold_preemption();
+                    let preemption_guard = hold_preemption_no_timer_disable();
                     task.block().unwrap();
                     Err(preemption_guard)
                 }
--- a/kernel/sync_block/src/condvar.rs
+++ b/kernel/sync_block/src/condvar.rs
@@ -1,0 +1,107 @@
+use crate::MutexGuard;
+use cpu_local_preemption::hold_preemption;
+use mpmc_queue::Queue;
+use sync::DeadlockPrevention;
+use sync_spin::Spin;
+use task::{get_my_current_task, TaskRef};
+
+/// A condition variable.
+// TODO: Is there even a point to exposing this generic?
+pub struct Condvar<P = Spin>
+where
+    P: DeadlockPrevention,
+{
+    inner: Queue<TaskRef, P>,
+}
+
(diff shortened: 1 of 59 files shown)


tsoutsman:
Now this PR only depends on #942.

kevinaboos:
@tsoutsman #942 has been merged in. thanks!

kernel/sync_block/src/condvar.rs
```diff
@@ -0,0 +1,107 @@
+use crate::MutexGuard;
+use cpu_local_preemption::hold_preemption;
+use mpmc_queue::Queue;
+use sync::DeadlockPrevention;
+use sync_spin::Spin;
+use task::{get_my_current_task, TaskRef};
+
+/// A condition variable.
+// TODO: Is there even a point to exposing this generic?
```
tsoutsman, line 12: Bump.

tsoutsman:
Should be good to go aside from the one comment I've left.

kernel/sync_block/src/condvar.rs
kevinaboos: see discord #sync-primitives discussion, but yea, we can remove it.

kernel/sync_block/src/condvar.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// protect against spurious wakeups.
+    pub fn wait<'a, T>(&self, guard: MutexGuard<'a, T>) -> MutexGuard<'a, T> {
+        let task = get_my_current_task().unwrap();
+        let mutex = guard.mutex();
+
+        let preemption_guard = self
+            .inner
+            .push_if_fail(task.clone(), || {
+                drop(guard);
+                let preemption_guard = hold_preemption();
```
kevinaboos: this and other calls to `hold_preemption()` are likely good candidates for using `hold_preemption_no_timer_disable()` instead. lmk if you disagree.

kevinaboos (commented):
Generally it looks pretty good. Will approve pending some discussion, one point of which is...

Could we potentially simplify the implementation using a waitqueue impl based on this new set of sync primitives? I only ask because it seems like there's a fair amount of overlap between the inner impl of Condvar (in the way that it uses the mpmc_queue) and what a Waitqueue offers.

kernel/sync_block/src/condvar.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /// Blocks the current thread until this condition variable receives a
+    /// notification.
+    ///
+    /// Note that this function is susceptible to spurious wakeups. Condition
+    /// variables normally have a boolean predicate associated with them, and
+    /// the predicate must always be checked each time this function returns to
+    /// protect against spurious wakeups.
+    pub fn wait<'a, T>(&self, guard: MutexGuard<'a, T>) -> MutexGuard<'a, T> {
```
kevinaboos: I think we can relax the implicit Sized bound on T here, as Mutexes allow for that too.
```suggestion
    pub fn wait<'a, T: ?Sized>(&self, guard: MutexGuard<'a, T>) -> MutexGuard<'a, T> {
```

tsoutsman:
> Could we potentially simplify the implementation using a waitqueue impl

No I don't think so, because we perform a different action on the first `push_if_fail` and a wait queue doesn't allow for that.

kernel/sync_block/src/condvar.rs
tsoutsman: Depends on #962.

kevinaboos: merged in #962, thanks.

tsoutsman:
@kevinaboos should be good to merge now

kevinaboos (approved):

kevinaboos:
> @kevinaboos should be good to merge now

Thanks, approved. If you want to remove that "TODO" item about the `P = Spin` parameter on `Condvar`, feel free to follow-up with another PR. I'm ok with it being there or not, either is fine.

PR #961, opened 2023-05-23
Fix `try_*` methods for deadlock prevention lock flavours

tsoutsman:
Previously they used relaxed loads which meant that they could fail even
if the lock was unlocked.

Depends on #942 and theseus-os/spin-rs#4.


kevinaboos (approved):
great, thanks!

PR #966, opened 2023-05-31
Switch to `sync_preemption` instead of `mutex_preemption`

tsoutsman:
Not sure if I like the reexports of `sync_preemption as preempt` everywhere. Maybe we should rename the `sync_preemption` package in `Cargo.toml` to be `preempt`?


tsoutsman:
Should be good, although should I remove the `Mutex`/`RwLock`/etc. type definitions altogether (only leaving `IrqSafeMutex`/`IrqSafeRwLock`/etc.?).

Also I didn't rename the blocking primitives because they are the "default".

kevinaboos:
I think this changeset is fine, the only decision yet to be made is whether we want to establish a nice shorter naming convention for the all the sync types. I think we should, but I'm not 100% settled on what the best naming convention is. Here are some ideas based on what we currently have or have had in the past:
* MutexIrqSafe
* MutexInterruptSafe
* MutexInterrupt
* MutexIntSafe
* MutexInt
* MutexIrq
* MutexPreempt
* MutexPreemptSafe
* MutexPreemptionSafe
* MutexNoPreempt
* MutexNoIrq
* MutexHoldInts, HoldPreempt[ion]

I dislike overly long type names but I dislike vagueness even more. Lmk if you have any thoughts or more clever naming ideas.

tsoutsman:
I prefer the term `Safe` because it's a "positive" term. Using a `ThingySafeMutex` is strictly more correct than using a `Mutex`. I think `No` doesn't correctly convey the meaning of the mutex; `NoIrq` sounds like it shouldn't be used in interrupts.

Also, I prefer the "modifier" be before "Mutex" e.g. `IrqSafeMutex` rather than `MutexIrqSafe`. This is consistent with `std`'s word ordering and is closer to normal English (`IrqSafe` is an adjective).

kevinaboos (approved):
Works for me, I don't particularly care about the naming convention as long as it's consistent, which this is. I'll go ahead and merge it in, thanks!

Issue #963, opened 2023-05-28
Switch all bitfields to bilge, take notes


hecatia-elegua:
This is good for converging on one thing instead of the 3-4 we currently use and for experimenting and finding usecases for `bilge`.
Really complicated ones could get their own issue for later.
There're also some places with manual bit-shifting to get at fields, which we can convert to real bitfields.

Let's crudely measure build time before and after as well?
I'll do `time make build` like trice, on linux, with nothing besides console open, with `make clean` inbetween.
Then again trice, to also look at hot builds.

I'll be working on this with @pickx, so here's a list with all crates.
Remove the crates which have nothing to change in them and tick all items which are done.
in kernel:
- [ ] acpi (seemed like a lot, might be good to do this in steps if there are too many changes)
- [ ] apic (good one)
- [ ] ata (good one)
- [ ] boot_info (bitflags)
- [ ] e1000 (medium amount)
- [ ] gdt (small amount)
~~- [ ] gic (wip/later)~~
- [ ] intel_ethernet (medium amount)
- [ ] interrupts (uhhh, this is rtc again? for testing?)
- [ ] ioapic (a bit special)
- [ ] iommu (medium amount)
- [ ] ixgbe (looks like a lot)
- [ ] keyboard (wip)
- [ ] libtest (bit_field, but actually not? run `cargo udeps` on all crates or sth)
- [ ] memory (bitflags, but actually not? probably removed dependency)
~~- [ ] memory_aarch64 (`cortex_a` crate, we won't do this)~~
- [ ] memory_structs (maybe this could be a usecase for `#[level]` or sth else)
- [ ] mlx_ethernet (with zerocopy byteorder usecase - we'd have to see how to test this and then look for how to best handle alignment 1 and so on, hopefully outside of the bitfields)
~~- [ ] mod_mgmt (`parse_nano_core_binary` acts on some flags from the crate `xmas_elf`, can't change things here, but the usecase is interesting)~~
- [ ] mouse (wip)
- [ ] multicore_bringup (icr and GraphicInfo, even if unused, having real names there and printing the info helps with bugs, such as #952)
- [ ] page_attribute_table (another nice example)
- [ ] page_table_entry (works a lot with its inner value, note usecases here)
- [ ] pci (wip, classes are interesting)
- [ ] pic (enums/intralingual stuff, not sure if bitfields)
- [ ] pit_clock (small changes)
- [ ] pit_clock_basic (small changes)
[...]

PR #967, opened 2023-06-02
Update signal_handler, removing bitflags

hecatia-elegua:
First I wanted to change that piece of code to use bilge, but I'm not even sure why this wasn't an enum in the first case.
It's just some commented out code, btw.
Also using `variant_count`.
#963

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3393,7 +3393,6 @@ dependencies = [
 name = "signal_handler"
 version = "0.1.0"
 dependencies = [
- "bitflags",
  "log",
  "memory",
  "spin 0.9.4",
--- a/kernel/signal_handler/Cargo.toml
+++ b/kernel/signal_handler/Cargo.toml
@@ -3,19 +3,13 @@ authors = ["Kevin Boos <kevinaboos@gmail.com>"]
 name = "signal_handler"
 description = "Support for registering POSIX-style signal handlers or trap handlers"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
 
 [dependencies]
 x86_64 = "0.14.8"
 log = "0.4.8"
-bitflags = "1.3"
 spin = "0.9.4"
 
-[dependencies.memory]
-path = "../memory"
-
-[dependencies.task]
-path = "../task"
-
-[dependencies.thread_local_macro]
-path = "../thread_local_macro"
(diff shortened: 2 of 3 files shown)


kevinaboos:
`variant_count` is cool, thanks for informing me about that feature! Didn't know that existed; we should definitely be using that.

Regarding the set of exception vectors, it's purposefully not an enum because we want to be able to specify a union of multiple variants at the same time, mostly for the purpose of registering a single handler that covers multiple distinct exception variants. While I know that code isn't used yet, that's why we used bitflags there. In fact, the only time we prefer bitflags is in this kind of scenario where we want to allow multiple variants to be selectable using only the space of a single primitive type, e.g., `u32`.

In general, I'm excited to experiment with `bilge` and highly in favor of replacing bitflags (and other related crates). This PR doesn't seem to include bilge; was that planned for an upcoming changeset?

hecatia-elegua:
Well I can change it to use a struct with bools instead, just didn't seem like the right thing to do

kevinaboos:
> Well I can change it to use a struct with bools instead, just didn't seem like the right thing to do

Do you mean a struct of bools that uses bilge (or another crate) for bitpacking? Because a regular struct of bools would be much larger -- at one byte per bool, that's 32 _bytes_ instead of 32 bits. That's also another reason why we use bitflags, to pack a lot of booleans into a small space.

hecatia-elegua:
Edit: yes, I mean a bilge struct, which is bitpacked

kevinaboos:
> Edit: yes, I mean a bilge struct, which is bitpacked

Ah i see, ok that does make sense.

For this PR, I think it's best to accept it with just the `variant_count` addition. I do want to use `bilge`, but the right place to experiment with that is the `pte_flags` crate, which is actually used in tons of other places and would be a much better proving ground for `bilge`. In contrast, if we add usage of `bilge` here in the signal handler crates, there's no easy way to actually test it thoroughly because this code isn't used and it also doesn't require any complex features like `pte_flags` do.

hecatia-elegua:
Please look at #963. It lists all places with bitfiddle/field/flags, some of those are just tiny changes or notes, though.
As discussed shortly on discord, `pte_flags` is a bit special due to how bitflags works, so we want to look into other usecases too.

kevinaboos (approved):
Thanks!

PR #968, opened 2023-06-06
Add middle path for blocking mutex

### LINKED ISSUE #577: Improve mutex performance
#576 provides a simple mutex implementation; ideally, we'd use a more performant one. Currently, `raw_mutex` only has a fast path (mutex is unlocked) and a slow path (mutex is locked - block current task). We should also have a middle path, which spins for a bit, hoping the mutex is unlocked soon.

Some useful links:
- How Linux does it: https://0xax.gitbooks.io/linux-insides/content/SyncPrim/linux-sync-4.html
- Another possible implementation: https://arxiv.org/pdf/1906.00490.pdf
- The concept of a backoff could be useful in some capacity: https://docs.rs/crossbeam/latest/crossbeam/utils/struct.Backoff.html

tsoutsman:
Starting the task ID counters at 1 rather than 0 also lets us represent task IDs using NonZeroUsize permitting niche optimisations e.g. `Option<TaskId>` taking up as much space as a `usize`. This would however involve some unsafe and even if we do decide to implement it, is better left for another PR.

Closes #577.

### DIFF
--- a/kernel/sync_block/src/lib.rs
+++ b/kernel/sync_block/src/lib.rs
@@ -83,18 +83,20 @@ impl MutexFlavor for Block {
             // likelihood of this is infinitesimally small and the code, is still correct as
             // once the lock is released the holder will still set data.holder to 0 and we
             // will exit the loop.
-            while holder_task.is_running()
-                && data.holder.load(Ordering::Acquire) == holder_id
-            {
+            while holder_task.is_running() && data.holder.load(Ordering::Acquire) == holder_id {
                 core::hint::spin_loop();
             }
-            if let Some(guards) = Self::try_lock(mutex, data) {
-                return guards;
-            }
+            // Holder is either no longer running, or has released the lock.
+            // Either way we will try the fast path one more time before moving
+            // onto the slow path.
         } else {
             // Unlikely case that another thread just acquired the lock, but hasn't yet set
             // data.holder.
             log::warn!("could not get holder task for mutex middle path");
+        }
+
+        if let Some(guards) = Self::try_lock(mutex, data) {
+            return guards;
         }
 
         // Slow path
--- a/kernel/sync_block/src/lib.rs
+++ b/kernel/sync_block/src/lib.rs
@@ -1,4 +1,4 @@
-#![feature(negative_impls)]
+#![feature(negative_impls, let_chains)]
 #![no_std]
 
 mod condvar;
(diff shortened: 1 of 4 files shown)


tsoutsman:
I left the `log::warn` in there so that when we switch to using `sync_block` we can make sure that the problematic interleaving is in fact incredibly rare. After that we can remove it.

tsoutsman:
Closes #576.

kernel/sync_block/src/lib.rs
```diff
@@ -47,21 +66,55 @@ impl MutexFlavor for Block {
...
         // when the mutex is unlocked and never be unblocked.
         if let Some(guards) = Self::try_lock(mutex, data) {
-            guards
+            return guards;
+        }
+
+        // Middle path: Wait for one timeslice (see below) of the mutex holder to see if
+        // they release the lock.
+        let holder_id = data.holder.load(Ordering::Acquire);
+        if let Some(holder_task) = task::get_task(holder_id).and_then(|task| task.upgrade()) {
```
kevinaboos: we can make this conditional short-circuit to false if the holder_id is `0`.  (not valid syntax unless let-chains now support this)
```rust
if holder_id != 0 && let Some(holder_task) = task::get_task(holder_id).and_then(|task| task.upgrade()) {
```

kernel/sync_block/src/lib.rs
```diff
@@ -47,21 +66,55 @@ impl MutexFlavor for Block {
...
+            if let Some(guards) = Self::try_lock(mutex, data) {
+                return guards;
+            }
         } else {
-            data.wait_until(|| Self::try_lock(mutex, data))
+            // Unlikely case that another thread just acquired the lock, but hasn't yet set
+            // data.holder.
+            log::warn!("could not get holder task for mutex middle path");
         }
+
```
kevinaboos, line 97: shall we try the fast path again here before blocking this task, considering that we just executed a series of "long-running" operations above as part of the middle path? I think so.

In other words, the lock may have been released whilst we were running `data.holder.load()` and `task::get_task(holder_id).and_then(|task| task.upgrade())`, so we can try to grab it before sleeping here.

kernel/sync_block/src/lib.rs
```diff
@@ -47,21 +66,55 @@ impl MutexFlavor for Block {
...
+        // they release the lock.
+        let holder_id = data.holder.load(Ordering::Acquire);
+        if let Some(holder_task) = task::get_task(holder_id).and_then(|task| task.upgrade()) {
+            // Hypothetically, if holder_task is running on another core and is perfectly in
+            // sync with us, we would only ever check if they are running when we are also
+            // running and so we wouldn't detect when their timeslice is over. However, the
+            // likelihood of this is infinitesimally small and the code, is still correct as
+            // once the lock is released the holder will still set data.holder to 0 and we
+            // will exit the loop.
+            while holder_task.running_on_cpu().is_some()
```
kevinaboos: if you just want to know whether it's running, you can use this
```suggestion
            while holder_task.is_running()
```

kernel/sync_block/src/lib.rs
kevinaboos: actually yeah, this works, all you need to do is move L91-93 to here instead such that it runs unconditionally after the middle path attempt.

kevinaboos (changes requested):
after more consideration and study, i think this is a good approach. I left a few small comments.

kevinaboos (approved):

PR #969, opened 2023-06-06
`interrupt_controller` crate: an abstraction over Arm GIC & x86 APIC+IOAPIC

NathanRoyer:
This PR introduces the `interrupt_controller` crate, which acts as an abstraction over:
- Arm Generic Interrupt Controllers (GIC), on aarch64
- the APIC & IOAPIC, on x86_64.

The PR doesn't include any use of the crate, as this is a first draft.
Please share your opinion on this before I continue.

### DIFF
--- a/kernel/gic/src/gic/dist_interface.rs
+++ b/kernel/gic/src/gic/dist_interface.rs
@@ -27,6 +27,7 @@ use cpu::MpidrValue;
 mod offset {
     use crate::{Offset32, Offset64};
     pub(crate) const CTLR:       Offset32 = Offset32::from_byte_offset(0x000);
+    pub(crate) const IIDR:       Offset32 = Offset32::from_byte_offset(0x008);
     pub(crate) const IGROUPR:    Offset32 = Offset32::from_byte_offset(0x080);
     pub(crate) const ISENABLER:  Offset32 = Offset32::from_byte_offset(0x100);
     pub(crate) const ICENABLER:  Offset32 = Offset32::from_byte_offset(0x180);
@@ -136,6 +137,12 @@ pub fn send_ipi_gicv2(registers: &mut GicRegisters, int_num: u32, target: IpiTar
     registers.write_volatile(offset::SGIR, value);
 }
 
+pub struct Implementer {
+    pub product_id: u8,
+    pub version: u8,
+    pub implementer_jep106: u16,
+}
+
 impl super::ArmGic {
     pub(crate) fn distributor(&self) -> &GicRegisters {
         match self {
@@ -148,6 +155,15 @@ impl super::ArmGic {
         match self {
             Self::V2(v2) => &mut v2.distributor,
             Self::V3(v3) => &mut v3.distributor,
+        }
+    }
+
+    pub fn implementer(&self) -> Implementer {
+        let raw = self.distributor().read_volatile(offset::IIDR);
+        Implementer {
+            product_id: (raw >> 24) as _,
+            version: ((raw >> 12) & 0xff) as _,
+            implementer_jep106: (raw & 0xfff) as _,
         }
     }
 
--- a/kernel/gic/src/gic/dist_interface.rs
+++ b/kernel/gic/src/gic/dist_interface.rs
@@ -137,9 +137,13 @@ pub fn send_ipi_gicv2(registers: &mut GicRegisters, int_num: u32, target: IpiTar
     registers.write_volatile(offset::SGIR, value);
 }
 
+/// Deserialized content of the `IIDR` distributor register
 pub struct Implementer {
+    /// Product Identifier of this distributor.
     pub product_id: u8,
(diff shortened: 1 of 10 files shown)


kernel/interrupt_controller/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+
+            (LocalInterruptNumber(num), prio)
+        }
+
+        #[cfg(target_arch = "x86_64")] {
+            panic!("This is an aarch64-only mechanism.");
+        }
+    }
+
+    /// Tell the interrupt controller that the current interrupt handling as ended.
```
kevinaboos: small typo
```suggestion
    /// Tell the interrupt controller that the current interrupt has been handled.
```

kernel/interrupt_controller/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+    pub fn end_of_interrupt(&self, number: LocalInterruptNumber) {
+        get_int_ctlr!(int_ctlr, end_of_interrupt);
+
+        #[cfg(target_arch = "aarch64")] {
+            int_ctlr.end_of_interrupt(number.0)
+        }
+
+        #[cfg(target_arch = "x86_64")] {
+            // what do we do with the IRQ number?
+            int_ctlr.eoi();
```
kevinaboos: on x86, we pass in an `Option<InterruptNumber>` to the `eoi()` function, because the interrupt handler is the one with the proper knowledge to determine whether the EOI signal requires an interrupt number. 

The interrupt handler doesn't always know whether the system is using PIC or APIC, but `eoi()` will figure that out when determining how to actually send the EOI signal. Basically, the older style PIC interrupt chip requires passing the interrupt number, but the newer APIC does not.

kernel/interrupt_controller/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        #[cfg(target_arch = "aarch64")] {
+            let (num, prio) = int_ctlr.acknowledge_interrupt();
+
+            (LocalInterruptNumber(num), prio)
+        }
+
+        #[cfg(target_arch = "x86_64")] {
+            panic!("This is an aarch64-only mechanism.");
+        }
```
kevinaboos: since the whole function is aarch64-only, you can remove this block.

kernel/interrupt_controller/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+            let _ = priority;
+        }
+    }
+
+    /// Aarch64-specific way to read the current pending interrupt number & priority.
+    #[cfg(target_arch = "aarch64")]
+    pub fn acknowledge_interrupt(&self) -> (LocalInterruptNumber, Priority) {
+        get_int_ctlr!(int_ctlr, acknowledge_interrupt);
+
+        #[cfg(target_arch = "aarch64")] {
```
kevinaboos: since the whole function is aarch64-only, you can remove this redundant cfg line.

kernel/gic/src/gic/dist_interface.rs
```diff
@@ -136,6 +137,12 @@ pub fn send_ipi_gicv2(registers: &mut GicRegisters, int_num: u32, target: IpiTar
     registers.write_volatile(offset::SGIR, value);
 }
 
+pub struct Implementer {
+    pub product_id: u8,
+    pub version: u8,
+    pub implementer_jep106: u16,
```
kevinaboos, line 147: can you document what this is? and the `Implementer` struct in general

kernel/interrupt_controller/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+    ($name:ident, $func:ident, $this:expr) => {
+        let mut $name = get_ioapic($this.id).expect(concat!("BUG: ", stringify!($func), "(): get_ioapic() returned None"));
+    };
+    ($name:ident, $func:ident) => {
+        let mut $name = get_my_apic().expect(concat!("BUG: ", stringify!($func), "(): get_my_apic() returned None"));
+        let mut $name = $name.write();
+    };
+}
+
+/// The Cpu were this interrupt should be handled, as well as
```
kevinaboos: typo
```suggestion
/// The Cpu where this interrupt should be handled, as well as
```

kevinaboos (changes requested):
Overall this looks really great! 

I think it might be a bit easier to read if we take the standard approach of splitting each arch's implementation into a separate module file.

Aside from that, the abstractions look good to me, so I'd say to go ahead and continue fleshing out the implementation. I'm sure we'll identify some other problems or improvements once we actually start using these abstractions, but for now I don't see anything wrong with them. :+1:

kevinaboos:
lmk when you're ready for another review, e.g., after you've split the arch-specific functions into separate module files.

kevinaboos (approved):

PR #970, opened 2023-06-08
page allocator: support allocating pages within an address range

kevinaboos:
This is only needed on aarch64 for a workaround to allow #940 to work.

Currently, based on the needs of aarch64, we reserve 128MiB of virtual address space for executable text sections. This region is contiguous with the base kernel image's .text section.

### DIFF
--- a/kernel/page_allocator/src/lib.rs
+++ b/kernel/page_allocator/src/lib.rs
@@ -476,13 +476,15 @@ fn find_any_chunk(
 		Inner::Array(ref mut arr) => {
 			for elem in arr.iter_mut() {
 				if let Some(chunk) = elem {
-					let lowest_start_page = *max(chunk.start(), range.start());
-					let highest_end_page  = *min(chunk.end(), range.end());
-					warn!("Chunk: {:?}, low_start: {:?}, num_pages: {}, high_end: {:?}", chunk, lowest_start_page, num_pages, highest_end_page);
-					if lowest_start_page + num_pages <= highest_end_page {
+					// Use max and min below to ensure that the range of pages we allocate from
+					// is within *both* the current chunk's bounds and the range's bounds.
+					let lowest_possible_start_page = *max(chunk.start(), range.start());
+					let highest_possible_end_page  = *min(chunk.end(), range.end());
+					warn!("Chunk: {:?}, low_start: {:?}, num_pages: {}, high_end: {:?}", chunk, lowest_possible_start_page, num_pages, highest_possible_end_page);
+					if lowest_possible_start_page + num_pages <= highest_possible_end_page {
 						warn!("\t\t yes");
 						return adjust_chosen_chunk(
-							lowest_start_page,
+							lowest_possible_start_page,
 							num_pages,
 							&chunk.clone(),
 							ValueRefMut::Array(elem),
@@ -512,11 +514,13 @@ fn find_any_chunk(
 			// This results in an O(1) allocation time in the general case, until all address ranges are already in use.
 			let mut cursor = tree.upper_bound_mut(Bound::Included(range.end()));
 			while let Some(chunk) = cursor.get().map(|w| w.deref()) {
-				let lowest_start_page = *max(chunk.start(), range.start());
-				let highest_end_page  = *min(chunk.end(), range.end());
-				if lowest_start_page + num_pages <= highest_end_page {
+				// Use max and min below to ensure that the range of pages we allocate from
+				// is within *both* the current chunk's bounds and the range's bounds.
+				let lowest_possible_start_page = *max(chunk.start(), range.start());
+				let highest_possible_end_page  = *min(chunk.end(), range.end());
+				if lowest_possible_start_page + num_pages <= highest_possible_end_page {
 					return adjust_chosen_chunk(
-						lowest_start_page,
+						lowest_possible_start_page,
 						num_pages,
 						&chunk.clone(),
 						ValueRefMut::RBTree(cursor)
--- a/kernel/page_allocator/src/lib.rs
+++ b/kernel/page_allocator/src/lib.rs
@@ -682,15 +682,6 @@ pub fn allocate_pages_deferred(
 }
 
 
-/// Allocates the given number of pages with the constraint they they must be
-/// within the given inclusive `range` of pages.
-pub fn allocate_pages_in_range(
-	num_pages: usize,
-	range: &PageRange,
-) -> Result<(AllocatedPages, DeferredAllocAction<'static>), &'static str> {
-	allocate_pages_deferred(None, num_pages, Some(range))
-}
-
 /// Similar to [`allocated_pages_deferred()`](fn.allocate_pages_deferred.html),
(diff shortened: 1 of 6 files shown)


kevinaboos:
@tsoutsman @NathanRoyer this isn't yet ready to merge (still has extra log statements and such), but the logic should be complete. I've tested it every way I can think of, but am still looking for reviews just to check the logic/math since it's such a core subsystem. Always helps to have an extra set of eyes on things, thanks!

kernel/page_allocator/src/lib.rs
```diff
@@ -429,38 +451,56 @@ fn find_specific_chunk(
...
+					if lowest_start_page + num_pages <= highest_end_page {
+						warn!("\t\t yes");
+						return adjust_chosen_chunk(
+							lowest_start_page,
+							num_pages,
+							&chunk.clone(),
+							ValueRefMut::Array(elem),
+						);
+					} else {
+						warn!("\t\t no");
```
NathanRoyer: I suspect these warn! calls were meant for temporary debugging, maybe not.
I thought I'd signal this.

NathanRoyer (approved):
I can confirm nothing seems weird here (I spent extra time in `find_any_chunk` to make sure that function was good).

Things I would change but you don't have to:
- this file seems to use hard tabs for some reason
- explicit imports were replaced with wildcard imports :(
- some warn! calls look like debugging statements, dunno if you want to keep them

kernel/page_allocator/src/lib.rs
kevinaboos: Yep, you're right. Will remove

kevinaboos:
> I can confirm nothing seems weird here (I spent extra time in `find_any_chunk` to make sure that function was good).

Much appreciated! 

> Things I would change but you don't have to:
> 
> * this file seems to use hard tabs for some reason

Oh weird, I'll look at that in another PR. 

> * explicit imports were replaced with wildcard imports :(

Yeah, I can do this. That's just me being lazy temporarily. 

> * some warn! calls look like debugging statements, dunno if you want to keep them

Indeed, will remove before merging.

kernel/page_allocator/src/lib.rs
```diff
@@ -624,6 +677,7 @@ fn adjust_chosen_chunk(
 pub fn allocate_pages_deferred(
 	requested_vaddr: Option<VirtualAddress>,
 	num_pages: usize,
+	within_range: Option<&PageRange>,
```
tsoutsman, line 677: Currently it's possible to request both a specific `vaddr` and a specific `within_range` even if the two values contradict.

May be better to do something like
```rust
pub enum MemoryRequest {
    Address(VirtualAddress),
    Range(PageRange),
}

pub fn allocate_pages_deferred(num_pages: usize, request: MemoryRequest);
```
(probably with a name better than `MemoryRequest`).

kernel/page_allocator/src/lib.rs
```diff
@@ -470,20 +510,33 @@ fn find_any_chunk(
 			// Because we allocate new pages by peeling them off from the beginning part of a chunk, 
 			// it's MUCH faster to start the search for free pages from higher addresses moving down. 
 			// This results in an O(1) allocation time in the general case, until all address ranges are already in use.
-			let mut cursor = tree.upper_bound_mut(Bound::Excluded(&DESIGNATED_PAGES_HIGH_START));
+			let mut cursor = tree.upper_bound_mut(Bound::Included(range.end()));
 			while let Some(chunk) = cursor.get().map(|w| w.deref()) {
-				if chunk.start() <= designated_low_end {
-					break; // move on to searching through the designated regions
+				let lowest_start_page = *max(chunk.start(), range.start());
+				let highest_end_page  = *min(chunk.end(), range.end());
```
tsoutsman: Same as above
```suggestion
				let lowest_start_page = *min(chunk.start(), range.start());
				let highest_end_page  = *max(chunk.end(), range.end());
```

kernel/page_allocator/src/lib.rs
```diff
@@ -429,38 +451,56 @@ fn find_specific_chunk(
...
-					if  chunk.size_in_pages() < num_pages || 
-						chunk.start() <= designated_low_end || 
-						chunk.end() >= &DESIGNATED_PAGES_HIGH_START
-					{
-						continue;
-					} 
-					else {
-						return adjust_chosen_chunk(*chunk.start(), num_pages, &chunk.clone(), ValueRefMut::Array(elem));
+					let lowest_start_page = *max(chunk.start(), range.start());
+					let highest_end_page  = *min(chunk.end(), range.end());
```
tsoutsman: I think it should be
```suggestion
					let lowest_start_page = *min(chunk.start(), range.start());
					let highest_end_page  = *max(chunk.end(), range.end());
```

tsoutsman (changes requested):
A potential way to simplify this in the future would be to move `UPCOMING_PAGE_TABLE_RECURSIVE_P4_INDEX` and `RECURSIVE_P4_INDEX` to the last entries in the page table so we can just chop off the last two pages rather than allocating around them.

kevinaboos:
> A potential way to simplify this in the future would be to move `UPCOMING_PAGE_TABLE_RECURSIVE_P4_INDEX` and `RECURSIVE_P4_INDEX` to the last entries in the page table so we can just chop off the last two pages rather than allocating around them.

yes indeed, i plan to do something similar to this once we refactor the heap a bit. We can't actually use entry 511, but we can use two contiguous entries like 509 and 510 instead of 508 and 510 like we have now.

kernel/page_allocator/src/lib.rs
kevinaboos: ha, yea i have that lined up for a future PR already. Great minds think alike.

kernel/page_allocator/src/lib.rs
kevinaboos: Hmm. My logic here is that the the range of pages we end up allocating needs to be within _both_ the chunk and the range. Looking at the first one, the lowest possible start page must be within both the chunk and the specified range, meaning that it has to be greater than both the chunk's start page and range's start part. If we use `min` for `lowest_start_page`, the Chunk's starting page might be below (outside of) the given range, or the range's starting page might also be below the chunk. So that wouldn't work.

The same is true for the highest-possible ending page, just inverted. That seems correct to me; did I make a mistake?

kernel/page_allocator/src/lib.rs
kevinaboos: see discussion on other comment.

kernel/page_allocator/src/lib.rs
tsoutsman: Yea ok you're right. I think the variable names also tripped me up because the lowest start page is actually the largest of the two start pages.

tsoutsman (approved):

kernel/page_allocator/src/lib.rs
kevinaboos: Ah I see. Will add some comments to that effect

PR #972, opened 2023-06-13
Use `interrupt_controller` in `interrupts::aarch64`

NathanRoyer:
Changes:
- `BOARD_CONFIG` becomes const so that I can reference it in other consts
- `BOARD_CONFIG` now contains board-specific interrupt numbers
- `interrupt_controller`'s initializer is called from captain
- `interrupt_controller::InterruptDestination` no longer contains a local_number on aarch64
- `interrupt_controller::send_ipi` allows sending to AllButMe
- new `interrupt_controller::InterruptNumber` type, wrapping a `u8` on all platforms
- aarch64-specific `interrupt_controller::init_secondary_cpu_interface`; panics on x86_64 (documented)
- don't enable timer/shootdown/pl011 interrupts from `interrupts::init` as they're enabled later


NathanRoyer:
todo: implement Binary, Octal, LowerHex, UpperHex for *InterruptNumber

NathanRoyer:
todo: Convert get_int_ctlr to a private util method

NathanRoyer:
This PR is paused until we find a way to represent both x86's interrupt numbers (a u8) and Arm GIC's interrupt number types (local peripheral, IPI, shared peripheral)

kernel/interrupt_controller/src/lib.rs
```diff
@@ -18,22 +18,24 @@ pub use arch::{
     SystemInterruptControllerVersion,
     SystemInterruptControllerId,
     LocalInterruptControllerId,
-    SystemInterruptNumber,
-    LocalInterruptNumber,
     Priority,
     SystemInterruptController,
     LocalInterruptController,
+    init,
```
kevinaboos, line 24: what's this for?

kernel/interrupt_controller/src/aarch64.rs
```diff
@@ -145,71 +112,129 @@ impl SystemInterruptControllerApi for SystemInterruptController {
...
+            .as_mut()
+            .expect("BUG: enable_local_interrupt(): INTERRUPT_CONTROLLER was uninitialized");
+
+        int_ctlr.set_interrupt_state(num as _, enabled);
     }
 
-    fn send_ipi(&self, destination: InterruptDestination) {
-        get_int_ctlr!(int_ctlr, send_ipi);
-        int_ctlr.send_ipi(destination.local_number.0, IpiTargetCpu::Specific(destination.cpu));
+    fn send_ipi(&self, num: InterruptNumber, dest: Option<CpuId>) {
```
kevinaboos: why did this function's parameters change? It seemed to be good previously -- the `InterruptDestination` type is much clearer than `Option<CpuId>`. 

It's not clear (and also surprising, especially without documentation) that passing in `dest = None` sends the IPI to all other CPUs. Seems like we should just use the `InterruptDestination` enum.

kernel/interrupt_controller/src/x86_64.rs
```diff
@@ -12,10 +12,6 @@ pub struct SystemInterruptControllerId(pub u32);
 #[derive(Debug, Copy, Clone)]
 pub struct LocalInterruptControllerId(pub u32);
 #[derive(Debug, Copy, Clone)]
-pub struct SystemInterruptNumber(pub(crate) u8);
-#[derive(Debug, Copy, Clone)]
-pub struct LocalInterruptNumber(pub(crate) u8);
-#[derive(Debug, Copy, Clone)]
 pub struct Priority;
 
     /// Initializes the interrupt controller, on aarch64
```
kevinaboos: indentation issue here

kevinaboos (changes requested):
looks decent, i left a few small comments. Also there is a conflict that needs to be resolved.

kernel/interrupt_controller/src/lib.rs
NathanRoyer: This re-exposes the init function, which is called from captain

kernel/interrupt_controller/src/aarch64.rs
NathanRoyer: It was a wrapping struct before, but an enum would be better indeed. I'll see if I can do that

kernel/interrupt_controller/src/lib.rs
kevinaboos: ah ok, nevermind, my github display was confusing. All good.

kevinaboos (approved):

PR #973, opened 2023-06-13
Make most basic memory type functions `const`

kevinaboos:


kevinaboos:
@NathanRoyer requesting your review for a double-check on the canonicalization functions.

PR #975, opened 2023-06-14
Merge deallocated pages into contiguously-adjacent chunks

### LINKED ISSUE #971: page_allocator: use `RbTree`'s `Entry` API for deallocation
Rather than always inserting a new `Chunk` for a dropped `AllocatedPages`, we should use the [`Entry` API](https://docs.rs/intrusive-collections/0.9.5/intrusive_collections/rbtree/struct.RBTree.html#method.entry) instead, which is faster and allows for basically free checks to see if we can eagerly merge the deallocated pages with an existing chunk of pages. This is "free" because it would add no additional overhead beyond inserting a new chunk. In fact, in the case of being able to merge with an existing (contiguously-adjacent) free chunk, it would be significantly faster because no heap allocation would be required for the new chunk.

Relevant code: https://github.com/theseus-os/Theseus/blob/e43c4e9ff4b677878a4d3f5ba5fd2719c0670f62/kernel/page_allocator/src/lib.rs#L285

kevinaboos:
Closes #971

### DIFF
--- a/kernel/page_allocator/src/lib.rs
+++ b/kernel/page_allocator/src/lib.rs
@@ -10,12 +10,11 @@
 //! but there are several convenience functions that offer simpler interfaces for general usage. 
 //!
 //! # Notes
-//! This allocator only makes one attempt to merge freed chunks (for de-fragmentation)
-//! upon deallocation. It does not iteratively merge adjacent chunks in order to
-//! Free chunks are lazily merged more aggressively upon running out of address space
+//! This allocator only makes one attempt to merge deallocated pages into existing
+//! free chunks for de-fragmentation. It does not iteratively merge adjacent chunks in order to
+//! maximally combine separate chunks into the biggest single chunk.
+//! Instead, free chunks are lazily merged only when running out of address space
 //! or when needed to fulfill a specific request.
-//! It only merges free chunks lazily upon request, i.e., when we run out of address space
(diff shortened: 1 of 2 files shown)


kernel/page_allocator/src/lib.rs
```diff
@@ -9,8 +9,11 @@
 //! The core allocation function is [`allocate_pages_deferred()`](fn.allocate_pages_deferred.html), 
 //! but there are several convenience functions that offer simpler interfaces for general usage. 
 //!
-//! # Notes and Missing Features
-//! This allocator currently does **not** merge freed chunks (de-fragmentation) upon deallocation. 
+//! # Notes
+//! This allocator only makes one attempt to merge freed chunks (for de-fragmentation)
+//! upon deallocation. It does not iteratively merge adjacent chunks in order to
```
hecatia-elegua, line 14: the code looks good to me, only here the end of sentence is missing

kernel/page_allocator/src/lib.rs
```diff
@@ -9,8 +9,11 @@
...
+//! Free chunks are lazily merged more aggressively upon running out of address space
```
hecatia-elegua: the code looks good to me (besides clippy's suggestion), only here the end of sentence is missing

kevinaboos: Oh thanks, that's what I get for trying to code while gaming 😅

PR #977, opened 2023-06-14
range_inclusive crate

Ramla-I:
Adds a range inclusive type that can create an iterator which doesn't mutate the original range.


libs/range_inclusive/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+    type Item = Idx;
+    type IntoIter = RangeInclusiveIterator<Idx>;
+
+    fn into_iter(self) -> RangeInclusiveIterator<Idx> {
+        self.iter()
+    }
+}
+
+pub struct RangeInclusiveIterator<Idx> {
+    offset: Idx,
```
kevinaboos: `offset` is a tiny bit misleading, name-wise. I'd vote for `current` or something similar.

libs/range_inclusive/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+#![no_std]
+#![feature(step_trait)]
+
+#[cfg(test)]
+mod test;
+
+use core::iter::Step;
+use core::ops::{RangeBounds, Bound, Bound::Included};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
kevinaboos: add `Hash` here too, as per `std::ops::RangeInclusive`

kevinaboos (changes requested):
Functionality looks good. Can you add docs to all structs and methods? Most can just be copied from `std::ops::RangeInclusive`, with the obvious exception of iterator-related stuff.

Also, please add implementations of the other appropriate iterator traits:
* `ExactSizeIterator`
* `DoubleEndedIterator`

Food for thought: (no need to do this now) we could probably remove the `Clone` bound on `Idx` by using an iterator type that takes a reference to the original `RangeInclusive` rather than a clone of it. Cloning isn't free, or at least not as cheap as copying in the vast majority of cases.

kevinaboos (approved):

PR #981, opened 2023-06-19
Rename `scheduler_realtime` to `scheduler_priority`

tsoutsman:
Depends on https://github.com/theseus-os/Theseus/pull/980.

The realtime scheduler doesn't actually implement rate monotonic
scheduling as the period is just used as a glorified inverse priority.

Next steps in the scheduler saga:
- Allow changing scheduler policy live
- Merge `set_priority` and `set_periodicity` (as the periodicity isn't
  properly used)
- Merge scheduler and run queue crates (pending discussion on Discord)
- Update docs to better explain crates, particularly `scheduler_epoch`
- Don't store idle task in queue
- Use binary heap for priority queue

I've implemented some of these and will open a PR once https://github.com/theseus-os/Theseus/pull/980 and #981 are
merged.

Signed-off-by: Klimenty Tsoutsman <klim@tsoutsman.com>

PR #982, opened 2023-06-19
Move idle task out of run queue

tsoutsman:
Depends on #980 and #981.

Moves `state_transfer` to `old_crates` because it no longer compiles and would need to be reworked, but it wasn't functional in the first place.


kevinaboos (approved):
Looks good!

Regarding the "TODO"s/questions you left, I think it's probably best to continue to return an `Option<TaskRef>` from `select_next_task()`, and then have the `schedule()` function select an idle task independently of the runqueue. This goes hand-in-hand with another change we can make, which is to move the idle task into CPU-local storage, since it "belongs" a CPU instance rather than a runqueue instance. This also makes the runqueue design have fewer concerns and also more flexible, since we may not always want to create exactly one runqueue per CPU.

I'll likely make some of those changes later, or you can do it as part of your series of scheduler PRs; your call.

PR #983, opened 2023-06-19
Merge priority and periodicity scheduler functions

tsoutsman:
Depends on #980 #981 #982.

The priority (realtime) scheduler's previous use of periodicity was just as an inverse priority.


kevinaboos (approved):
LGTM, thanks!

PR #985, opened 2023-06-22
Add priority inheritance for blocking mutex

tsoutsman:
Depends on #981 #982 #983.

The code duplication across `runqueue_epoch` and `runqueue_priority` is temporary before we refactor schedulers and runqueues to not use `cfg` blocks.


kevinaboos:
all dependency PRs have been merged in now :+1:

tsoutsman:
Should be good to merge.

kernel/sync_block/src/lib.rs
```diff
@@ -89,18 +89,35 @@ impl MutexFlavor for Block {
...
             // Either way we will try the fast path one more time before moving
             // onto the slow path.
+
+            if let Some(guards) = Self::try_lock(mutex, data) {
+                return guards;
+            }
+
+            // Slow path
+            #[cfg(priority_inheritance)]
+            let deinherit_priority = scheduler::inherit_priority(&holder_task);
```
kevinaboos: This is a nice design for undoing the effects of priority inheritance. It'd be even nicer to use a guard type such that the effects of priority inheritance can also be undone in the case of exceptional execution (unwinding after a panic). Should be easy enough to wrap up the returned closure in a struct that implements `Drop` too.

kevinaboos (changes requested):
Looks good overall, just left one suggestion about using a guard-like type to ensure priority inheritance gets undone even in the case of a panic.

kevinaboos (approved):

PR #986, opened 2023-06-23
Add `graphic=no` and `terminal` auto-launch options for QEMU

tsoutsman:
`graphic=no` adds the `-nographic` flag to the QEMU invocation. Supporting this required a few changes to `captain`.

`terminal=(name of terminal)` e.g. `terminal=alacritty` will launch a new terminal window connected to `SERIAL2`. We do this by testing the next free TTY with `tools/free_tty` and assuming QEMU will use the same TTY. This may seem jank, but I think it's the only feasible solution.

> Note: for a more typical terminal emulator like GNOME's, you need a more complicated terminal command, i.e., `make run terminal="gnome-terminal -- screen"`

I've tested this on MacOS (side note: macOS makes it very hard to launch a new terminal window from the terminal) so if someone could test this on Linux as well that would be good. Also, I wasn't able to build for `aarch64` so the changes to `captain` need to be tested to ensure builds for `aarch64` didn't break.
> Note: Kevin tested this and it worked properly on aarch64 and all other build configurations.


kernel/captain/src/lib.rs
```diff
@@ -175,14 +175,15 @@ pub fn init(
...
+        Err(error) => info!("Window manager initialisation failure: {error}"),
+    }
 
-    device_manager::init(
-        #[cfg(target_arch = "x86_64")]
-        key_producer,
-        #[cfg(target_arch = "x86_64")]
-        mouse_producer,
-    )?;
+    #[cfg(not(target_arch = "x86_64"))]
```
kevinaboos: nit: prefer aarch64 instead of not x86 for clarity

Makefile
```diff
@@ -948,15 +953,37 @@ loadable: run
...
 ### builds and runs Theseus in QEMU, but pauses execution until a GDB instance is connected.
-run_pause: $(iso)
+run_pause: $(iso) $(free_tty_bin)
+ifdef terminal
+# Check which TTY the OS would give a process requesting a TTY
+	$(eval temp_tty += $(shell $(free_tty_binary)))
+# If another process snags a new TTY here, the command below will connect to the wrong TTY, but
+# there's not much we can do about it.
+#
+# The slee is to give QEMU time to create the TTY.
```
kevinaboos: typo

Makefile
```diff
@@ -948,15 +953,37 @@ loadable: run
...
 ### builds and runs Theseus in QEMU
-run: $(iso)
+run: $(iso) $(free_tty_binary)
+ifdef terminal
+# Check which TTY the OS would give a process requesting a TTY
+	$(eval temp_tty += $(shell $(free_tty_binary)))
+# If another process snags a new TTY here, the command below will connect to the wrong TTY, but
+# there's not much we can do about it.
+#
+# The slee is to give QEMU time to create the TTY.
```
kevinaboos: typo

kernel/captain/src/lib.rs
```diff
@@ -175,14 +175,15 @@ pub fn init(
 
     // arch-gate: no windowing/input support on aarch64 at the moment
     #[cfg(target_arch = "x86_64")]
-    let (key_producer, mouse_producer) = window_manager::init()?;
+    match window_manager::init() {
+        Ok((key_producer, mouse_producer)) => {
+            device_manager::init(key_producer, mouse_producer)?;
+        },
+        Err(error) => info!("Window manager initialisation failure: {error}"),
```
kevinaboos: might want to mention that this is expected if nographic was specified.

kevinaboos (changes requested):
Looks good. A few minor points:
1. The name `free_tty` could be misinterpreted as freeing (releasing) a TTY, so I'd vote changing it to something more specific like `available_tty` or `get_tty`.
2. Can we use `cargo run` to invoke the tty getter application, or does that interfere with the `fork()` call therein? If possible, I'd prefer that as it makes the makefile dependency set cleaner.

Two general comments:
* I wonder if there is a proper way is to actually obtain the QEMU-created PTY from  QEMU itself. That'd be nice.
* Do we always want "nographic" specifically? I'm unsure if that will interfere with our `mon:stdio` settings.

tsoutsman:
> - I wonder if there is a proper way is to actually obtain the QEMU-created PTY from QEMU itself. That'd be nice.

I'm pretty sure there isn't aside from grepping the output but that is non-trivial since it's a stream and we still want to display it to the user.

> - Do we always want "nographic" specifically? I'm unsure if that will interfere with our mon:stdio settings.

It shouldn't. The only thing `nographic` option changes is that the VGA text mode is redirected to stdio as well.

kevinaboos:
> The only thing `nographic` option changes is that the VGA text mode is redirected to stdio as well.

I see that from the QEMU manual, however it doesn't actually do that on my system -- i see no graphical output, even when forcing VGA text mode. Nbd, doesn't matter since we don't even use VGA text mode by default any more. It's also worth nothing that `nographic` doesn't remove the VGA device, so Theseus still sees and uses it. So it's not truly headless yet, it just hides the graphical window even though Theseus is still rendering content there. This is good for now since we don't actually yet support true headless operation where graphics doesn't exist, but it'd be easy to account for that possibility in the captain during initialization.

kevinaboos (approved):

PR #987, opened 2023-06-24
Implement `println!` using nested `format_args!`

tsoutsman:
`concat!` prevents `format_args!` from capturing variables from the surrounding scope. Implementing `println!` using a nested `format_args!` removes this limitation allowing us to do the following:
```rust
let x = 3;
println!("{x}");
```
Similar to `std::println!`.

Also, the change [does not impact performance][1].

[1]: https://github.com/rust-lang/rust/pull/106824


kevinaboos (approved):
Nice, I was wondering about doing that sort of variable capturing in custom format macros! 

There are a few other places where we do this similar `concat!()` pattern, a few I can think of off the top of my head are `early_printer`, `println_log!()` in the `deps` app, etc. Would you mind submitting similar PR(s) for those?

PR #989, opened 2023-06-24
task_switch: explicitly drop current task's TLS reference

kevinaboos:
* In the inner task switching routine, pass in access to the actual `RefMut<Option<TaskRef>>` object that was borrowed from the `CURRENT_TASK` TLS area, such that the actual task switch logic can explicitly drop it at the right point -- just before switching to the next task.

* This change ensures that the `CURRENT_TASK` TLS variable that is used in task switching is dropped before switching to the next task, in order to ensure that `RefCell`'s runtime borrowing mechanic always works across task switches.

### DIFF
--- a/kernel/task/src/lib.rs
+++ b/kernel/task/src/lib.rs
@@ -37,6 +37,7 @@ use alloc::{
 };
 use core::{
     any::Any,
+    cell::RefMut,
     fmt,
     hash::{Hash, Hasher},
     ops::Deref,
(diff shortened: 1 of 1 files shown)


tsoutsman (approved):
So just to be clear, the problem was that the lifetime returned by the implicit `deref_mut` on the `RefMut` was too long (i.e. after `next.set_as_current_task()` it became invalid)?

The PR itself looks good aside from the clippy thing. Not the biggest fan of how this can occur in seemingly safe code, but I don't see a good way to enforce it.

kevinaboos:
> So just to be clear, the problem was that the lifetime returned by the implicit `deref_mut` on the `RefMut` was too long (i.e. after `next.set_as_current_task()` it became invalid)?

Yep, once again lifetimes last longer than I expect. Since dropping a primitive reference type is now a no-op, the only way we can control its lifetime is by dropping the `RefMut` object. 

> The PR itself looks good aside from the clippy thing. Not the biggest fan of how this can occur in seemingly safe code, but I don't see a good way to enforce it.

Yeah, I hear that. It's mostly a correctness issue.

PR #994, opened 2023-06-28
Add `&` and `|` operators to `hull`

tsoutsman:
Adds `&` and `|` operators, and implements the `jobs` and `history` commands.


tsoutsman:
Depends on #1000.

kevinaboos (commented):
This looks generally good so far. The one thing I'm unsure of is spawning another entire task to watch a new app task, but I suppose we could accept that heavyweight approach for now and work on a different monitoring mechanism later, e.g., a select-like operator.

I'm ok with accepting it as is, though I'll wait for our meeting tmrw to discuss it in person.

kevinaboos (approved):

PR #996, opened 2023-06-29
`multicore_bringup::aarch64` - fix identity-mapped entrypoint

NathanRoyer:
This fixes the (currently disabled) secondary CPU entrypoint identity mapping issues:
- the mapping is now executable
- the hardcoded address is now `0x5001_0000`, which is currently entirely unused, as it seems.

This is still disabled, as we need a proper way in theseus to dynamically locate and map identity-mappable regions.

### DIFF
--- a/kernel/multicore_bringup/src/aarch64.rs
+++ b/kernel/multicore_bringup/src/aarch64.rs
@@ -57,10 +57,6 @@ pub fn handle_ap_cores(
         let mut kernel_mmi = kernel_mmi_ref.lock();
         let page_table = &mut kernel_mmi.page_table;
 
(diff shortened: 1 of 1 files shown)


kevinaboos (approved):
Thanks, i'll work on supporting true identity mappings sometime later next week. Shouldn't be hard to get one with the current set of features, it just wouldn't be guaranteed so the API is kinda shitty. Basically we'd just want to allocate any single frame and then allocate the page that matches that frame's address. If you ask for a frame at a specific address then you are implying you know that frame actually corresponds to real memory, which was one of the problems behind the original implementation.

PR #1000, opened 2023-07-04
Clone run queue in `rq` app

tsoutsman:
Avoids disabling preemption for the duration of the program which can interfere with printing to the terminal.


kevinaboos (approved):
Thanks!

PR #1001, opened 2023-07-05
Enable `BorrowedMappedPages` to use `Arc`, `Rc`, `&'static` `MappedPages`

kevinaboos:
* Previously, the `BorrowedMappedPages` and `BorrowedSliceMappedPages` types could only accommodate a plain `MappedPages` owned object, which made it impossible to have multiple different instances of a `BorrowedMappedPages` object that all refer to the same underlying `MappedPages` memory region.

* This is now possible by creating a `BorrowedMappedPages` based on any owned (non-lifetime) type that implements `Borrow<MappedPages>`, such as `Arc<MappedPages>`, which you can then clone and repeatedly pass to `BorrowedMappedPages::from()`.

* You can also obtain a reference to the inner `Borrow<MappedPages>` object via the `inner_ref()` method.

### DIFF
--- a/kernel/memory/src/paging/mapper.rs
+++ b/kernel/memory/src/paging/mapper.rs
@@ -981,12 +981,15 @@ impl Drop for MappedPages {
 /// 1. `T: FromBytes`: the same parameter used in [`MappedPages::as_type()`] functions.
 /// 2. `M: Mutability`: whether or not the `MappedPages` can be mutably borrowed. 
 ///    * For ease of use, the default is [`Immutable`].
-/// 3. `B: Borrow<MappedPages>`: enables the user of this type to use a wrapper type
-///    around `MappedPages`, such as `Arc`, `Rc`, or a `&'static` reference.
+/// 3. `B: Borrow<MappedPages>`: enables the user of this type to use a borrow-able
+///    wrapper type or reference around `MappedPages`.
 ///    * For ease of use, the default is just a plain `MappedPages` with no wrapper.
 ///      This default `MappedPages` type can be borrowed immutably or mutably.
-///    * An `Arc<MappedPages`, `Rc<MappedPages>`, and `&'static MappedPages`
-///      can only be borrowed immutably because these types do not permit mutable access.
+///    * Example types that satisfy the bounds for `B` are `Arc<MappedPages`,
+///      `Rc<MappedPages>`, `&MappedPages`, etc, all of which can only be
+///      borrowed immutably via the `from()` method, as these types do not permit mutable access.
+///      For mutable borrows, see the `from_mut()` method, which works with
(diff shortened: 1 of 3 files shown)


kevinaboos:
@tsoutsman requesting your review since you previously helped review the original `BorrowedMappedPages` implementation.

kernel/memory/src/paging/mapper.rs
```diff
@@ -1035,15 +1064,20 @@ impl<T: FromBytes> BorrowedMappedPages<T, Mutable> {
...
 
-impl<T: FromBytes, M: Mutability> BorrowedMappedPages<T, M> {
+impl<T: FromBytes, M: Mutability, B: Borrow<MappedPages>> BorrowedMappedPages<T, M, B> {
     /// Consumes this object and returns the inner `MappedPages`.
-    pub fn into_inner(self) -> MappedPages {
+    pub fn into_inner(self) -> B {
         self.mp
     }
+
+    /// Returns a reference to the inner borrowed `MappedPages`.
```
tsoutsman: It doesn't return a reference to the inner borrowed `MappedPages`, but rather to the container holding the `MappedPages`.

kernel/memory/src/paging/mapper.rs
```diff
@@ -1160,15 +1225,21 @@ impl<T: FromBytes> BorrowedSliceMappedPages<T, Mutable> {
...
 
-impl<T: FromBytes, M: Mutability> BorrowedSliceMappedPages<T, M> {
+impl<T: FromBytes, M: Mutability, B: Borrow<MappedPages>> BorrowedSliceMappedPages<T, M, B> {
     /// Consumes this object and returns the inner `MappedPages`.
-    pub fn into_inner(self) -> MappedPages {
+    pub fn into_inner(self) -> B {
         self.mp
     }
+
+    /// Returns a reference to the inner borrowed `MappedPages`.
```
tsoutsman, line 1236: Same as above.

tsoutsman (changes requested):
We can use the [`BorrowMut`](https://doc.rust-lang.org/std/borrow/trait.BorrowMut.html) trait for `BorrowedMappedPages<T, Mutable, B>`.

kernel/memory/src/paging/mapper.rs
kevinaboos: true, thanks.

kevinaboos:
> We can use the [`BorrowMut`](https://doc.rust-lang.org/std/borrow/trait.BorrowMut.html) trait for `BorrowedMappedPages<T, Mutable, B>`.

Good idea! that didn't cross my mind, thanks.

kevinaboos:
Changes are in.

tsoutsman (approved):

PR #1004, opened 2023-07-06
Frame allocator changes

Ramla-I:
- Merged `Chunk` and `AllocatedFrames` into a single type `Frames` with two states: Unmapped and Mapped. The frame allocator is responsible for make sure a `Frames` object has no overlap.
- Created a `Region` type used just for bookkeeping.
- Changed the core allocation functions to remove `Frames` objects from the free list before allocating from it (since we can no longer clone a `Frames`). 
- For code local to the frame allocator, I changed all references to `AllocatedFrames` to `UnmappedFrames`, but for all public functions I kept the `AllocatedFrames` terminology.
- I also kept the word "chunk" in comments and function definitions rather than changing it everywhere to "frames" to keep the changes minimal.

Questions:
1. Can we combine `PhysicalMemoryRegion` and `Region`? They contain the same information, but `Region` has additional trait implementations.
2. I've added comments in the code for a few specific questions.


kernel/frame_allocator/src/lib.rs
```diff
@@ -262,113 +278,287 @@ pub enum MemoryRegionType {
...
+        UnmappedFramesIter {
+            _owner: self,
+            range: self.frames.iter(),
+        }
+    }
+}
+
+/// An iterator over each [`UnmappedFrame`] in a range of [`Frames<{FrameState::Unmapped}>`].
+/// 
+/// To Do (get Kevin's input): Description is no longer valid, since we have an iterator for RangeInclusive now.
```
Ramla-I: I think this comment can be made smaller, just say that we implement out iterator to tie the lifetime of the `UnmappedFrame` to the original object. Remove the portions about `RangeInclusive` iterator since we have our own implementation now.
Or should we keep this in, and mention that we created our own `RangeInclusive` to deal with this issue?

kernel/frame_allocator/src/lib.rs
```diff
@@ -421,120 +648,56 @@ impl AllocatedFrames {
...
-
-impl<'f> IntoIterator for &'f AllocatedFrames {
-    type IntoIter = AllocatedFramesIter<'f>;
-    type Item = AllocatedFrame<'f>;
-    fn into_iter(self) -> Self::IntoIter {
-        AllocatedFramesIter {
-            _owner: self,
-            range_iter: self.frames.iter(),
-        }
+// To Do: will this be an issue as now this applies to Chunk as well as AllocatedFrames?
```
Ramla-I: previously AllocatedFrames didn't have an implementation for PartialEq, only Chunk did. Will this definition of PartialEq (and the other traits) for AllocatedFrames lead to any incorrectness?

kernel/frame_allocator/src/lib.rs
```diff
@@ -870,8 +1030,62 @@ fn contains_any(
...
+/// Adds the given `frames` to the given `list` as a Region of reserved frames. 
+/// 
+/// Returns the range of **new** frames that were added to the list, 
+/// which will be a subset of the given input `frames`.
+///
+/// Currently, this function adds no new frames at all if any frames within the given `frames` list
+/// overlap any existing regions at all. 
+/// TODO: handle partially-overlapping regions by extending existing regions on either end.
+/// 
+/// TODO: combine both functions for adding reserved regions/ frames using a trait OR make one function which adds to both lists, since they are always called together.
```
Ramla-I: Right now I've duplicated this function for the free list containing Frames and the regions list.
Since they're always called together, should we just have one function which checks both the RESERVED_REGIONS list and the free_reserved_frames_list?

kernel/frame_allocator/src/lib.rs
kevinaboos: The comment is still correct, and in fact I recently updated it to reflect the new `RangeInclusive` type that you contributed. The key aspect is that the iterator clones the original range instead of taking a reference to it, so we still need this custom iterator type hack.

kevinaboos: It seems like you overwrote my recent changes. See the latest version from the main branch, it's more accurate.

kernel/frame_allocator/src/lib.rs
kevinaboos: no worries, that should be fine.

kernel/frame_allocator/src/lib.rs
```diff
@@ -421,120 +648,56 @@ impl AllocatedFrames {
...
-impl<'f> IntoIterator for &'f AllocatedFrames {
-    type IntoIter = AllocatedFramesIter<'f>;
-    type Item = AllocatedFrame<'f>;
-    fn into_iter(self) -> Self::IntoIter {
-        AllocatedFramesIter {
-            _owner: self,
-            range_iter: self.frames.iter(),
-        }
+// To Do: will this be an issue as now this applies to Chunk as well as AllocatedFrames?
+#[cfg(not(test))]
```
kevinaboos: why define different behavior for tests vs actual runtime? IMO we shouldn't do that.

kernel/frame_allocator/src/lib.rs
```diff
@@ -606,7 +772,11 @@ enum AllocationError {
     /// or enough remaining chunks that could satisfy the requested allocation size.
     OutOfAddressSpace(usize),
     /// The starting address was found, but not all successive contiguous frames were available.
-    ContiguousChunkNotFound(Frame, usize)
+    ContiguousChunkNotFound(Frame, usize),
+    /// Failed to remove a chunk from the free list given a reference to it.
+    ChunkRemovalFailed,
+    /// Failed to merge or split a Chunk.
+    ChunkOperationFailed,
```
kevinaboos: these two errors are of a different class than the other four variants. These are logic errors, which should probably throw a panic with a "BUG: ..." message rather than returning an error. I don't think the user/caller of an `allocate_frames` function needs to be concerned with logic errors internal to the frame allocator; in fact, the best case would be to avoid these conditions completely.

kernel/frame_allocator/src/lib.rs
kevinaboos: yeah, that's a good idea, go ahead and make that change :+1:

kernel/frame_allocator/src/lib.rs
```diff
@@ -750,88 +922,76 @@ fn find_any_chunk(
...
-/// Returns a tuple of three chunks:
-/// 1. The `Chunk` containing the requested range of frames starting at `start_frame`.
-/// 2. The range of frames in the `chosen_chunk` that came before the beginning of the requested frame range.
-/// 3. The range of frames in the `chosen_chunk` that came after the end of the requested frame range.
-fn split_chosen_chunk(
+/// The final part of the main allocation routine that splits the given chosen chunk
+/// into multiple smaller chunks, thereby "allocating" frames from it.
+///
+/// This function breaks up that chunk into multiple ones and returns an `AllocatedFrames` 
+/// from (part of) that chunk, ranging from `start_frame` to `start_frame + num_frames`.
```
kevinaboos: this is an inclusive range, right? If so, let's use that specific `FrameRange` type so we don't introduce too many new concepts.

kernel/frame_allocator/src/lib.rs
```diff
@@ -382,24 +572,61 @@ impl AllocatedFrames {
...
+    /// 1. The `Frames` containing the requested range of frames starting at `start_frame`.
+    /// 2. The range of frames in the `self` that came before the beginning of the requested frame range.
+    /// 3. The range of frames in the `self` that came after the end of the requested frame range.
+    /// 
+    /// If `start_frame` is not contained within `self` or `num_frames` results in an end frame greater than the end of `self`,
+    /// then `self` is not changed and we return (self, None, None).
+    pub fn split(
+        self,
+        start_frame: Frame,
+        num_frames: usize,
```
kevinaboos: as per my other comment, let's use `FrameRange` here for the sake of clarity.

kernel/frame_allocator/src/lib.rs
```diff
@@ -382,24 +572,61 @@ impl AllocatedFrames {
...
+    /// Splits up the given `Frames` into multiple smaller `Frames`.
+    /// 
+    /// Returns a tuple of three `Frames`:
+    /// 1. The `Frames` containing the requested range of frames starting at `start_frame`.
+    /// 2. The range of frames in the `self` that came before the beginning of the requested frame range.
+    /// 3. The range of frames in the `self` that came after the end of the requested frame range.
+    /// 
+    /// If `start_frame` is not contained within `self` or `num_frames` results in an end frame greater than the end of `self`,
+    /// then `self` is not changed and we return (self, None, None).
+    pub fn split(
```
kevinaboos: The name `split` isn't clear enough. I'm not sure of the exact best name (we can bikeshed it a bit), but it should include the word "range", and maybe "extract".

split_around_range? extract_range? split_range?

kernel/frame_allocator/src/lib.rs
```diff
@@ -382,24 +572,61 @@ impl AllocatedFrames {
...
+        self,
+        start_frame: Frame,
+        num_frames: usize,
+    ) -> (Self, Option<Self>, Option<Self>) {
```
kevinaboos: imo, the order of the returned tuple should match the order of the frame ranges: Option<Self>, Self, Option<Self>. 

It would be best to return a struct that had named fields, which would make things exceedingly clear:
```rust
struct SplitRange {
    before_start: Option<Frames<...>>,
    start_to_end: Frames<...>,
    after_end: Option<Frames<...>>,
}
```

kernel/frame_allocator/src/lib.rs
```diff
@@ -382,24 +572,61 @@ impl AllocatedFrames {
...
-    /// Splits this `AllocatedFrames` into two separate `AllocatedFrames` objects:
+    /// Splits up the given `Frames` into multiple smaller `Frames`.
+    /// 
+    /// Returns a tuple of three `Frames`:
+    /// 1. The `Frames` containing the requested range of frames starting at `start_frame`.
+    /// 2. The range of frames in the `self` that came before the beginning of the requested frame range.
+    /// 3. The range of frames in the `self` that came after the end of the requested frame range.
+    /// 
+    /// If `start_frame` is not contained within `self` or `num_frames` results in an end frame greater than the end of `self`,
+    /// then `self` is not changed and we return (self, None, None).
```
kevinaboos: this seems like a code smell (design mistake) in that you're hiding errors. This condition should be returned as an explicit error, with a function signature like
```rust
pub fn split_range(self, range: FrameRange) -> Result<SplitRange, Self> { ... }
```

kernel/frame_allocator/src/lib.rs
```diff
@@ -382,24 +572,61 @@ impl AllocatedFrames {
...
+    pub fn split(
+        self,
+        start_frame: Frame,
+        num_frames: usize,
+    ) -> (Self, Option<Self>, Option<Self>) {
+        if (start_frame < *self.start()) || (start_frame + (num_frames - 1) > *self.end()) || (num_frames == 0) {
+            return (self, None, None);
+        }
```
kevinaboos: if you take a `FrameRange` parameter, the logic becomes super simple here -- just use the `overlap()` method

kevinaboos (changes requested):
This is just a partial review so you can get started addressing some of my comments without having to wait too long for more feedback.

My main concern is that the term `UnmappedFrames` or `Frames<{FrameState::Unmapped}>` is misleading, or at the very least, quite easy for readers to misunderstand. It makes it sound like those frames were previously mapped and we _just_ unmapped them; in fact, that's how the term has been used thus far, e.g., in the `page_table_entry` crate. I'd be interested to hear your thoughts on this; perhaps we can brainstorm another term.

Similarly, there isn't a `FrameState` variant that differentiates between "free" (un-allocated) and "unmapped" (allocated). I find that to be confusing, as one would expect there to be a series of changes in state when going from Mapped --> Unmapped --> Free.

Another minor concern is that `Frames<{FrameState::Unmapped}>` is ugly to read and definitely hard for newcomers to understand, especially considering that that const enum variant syntax isn't even stable. Let's define typedef aliases for these states and use them everywhere, including within the frame allocator itself.

> * Merged `Chunk` and `AllocatedFrames` into a single type `Frames` with two states: Unmapped and Mapped. The frame allocator is responsible for make sure a `Frames` object has no overlap.

This is generally fine. It's a bit weird that "chunk" is still used prolifically throughout the frame allocator, but I see that you did that on purpose to minimize changes. We can always refactor that later.

> * Created a `Region` type used just for bookkeeping.

Nice, I like this.

> * Changed the core allocation functions to remove `Frames` objects from the free list before allocating from it (since we can no longer clone a `Frames`).

This is inefficient, right? In my mind we should be able to avoid this, since we only need to split it rather than clone it... right?

> * For code local to the frame allocator, I changed all references to `AllocatedFrames` to `UnmappedFrames`, but for all public functions I kept the `AllocatedFrames` terminology.

Sure, fine for now I guess. I'd still like to decide on one term or typedef alias that we can use everywhere, as I said above.


> Questions:
> 
> 1. Can we combine `PhysicalMemoryRegion` and `Region`? They contain the same information, but `Region` has additional trait implementations.

Yep, good idea! Let's use the term `PhysicalMemoryRegion` since it's much more descriptive than `Region`.

> 2. I've added comments in the code for a few specific questions.

Thanks, I responded to them.

kernel/frame_allocator/src/lib.rs
Ramla-I: we're testing the split and merge functions, and for those we're concerned about the range end value as well. We don't have to redefine the Eq trait to test that though.

Ramla-I:
> > * Changed the core allocation functions to remove `Frames` objects from the free list before allocating from it (since we can no longer clone a `Frames`).
> 
> This is inefficient, right? In my mind we should be able to avoid this, since we only need to split it rather than clone it... right?

In the original design, a clone of the chosen Chunk and the reference to the Chunk in the rb-tree would be passed to the `allocate_from_chosen_chunk` function, where the cloned Chunk would be split and then the Chunk in the rb-tree would be removed.
The way I see it, I'm simply reversing the steps: removing the Frames object first and then splitting it, with additional steps to extract the Frames from its wrapper.

Also, I did consider the design where we split/ merged the chunk while it's still stored in the tree as that could be useful when merging contiguous `Frames` objects, but I don't think we can gain mutable access to an element in the tree. The `CursorMut` only has a `get()` method, no `get_mut()` method. I assume that's because changing the element can mess up the ordering of the rb-tree.

Though, rb-tree already has functions like `replace_with`, `insert_after`, etc, that rely on the user to maintain ordering, so we could add our own `get_mut` function.

kernel/frame_allocator/src/lib.rs
kevinaboos: ok yeah, just check the actual fields in the tests then. Redefining the trait (and in general, using `#[cfg(test)]`) is kinda odd.

kevinaboos:
ah, oops -- i made a change to the page allocator to use  `replace_with()` for that purpose, as in 99% of cases, we can just add on the deallocated page range to an existing chunk.

I just didn't apply that to the frame allocator yet, so I was mis-remembering what the Drop impl of `AllocatedFrames` did. Sorry! 

Anyway, take a look at the Drop implementation for `AllocatedPages`, you'll see what I did there and how it can be much more efficient. Do let me know if that approach isn't compatible with the new `Frames` type due to trait limitations or linear typing restrictions.

kernel/frame_allocator/src/lib.rs
Ramla-I: it's not really testing for overlap, more to see if the frame range parameter is contained within self.

kevinaboos: isn't that the same as what `FrameRange::overlap()` does? at least the logic looks quite similar to me

Ramla-I: It's one case of an overlap where a range is not the subset of another. The expectation of the split function is that the frames that are passed are a subset of self which is another case of an overlap.

e.g. take 2 ranges [0,4] and [1,3]. The overlap function will return Some() to indicate there's an overlap whereas our check will return false.

Ramla-I: where self = [0, 4] and the frame range parameter is [1,3]

kernel/frame_allocator/src/lib.rs
Ramla-I: extract_frames_with_range(self, frames: FrameRange)?

kevinaboos: Sure, that's a bit long though. I'd say to let the name of the below `SplitRange` struct define the name of the function (based on whatever you end up calling that type).

kernel/frame_allocator/src/lib.rs
kevinaboos: oh ok, so this is `contains()` then, right? sorry i misunderstood.

basically i'm trying to figure out how we can refactor and reuse this logic elsewhere, perhaps within `RangeInclusive` itself? e.g.:
```rust
impl RangeInclusive {
    pub fn contains_range(&self, other: &RangeInclusive) -> bool { ... }
}
```

kernel/frame_allocator/src/lib.rs
kevinaboos: ok so i skimmed the rust std docs for Vec and slice, and on second thought I don't think `extract` is a good choice because it has a bit of a name conflict with `extract_if` (which is the new name of the feature previously known as `drain_filter`).

`split_*` seems like the best, so let's just go with `split_range`. The `Vec` type also has a function called `split_off()`, which we could follow, but in my mind that invokes an idea of just one chunk being split off into two, whereas we have three.

kevinaboos:
Can we move the various new functions from `Frames` into a "lower-level" type, perhaps `FrameRange` or maybe even `RangeInclusive` (assuming you can still get Prusti to work therein)?  The functions I'm referring to are things like `merge`, `split`, `split_at`, `split_range`, `contains_range`, etc.

kernel/frame_allocator/src/lib.rs
Ramla-I: agree about split_off, I think split_range is good.

kernel/frame_allocator/src/lib.rs
Ramla-I: yeah we can do a contains function, but I guess do it at the `FrameRange` and `PageRange` level since that's where overlap is defined as well, and as you mentioned in another place, we might end up changing the `RangeInclusive` we use in the future.

kernel/frame_allocator/Cargo.toml
```diff
@@ -8,10 +8,12 @@ edition = "2021"
...
-intrusive-collections = "0.9.0"
 static_assertions = "1.1.0"
 
 range_inclusive = { path = "../../libs/range_inclusive" }
 
 kernel_config = { path = "../kernel_config" }
 memory_structs = { path = "../memory_structs" }
+
+[dependencies.intrusive-collections]
+git = "https://github.com/Ramla-I/intrusive-rs"
```
Ramla-I: added `get_mut`, and other minimal changes required to support that function.

Ramla-I:
New Changes:
- 4 typestates (`Free, Allocated, Mapped, Unmapped`)

- added `get_mut` to the rbtree implementation, which allows use to merge frames in the drop handler... we could use `replace_with` but then we'd have to replace with an empty frames, merge, and then replace the merged chunk back in.

- changed the name of `page_table_entry::UnmappedFrames` to `page_table_entry::UnmappedFramesInfo` since `UnmappedFrames` is now a typestate. Now the call back creates an `UnmappedFrames` from `UnmappedFramesInfo`

kernel/frame_allocator/Cargo.toml
kevinaboos: I don't know if implementing `get_mut()` on an RBTree's `CursorMut` is sound. We'd need to submit that as a separate PR to the upstream `intrusive-rs` crate to get feedback from the maintainers to see if doing that is correct. I have a sneaking suspicion that it's not sound to do so, otherwise Amanieu likely would've provided that API.

kernel/frame_allocator/src/lib.rs
```diff
@@ -263,112 +322,291 @@ pub enum MemoryRegionType {
...
-/// its **starting** `Frame`. This is useful so we can store `Chunk`s in a sorted collection.
+/// `Frames` implements the `Ord` trait, and its total ordering is ONLY based on
+/// its **starting** `Frame`. This is useful so we can store `Frames` in a sorted collection.
 ///
-/// Similarly, `Chunk` implements equality traits, `Eq` and `PartialEq`,
-/// both of which are also based ONLY on the **starting** `Frame` of the `Chunk`.
-/// Thus, comparing two `Chunk`s with the `==` or `!=` operators may not work as expected.
+/// Similarly, `Frames` implements equality traits, `Eq` and `PartialEq`,
+/// both of which are also based ONLY on the **starting** `Frame` of the `Frames`.
+/// Thus, comparing two `Frames` with the `==` or `!=` operators may not work as expected.
```
kevinaboos, line 356: This is a public type, so I'm concerned about the way we're implementing `Eq`/`PartialEq` for it. That would very likely lead to some usage mistakes in foreign code using this type.

I think implementing `Ord` based on just the starting frame is okay here, but not `Eq`. Do we need `Eq` for this to work in some particular collection or use context?

kernel/frame_allocator/src/lib.rs
```diff
@@ -263,112 +322,291 @@ pub enum MemoryRegionType {
...
+/// In the mapped state, frames have been mapped to a virtual memory page and are in use.
+/// In the unmapped state, frames have been unmapped and can be retured to the frame allocator.
+/// 
+/// When a `Frames` object in a free state is dropped, it will be added back to the free list.
+/// When a `Frames` object in an allocated state is dropped, it is transitioned to a free state. 
+/// When a `Frames` obect in an unmapped state is dropped, it is transitioned to an allocated state.
+/// We expect that `Frames` in a mapped state will never be dropped, but instead will be forgotten and then 
+/// recreated in an unmapped state when it's frames are cleared from the page table.
 ///
+///     (Free) <--> (Allocated) -> (Mapped) -> (Unmapped) -> (Allocated)
```
kevinaboos: I rewrote these docs a bit to clarify them. If you like these changes, you should just be able to accept this suggestion and have github auto-commit it to this PR branch.

```suggestion
///
/// Owning a `Frames` object implies globally-exclusive ownership of and access to
/// the range of frames it contains.
/// 
/// A `Frames` object can be in one of four states:
/// * `Free`: frames are owned by the frame allocator and have not been allocated for any use.
/// * `Allocated`: frames have been removed from the allocator's free list and are owned elsewhere;
///    they can now be used for mapping purposes.
/// * `Mapped`: frames have been (and are currently) mapped by a range of virtual memory pages.
/// * `Unmapped`: frames have been unmapped and can be returned to the frame allocator.
///
/// The drop behavior for a `Frames` object is based on its state:
/// * `Free`:  the frames will be added back to the frame allocator's free list.
/// * `Allocated`: the frames will be transitioned into the `Free` state.
/// * `Unmapped`: the frames will be transitioned into the `Allocated` state.
/// * `Mapped`: currently, Theseus does not actually drop mapped `Frames`, but rather they are forgotten
///    when they are mapped by virtual pages, and then re-created in the `Unmapped` state
///    after being unmapped from the page tables.
///
/// As such, one can visualize the `Frames` state diagram as such:
/// ```
/// (Free) <---> (Allocated) --> (Mapped) --> (Unmapped) --> (Allocated) <---> (Free)
/// ```
```

kernel/frame_allocator/src/lib.rs
kevinaboos: Also, excluding Prusti's needs, I can't think of any reason why we'd need to implement `Eq` for this type, especially considering that a `Frames` type is supposed to be globally exclusive/unique... which means that `Eq` would never return true for any two `Frames` objects.

kernel/frame_allocator/src/lib.rs
```diff
@@ -263,112 +322,291 @@ pub enum MemoryRegionType {
...
-impl Chunk {
-    fn as_allocated_frames(&self) -> AllocatedFrames {
-        AllocatedFrames {
-            frames: self.frames.clone(),
+
+// Type aliases for the Frames in each state
+pub type FreeFrames = Frames<{MemoryState::Free}>;
+pub type AllocatedFrames = Frames<{MemoryState::Allocated}>;
+pub type MappedFrames = Frames<{MemoryState::Mapped}>;
+pub type UnmappedFrames = Frames<{MemoryState::Unmapped}>;
```
kevinaboos: for correct docs appearance, we should use `///` doc comments for each type alias.

```suggestion
/// A type alias for `Frames` in the `Free` state.
pub type FreeFrames = Frames<{MemoryState::Free}>;
/// A type alias for `Frames` in the `Allocated` state.
pub type AllocatedFrames = Frames<{MemoryState::Allocated}>;
/// A type alias for `Frames` in the `Mapped` state.
pub type MappedFrames = Frames<{MemoryState::Mapped}>;
/// A type alias for `Frames` in the `Unmapped` state.
pub type UnmappedFrames = Frames<{MemoryState::Unmapped}>;
```

kernel/frame_allocator/src/lib.rs
```diff
@@ -263,112 +322,291 @@ pub enum MemoryRegionType {
...
         }
     }
 
-    /// Returns a new `Chunk` with an empty range of frames. 
-    const fn empty() -> Chunk {
-        Chunk {
-            typ: MemoryRegionType::Unknown,
-            frames: FrameRange::empty(),
-        }
+    /// Consumes the `Frames` in an free state and converts them to `Frames` in a allocated state.
```
kevinaboos:
```suggestion
    /// Consumes this `Frames` in the `Free` state and converts them into the `Allocated` state.

```

kernel/frame_allocator/src/lib.rs
```diff
@@ -263,112 +322,291 @@ pub enum MemoryRegionType {
...
+
+// Frames must not be Cloneable, and it must not expose its inner frames as mutable.
+assert_not_impl_any!(Frames<{MemoryState::Free}>: DerefMut, Clone);
+assert_not_impl_any!(Frames<{MemoryState::Allocated}>: DerefMut, Clone);
+assert_not_impl_any!(Frames<{MemoryState::Mapped}>: DerefMut, Clone);
+assert_not_impl_any!(Frames<{MemoryState::Unmapped}>: DerefMut, Clone);
+
+
+impl FreeFrames {
+    /// Creates a new `Frames` object in an free state.
```
kevinaboos:
```suggestion
    /// Creates a new `Frames` object in the `Free` state.
```

kernel/frame_allocator/src/lib.rs
```diff
@@ -263,112 +322,291 @@ pub enum MemoryRegionType {
...
+        f
     }
 }
-impl Deref for Chunk {
-    type Target = FrameRange;
-    fn deref(&self) -> &FrameRange {
-        &self.frames
+
+impl AllocatedFrames {
+    /// Consumes the `Frames` in an allocated state and converts them to `Frames` in a mapped state.
```
kevinaboos:
```suggestion
    /// Consumes this `Frames` in the `Allocated` state and converts them into the `Mapped` state.
```

kernel/frame_allocator/src/lib.rs
```diff
@@ -263,112 +322,291 @@ pub enum MemoryRegionType {
...
+            _phantom: core::marker::PhantomData,
+        }
     }
 }
-impl PartialOrd for Chunk {
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        Some(self.cmp(other))
+
+impl UnmappedFrames {
+    /// Consumes the `Frames` in an unmapped state and converts them to `Frames` in an allocated state.
```
kevinaboos:
```suggestion
    /// Consumes this `Frames` in the `Unmapped` state and converts them into the `Allocated` state.
```

kernel/frame_allocator/src/lib.rs
```diff
@@ -263,112 +322,291 @@ pub enum MemoryRegionType {
...
+        MemoryRegionType::Reserved
+    } else {
+        MemoryRegionType::Free
+    };
+    Frames{ typ, frames }
+}
+
+impl<const S: MemoryState> Drop for Frames<S> {
+    fn drop(&mut self) {
+        match S {
```
kevinaboos, line 471: is this match statement elided during code generation, or does it appear as a series of conditionals at runtime?

If it's not elided, we may need to implement it manually as 4 different `drop` impl blocks, one for each state, in order to avoid runtime overhead.

kevinaboos: Actually, I think you should just split it into four `impl Drop` blocks anyway, as i think it'd be easier to read/understand since it'd reduce nesting and conditional complexity. It might be even clearer if you used the type aliases.

(assuming Rust actually lets you do that as four separate impl blocks)

kernel/frame_allocator/src/lib.rs
```diff
@@ -263,112 +322,291 @@ pub enum MemoryRegionType {
...
 }
+assert_not_impl_any!(AllocatedFrame: DerefMut, Clone);
 
-// AllocatedFrames must not be Cloneable, and it must not expose its inner frames as mutable.
-assert_not_impl_any!(AllocatedFrames: DerefMut, Clone);
+/// The result of splitting a `Frames` object into multiple smaller `Frames` objects.
+pub struct SplitFrames<const S: MemoryState>  {
+    before_start: Option<Frames<S>>,
+    start_to_end:  Frames<S>,
+    after_end:    Option<Frames<S>>,
```
kevinaboos: nit: minor spacing/indent issues

kernel/frame_allocator/src/lib.rs
```diff
@@ -263,112 +322,291 @@ pub enum MemoryRegionType {
...
+}
 
-impl Deref for AllocatedFrames {
-    type Target = FrameRange;
-    fn deref(&self) -> &FrameRange {
-        &self.frames
+impl<const S: MemoryState> Frames<S> {
+    #[allow(dead_code)]
+    pub(crate) fn frames(&self) -> FrameRange {
+        self.frames.clone()
```
kevinaboos: what's the purpose of this, just to avoid having to call frames.clone()? seems unnecessary.

If you do actually need it, the standard Rust naming convention for a method like this is `to_*()` since it takes a reference and produces an owned object without actually cloning the outer object. Examples are `to_vec()`. So you could call it something like `to_range()`

kernel/frame_allocator/src/lib.rs
```diff
@@ -382,24 +620,62 @@ impl AllocatedFrames {
...
 
-    /// Splits this `AllocatedFrames` into two separate `AllocatedFrames` objects:
+    /// Splits up the given `Frames` into multiple smaller `Frames`.
+    /// 
+    /// Returns a `SplitFrames` instance containing three `Frames`:
+    /// 2. The range of frames in the `self` that came before the beginning of the requested frame range.
+    /// 1. The `Frames` containing the requested range of frames, `frames_to_extract`.
+    /// 3. The range of frames in the `self` that came after the end of the requested frame range.
+    /// 
+    /// If `frames_to_extract` is not contained within `self`, then `self` is not changed and we return it within an Err.
```
kevinaboos: some doc fixes

```suggestion
    /// 1. The range of frames in `self` that are before the beginning of `frames_to_extract`.
    /// 2. The `Frames` containing the requested range of frames, `frames_to_extract`.
    /// 3. The range of frames in `self` that are after the end of `frames_to_extract`.
    /// 
    /// If `frames_to_extract` is not contained within `self`, then `self` is returned unchanged within an `Err`.
```

kernel/frame_allocator/src/lib.rs
```diff
@@ -382,24 +620,62 @@ impl AllocatedFrames {
...
+        };
+
+        let after_end = if *start_to_end.end() == MAX_FRAME || *start_to_end.end() == *self.end(){
+            None
+        } else {
+            Some(Frames{ frames: FrameRange::new(*start_to_end.end() + 1, *self.end()), ..self })
+        };
+
+        core::mem::forget(self);
+        Ok( SplitFrames{ before_start, start_to_end, after_end} )
```
kevinaboos: formatting nits:
```suggestion
        let start_to_end = Frames { frames: frames_to_extract, ..self };
        
        let before_start = if start_frame == MIN_FRAME || start_frame == *self.start() {
            None
        } else {
            Some(Frames { frames: FrameRange::new(*self.start(), *start_to_end.start() - 1), ..self })
        };

        let after_end = if *start_to_end.end() == MAX_FRAME || *start_to_end.end() == *self.end() {
            None
        } else {
            Some(Frames { frames: FrameRange::new(*start_to_end.end() + 1, *self.end()), ..self })
        };

        core::mem::forget(self);
        Ok(SplitFrames { before_start, start_to_end, after_end })
```

kernel/frame_allocator/src/lib.rs
```diff
@@ -421,120 +697,48 @@ impl AllocatedFrames {
...
-    type Item = AllocatedFrame<'f>;
-    fn next(&mut self) -> Option<Self::Item> {
-        self.range_iter.next().map(|frame|
-            AllocatedFrame {
-                frame, _phantom: PhantomData,
-            }
-        )
+impl<const S: MemoryState> Borrow<Frame> for &'_ Frames<S> {
+    fn borrow(&self) -> &Frame {
+        self.frames.start()
```
kevinaboos, line 778: hmm, this is also an interesting choice. One generally expects `Borrow` and `Deref` to have similar behavior, i.e., here `Borrow` would return a `FrameRange` rather than just the starting Frame. What's the rationale behind this?

kevinaboos: Also if this behavior is necessary, we should explicitly mention it in the `Frames` docs alongside the disclaimers about `Ord`.

kernel/frame_allocator/src/lib.rs
```diff
@@ -421,120 +697,48 @@ impl AllocatedFrames {
...
-    frame: Frame,
-    _phantom: PhantomData<&'f Frame>,
-}
-impl<'f> Deref for AllocatedFrame<'f> {
-    type Target = Frame;
-    fn deref(&self) -> &Self::Target {
-        &self.frame
+impl<const S: MemoryState> fmt::Debug for Frames<S> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Frames({:?}, {:?})", self.typ, self.frames)
```
kevinaboos: i apologize for this super-duper nitpicky thing, but I'm used to reading frames/pages with the range first. sorry, lol

```suggestion
        write!(f, "Frames({:?}, {:?})", self.frames, self.typ)
```

kernel/frame_allocator/src/lib.rs
```diff
@@ -861,56 +1038,66 @@ fn contains_any(
...
-    match &mut list.0 {
-        Inner::Array(ref mut arr) => {
-            for chunk in arr.iter().flatten() {
-                if let Some(_overlap) = chunk.overlap(&frames) {
-                    // trace!("Failed to add reserved region {:?} due to overlap {:?} with existing chunk {:?}",
-                    //     frames, _overlap, chunk
-                    // );
-                    return Err("Failed to add reserved region that overlapped with existing reserved regions (array).");
+    // first check the regions list for overlaps and proceed only if there are none.
+    if !contains_any(&regions_list, &frames){
```
kevinaboos: it'd be nicer to read and produce less of a large diff if you did an early return pattern, e.g.,
```rust
if contains_any(...) {
    return Err(...);
}
```

kernel/frame_allocator/src/lib.rs
```diff
@@ -861,56 +1038,66 @@ fn contains_any(
...
 /// which will be a subset of the given input `frames`.
 ///
 /// Currently, this function adds no new frames at all if any frames within the given `frames` list
 /// overlap any existing regions at all. 
 /// TODO: handle partially-overlapping regions by extending existing regions on either end.
-fn add_reserved_region(
-    list: &mut StaticArrayRBTree<Chunk>,
+fn add_reserved_region_to_lists(
+    regions_list: &mut StaticArrayRBTree<PhysicalMemoryRegion>,
+    frames_list: &mut StaticArrayRBTree<FreeFrames>,
```
kevinaboos, line 1097: nice, i really like this redesign. :+1:

kernel/frame_allocator/src/lib.rs
```diff
@@ -861,56 +1038,66 @@ fn contains_any(
...
+                    cursor_mut.move_next();
                 }
-                cursor_mut.move_next();
             }
         }
+        
+        regions_list.insert(PhysicalMemoryRegion {
+                typ: MemoryRegionType::Reserved,
+                frames: frames.clone(),
+        }).map_err(|_c| "BUG: Failed to insert non-overlapping frames into list.")?;
```
kevinaboos: minor semantic correction for accuracy:
```suggestion
        }).map_err(|_c| "BUG: Failed to insert non-overlapping physical memory region into reserved regions list.")?;
```

kernel/memory_structs/src/lib.rs
```diff
@@ -7,19 +7,35 @@
 
 #![no_std]
 #![feature(step_trait)]
+#![allow(incomplete_features)]
```
kevinaboos, line 10: whoa, what's this? haven't seen this before. Is it related to using enum variants as const generic params?

kernel/memory_structs/src/lib.rs
```diff
@@ -7,19 +7,35 @@
...
+#![feature(adt_const_params)]
 
 use core::{
     cmp::{min, max},
     fmt,
     iter::Step,
-    ops::{Add, AddAssign, Deref, DerefMut, Sub, SubAssign}
+    ops::{Add, AddAssign, Deref, DerefMut, Sub, SubAssign},
+    marker::ConstParamTy
```
kevinaboos: alphabetical
```suggestion
    marker::ConstParamTy,
    ops::{Add, AddAssign, Deref, DerefMut, Sub, SubAssign},
```

kernel/memory_structs/src/lib.rs
```diff
@@ -7,19 +7,35 @@
...
 };
 use kernel_config::memory::{MAX_PAGE_NUMBER, PAGE_SIZE};
 use zerocopy::FromBytes;
 use paste::paste;
 use derive_more::*;
 use range_inclusive::{RangeInclusive, RangeInclusiveIterator};
 
+#[derive(PartialEq, Eq, ConstParamTy)]
+/// The possible states of a page or frame in virtual and physical memory.
```
kevinaboos:
```suggestion
/// The possible states that a range of exclusively-owned pages or frames can be in.
#[derive(PartialEq, Eq, ConstParamTy)]
```

kernel/memory_structs/src/lib.rs
```diff
@@ -470,6 +486,16 @@ macro_rules! implement_page_frame_range {
...
                 }
+
+                #[doc = "Returns true if `other` is contained within the `" $TypeName "`."]
+                pub fn contains_range(&self, other: &$TypeName) -> bool {
+                    if !other.is_empty() && (other.start() >= self.start()) && (other.end() <= self.end()) 
+                    {
+                        true
+                    } else {
+                        false
+                    }
```
kevinaboos: surprised that clippy didn't catch this, but you can just return the boolean conditional directly instead of doing `if (...) { true } else { false }`

kevinaboos (changes requested):
Based on our slack discussion, let's revert back to using `replace_with()` until we have confirmation from `intrusive-rs` maintainers that adding a `get_mut()` cursor API is okay/sound.

I love the new naming conventions in the frame allocator, they're much clearer! Nice. (I'm referring to things like `UnmappedFrames`, `FreeFrames` instead of `Chunk`, `PhysicalMemoryRegion`, etc.

`PhysicalMemoryRegion` and its docs are perfectly done, very nice!

One thing about your comment:
> * changed the name of `page_table_entry::UnmappedFrames` to `page_table_entry::UnmappedFramesInfo` since `UnmappedFrames` is now a typestate. Now the call back creates an `UnmappedFrames` from `UnmappedFramesInfo`

Agreed on the need for this name change; however, let's go with something like `UnmappedFrameRange` instead of "info" because "info" implies it's some kind of contextual metadata or something, not the actual frames that were unmapped.

-------

Note that I haven't yet deeply reviewed the changes to the internal frame allocator logic itself, i.e., `find_specific_chunk()`, `find_any_chunk()`, `allocate_from_chosen_chunk()`, etc.

I'm still not a huge fan of the redesign regarding `allocate_from_chosen_chunk()` in which you always remove a chunk from the list and then re-add it.... I know we previously discussed this briefly during a meeting, but I can't remember what the resolution was.
We can re-discuss this again once your changes to remove `get_mut()` usage are in. At that point, I will do a deep review of the actual logic changes in frame allocation, and hopefully we can go through them together in person to speed up my understanding of the changes.

kernel/frame_allocator/Cargo.toml
Ramla-I: ok, so there have been multiple issues filed about this ([link](https://github.com/Amanieu/intrusive-rs/issues?q=is%3Aissue+get_mut+is%3Aclosed))
It's considered unsound behavior because there can be multiple references to an item, and the recommendation is to use Cell/RefCell

kevinaboos: Ah yep, there it is. I knew I remembered seeing discussions about it in the past.

I think replace_with() is good enough for our needs here.

kernel/memory_structs/src/lib.rs
Ramla-I: right, it's to suppress the compiler warning `"the feature `adt_const_params` is incomplete and may not be safe to use and/or cause compiler crashes"`

kernel/frame_allocator/src/lib.rs
Ramla-I: To implement `Ord` and `PartialOrd,` we need an implementation of `PartialEq`

kernel/frame_allocator/src/lib.rs
Ramla-I: yeah, I'm removing this function. It's not necessary at all.

kernel/frame_allocator/src/lib.rs
Ramla-I: yes, this is necessary because when we create a cursor for the `RBTree`, we search using a `Frame` value, and so we need `Frames` to implement `Borrow<Frame>`

Ramla-I: added a disclaimer

kernel/frame_allocator/src/lib.rs
Ramla-I: That was actually my first attempt but Rust doesn't allow for specialized Drop implementations :( ([link](https://github.com/rust-lang/rust/pull/23638))

Also, I checked the binary for a simplified version of the Frames tyepstates, and it doesn't seem like the match statement is elided. A generic drop function is called for all Frames objects.

kernel/frame_allocator/src/lib.rs
kevinaboos: ah ok, right. I suppose the comment will have to suffice then.

kevinaboos:
Thanks, looks good. I can confirm that everything except the changes to the internal frame allocation logic are approved now. I'll review the inner details of the frame allocation logic changes soon.

kevinaboos:
(Also i made a few minor changes here and there, make sure to pull those down before you push any other changes)

Ramla-I:
> Thanks, looks good. I can confirm that everything except the changes to the internal frame allocation logic are approved now. I'll review the inner details of the frame allocation logic changes soon.

Thanks for the review. lmk if you would like to go through those changes together over zoom.

kernel/frame_allocator/src/lib.rs
kevinaboos: The match statement is in fact elided, see #1017. So we can resolve this discussion with no further action needed.

kernel/frame_allocator/src/lib.rs
```diff
@@ -262,113 +321,330 @@ pub enum MemoryRegionType {
...
+                } else {
+                    FREE_GENERAL_FRAMES_LIST.lock()
+                };        
+            
+                match &mut list.0 {
+                    // For early allocations, just add the deallocated chunk to the free pages list.
+                    Inner::Array(_) => {
+                        if list.insert(free_frames).is_ok() {
+                            return;
+                        }
```
kevinaboos: should log an error here, as that would indicate that we need to use a larger early static array.

kernel/frame_allocator/src/lib.rs
```diff
@@ -262,113 +321,330 @@ pub enum MemoryRegionType {
...
+                    }
+                    
+                    // For full-fledged deallocations, use the entry API to efficiently determine if
```
kevinaboos: this is probably copied from my old code (so my fault), but we're not actually using the entry API any more, so you should update this comment to reflect the new behavior.

kernel/frame_allocator/src/lib.rs
```diff
@@ -262,113 +321,330 @@ pub enum MemoryRegionType {
...
+impl<const S: MemoryState> Drop for Frames<S> {
+    fn drop(&mut self) {
+        match S {
+            MemoryState::Free => {
+                if self.size_in_frames() == 0 { return; }
+        
+                let mut free_frames: FreeFrames = Frames {
+                    typ: self.typ,
+                    frames: self.frames.clone(),
+                };
```
kevinaboos: why do we need to re-instantiate a new `Frames` object here? Isn't this the same as `self`?

kevinaboos: I think this is actually a dangerous bug lurking -- this could result in a double-free if a panic occurs 

Instead, use the trick to `mem::replace` `self` with an empty `Frames` object, which avoids creating another duplicate instance out of the blue.

In general, you now need to go through _every single_ place in the code that instantiates `Frames` objects to ensure you're not creating a duplicate instance that could be dropped and result in incorrectly adding a `FreeFrames` object back to the allocator's free list _twice_.

kernel/frame_allocator/src/lib.rs
```diff
@@ -262,113 +321,330 @@ pub enum MemoryRegionType {
...
+                                let mut next_frames = cursor_mut
+                                    .replace_with(Wrapper::new_link(Frames::empty()))
+                                    .expect("BUG: couldn't remove next frames from free list in drop handler")
+                                    .into_inner();
+
+                                // trace!("Prepending {:?} onto beg of next {:?}", free_frames, next_frames);
+                                if next_frames.merge(free_frames).is_ok() {
+                                    // trace!("newly merged next chunk: {:?}", next_frames);
+                                    // now return newly merged chunk into list
+                                    match cursor_mut.replace_with(Wrapper::new_link(next_frames)) {
```
kevinaboos: this whole sequence is quite inefficient, in two main ways:

1. We replace the frames at the cursor with an empty frames. There's no real point in doing this; we might as well just remove the frames from the list, modify (extend) them, and then re-insert them at the current cursor. That's as efficient as we can get.
2. Using `merge` is redundant -- we have already checked the merge condition here in L497, so it's a redundant check. The corresponding panic in L513 is also redundant and unnecessary, since we can directly modify the inner range of the frames object that we removed from the list/cursor above.

To summarize, we should just do the following:
1. Remove the frames in L499.
2. Directly modify those frames instead of redundantly calling `merge` on those frames.
3. Re-insert the modified frames object back into the rbtree using the current cursor.

Please also apply this change to the code block starting on L517.

kernel/frame_allocator/src/lib.rs
```diff
@@ -861,29 +1077,34 @@ fn contains_any(
...
+fn add_reserved_region_to_lists(
+    regions_list: &mut StaticArrayRBTree<PhysicalMemoryRegion>,
+    frames_list: &mut StaticArrayRBTree<FreeFrames>,
     frames: FrameRange,
 ) -> Result<FrameRange, &'static str> {
 
+    // first check the regions list for overlaps and proceed only if there are none.
+    if contains_any(regions_list, &frames){
+        return Err("Failed to add reserved region that overlapped with existing reserved regions.");
+    }
```
kevinaboos, line 1104: (this comment is for future reference)

you're close to getting the early return pattern right -- the point is to avoid the big `else` block, which is now unnecessary. See the changes I made (which also significantly reduces the noisy diff on this PR)

kevinaboos (changes requested):
Full review of internal logic is complete. I left some comments.

There are still some more efficiency improvements that can be made, but I didn't want to overburden you with too many nitpicky comments. I don't think we should bother addressing them until _after_ the verification-related changes are merged in, since that will likely change the code structure even further.

kernel/frame_allocator/src/lib.rs
Ramla-I: Point 1 makes sense, thanks for pointing that out.
For point 2, I can replace it for now, but once a `Frames` object is composed of a `TrustedChunk`, then we will have to use the merge function since we cannot directly set the range of a `TrustedChunk`.

kevinaboos: Okay, if that's the case for point 2, then don't bother with it until we are reviewing the `TrustedChunk` PR.

However, I'd question why that is _required_, or if it's just an artifact of the current `TrustedChunk` design. Regardless, it should be possible to refactor that code sequence to avoid the redundant conditional (in L497 and in `merge()`), which then will subsequently avoid the unreachable error condition in L513.

kernel/frame_allocator/src/lib.rs
Ramla-I: what is your opinion on using transmute to transition Frames objects between states?  It seems more efficient and safer than first replacing the previous frames object with an empty range and then forgetting it. I believe the session types in rust work uses transmute for this purpose.

Of course in this particular example, we'd still have to use mem::replace since we only have a mutable reference to work with.

kevinaboos: My opinion is that transmute is to be avoided at all costs. But yea, here the `mem::replace` trick is the perfect solution.

kevinaboos: And also how would you account for transmution in a verification process? That seems impossible

Ramla-I: yeah, there's no way to verify that.

kernel/frame_allocator/src/lib.rs
Ramla-I: I've replaced `replace_with` with `remove`. For now I'll keep it at that.

But for future discussion on the `TrustedChunk` design:
Checking the merge conditions and performing the merge needs to take place in the same function, as that's the granularity of verification. So for that, we need mutable access to the contiguous chunk in the rbtree which is only possible by removing it from the tree. So the sequence could be to `remove frames from the tree -> merge -> reinsert into tree `rather than `check merge conditions -> remove from tree -> merge -> reinsert`, but that would lead to the overhead of removing an object from the rbtree in every drop handler.

kernel/frame_allocator/src/lib.rs
```diff
@@ -469,10 +472,8 @@ impl<const S: MemoryState> Drop for Frames<S> {
             MemoryState::Free => {
                 if self.size_in_frames() == 0 { return; }
         
-                let mut free_frames: FreeFrames = Frames {
-                    typ: self.typ,
-                    frames: self.frames.clone(),
-                };
+                let frames = core::mem::replace(&mut self.frames, FrameRange::empty());
```
Ramla-I, line 475: we have to replace the `FrameRange` here rather than the whole `Frames` because trying to replace the `Frames` object leads to a compiler error for mismatched types, the dest is `&mut Frames<S>` and the src is a `FreeFrames`.

kernel/frame_allocator/src/lib.rs
kevinaboos: Okay cool.

There's an easy way to get around that, just have another method for `TrustedChunk` that acts as proof-of-work, sort of like a `VerifiedMergeable` type that is returned from that method that just performs the check. Then, when you pass that into another `merge_verified()`/pre-checked function, you know you don't need to re-evaluate the conditional. But we can get to that later.

kernel/frame_allocator/src/lib.rs
kevinaboos: got it. Hope that's not too hard to verify, seems tough but it works for me here.

kernel/frame_allocator/src/lib.rs
```diff
@@ -469,10 +472,8 @@ impl<const S: MemoryState> Drop for Frames<S> {
...
                 if self.size_in_frames() == 0 { return; }
         
-                let mut free_frames: FreeFrames = Frames {
-                    typ: self.typ,
-                    frames: self.frames.clone(),
-                };
+                let frames = core::mem::replace(&mut self.frames, FrameRange::empty());  
+                let free_frames: FreeFrames = Frames { typ: self.typ, frames };
         
                 let mut list = if free_frames.typ == MemoryRegionType::Reserved {
```
kevinaboos, line 478: this is fine for now, but in a future PR we should log an error and return early if we encounter an `Unknown` type here.

kevinaboos (approved):
woot, finally made it! The first step to pushing verified code up to the mainline!

Thanks very much! :tada:

PR #1007, opened 2023-07-11
Remove `mutex_sleep`

tsoutsman:
Should be merged in at the same time/depends on theseus-os/backtrace-rs#1 and theseus-os/wasmtime#3.

Not sure what the best way to go about that is because until they are merged CI will fail, but they also depend on this PR.


kevinaboos:
> Not sure what the best way to go about that is because until they are merged CI will fail, but they also depend on this PR.

Thanks, yes this is the best way to go about it. I will merge in those PRs first and then you can update this to use the latest commits from those submodule projects.

kevinaboos:
@tsoutsman the two other submodule PRs have been merged in. Thanks!

tsoutsman:
Should be ready.

kevinaboos (approved):

PR #1008, opened 2023-07-12
Replace `ping` with `ping_2`

tsoutsman:


applications/ping/src/lib.rs
```diff
@@ -1,356 +1,184 @@
-//! This application pings a specific IPv4 address and gets ping statistics.
-//! Important: QEMU does not support the ICMP protocol by default so it's important to 
-//! run this command: sudo sh -c "echo \"0 2147483647\" > /proc/sys/net/ipv4/ping_group_range"
-//! in the environment prior to running this application
```
kevinaboos, line 4: it'd be nice to retain this info. Can you move it into the README (and/or the relevant book section about running on QEMU)?

kevinaboos (changes requested):
Looks good, thanks! One small request before merging.

applications/ping/src/lib.rs
tsoutsman: 👍 I actually don't think I never needed to run that command on fedora or macos but I'll add it anyway.

kevinaboos (approved):
Excellent, thanks!

PR #1009, opened 2023-07-13
Port `upd` to new `net` API

tsoutsman:
Ports `upd` (and by extension `ota_update_client` and `http_client`) to the new `net` API.

A couple of points:
- Currently `net::get_ephemeral_port` doesn't keep track of used ports, but that's no better than it was before. I'll leave that for a future PR when I have a better idea of what information we need to track about ports.
- The `LockedSocket::connect` method is a temporary fix until we introduce our own socket types. After this PR I'll have a look at the API `net` needs to expose to `std` and rework our socket abstractions to better reflect this.
- I changed the logic in `ota_update_client` to not reconnect on every loop iteration. Previously, it would open a new connection for every file, but only waited on the last connection. Now it only opens one connection and waits on that connection.
- I'm not sure how I can test the changes to `upd`.

Next steps are:
- Remove `network_manager` and `ethernet_smoltcp_device`
- Tweak `net` interface to better reflect usage patterns and `std` needs
- Track used ports


kernel/http_client/src/lib.rs
```diff
@@ -1,56 +1,47 @@
...
     Requesting,
-    /// The HTTP request has been sent, but the response has not yet been fully received.
+    /// The HTTP request has been sent, but the response has not yet been fully
+    /// received.
     ReceivingResponse,
-    /// The response has been received in full, including the headers and the entire content.
-    Responded
+    /// The response has been received in full, including the headers and the
+    /// entire content.
+    Responded,
```
kevinaboos: it looks like you're running `rustfmt` or something, which is responsible for these changes. I don't mind this, but it appears that the default setting is to add hard line breaks at the 80th column (?).
I eventually do want to enforce this, but for now you can definitely increase that limit to 100 characters, 80 is unnecessarily short imo on modern screens. 100 characters is the max (soft) column length I've been using in all of my recent changesets over the past year or so.

kernel/net/src/interface.rs
```diff
@@ -56,14 +56,14 @@ impl NetworkInterface {
     }
 
     /// Adds a socket to the interface.
-    pub fn add_socket<T>(&self, socket: T) -> Socket<T>
+    pub fn add_socket<T>(self: &Arc<Self>, socket: T) -> Socket<T>
```
kevinaboos: if you always clone it, it's likely better to just accept `Arc<Self>` instead of a reference to make it clearer how the internal socket ownership is handled.

kernel/net/src/lib.rs
```diff
@@ -62,7 +62,22 @@ pub fn get_interfaces() -> &'static Mutex<Vec<Arc<NetworkInterface>>> {
...
+/// ports.
+pub fn get_ephemeral_port() -> u16 {
+    use rand::Rng;
+
+    // FIXME: Keep track of taken ports.
+
+    const RANGE_START: u16 = 49152;
+    const RANGE_END: u16 = u16::MAX;
+
+    // TODO: Does this need to be cryptographically secure?
```
kevinaboos: nah, it's just a random port, doesn't matter how it's chosen. (in my view)

kernel/http_client/src/lib.rs
```diff
@@ -89,237 +81,261 @@ impl HttpResponse {
...
+impl<'a> HttpClient<'a> {
+    // TODO: Use per-request destination rather than per-client.
+    /// Creates a new HTTP client connected to the given remote endpoint.
+    pub fn new(
+        interface: &'a Arc<NetworkInterface>,
+        local_port: u16,
+        remote_endpoint: IpEndpoint,
+    ) -> Result<Self, &'static str> {
+        let rx_buffer = tcp::SocketBuffer::new(vec![0; 256]);
+        let tx_buffer = tcp::SocketBuffer::new(vec![0; 256]);
```
kevinaboos, line 93: does the newer smoltcp API require passing in growable storage like a Vec, or can we use a fixed-size slice/array?
If it does require or support a growable collection type, is there a reason we choose a 256-byte zeroed vector rather than an empty one?

kernel/http_client/src/lib.rs
```diff
@@ -89,237 +81,261 @@ impl HttpResponse {
...
+            let may_recv = locked.may_recv();
+            drop(locked);
+
+            state = match state {
+                HttpState::Requesting if can_send => {
+                    debug!("http_client: sending HTTP request: {:?}", request);
+                    socket
+                        .lock()
+                        .send_slice(request.as_ref())
+                        .expect("http_client: cannot send request");
```
kevinaboos: is there a reason why we choose to panic in this particular case rather than return an error? IOW, is this is an unexpected error here that should never happen?

kevinaboos (changes requested):
The code itself looks great, thanks so much for undergoing the tedious effort of transforming this to the new net/smoltcp API.

I left a few tiny comments, nothing major.

> A couple of points:
> 
> * Currently `net::get_ephemeral_port` doesn't keep track of used ports, but that's no better than it was before. I'll leave that for a future PR when I have a better idea of what information we need to track about ports.

yep, no problem with that. The whole reason I introduced the idea of random port assignment is cuz i kept encountering constant failures with reusing ports -- they wouldn't be reopened properly even after closing them. This was like 4 years ago; not sure if it was a smoltcp error, a QEMU problem, or a problem with how we were using them, but either way it's probably irrelevant now.

> * The `LockedSocket::connect` method is a temporary fix until we introduce our own socket types. After this PR I'll have a look at the API `net` needs to expose to `std` and rework our socket abstractions to better reflect this.

Sounds good, I'm happy to see that the new smoltcp design permits avoiding that kind of lock-based design.

> * I changed the logic in `ota_update_client` to not reconnect on every loop iteration. Previously, it would open a new connection for every file, but only waited on the last connection. Now it only opens one connection and waits on that connection.

Yeah, see the above response about port-related reopening failures, that's likely why that dumb logic was there, i.e., just to get around countless errors.

> * I'm not sure how I can test the changes to `upd`.

Gonna be hard since I haven't used that in years. The way to generate the files for it is to use the `make build_server` target, but it likely doesn't work any more due to so many massive structural changes. Even in the actual crate live swap doesn't work, we should still be able to test whether the actual new crate object files and the update manifest file all get downloaded correctly. That being said, I wouldn't consider proper testing a barrier to accepting this PR, just in this one case.


> Next steps are:
> 
> * Remove `network_manager` and `ethernet_smoltcp_device`
> * Tweak `net` interface to better reflect usage patterns and `std` needs
> * Track used ports

All sounds good to me. Appreciate your efforts here!

kernel/http_client/src/lib.rs
tsoutsman: The storage type can either be a [mutable slice or a vec](https://docs.rs/managed/0.8.0/managed/enum.ManagedSlice.html).

Using a slice wouldn't work because then the lifetime of the socket (and by extension `HttpClient`) would be tied to the lifetime of the array created in the function, and so we wouldn't be able to return it.

The fact that `Vec` is growable isn't important, AFAICT it's only used as a fixed-size owned storage of bytes, so we need to initialise it with a specific size.

kernel/http_client/src/lib.rs
tsoutsman: No, just a remnant of me copy-pasting the previous `send` implementation. Will fix.

kernel/net/src/interface.rs
tsoutsman: Yea that's something I initially did, but switched it because cloning an interface to add a socket didn't really make sense. I'll change it back because this is something I hope to eventually change anyway when I update the `net` API.

kernel/http_client/src/lib.rs
kevinaboos: ah yes, ManagedSlice. Okay thanks, we'll keep it as is then.

kevinaboos (approved):

PR #1010, opened 2023-07-14
Implement `time::ClockSource` for TSC

tsoutsman:
Also removes `libterm`'s direct dependency on `tsc`.

### DIFF
--- a/kernel/tsc/src/lib.rs
+++ b/kernel/tsc/src/lib.rs
@@ -1,78 +1,38 @@
 #![no_std]
 
-#[macro_use] extern crate log;
-extern crate pit_clock_basic;
+use log::info;
+use time::{Instant, Period};
 
-use core::sync::atomic::{AtomicUsize, Ordering};
+pub struct Tsc;
 
+impl time::ClockSource for Tsc {
+    type ClockType = time::Monotonic;
 
-#[derive(Debug)]
-pub struct TscTicks(u128);
-
-impl TscTicks {
-    /// Converts ticks to nanoseconds. 
-    ///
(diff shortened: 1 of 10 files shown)


kernel/tsc/src/lib.rs
```diff
@@ -1,78 +1,38 @@
...
-        TSC_FREQUENCY.store(tsc_freq as usize, Ordering::Release);
-        Ok(tsc_freq)
-    }
+    Some(tsc_period)
+}
+
+#[doc(hidden)]
+pub fn tsc_value() -> u64 {
+    let mut _aux = 0;
+    unsafe { core::arch::x86_64::__rdtscp(&mut _aux) }
```
kevinaboos, line 40: restore safety comment; at some point we will implement a linter pass as a CI check that checks unsafe blocks for such a comment.
```suggestion
    // SAFETY: reading the TSC value is a platform-specific intrinsic
    // that has no side effects or dangerous behavior, and is supported
    // on all modern x86_64 hardware.
    unsafe { core::arch::x86_64::__rdtscp(&mut _aux) }
```

kevinaboos: Theoretically here we could insert a check (using the raw_cpuid crate) to verify that this instruction is supported on the current hardware, but i don't think it's really that important.

kevinaboos (changes requested):
LGTM, thanks!

> Also removes `libterm`'s direct dependency on `tsc`.

Great, always nice to remove the arch-specific parts in crates that don't need them, even though we may likely remove `libterm` at some point.

kevinaboos (approved):

PR #1011, opened 2023-07-20
Remove old networking API

tsoutsman:
Also implicitly updates to `smoltcp v0.10` by removing the patch which brings some nice changes to error handling.

I've made some changes to `virtual_nic` and would like to run `test_ixgbe` but I'm unable to emulate an `ixgbe82599` NIC as [QEMU doesn't support it](https://en.wikibooks.org/wiki/QEMU/Devices/Network).

Would also progress #587.


kernel/net/src/device.rs
```diff
@@ -94,18 +98,18 @@ pub(crate) struct TxToken<'a> {
...
+        F: FnOnce(&mut [u8]) -> R,
     {
-        let mut buf = vec![0; len];
-        let ret = f(&mut buf)?;
-        self.device.send(&buf)?;
-        Ok(ret)
+        // This will only fail if the underlying memory allocation fails.
+        //
+        // TODO: Arguably `TransmitBuffer::new` should panic instead, similar to other
+        // data structures (e.g. `Vec`).
```
tsoutsman: bump

kernel/net/src/device.rs
```diff
@@ -94,18 +103,27 @@ pub(crate) struct TxToken<'a> {
...
-        let mut buf = vec![0; len];
-        let ret = f(&mut buf)?;
-        self.device.send(&buf)?;
-        Ok(ret)
+        match u16::try_from(len) {
+            Ok(len) => {
+                // This will only fail if the underlying memory allocation fails.
+                //
+                // TODO: Arguably `TransmitBuffer::new` should panic instead, similar to other
+                // data structures (e.g. `Vec`).
```
tsoutsman: bump

kernel/device_manager/src/lib.rs
```diff
@@ -11,25 +11,13 @@ use {
...
 
-/// A randomly chosen IP address that must be outside of the DHCP range.
-/// TODO: use DHCP to acquire an IP address.
-#[cfg(target_arch = "x86_64")]
-const DEFAULT_LOCAL_IP: &str = "10.0.2.15/24"; // the default QEMU user-slirp network gives IP addresses of "10.0.2.*"
-
-/// Standard home router address.
-/// TODO: use DHCP to acquire gateway IP
-#[cfg(target_arch = "x86_64")]
-const DEFAULT_GATEWAY_IP: [u8; 4] = [10, 0, 2, 2]; // the default QEMU user-slirp networking gateway IP
```
kevinaboos, line 31: if this stuff is no longer required, that'd be awesome. Does smoltcp now support DHCP by default, or if not, where else are we defining the gateway IP for smoltcp to use?

kernel/net/src/device.rs
kevinaboos: resolved in discussion, panicking here is clearer than within `TransmitBuffer::new()`

kernel/e1000/src/lib.rs
```diff
@@ -272,7 +247,10 @@ impl E1000Nic {
         let deferred_task = deferred_interrupt_tasks::register_interrupt_handler(
             self.interrupt_num,
             e1000_handler,
-            poll_interface,
+            |interface| {
```
kevinaboos: I actually think the old way was clearer and easier to grep for / understand. You can just make that function return OK unconditionally

kernel/net/src/device.rs
```diff
@@ -21,9 +18,14 @@ const STANDARD_MTU: usize = 1500;
 ///
 /// [`register_device`]: crate::register_device
 /// [`NetworkInterface`]: crate::NetworkInterface.
-pub trait NetworkDevice: Send + Sync + Any {
-    fn send(&mut self, buf: &[u8]) -> core::result::Result<(), crate::Error>;
-
+pub trait NetworkDevice: Send + Sync {
```
kevinaboos, line 21: Is Any no longer needed? Is there another means of downcasting this kind of trait object?

old_crates/ethernet_smoltcp_device/Cargo.toml
kevinaboos, line 1: I don't think there's any real reason to keep these crates, you can just delete em.

kevinaboos (changes requested):

kevinaboos:
> I've made some changes to `virtual_nic` and would like to run `test_ixgbe` but I'm unable to emulate an `ixgbe82599` NIC as [QEMU doesn't support it](https://en.wikibooks.org/wiki/QEMU/Devices/Network).

@Ramla-I would you be able to test this IXGBE changeset for us? or do you have access to hardware with an IXGBE card?

Ramla-I:
> > I've made some changes to `virtual_nic` and would like to run `test_ixgbe` but I'm unable to emulate an `ixgbe82599` NIC as [QEMU doesn't support it](https://en.wikibooks.org/wiki/QEMU/Devices/Network).
> 
> @Ramla-I would you be able to test this IXGBE changeset for us? or do you have access to hardware with an IXGBE card?

Yes I do. I can test it. I just need to check that the ixgbe rx/tx is still working when using the virtual nic interface, right?

kevinaboos:
> Yes I do. I can test it. I just need to check that the ixgbe rx/tx is still working when using the virtual nic interface, right?

Great, thanks! Yes, that's correct, just whatever networking tests you normally do. `ping`, at a bare minimum.

kernel/net/src/device.rs
tsoutsman: No, I'm pretty sure it was a leftover from a previous version of the net API. I just forgot to remove the bound previously.

kernel/device_manager/src/lib.rs
tsoutsman: > Does smoltcp now support DHCP by default

Yes, but it's not as simple as creating a DHCP socket and polling the interface as that returns [`Event::Deconfigured`](https://docs.rs/smoltcp/latest/smoltcp/socket/dhcpv4/enum.Event.html).

I'll look into it more but I think it's best left for a future PR.

> where else are we defining the gateway IP for smoltcp to use?

`net/src/lib.rs`

kevinaboos: Ok great, I missed that those IP addrs were already moved into `net` in a previous changeset; I was looking for them as part of this PR.

kevinaboos (approved):
Gonna go ahead and approve/merge this PR; if `ixgbe` testing turns up an error we can address that separately.

Regarding the comment about #587, I don't think that's too likely to actually happen, but it's noted that this removes one of those patches, which is nice for myriad other reasons.

Also see my discussion starter about the design of the `NetworkDevice` trait on discord #networking.

PR #1012, opened 2023-07-21
Implement more ergonomic CPU-local storage

tsoutsman:
MVP of a `cpu_local` attribute.

Limitations that will be addressed in a future PR:
- Only integer types supported
- CLS section offsets must be manually specified

~Draft PR because the macro needs to be implemented for `aarch64`.~


kernel/cls/cls_macros/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+    // if !args.is_empty() {
+    //     Diagnostic::spanned(
+    //         Span::call_site(),
+    //         Level::Error,
+    //         "malformed `cpu_local` attribute input",
+    //     )
+    //     .help("must be of the form `#[cpu_local]`")
+    //     .emit();
+    //     return TokenStream::new();
+    // }
```
tsoutsman, line 61: Left this in for when we remove manual offsets.

Cargo.toml
```diff
@@ -12,6 +12,7 @@ resolver = "2"
 members = [
 	"theseus_features",  ## Must be included to realize global Cargo features across Theseus.
 	"kernel/[!.]*/",
+	"kernel/cls/cls_macros",
```
kevinaboos: why is this needed? Looks like `cls_macros` is an unconditional dependency of the `kernel/cls` crate, so any time someone depends on `cls` it will get pulled in there.

If it's needed here, should it also go in `default-members` below?

kernel/cls/cls_macros/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        #[cfg(target_arch = "aarch64")]
+        impl #struct_name {
+            #[inline]
+            pub fn load(&self) -> #ty {
+                let ret;
+                unsafe {
+                    ::core::arch::asm!(
+                        "1:",
```
kevinaboos: i don't recall where i saw this, but some aarch64 asm guide somewhere stated that we shouldn't use the integer "1" as a branch label, and that you're supposed to start at "2". I think it was due to some legacy asm parser not being able to handle `1` as a label, but other than that I don't know exactly why....  but i guess it's better to be safe than sorry.

kernel/cls/cls_macros/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+                return TokenStream::new();
+            }
+        };
+
+    let x64_width_modifier = format!("{x64_asm_width} ptr ");
+    let x64_cls_location = format!("gs:[{offset}]");
+
+    quote! {
+        #(#attributes)*
+        #[thread_local]
```
kevinaboos, line 109: i see this is now a ZST, so it shouldn't affect the TLS offsets calculated by the static linker, but does it still appear as a TLS symbol in the outputted binary? If so, does adding the `link_section` attribute back in avoid that?

Ideally we don't want these CLS vars to end up as TLS sections in the binary.

kernel/cls/cls_macros/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+        // #[link_section = ".cls"]
+        #visibility static #name: #struct_name = #struct_name;
+
+        #[repr(transparent)]
+        #[non_exhaustive]
+        #visibility struct #struct_name;
+
+        #[cfg(target_arch = "x86_64")]
+        impl #struct_name {
```
kevinaboos: is the API for x86_64 and aarch64 intentionally different? e.g., there's no `add` or `sub` impl'd for x86.

One way to avoid this is to restructure the code, with a single `impl` block for all functions and then arch-specific blocks inside each function. I don't necessarily think that's mandatory here, but it might be nice to ensure a consistent API across both archs.

kernel/cls/cls_macros/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+
+            pub fn sub(&self, operand: #ty) {
+                unsafe {
+                    ::core::arch::asm!(
+                        "1:",
+                        // Load value.
+                        // TODO: Can we add offset and load in one instruction?
+                        "mrs {tp_1}, tpidr_el1",
+                        concat!("add {ptr}, {tp_1}, ", stringify!(#offset)),
+                        concat!("ldxr", #aarch64_instr_width, " {value", #aarch64_reg_modifier,"}, [{ptr}]"),
```
kevinaboos: > // TODO: Can we add offset and load in one instruction?

Yes, but I don't think it's doable with the LDXR/STXR pair. The question is do we actually _need_ to use that pair, or can we get by with regular load/store? In my view, it shouldn't be strictly necessary to use the synchronization primitive-focused LDXR/STXR pairing since we don't really need the implied/silent locking behavior for accessing CPU-local variables.
Anyway, if we don't use that, you can do so with `LDR DEST_REG, [SRC_REG, IMM_OFFSET]` as shown in this link:

https://developer.arm.com/documentation/den0024/a/The-A64-instruction-set/Memory-access-instructions/Specifying-the-address-for-a-Load-or-Store-instruction?lang=en

kernel/cls/cls_macros/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+                        "mrs {tp_2}, tpidr_el1",
+                        "cmp {tp_1}, {tp_2}",
+                        "b.ne 1b",
+
+                        // Compute and store value.
+                        "sub {value}, {value}, {operand}",
+                        concat!("stxr", #aarch64_instr_width, " {cond:w}, {value", #aarch64_reg_modifier,"}, [{ptr}]"),
+
+                        // Make sure task wasn't migrated between ldxr and stxr.
+                        "cbnz {cond}, 1b",
```
kevinaboos: (this is a question for my understanding, not a question of whether you're correct)

how does this work? I expected to see another set of instructions here (right before this check) that load TPIDR_EL1 again to check for a migration to a different CPU, much like L225-227. Why don't we have to do that again?

kernel/cls/cls_macros/src/lib.rs
kevinaboos: Another option would be to just make `add` and `sub` private for now, at least until we choose to implement them for x86.

kevinaboos (changes requested):
Generally looks good, i left a few questions

kernel/cls/cls_macros/src/lib.rs
tsoutsman: Yep. https://doc.rust-lang.org/nightly/rust-by-example/unsafe/asm.html#labels

kernel/cls/cls_macros/src/lib.rs
tsoutsman: We do need `LDXR`/`STXR`. Otherwise if the task was migrated between `b.ne 2b` and `stxr`, we wouldn't know.

> implied/silent locking

It's far from locking. `LDXR` and `STXR` are used for relaxed accesses of atomic variables.

kernel/cls/cls_macros/src/lib.rs
tsoutsman: So the `stxr` line works out to something like:
```armasm
stxr {cond:w} {value} [{ptr}]
```
which tries to move `value` into `[ptr]`, and if it fails (i.e. task was migrated), it sets `cond` to `1`.

The `cbnz` checks if `cond` is non-zero (i.e. the task was migrated), and if so, restarts the process.

kernel/cls/cls_macros/src/lib.rs
tsoutsman: No. This is the list of symbols in the `preemption` elf file
```
Symbol table '.symtab' contains 24 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND
     1: 0000000000000000     0 SECTION LOCAL  DEFAULT    2 .text
     2: 0000000000000000     0 SECTION LOCAL  DEFAULT    4 .rodata
     3: 0000000000000000     0 SECTION LOCAL  DEFAULT    8 .gcc_except_table
     4: 0000000000000000    12 FUNC    LOCAL  DEFAULT    2 <cpu::CpuId as core::fmt::Display>::fmt
     5: 0000000000000010    12 FUNC    LOCAL  DEFAULT    2 core::ptr::drop_in_place<preemption::PreemptionGuard>
     6: 0000000000000020    24 FUNC    LOCAL  DEFAULT    2 core::ptr::drop_in_place<sync::rw_lock::RwLockWriteGuard<apic::LocalApic,sync_irq::DisableIrq>>
     7: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND cpu::arch::current_cpu
     8: 0000000000000240   219 FUNC    GLOBAL DEFAULT    2 preemption::hold_preemption_no_timer_disable
     9: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND core::panicking::panic_cannot_unwind
    10: 0000000000000560    14 FUNC    GLOBAL DEFAULT    2 preemption::preemption_enabled
    11: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND apic::LocalApic::enable_lvt_timer
    12: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND apic::get_my_apic
    13: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _Unwind_Resume
    14: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND core::panicking::panic_fmt
    15: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND <irq_safety::held_interrupts::HeldInterrupts as core::ops::drop::Drop>::drop
    16: 0000000000000330     3 FUNC    GLOBAL DEFAULT    2 preemption::PreemptionGuard::cpu_id
    17: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND core::option::expect_failed
    18: 0000000000000320     5 FUNC    GLOBAL DEFAULT    2 preemption::PreemptionGuard::preemption_was_enabled
    19: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND core::fmt::num::imp::<impl core::fmt::Display for u32>::fmt
    20: 0000000000000040   511 FUNC    GLOBAL DEFAULT    2 preemption::hold_preemption
    21: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND irq_safety::held_interrupts::hold_interrupts
    22: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND rust_eh_personality
    23: 0000000000000340   543 FUNC    GLOBAL DEFAULT    2 <preemption::PreemptionGuard as core::ops::drop::Drop>::drop
```

kernel/cls/cls_macros/src/lib.rs
kevinaboos: ah ok, yeah it was for x86 but it wouldn't hurt to apply that here, as we may end up reusing some of that code flow for a future more complex x86 primitive operation.

kernel/cls/cls_macros/src/lib.rs
kevinaboos: Awesome, thanks for confirming. I would presume the symbol is completely removed since it's a ZST, and only appears by means of its functions that get inlined into the callsites

kernel/cls/cls_macros/src/lib.rs
kevinaboos: Ok thanks for confirming that the LD**X**R variants are needed. Continuing to read the [ARMv8 docs about that](https://developer.arm.com/documentation/den0024/a/Multi-core-processors/Multi-processing-systems/Synchronization?lang=en) helped clarify that for me, specifically this:
> `LDXR` performs a load of memory, but also tags the Physical Address to be monitored for exclusive access by that core. `STXR` performs a conditional store to memory, succeeding only if the target location is tagged as being monitored for exclusive access by that core. This instruction returns non-zero in the general-purpose register `Ws` if the store does not succeed, and a value of 0 if the store is successful. In the assembler syntax, it is always specified as a `W` register, that is, not an `X` register. In addition, the `STXR` clears the exclusive tag.

My usage of the term locking comes directly from their docs, which is to say that it's an implicit form of locking invisible to systems software that is indeed (as you said) used for relaxed atomic access.
https://developer.arm.com/documentation/den0024/a/The-A64-instruction-set/Memory-access-instructions/Synchronization-primitives

kernel/cls/cls_macros/src/lib.rs
kevinaboos: thanks, yeah reading about the semantics of `stxr` answered my question.

kernel/preemption/src/lib.rs
```diff
@@ -3,30 +3,14 @@
...
-    }
-}
-// SAFETY: The `PreemptionCount` type corresponds to a field in `PerCpuData`
-//         with the offset specified by `PerCpuField::PreemptionCount`.
-unsafe impl CpuLocalField for PreemptionCount {
-    const FIELD: PerCpuField = PerCpuField::PreemptionCount;
-}
+// NOTE: This offset must be kept in sync with `cpu_local::PerCpuField`.
+#[cls::cpu_local(12)]
+static PREEMPTION_COUNT: u8 = 0;
```
kevinaboos, line 14: this uses a `u8` here but in the `PerCpuData` struct you use a `u16`. I don't particularly mind which one we pick, but let's keep it consistent.

kernel/preemption/src/lib.rs
```diff
@@ -58,25 +42,19 @@ pub fn hold_preemption_no_timer_disable() -> PreemptionGuard {
 /// will be disabled upon a transition from preemption being enabled to being disabled.
 fn hold_preemption_internal<const DISABLE_TIMER: bool>() -> PreemptionGuard {
     let cpu_id = cpu::current_cpu();
+
     // Create an initial preemption guard such that we can call `CpuLocal::with_preempt()`,
     // but don't allow it to be dropped until we actually disable preemption below.
-    let mut guard_placeholder = NoDrop::new(PreemptionGuard {
+    PREEMPTION_COUNT.increment();
+    // Not a race condition because preemption is disabled.
+    let prev_val = PREEMPTION_COUNT.load() - 1;
```
kevinaboos: hmm, that's true in the simple case, but that only holds if we do not ever mess with preemption in an interrupt handler.

The increment and subsequent load are still not atomic, unlike the prior approach that used actual atomics (`fetch_add`). It's possible for an interupt to occur after the `increment` but before the `load`, and if that interrupt handler obtains or drops a `PreemptionGuard` and then returns, then the `prev_val` will be incorrect.

I assume there is no such single instruction that can increment a register _and_ return its prior value.... so without that, we have to disable interrupts around this sequence of two operations, right?

kernel/preemption/src/lib.rs
```diff
@@ -146,10 +124,10 @@ impl Drop for PreemptionGuard {
             This indicates an unexpected task migration across CPUs."
         );
 
-        let prev_val = PREEMPTION_COUNT.with_preempt(
-            self,
-            |count| count.0.fetch_sub(1, Ordering::Relaxed)
-        );
+        // Not a race condition because preemption is disabled.
+        let prev_val = PREEMPTION_COUNT.load();
+        PREEMPTION_COUNT.decrement();
```
kevinaboos: same question as above

kevinaboos (changes requested):
Left one major question about the atomicity of increment+load and load+decrement.

Also, can you add (or restore) typical things like crate-level docs, authorship/description fields in the toml, etc? thanks.

Everything else looks good.

kernel/preemption/src/lib.rs
kevinaboos: Note: resolved via discord discussion, we'll use an atomic instruction `xadd` that returns the prior value instead of `inc`

kernel/preemption/src/lib.rs
```diff
@@ -3,30 +3,14 @@
 //! Supports enabling and disabling preemption for the purpose of 
 //! safe task state management, e.g., through preemption-safe locks.
 
-use core::sync::atomic::{AtomicU8, Ordering};
-use cpu::CpuId;
-use crate::cpu_local::{CpuLocal, CpuLocalField, PerCpuField};
-use no_drop::NoDrop;
+#![no_std]
+#![feature(negative_impls, thread_local)]
```
kevinaboos, line 7: note for next PR: i think we can avoid the need for every user of `cls_macro` to declare `#![feature(thread_local)]` by using the `#[allow_internal_unstable(...)]` attribute in the `cls_macro` crate, either on the proc macro definition or as a crate-level attribute. I recall using that attribute in the `thread_local_macro` crate, based on `std`'s `thread_local!()` macro trickery.

This isn't a blocker to merging/approving this PR, but we should try to improve that rough edge, especially once we have dozens of CPU local vars everywhere.

kevinaboos: (i just realized this is intended for regular macros, not procedural macros, but you should try it anyway)

kevinaboos (approved):
left one comment as a suggestion to try out in #1015

PR #1016, opened 2023-07-24
Huge pages support for x86

NIMogen:
PR as requested. Higher level page table entries (such as p2) are always marked as used for some reason. 
Just a draft as this is a WIP and not actually ready yet.


kevinaboos:
As discussed on discord, I can reproduce the error, which is caused by not using a range of pages and frames that are properly aligned to a 2MiB boundary.

kernel/pte_flags/src/pte_flags_x86_64.rs
```diff
@@ -223,6 +223,12 @@ impl PteFlagsX86_64 {
         self.set(Self::DIRTY, enable);
         self
     }
+    
+    #[must_use]
```
kevinaboos, line 232: please add docs in the same style of the functions above it

kernel/pte_flags/src/pte_flags_x86_64.rs
```diff
@@ -271,10 +277,11 @@ impl PteFlagsX86_64 {
...
     ///     only the lowest-level P1 entry should.
     /// * Clears the `EXCLUSIVE` bit.
-    ///   * Currently, we do not use the `EXCLUSIVE` bit for P4, P3, or P2 entries,
+    ///   * Generally, we do not use the `EXCLUSIVE` bit for P4, P3, or P2 entries,
     ///     because another page table frame may re-use it (create another alias to it)
-    ///     without our page table implementation knowing about it.
-    ///   * Only P1-level PTEs can map a frame exclusively.
+    ///     without our page table implementation knowing about it. However, the `EXCLUSIVE` bit
+    ///     will be set on P3 and P2 entries for huge pages (P2 for 2MiB pages, and P3 for 1GiB pages).
+    ///   * Only P1-level PTEs, and in the case of huge pages, P2 and P3-level PTEs, can map a frame exclusively.
```
kevinaboos: Good correction here. I have a further suggestion to better generalize when we use the `EXCLUSIVE` flag: we use it only for leaf nodes in the page tables, i.e., those that directly map a frame. That occurs only for:
1. all P1-level page table frames, since they can only directly map 4K frames.
2. only huge P2-level or P3-level page tables frames, which directly map 1G and 2M frames.

Kindly update this comment block to include those details.

kernel/page_allocator/src/lib.rs
```diff
@@ -1,21 +1,3 @@
-//! Provides an allocator for virtual memory pages.
```
kevinaboos: please restore this comment block.

Also, there are thousands of whitespace changes; please remove those too. I don't mind them, but they should be made in a separate PR in order to avoid polluting this PR with many unrelated changed lines.

kernel/page_allocator/src/lib.rs
```diff
@@ -139,6 +121,222 @@ pub fn init(end_vaddr_of_low_designated_region: VirtualAddress) -> Result<(), &'
...
+
+// These methods mostly destructure the enum in order to call internal methods
+impl PageRangeSized {
+	/// Get the size of the pages for the contained PageRange
+	pub fn page_size(&self) -> usize {
+		match self {
+			PageRangeSized::Normal4KiB(pr) => {
+				pr.start().page_size()
+			}
+			// Pages of other sizes should not be used
```
kevinaboos: i don't understand this comment

kernel/page_allocator/src/lib.rs
```diff
@@ -139,6 +121,222 @@ pub fn init(end_vaddr_of_low_designated_region: VirtualAddress) -> Result<(), &'
...
+/// An enum used to wrap the generic PageRange variants corresponding to different page sizes.
+/// Additional methods corresponding to PageRange methods are provided in order to destructure the enum variants.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PageRangeSized {
+	Normal4KiB(PageRange),
+	Huge2MiB(PageRange<Page2MiB>),
+	Huge1GiB(PageRange<Page1GiB>),
+}
+
+// These methods mostly destructure the enum in order to call internal methods
```
kevinaboos: i don't understand the need for the following methods on this enum:
* `range`
* `into_range_4k`
* `range_2mb`
* `range_1gb`

None of them seem to be necessary?

kernel/page_allocator/src/lib.rs
```diff
@@ -139,6 +121,222 @@ pub fn init(end_vaddr_of_low_designated_region: VirtualAddress) -> Result<(), &'
...
+	pub fn end(&self) -> &Page {
+		match self {
+			PageRangeSized::Normal4KiB(pr) => {
+				pr.end()
+			}
+			_ => {
+				panic!("Attempt to get the end of a huge page range as a 4KiB page.");
+			}
+		}
+	}
```
kevinaboos: these are just for temporary compatibility, right? If so, let's indicate that as such in the comments, because otherwise I don't think these should necessarily exist since they invite panicking behavior.

At the very least, the comments need to indicate that they are only to be used to 4K pages,  and will panic otherwise.

kernel/page_allocator/src/lib.rs
```diff
@@ -94,40 +76,40 @@ pub fn init(end_vaddr_of_low_designated_region: VirtualAddress) -> Result<(), &'
 		// The first region contains all pages from address zero to the end of the low designated region,
 		// which is generally reserved for identity-mapped bootloader stuff and base kernel image sections.
 		Some(Chunk {
-			pages: PageRange::new(
+			pages: PageRangeSized::Normal4KiB(PageRange::new(
```
kevinaboos: the page allocator deals exclusively with 4KiB normal pages, so there is no need to change its `Chunk` type at all, since it will never store huge pages. So you can remove this and the other modifications in the 40 lines below.

kernel/nano_core/src/lib.rs
```diff
@@ -115,6 +115,8 @@ where
     }
 
     let rsdp_address = boot_info.rsdp();
+
+
```
kevinaboos: please remove unnecessary changes to this file

kernel/memory_structs/src/lib.rs
```diff
@@ -407,44 +518,22 @@ macro_rules! implement_page_frame_range {
...
-                    and `address_at_offset(0x2000)` would return `None`."]
-                pub const fn address_at_offset(&self, offset: usize) -> Option<$address> {
-                    if offset < self.size_in_bytes() {
-                        Some($address::new_canonical(self.start_address().value() + offset))
-                    }
-                    else {
-                        None
-                    }
+                pub fn into_1gb_range(&self) -> $TypeName<Page1GiB> {
+                    $TypeName::<Page1GiB>(RangeInclusive::new(self.start().as_1gb(), self.end().as_1gb()))
```
kevinaboos: can you move all of these new functions to the end of this impl block, in order to avoid a massive set of mistakenly-changed lines? it'll make the  diff much cleaner.

kevinaboos: Also, these functions aren't correct and/or don't quite make sense. 

It makes sense to infallibly convert from a huger page to a smaller/normal page, but not the other way around:
* a 1GiB huge page can always be converted into a range of 512 2MiB huge pages, or a range of 512*512 4KiB normal pages
* a 2MiB huge page can always be converted into a range of 512 4KiB normal pages

However, converting a range of 4KiB normal pages into a range of 2MiB or 1GiB huge pages is _not_ infallible -- they need to have the correct alignment and size in pages that is a correct multiple of the desired huge page size. Same goes for converting a range of 2MiB huge pages into a range of 1GiB huge pages.
Thus, the conversion functions for those must return a `Result` to indicate the possibility of failure.

You could also use `Into` for the infallible conversions and `TryInto` for the fallible ones; that'd be the most canonically Rusty approach.

kernel/memory_structs/src/lib.rs
```diff
@@ -471,14 +560,84 @@ macro_rules! implement_page_frame_range {
...
+
+                #[doc = "Creates a new range of [`" $chunk "`]s that spans from `start` to `end`, both inclusive bounds."]
+                pub const fn new_2mb(start: $chunk<Page2MiB>, end: $chunk<Page2MiB>) -> $TypeName<Page2MiB> {
+                    $TypeName::<Page2MiB>(RangeInclusive::new(start, end))
+                }
+
+                #[doc = "Creates a new range of [`" $chunk "`]s that spans from `start` to `end`, both inclusive bounds."]
+                pub const fn new_1gb(start: $chunk<Page1GiB>, end: $chunk<Page1GiB>) -> $TypeName<Page1GiB> {
+                    $TypeName::<Page1GiB>(RangeInclusive::new(start, end))
+                }
```
kevinaboos: these two functions aren't needed, that's the point of generics. Instead, you just use `new` as normal:
```rust
let range_2mib = PageRange::<PageSize2MiB>::new(start, end):
let range_1gib = PageRange::<PageSize1GiB>::new(start, end):
```

kevinaboos: alternatively you can do this:
```rust
let range_2mib: PageRange<PageSize2MiB> = PageRange::new(start, end):
let range_1gib: PageRange<PageSize1GiB> = PageRange::new(start, end):
```

nothing special, that's just how Rust type inference rules work.

Makefile
```diff
@@ -882,7 +882,7 @@ ifeq ($(graphic), no)
 endif
 
 ## Set the amount of system memory (RAM) provided to the QEMU guest OS
-QEMU_MEMORY ?= 512M
+QEMU_MEMORY ?= 3G
```
kevinaboos: we can't merge this change into the mainline, so please remove it.

But, what you can and should do is add a statement to the test_huge_pages apps that, upon an error occurring, prints a suggestions error message about needing to have large amounts of physical memory to actually test this, and that they can re-run Theseus in QEMU with something like:
```sh
make orun QEMU_MEMORY=2G
```
or the like

kernel/frame_allocator/src/lib.rs
kevinaboos, line 1: please undo changes to this file

kernel/memory/src/paging/mapper.rs
```diff
@@ -218,18 +219,55 @@ impl Mapper {
...
+                    if !p1[page.p1_index()].is_unused() {
+                        error!("map_allocated_pages_to(): page {:#X} -> frame {:#X}, page was already in use!", page.start_address(), frame.start_address());
+                        return Err("map_allocated_pages_to(): page was already in use");
+                    } 
+
+                    p1[page.p1_index()].set_entry(frame, actual_flags);
+                }
+            }
+            PAGE_2MB_SIZE => {
+                for (page, frame) in pages.range_2mb().clone().into_iter().step_by(512).zip(frames.borrow().into_iter().step_by(512)) {
```
kevinaboos: this is odd, and doesn't really make sense. It looks like you're trying to convert the range of 2MiB huge pages into a bunch of 4KiB normal pages, and _then_ do the mapping procedure for every 512th normal page.

What we should do instead is just iterate over 2MiB pages directly -- that's the whole point of introducing the `PageSize` trait into the `PageRange` and `Page` types. Then you won't need any of this `step_by()` stuff (which is also very, very slow).

kernel/memory/src/paging/mapper.rs
```diff
@@ -218,18 +219,55 @@ impl Mapper {
...
+                        error!("map_allocated_pages_to(): page {:#X} -> frame {:#X}, page was already in use!", page.start_address(), frame.start_address());
+                        return Err("map_allocated_pages_to(): page was already in use");
+                    }
 
-            p1[page.p1_index()].set_entry(frame, actual_flags);
+                    p2[page.p2_index()].set_entry(frame, actual_flags);
+                }
+            }
+            PAGE_1GB_SIZE => {
+                for (page, frame) in pages.range_1gb().clone().into_iter().step_by(512*512).zip(frames.borrow().into_iter().step_by(512*512)) {
```
kevinaboos: same comment here as above

kernel/memory/src/paging/mapper.rs
```diff
@@ -281,24 +319,70 @@ impl Mapper {
...
+                        error!("map_allocated_pages(): page {:#X} -> frame {:#X}, page was already in use!",
+                            page.start_address(), af.start_address()
+                        );
+                        return Err("map_allocated_pages(): page was already in use");
+                    } 
+        
+                    p3[page.p3_index()].set_entry(af.as_allocated_frame(), actual_flags);
+                    core::mem::forget(af); // we currently forget frames allocated here since we don't yet have a way to track them.
+                }
+            }
```
kevinaboos: since these aren't yet implementable due to lack of support in the frame allocator, we shouldn't allow it to proceed. Instead, do this:

```suggestion
                todo!("Mapping 2MiB huge pages to randomly-allocated huge frames is not yet supported")
            }
            PAGE_1GB_SIZE => {
                todo!("Mapping 1GiB huge pages to randomly-allocated huge frames is not yet supported")
            }
```

kernel/memory/src/paging/mapper.rs
```diff
@@ -281,24 +319,70 @@ impl Mapper {
...
-            if !p1[page.p1_index()].is_unused() {
-                error!("map_allocated_pages(): page {:#X} -> frame {:#X}, page was already in use!",
-                    page.start_address(), af.start_address()
-                );
-                return Err("map_allocated_pages(): page was already in use");
-            } 
-
-            p1[page.p1_index()].set_entry(af.as_allocated_frame(), actual_flags);
-            core::mem::forget(af); // we currently forget frames allocated here since we don't yet have a way to track them.
+        match pages.page_size() {
```
kevinaboos, line 402: this should return an enum variant rather than an integer. That allows you to avoid the need for the trailing match statement (with the `return Err("map_allocated_pages(): page has incorrect size");`), as it would be known unreachable.

kernel/memory/src/paging/mapper.rs
```diff
@@ -537,18 +621,45 @@ impl MappedPages {
...
 
-        for page in self.pages.range().clone() {
-            let p1 = active_table_mapper.p4_mut()
-                .next_table_mut(page.p4_index())
-                .and_then(|p3| p3.next_table_mut(page.p3_index()))
-                .and_then(|p2| p2.next_table_mut(page.p2_index()))
-                .ok_or("mapping code does not support huge pages")?;
-            
-            p1[page.p1_index()].set_flags(new_flags);
+        match self.pages.page_size() {
```
kevinaboos, line 745: see above statement about returning an enum here.

kernel/memory/src/paging/mapper.rs
```diff
@@ -537,18 +621,45 @@ impl MappedPages {
...
+            PAGE_4KB_SIZE => {
+                for page in self.pages.range().clone() {
+                    let p1 = active_table_mapper.p4_mut()
+                        .next_table_mut(page.p4_index())
+                        .and_then(|p3| p3.next_table_mut(page.p3_index()))
+                        .and_then(|p2| p2.next_table_mut(page.p2_index()))
+                        .ok_or("remap: error getting p1 entry for 4kb page")?;
```
kevinaboos: please add "BUG" to the beginning of the three error message variants here, as that should never happen. thanks

kernel/memory/src/paging/mapper.rs
```diff
@@ -608,83 +719,198 @@ impl MappedPages {
...
+        // The different branches mostly have the same logic, differing 
+        // only in what level is unmapped and what unmapping function is used.
+        match self.pages.page_size() {
+            PAGE_4KB_SIZE => {
+                for page in self.pages.range().clone() {
+                    let p1 = active_table_mapper.p4_mut()
+                        .next_table_mut(page.p4_index())
+                        .and_then(|p3| p3.next_table_mut(page.p3_index()))
+                        .and_then(|p2| p2.next_table_mut(page.p2_index()))
+                        .ok_or("mapping code does not support huge pages")?;
```
kevinaboos: as with above, please update this error message to reflect a BUG.

kernel/page_allocator/src/lib.rs
```diff
@@ -810,46 +1050,66 @@ pub fn allocate_pages_by_bytes_deferred(
...
+pub fn allocate_1gb_pages(num_pages: usize) -> Option<AllocatedPages> {
+	let num_pages = num_pages * 512 * 512;
+	if let Some(mut ap) = allocate_pages_deferred(AllocationRequest::Any, num_pages)
+		.map(|(ap, _action)| ap)
+		.ok() {
+			ap.to_1gb_allocated_pages();
+			return Some(ap);
+		}
+	None
+}
```
kevinaboos: These functions aren't correct due to lack of checking alignment, and also aren't necessary. Kindly remove them. 

Instead, what we should do (for now, until we have a proper `AllocationRequest::AlignedAt` variant) is to implement these as a fallible conversion method on the `AllocatedPages` type. This is the easiest way, as all we'll need to do is:
1. call a regular `allocate_pages` function to obtain a sufficiently large and properly aligned `AllocatedPages` consisting of 4KiB normal pages
2. call the conversion method on the returned pages, which will check the alignment conditions for both the start and end page, and adjust the internal `PageRangeSized` object to be the correct variant.

To give you a more specific example, here's some pseudo-ish code for what i mean:
```rust
impl AllocatedPages {
    ...

    pub fn try_into_2mib_huge_pages(self) -> Result<Self, ErrorOfSomeKind> {
        match self.pages {
            PageRangeSized::Normal4KiB(range) => {
                if range.start_address().is_aligned_to(PAGE_SIZE * 512) 
                   && range.end().start_address().is_aligned_to(PAGE_SIZE * 512)
                {
                    Ok(AllocatedPages {
                        pages: PageRangeSized::Huge2MiB(PageRange::new(
                            Page::containing_address(range.start_address()),
                            Page::containing_address(range.end_address()), // need to add this function to PageRange, or you can calculate it here
                        ),
                    })
                }
            }
            PageRangeSized::Huge2MiB(range) => Ok(range),
            PageRangeSized::Huge1GiB(range) => {
                // todo: infallible conversion from 1G to 2M huge pages
                ...
            }
        }
    }
}
```

kernel/memory/src/paging/mapper.rs
```diff
@@ -537,18 +621,45 @@ impl MappedPages {
...
+                        .and_then(|p2| p2.next_table_mut(page.p2_index()))
+                        .ok_or("remap: error getting p1 entry for 4kb page")?;
+
+                    p1[page.p1_index()].set_flags(new_flags);
+
+                    tlb_flush_virt_addr(page.start_address());
+                }
+            }
+            PAGE_2MB_SIZE => {
+                for page in self.pages.range_2mb().clone().into_iter().step_by(512) {
```
kevinaboos: the same comment from above applies to all of the blocks where you iterate over the huge page ranges -- there shouldn't be any need to use `step_by` or a 2MiB-specific function like `range_2mb()`

applications/test_1gb_huge_pages/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "test_1gb_huge_pages"
+version = "0.1.0"
+authors = ["Noah Mogensen <mogensenn@sou.edu>"]
```
kevinaboos: for this and the other crate, please add a description field too.

Also, can you use a more permanent email address such in case a future contributor needs to get in contact with you?

kevinaboos (changes requested):
This is a great start, thanks Noah! Now that the major functionality is working, I left lots of comments about style, correctness, using the type system & generics to their fullest effect, etc. 

Please feel free to ask follow-up questions if you're confused or disagree with my comments, either here or on discord. 

Also, I know this is a volunteer project for you, so if you don't have time or the desire to make some or all of these changes on your own, please let me know what you don't want to tackle, and I can definitely take over those changes (or a subset of them) myself.

kernel/page_allocator/src/lib.rs
NIMogen: The range() and range_*() are present on `PageRangeSized` because the enum needs to be destructured before you can call methods on the internal `PageRange,` which could be one of several types. I didn't really see a way to get around adding these extra methods for this reason. Although dynamic dispatch could be one possible alternative, I figured that wasn't appropriate here.


`into_range_4k()` is for `Drop for AllocatedPages,` where a huge `PageRange` will be converted to a 4kb `PageRange` before anything. This is so that the allocator doesn't need any new logic for dropping huge pages.
I suppose that you could get the `PageRange` from `PageRangeSized` and then convert to a 4kb page range, but that means you would have to check page size and call the correct range method, then convert to a 4kb range and make a new `PageRangeSized`,  whereas with `into_range_4k()` you just convert directly if the `PageRange` is huge. Either method works though, and I don't know if either would be better.

kernel/memory/src/paging/mapper.rs
NIMogen: Yeah I'm not a fan of using `into_iter` and `step_by` here.

But the reason I went with this is because the `PageRange` is ultimately just a large `RangeInclusive` of 4KiB pages. If you had 2 huge pages, the range that's looped over would cover all the pages in between the 2 huge pages that we actually want to map.

As an alternative to this, would it work to change `Step` implementation for `Page` and `Frame` in order to adapt to the different sizes?

PR #1019, opened 2023-07-25
Port of the `pci` crate

NathanRoyer:
This brings the following changes:
- `pci::scan_pci` maps the pci config space on aarch64, in addition to what it used to do.
- `pci::PciLocation` was modified so that it uses the mapping on aarch64
- `device_manager` calls it (via pci_device_iter) on both platforms
- I tried converting the `const` offsets to an enum but the capability "linked-list" from `find_pci_capability` needs the offset to be a raw u8, and I might also need custom offsets for USB support so I dropped this.
- `QEMU_FLAGS` starts the virt machine with new options: `highmem=on` & `highmem-mmio=on` because without them, the PCI configuration space lies in "low memory" but its base address points to a bunch of zeros for some reason.

Note: while `pci` currently uses port-io on x86 and mmio on aarch64, x86 may also support memory-based PCI configuration in the future; port-io is the legacy way to access the config space.


kernel/pci/src/lib.rs
```diff
@@ -65,18 +79,32 @@ const MAX_SLOTS_PER_BUS: u8 = 32;
 const MAX_FUNCTIONS_PER_SLOT: u8 = 8;
 
 /// Addresses/offsets into the PCI configuration space should clear the least-significant 2 bits.
-const PCI_CONFIG_ADDRESS_OFFSET_MASK: u8 = 0xFC; 
+const PCI_CONFIG_ADDRESS_OFFSET_MASK: u32 = !0x2;
```
kevinaboos: this value is 0xFFFFFFFD, which doesn't match up with the comment. This mask value only clears bit 1, not bit 0 and bit 1. What is the desired behavior; does it differ on aarch64/MMIO?

kernel/pci/src/lib.rs
```diff
@@ -129,7 +158,20 @@ pub struct PciBus {
...
+        if config_space.is_none() {
+            let config = BOARD_CONFIG.pci_ecam;
+            let flags = memory::PteFlags::new().valid(true).writable(true);
+            let mapped = memory::map_frame_range(config.base_address, config.size_bytes, flags)?;
+            match mapped.into_borrowed_slice_mut(0, config.size_bytes) {
+                Ok(bsm) => config_space.insert(bsm),
+                Err((_, msg)) => return Err(msg),
+            };
+        }
+    }
```
kevinaboos: we're using `try_call_once()` for the PCI Bus, shouldn't we also use it for this? Is there any other reason to re-initialize or otherwise modify the `PCI_CONFIG_SPACE` variable more than once?

kernel/pci/src/lib.rs
```diff
@@ -221,74 +263,98 @@ impl PciLocation {
...
-        ((self.bus  as u32) << 16) | 
-        ((self.slot as u32) << 11) | 
-        ((self.func as u32) <<  8) | 
-        ((offset as u32) & (PCI_CONFIG_ADDRESS_OFFSET_MASK as u32)) | 
-        0x8000_0000
+        BASE_OFFSET |
+        ((self.bus  as u32) << 16) |
+        ((self.slot as u32) << 11) |
+        ((self.func as u32) <<  8) |
+        (offset as u32)
```
kevinaboos: here you're no longer bit-masking out the least two significant bits. Is that correct? It doesn't seem right.

kernel/pci/src/lib.rs
```diff
@@ -221,74 +263,98 @@ impl PciLocation {
...
+            PCI_CONFIG_DATA_PORT.lock().read() >> shift
+        }
+
+        #[cfg(target_arch = "aarch64")] {
+            let config_space = PCI_CONFIG_SPACE.lock();
+            let config_space = config_space.deref().as_ref().unwrap();
+            let address = address as usize;
+            let slice = &config_space[address..][..size_of::<u32>()];
+            let mut value = [0; 4];
+            value.copy_from_slice(slice);
```
kevinaboos: hmm, i'm not sure that this is correct or even sound. Two things:
1. Don't you have to write to the ADDRESS port and then read from the DATA port, just like on x86 with port I/O?
2. You're accessing the cfg space like it's just a static chunk of memory, but it's not. At the very least, you need to use volatile operations to ensure the compiler doesn't remove/optimize out the memory reads here. The config space can be written to and read from interactively, i.e., its values may change depending on what you write to it. So you need to ensure that every read and write operation is done using the proper two-step sequence of instructions (write the address, then read/write the data).

kevinaboos: Also, I think this may be why you're getting invalid data out of the PCI cfg space in some cases ---- this shouldn't work, but it did happen to work sometimes. Fixing this may also fix the issue of `highmem=on` being required.

kernel/pci/src/lib.rs
```diff
@@ -221,74 +263,98 @@ impl PciLocation {
...
 
-    fn write_data_port(value: u32) {
-        unsafe {
-            PCI_CONFIG_DATA_PORT.lock().write(value);
+        #[cfg(target_arch = "aarch64")] {
+            let mut config_space = PCI_CONFIG_SPACE.lock();
+            let config_space = config_space.deref_mut().as_mut().unwrap();
+            let address = address as usize;
+            let slice = &mut config_space[address..][..size_of::<u32>()];
+            slice.copy_from_slice(&value.to_ne_bytes());
```
kevinaboos: the same thing applies here as my comment from `pci_read_32_addr`

kernel/pci/src/lib.rs
```diff
@@ -221,74 +263,98 @@ impl PciLocation {
...
     fn pci_read_8(&self, offset: u8) -> u8 {
         self.pci_read_32(offset) as u8
     }
 
-    /// Write 32-bit data to the specified `offset` for the PCI device.
-    fn pci_write(&self, offset: u8, value: u32) {
-        unsafe {
-            PCI_CONFIG_ADDRESS_PORT.lock().write(self.pci_address(offset)); 
-            Self::write_data_port((value) << ((offset & 2) * 8));
+    fn pci_write_addr(&self, address: u32, value: u32) {
```
kevinaboos: why do we need a new function for this? Perhaps I don't understand what this is for, but what's the difference between an "address" (`u32`) and and "offset" (`u8`)? 

This function seems like it overlaps with `pci_write_32`

kevinaboos (changes requested):

kernel/pci/src/lib.rs
NathanRoyer: You're completely right, I meant to write 0x3

kernel/pci/src/lib.rs
NathanRoyer: > Also, I think this may be why you're getting invalid data out of the PCI cfg space in some cases ---- this shouldn't work, but it did happen to work sometimes. Fixing this may also fix the issue of `highmem=on` being required.

I'll try.

> 1. Don't you have to write to the ADDRESS port and then read from the DATA port, just like on x86 with port I/O?

Nope

> 2. You're accessing the cfg space like it's just a static chunk of memory, but it's not. At the very least, you need to use volatile operations to ensure the compiler doesn't remove/optimize out the memory reads here.

You're fully right.

kernel/pci/src/lib.rs
NathanRoyer: Indeed; I moved this to the read & write methods, but I'll restore the original structure

kernel/pci/src/lib.rs
NathanRoyer: We need to modify it, when using `pci_write_*`, but I will replace the option with Once.

kernel/pci/src/lib.rs
NathanRoyer: Yep, I intended this to allow successive calls to pci_read/write to keep the translated address in their stack frame between calls, but it's probably already happening on optimized builds anyway. I'll restore the original API.

kernel/pci/src/lib.rs
NathanRoyer: Fun fact, I thought the high PCIE address was so high because of the highmem options but it's not. With and without the option, the address still is the same. So, even if it's unrelated to these reads and writes (still, thanks for the idea), we no longer need the highmem options x)

kevinaboos: oh interesting. Okay cool, lmk how it goes.

kevinaboos:
Not sure if you're ready for me to re-review it or if you're still in the process of making changes. Once you're done, click the re-request review button at the top right under "Reviewers" so I know for sure.

kevinaboos:
I merged in #972, please update this PR to incorporate those changes. Thanks!

kernel/device_manager/src/lib.rs
```diff
@@ -97,20 +97,20 @@ pub fn init(
         keyboard::init(ps2_controller.keyboard_ref(), key_producer)?;
         mouse::init(ps2_controller.mouse_ref(), mouse_producer)?;
     }
+    // Initialize/scan the PCI bus to discover PCI devices
+    for dev in pci::pci_device_iter()? {
+        debug!("Found PCI device: {:#X?}", dev);
+    }
```
kevinaboos, line 92: i can't comment on that line, but can you remove the outdated comment in L105 below?

kernel/pci/src/lib.rs
```diff
@@ -129,7 +157,18 @@ pub struct PciBus {
 
 /// Scans all PCI Buses (brute force iteration) to enumerate PCI Devices on each bus.
 /// Initializes structures containing this information. 
-fn scan_pci() -> Vec<PciBus> {
+fn scan_pci() -> Result<Vec<PciBus>, &'static str> {
+    #[cfg(target_arch = "aarch64")]
+    PCI_CONFIG_SPACE.lock().try_call_once(|| {
+        let config = BOARD_CONFIG.pci_ecam;
+        let flags = memory::PteFlags::new().valid(true).writable(true);
```
kevinaboos: do we need to map this as  device memory?

kernel/pci/src/lib.rs
```diff
@@ -217,78 +256,110 @@ impl PciLocation {
...
     /// The least two significant bits of offset are masked, so it's 4-byte aligned addressing,
     /// which makes sense since we read PCI data (from the configuration space) in 32-bit chunks.
-    fn pci_address(self, offset: u8) -> u32 {
-        ((self.bus  as u32) << 16) | 
-        ((self.slot as u32) << 11) | 
-        ((self.func as u32) <<  8) | 
-        ((offset as u32) & (PCI_CONFIG_ADDRESS_OFFSET_MASK as u32)) | 
-        0x8000_0000
+    /// The shift allows selecting a byte/word from the addressed value.
+    fn pci_address_and_shift(self, offset: u8) -> (u32, u32) {
```
kevinaboos: more docs would be welcome here, especially for how `shift` works, e.g., it's used to right-shift the PCI data value from pci_read_32

It'd also be a good idea to use a struct for the return type, and/or a newtype wrapper for the address and shift value too (e.g., `PciAddress` and `PciShift`)

kevinaboos (changes requested):
Just one small comment; most other changes look great!

How much have you tested this, e.g., does it fully work with all existing supported devices on aarch64 and x86_64?

kernel/pci/src/lib.rs
NathanRoyer: Indeed! since the PCI controller can be in a north bridge, which is a separate chip behind CPU caches.
I missed this, thank you for spotting it.

kernel/pci/src/lib.rs
NathanRoyer: New doc and struct:
```rust
struct PciAddress {
    /// Offset to a double word in a PCI configuration space
    dword_address: u32,
    /// Bit offset of the bytes of interest in this dword
    byte_shift: u32,
}

    /// Computes a [`PciAddress`] from bus, slot, func, and offset. 
    ///
    /// PCI configuration space addresses must be aligned to 4 bytes (double word)
    /// for reads and writes to succeed. The `dword_address` field of the returned
    /// structure will be aligned to 4 bytes accordingly.
    ///
    /// Offsets with the two least significant bits set are still valid (they point
    /// to words and bytes). This function computes a `byte_shift` value which should
    /// be used to shift the double word pointed to by `dword_address`.
```

kernel/pci/src/lib.rs
```diff
@@ -211,84 +249,133 @@ pub struct PciLocation {
...
+        #[cfg(target_arch = "aarch64")] {
+            let config_space = PCI_CONFIG_SPACE.lock();
+            let config_space = config_space.get()
+                .expect("PCI Config Space wasn't mapped yet");
+            value = unsafe {
+                config_space
+                    .as_ptr()
+                    .add(address.dword_address as usize)
+                    .cast::<u32>()
+                    .read_volatile()
```
kevinaboos: hmm, so if we map the config space as an array (i.e., `BorrowedSliceMappedPages`), why do we need to use unsafe pointer arithmetic here instead of just indexing into the array?

Seems like you ought to be able to do something like map the config space as a slice of `&mut [Volatile<u32>]` and then access it by indexing into it with the index being `address.dword_address / 4`.

As a bonus, this would also avoid unaligned reads/writes.

kernel/pci/src/lib.rs
```diff
@@ -211,84 +249,133 @@ pub struct PciLocation {
     func: u8,
 }
 
+struct PciAddress {
+    /// Offset to a double word in a PCI configuration space
+    dword_address: u32,
+    /// Bit offset of the bytes of interest in this dword
+    byte_shift: u32,
```
kevinaboos: this documentation is needed, but it's still a bit confusing. It's a bitwise shift value, so what does it have to do with "byte"s?

kernel/pci/src/lib.rs
```diff
@@ -211,84 +249,133 @@ pub struct PciLocation {
     func: u8,
 }
 
+struct PciAddress {
+    /// Offset to a double word in a PCI configuration space
+    dword_address: u32,
```
kevinaboos: this isn't really an address, it's an offset into the configuration space.

Also, if you adopt the redesign suggestion below, it would be an index into the slice of volatile `u32`s, so that might help with the naming convention and documentation clarity here.

kevinaboos (changes requested):
haven't yet tested it on my machine, but i did leave a few more suggestions.

kevinaboos:
Disk I/O works correctly with your changes 👍

kernel/pci/src/lib.rs
NathanRoyer: It's a bit offset to a specific byte (the offset can only be a multiple of 8)

kernel/pci/src/lib.rs
NathanRoyer: > this isn't really an address, it's an offset into the configuration space.

I was re-using the original terminology of the crate but I agree

kernel/pci/src/lib.rs
NathanRoyer: Indeed; will do

NathanRoyer:
I just noticed something weird that is also [present in the original implementation](https://github.com/theseus-os/Theseus/blob/theseus_main/kernel/pci/src/lib.rs#L250).

In `pci_write`, a shifted value is directly written to the data port; if the offset was to an unaligned byte, this can discard other bytes of the dword and replace them with zeros right?

kernel/pci/src/lib.rs
NathanRoyer: Do you prefer DwordOffset or AbsoluteOffset?

kernel/pci/src/lib.rs
kevinaboos: Then perhaps we should use an enum instead, especially if this field should be restricted to a certain subset of values. perhaps:
```rust
/// todo: Docs, plus think of a better name, lol
enum ByteSelector {
    Byte0, // = 0,
    Byte1, // = 8,
    Byte2, // = 16,
    Byte3, // = 32,
}
```

then each variant can either be given a `u8` value (as shown by the suggested comments) or you can implement a function on that enum that returns a bitshift value or extracts the actual byte(s) from an inputted `u32`.

kernel/pci/src/lib.rs
kevinaboos: the latter, i suppose.  I don't think `dword` is a great term, since folks readers may not know what that is. Documentation should resolve any ambiguity though -- you can describe what the value is and how it's derived from an offset passed into the PCI read/write functions

kevinaboos:
> I just noticed something weird that is also [present in the original implementation](https://github.com/theseus-os/Theseus/blob/theseus_main/kernel/pci/src/lib.rs#L250).
> 
> In `pci_write`, a shifted value is directly written to the data port; if the offset was to an unaligned byte, this can discard other bytes of the dword and replace them with zeros right?

That's true, I think that's intentional though, as it's a "raw" inner function that is designed to be used by other public PCI functions. I suppose the idea is that if you wanted to preserve those bits, you would have read the existing value right before that and then passed in the updated value to that write function.

kernel/pci/src/lib.rs
kevinaboos: Now that we represent the configuration space as an array of `u32`s, let's just call this what it is -- an index.

kernel/pci/src/lib.rs
kevinaboos: bump

kevinaboos (changes requested):
still waiting on the `PciAddress`-related fixes, some pending comments there. After that we can merge.

NathanRoyer:
@kevinaboos I replaced the previous pci read/writers with this:
```rust
impl PciLocation {
    /// Read a 4-bytes register from the PCI Configuration Space.
    ///
    /// Panics if the register isn't a [`FullDword`]
    fn pci_read_32(&self, register: PciRegister) -> u32

    /// Read a 2-bytes register from the PCI Configuration Space.
    ///
    /// Panics if the register isn't a [`Word0`] / [`Word1`]
    fn pci_read_16(&self, register: PciRegister) -> u16

    /// Read a one-byte register from the PCI Configuration Space.
    ///
    /// Panics if the register isn't a [`Byte0`] / [`Byte1`] / [`Byte2`] / [`Byte3`]
    fn pci_read_8(&self, register: PciRegister) -> u8;

    /// Write a 4-bytes register from the PCI Configuration Space.
    ///
    /// Panics if the register isn't a [`FullDword`]
    fn pci_write_32(&self, register: PciRegister, value: u32);
    /// Write a 2-bytes register from the PCI Configuration Space.
    ///
    /// Panics if the register isn't a [`Word0`] / [`Word1`]
    fn pci_write_16(&self, register: PciRegister, value: u16);

    /// Write a one-byte register from the PCI Configuration Space.
    ///
    /// Panics if the register isn't a [`Byte0`] / [`Byte1`] / [`Byte2`] / [`Byte3`]
    fn pci_write_8(&self, register: PciRegister, value: u8);
}

/// (PCI Config Space Dword Index, Register Span)
type PciRegister = (u8, RegisterSpan);

enum RegisterSpan {
    /// Bits [0:31]
    FullDword,
    /// Bits [0:15]
    Word0,
    /// Bits [16:31]
    Word1,
    /// Bits [0:7]
    Byte0,
    /// Bits [8:15]
    Byte1,
    /// Bits [16:23]
    Byte2,
    /// Bits [24:31]
    Byte3,
}

const PCI_VENDOR_ID:             PciRegister = (0x0, Word0);
const PCI_DEVICE_ID:             PciRegister = (0x0, Word1);
const PCI_COMMAND:               PciRegister = (0x1, Word0);
const PCI_STATUS:                PciRegister = (0x1, Word1);
const PCI_REVISION_ID:           PciRegister = (0x2, Byte0);
const PCI_PROG_IF:               PciRegister = (0x2, Byte1);
// ...
```

kevinaboos (approved):
I made some changes to clean things up a bit. I like the introduction of all the new types, but we can actually go further to use traits and trait bounds to move all of the assertions into compile time type errors. We'll save that for the next PR.

PR #1022, opened 2023-08-02
Remove `scheduler`'s dependency on `task`

tsoutsman:
Separates the `task` crate into:
- `task_struct`: Defines a `RawTaskRef` type that is a wrapper around `Arc<Task>` defining methods that don't use the scheduler e.g. `is_runnable`. This type is used in the scheduler subsystem.
- `task`: Defines the `TaskRef` type that is a wrapper around a `RawTaskRef`. This type defines methods that (will) depend on the scheduler e.g. `block`, `unblock`
- `task_init`: Defines the interrupt handler. Has to be a separate crate because the interrupt handler depends on `sleep` which depends on `task`/`task_struct`.

Things that ideally wouldn't be public:
- `task_struct::RawTaskRef::expose`
- `task_struct::ExitableTaskRef::from_raw`
- `task::post_context_switch_action`

TBH, I'm still of the opinion that the task and scheduler subsystems should be merged considering the fact that `schedule()` is defined in the `task` crate not the `scheduler` crate. It would avoid this whole mess of types and `.inner.inner.inner` statements.


kernel/task_struct/src/exposed.rs
```diff
@@ -0,0 +1,75 @@
...
+/// [`Task`], which makes it completely safe to be public because there is
+/// nowhere else besides within the `TaskRef::create()` constructor that one can
+/// obtain access to an owned `Task` value that is already registered/spawned
+/// (actually usable).
+///
+/// If another crate instantiates a bare `RawTaskRef` (not a `Taskref`) and then
+/// converts it into this `ExposedTask` type, then there's nothing they can do
+/// with that task because it cannot become a spawnable/schedulable/runnable
+/// task until it is passed into `TaskRef::create()`, so that'd be completely
+/// harmless.
```
tsoutsman, line 26: This isn't true anymore.

PR #1024, opened 2023-08-03
Add support for CLS variables with user-defined types

tsoutsman:
The `cpu_local` macro now generates four functions for user-defined types:
```rust
pub fn replace(&self, value: #ty) -> #ty;

pub fn replace_guarded<G>(&self, mut value: #ty, guard: &G) -> #ty
where
    G: ::cls::Guard;

pub fn set(&self, value: #ty);

pub fn set_guarded<G>(&self, mut value: #ty, guard: &G)
where
    G: ::cls::Guard;
```

`cls::Guard` is implemented for `preemption::PreemptionGuard` and `irq_safety::HeldInterrupts`. Passing a reference ensures that one or the other is disabled and so it is safe to access the CLS variable across multiple assembly instructions.

However, this doesn't work for all use cases so the PR also adds two optional args to the `cpu_local` macro:
- `cls_dep`: Prevents generating functions that depend on `cls`. Used by the `preemption` crate to avoid a circular dependency.
- `stores_guard`: Used for CLS variables that store an `Option<impl cls::Guard>` (i.e. `TASK_SWITCH_PREEMPTION_GUARD`). Modifies the signatures of `replace` and `set`:
  ```rust
  pub fn replace(&self, guard: #guard_type) -> #ty;
  pub fn set(&self, mut guard: #guard_type);
  ```
  As the additional guard reference isn't needed to ensure preemption/interrupts are disabled.

### Pain points

- `cpu_local` will generate a different API for integers, custom-types, and `Option<impl cls::Guard>` which can cause some confusion. In practice this shouldn't be a major problem because `Option<impl cls::Guard>` is a niche workaround used for a specific CLS variable, and `cargo doc` works as you'd expect, but still... Currently, the macro isn't actually necessary as we could recreate all of the functionality using `CpuLocalU*`/`CpuLocalCell` types which would dramatically improve the unpredictable API. But (I'm pretty sure) the macro is necessary for when we implement dynamic CLS sections so there's no point removing it now.

### DIFF
--- a/kernel/cls/cls_macros/src/lib.rs
+++ b/kernel/cls/cls_macros/src/lib.rs
@@ -150,27 +150,16 @@ pub fn cpu_local(args: TokenStream, input: TokenStream) -> TokenStream {
 
     let ref_expr = quote! {
         {
-            let mut ptr: usize;
             #[cfg(target_arch = "x86_64")]
-            {
-                unsafe {
-                    ::core::arch::asm!(
-                        "rdgsbase {}",
-                        out(reg) ptr,
-                        options(nomem, preserves_flags, nostack),
-                    )
-                };
-            }
+            let mut ptr = {
+                use cls::__private::x86_64::registers::segmentation::{GS, Segment64};
+                GS::read_base().as_u64()
+            };
             #[cfg(target_arch = "aarch64")]
-            {
-                unsafe {
-                    ::core::arch::asm!(
-                        "mrs {}, tpidr_el1",
-                        out(reg) ptr,
-                        options(nomem, preserves_flags, nostack),
-                    )
-                };
-            }
+            let mut ptr = {
+                use cls::__private::tock_registers::interfaces::Readable;
+                cls::__private::cortex_a::registers::TPIDR_EL1.get()
+            };
             ptr += #offset;
             unsafe { &mut*(ptr as *mut #ty) }
         }
--- a/kernel/cls/cls_macros/src/lib.rs
+++ b/kernel/cls/cls_macros/src/lib.rs
@@ -62,35 +62,34 @@ impl Parse for Args {
             if name.is_ident("cls_dep") {
                 cls_dependency = input.parse::<LitBool>()?.value();
             } else if name.is_ident("stores_guard") {
-                stores_guard = Some(input.parse()?);
+                stores_guard = Some((name, input.parse::<Type>()?));
             } else {
                 Diagnostic::spanned(
                     name.span().unwrap(),
                     Level::Error,
                     format!("invalid argument `{}`", name.to_token_stream()),
                 )
-                .help("valid arguments are: `cls_dep`")
+                .help("valid arguments are: `cls_dep`, `stores_guard`")
                 .emit();
                 return Err(syn::Error::new(name.span(), ""));
             }
         }
 
-        if stores_guard.is_some() && !cls_dependency {
-            todo!();
-            // Diagnostic::spanned(
-            //     name.span().unwrap(),
-            //     Level::Error,
-            //     format!("invalid argument `{}`", name.to_token_stream()),
-            // )
-            // .help("valid arguments are: `cls_dep`")
-            // .emit();
-            // return Err(syn::Error::new(name.span(), ""));
+        if let Some((ref name, ref value)) = stores_guard && !cls_dependency {
+            let span = name.span().join(value.span()).unwrap().unwrap();
+            Diagnostic::spanned(
+                span,
+                Level::Error,
+                "`stores_guard` requires `cls_dep`",
+            )
+            .emit();
+            return Err(syn::Error::new(name.span(), ""));
         }
 
         Ok(Self {
             offset,
             cls_dependency,
-            stores_guard,
+            stores_guard: stores_guard.map(|tuple| tuple.1),
         })
     }
 }
@@ -101,13 +100,38 @@ impl Parse for Args {
 ///
 /// The initialisation expression has no effect; to set the initial value,
 /// `per_cpu::PerCpuData::new` must be modified.
+///
+/// # Arguments
+///
+/// The macro supports additional named arguments defined after the offset (e.g.
+/// `#[cpu_local(0, cls_dep = false)]`):
+/// - `cls_dep`: Whether to define methods that depend on `cls` indirectly
+///   adding a dependency on `preemption` and `irq_safety`. This is only really
+///   useful for CPU locals defined in `preemption` to avoid a circular
+///   dependency. Defaults to true.
+/// - `stores_guard`: If defined, must be set to either `HeldInterrupts` or
+///   `PreemptionGuard` and signifies that the CPU local has the type
+///   `Option<Guard>`. This option defines special methods that use the guard
+///   being switched into the CPU local, rather than an additional guard
+///   parameter, as proof that the CPU local can be safely accessed.
 #[proc_macro_attribute]
 pub fn cpu_local(args: TokenStream, input: TokenStream) -> TokenStream {
     let Args {
         offset,
         cls_dependency,
         stores_guard,
-    } = parse_macro_input!(args as Args);
+    } = match syn::parse(args) {
+        Ok(args) => args,
+        Err(error) => {
+            if error.to_string() == "" {
+                // We've already emmited an error diagnostic.
+                return TokenStream::new();
+            } else {
+                Diagnostic::spanned(error.span().unwrap(), Level::Error, error.to_string()).emit();
+                return TokenStream::new();
+            }
+        }
+    };
 
     let CpuLocal {
         attributes,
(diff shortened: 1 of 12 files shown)


kernel/cls/cls_macros/src/lib.rs
```diff
@@ -87,22 +148,114 @@ pub fn cpu_local(args: TokenStream, input: TokenStream) -> TokenStream {
...
+                unsafe {
+                    ::core::arch::asm!(
+                        "mrs {}, tpidr_el1",
+                        out(reg) ptr,
+                        options(nomem, preserves_flags, nostack),
+                    )
+                };
+            }
+            ptr += #offset;
+            unsafe { &mut*(ptr as *mut #ty) }
```
kevinaboos: there are a few alternatives here:
1. rather than writing our own asm here, let's just use existing Rust-level crates to access GSBASE (`x86_64`) or TPIDR_EL1 (`cortex_a`).
2.  the `PerCpuData` struct starts with a self pointer at offset 0, so it should be much faster to obtain the pointer to a CLS variable by using that self pointer instead of the base register. At least on x86 it's generally faster to read from `%gs:0` than the `GSBASE` register; I suppose that it doesn't make a difference on aarch64 since we always have to read TPIDR_EL1 manually anyways.

kevinaboos: Obviously option 2 is only available if we continue to use a CLS self pointer, which we probably will for the foreseeable future.

kernel/cls/cls_macros/src/lib.rs
```diff
@@ -87,22 +148,114 @@ pub fn cpu_local(args: TokenStream, input: TokenStream) -> TokenStream {
...
+                    self.replace(guard);
+                }
+            }
+        } else {
+            quote! {
+                #guarded_functions
+
+                #[inline]
+                pub fn replace(&self, value: #ty) -> #ty {
+                    // TODO: Should this ever be `disable_interrupts` instead?
```
kevinaboos, line 230: not by default, no, but IIRC we had discussed a prior idea of using the `DeadlockPrevention` trait here to allow the user/definer of the CLS variable to specify

Instead of doing that, we can just use your new `Guard` trait (which could be further abstracted as an "atomicity-ensurer" trait or something) for that purpose, as using a guard for atomicity is more accurate for what we want to achieve here than preventing deadlock.

kernel/cls/src/lib.rs
```diff
@@ -7,3 +7,20 @@
 extern crate alloc;
 
 pub use cls_macros::cpu_local;
+
+pub trait Guard: sealed::Sealed {}
```
kevinaboos: Good solution! Just needs some docs.

Also, I'd vote to give it a more descriptive name, e.g., `AtomicityGuard` or something to do with ensuring atomicity (also mentioned in my other comment)

kernel/cpu_local/src/lib.rs
```diff
@@ -219,6 +219,13 @@ pub fn init<P>(
     size_of_per_cpu_data: usize,
     per_cpu_data_initializer: impl FnOnce(usize) -> P
 ) -> Result<(), &'static str> {
+    // Enable FSGSBASE instructions
+    #[cfg(target_arch = "x86_64")]
+    {
+        use x86_64::registers::control::{Cr4, Cr4Flags};
+        unsafe { Cr4::update(|flags| flags.insert(Cr4Flags::FSGSBASE)) };
+    }
+
```
kevinaboos, line 228: even after reading the intel SDM, i cannot tell if this CR4 bit is only required to enable these instructions for non-privileged execution modes (ring 3) or if it's always required.
Also, is this set of instructions supported on every processor generation or just newer ones? If it's not supported on all x86_64/amd64 hardware, i'd prefer to revert to using the old MSR-based `GsBase::write()`, especially since it doesn't need to be efficient here, as we only write it once upon initializing each CPU.

There's definitely an argument to be made for using this in `set_next_task()` to update `FSBASE`, since that runs on every context switch.

kernel/preemption/Cargo.toml
```diff
@@ -6,7 +6,7 @@ description = "Handles enabling and disabling preemption for each CPU core"
 edition = "2021"
 
 [dependencies]
-cls = { path = "../cls" }
+cls_macros = { path = "../cls/cls_macros" }
```
kevinaboos, line 9: any particular reason why we depend on `cls_macros` directly now? FWIW, i think it's cleaner to depend on `cls` directly and allow the `cls_macros` to be an internal implementation detail.

kernel/task/src/lib.rs
```diff
@@ -953,70 +953,32 @@ fn task_switch_inner(
...
+#[cls::cpu_local(16, stores_guard = PreemptionGuard)]
+pub(crate) static TASK_SWITCH_PREEMPTION_GUARD: Option<PreemptionGuard> = None;
+
+/// Data that should be dropped after switching away from a task that has exited.
+///
+/// Currently, this contains the previous Task's `TaskRef` removed from its TLS area;
+/// it is stored in a CPU-local variable because it's only related to
+/// a task switching operation on a particular CPU.
+#[cls::cpu_local(24)]
+pub(crate) static DROP_AFTER_TASK_SWITCH: Option<TaskRef> = None;
```
kevinaboos: nit: since we moved these out of an inner module, they don't need to have any visibility modifier like `pub(crate)`

kevinaboos (changes requested):
Very nice! love the API and design decisions. I left a few comments, mostly questions and nitpicky shit.

kernel/preemption/Cargo.toml
tsoutsman: Because depending on `cls` would lead to a circular dependency. That's why the `cls_dep` argument on the macro is necessary.

kernel/cpu_local/src/lib.rs
tsoutsman: Without setting the CR4 bit, the instruction was triggering an illegal instruction exception.

> Also, is this set of instructions supported on every processor generation or just newer ones?

Well we're using Broadwell and that's pretty old. FWIW `sel4` [uses `fsgsbase` instructions by default](https://github.com/seL4/seL4/blob/49f4ddfc8c20249b7a114a4cc3baa42f5f9eaa6c/src/arch/x86/config.cmake#L210) when compiling for `x86_64`.

> it doesn't need to be efficient here, as we only write it once upon initializing each CPU.

We call `rdgsbase` every time we access a CPU local with a user-defined type.

kernel/cls/cls_macros/src/lib.rs
tsoutsman: > rather than writing our own asm here, let's just use existing Rust-level crates to access GSBASE (x86_64) or TPIDR_EL1 (cortex_a).

The only reason I was hesitant to do that because it means I would have to reexport them in `cls::__private`, but it's not that big of a deal.

> At least on x86 it's generally faster to read from %gs:0 than the GSBASE register

So I'm assuming you're referring to [this optimisation](https://github.com/rust-osdev/x86_64/pull/257). The reason using a self ptr is faster, is because first loading the value of the pointer in to a register means Rust can reason about and optimise the memory access. We do the same thing except using the `rdgsbase` instruction, which is specifically designed to read the gs base register efficiently. So, I don't think this applies to our use case.

Also as a sidenote i'm pretty sure `rdgsbase` is faster than access via `msr`.

kernel/preemption/Cargo.toml
kevinaboos: Ah yup, ok.

kernel/cpu_local/src/lib.rs
kevinaboos: Ok yea that's what I was wondering, whether it'd fault if we didn't set that bit. All good.

kevinaboos: > We call `rdgsbase` every time we access a CPU local with a user-defined type

True, but should we be doing that? I know that reading an MSR is slow, but perhaps using `rdgsbase` is faster than using a CLS self pointer and accessing it via `[%gs:0]`?

kernel/cls/cls_macros/src/lib.rs
kevinaboos: I was just asking about doing `%gs:0` vs `rdgsbase` alone, I realize they both allow the compiler to perform the same reasoning. It's likely not a big deal, I honestly did not know about `rsgsbase` before, so if that's the preferred efficient way to do it then we can skip the CLS self pointer entirely (unless it ends up being required for a TLS-based CLS approach, but we still don't have to use it)

kernel/cpu_local/src/lib.rs
kevinaboos: Ignore this, we can discuss it on the other comment.

kernel/cls/cls_macros/src/lib.rs
tsoutsman: I doubt that `%gs:0` is faster than even `msr` access, because then other operating systems like Linux would be using that method.

kevinaboos: Can't argue with that, yeah

kevinaboos: So option 1 is good then as long as it compiles down to the same thing as our inline asm. Just trying to reduce inline asm or dedup it as much as possible, y'know

kernel/cls/cls_macros/src/lib.rs
tsoutsman: [We can't define default type for functions](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f070328297316f6ab023b579068c438) so I think for now it's best to leave as is because it would just add more complexity to the API for no apparent gain.

kevinaboos: Yeah, no problem. We could define it on the struct perhaps, but no need.

kevinaboos (approved):
Looks great, thanks as always! Tested and working for me.

PR #1025, opened 2023-08-04
Implement identity mapping convenience function

kevinaboos:
The new `create_identity_mapping()` function chooses random virtual and physical addresses (that are obviously the same) and maps the allocated virtual pages to the corresponding allocated physical frames.

The procedure is as follows:
> We first iterate over all free general-purpose frames,
> as there are far fewer available frames than available pages.
> Once we find a suitable free chunk of frames,
> we then attempt to allocate the corresponding identity pages,
> and if that succeeds, we allow the frame allocator to proceed
> with allocating the range of frames that matches those pages.

This is needed for proper/easier multicore CPU bringup on arrch64.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3806,6 +3806,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "test_identity_mapping"
+version = "0.1.0"
+dependencies = [
+ "app_io",
+ "log",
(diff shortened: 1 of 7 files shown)


kevinaboos:
~~Note: this is still untested.~~

Tested and fully working.

PR #1026, opened 2023-08-04
Update the GIC driver and interrupt controller for owned + per-CPU memory mappings

NathanRoyer:
Changes:
- The GIC driver has two main structures: `ArmGicDistributor` & `ArmGicCpuComponents` which map to `SystemInterruptController` & `LocalInterruptController`. There's one `ArmGicCpuComponents` per CPU. This is a cleaner ownership distribution.
- I had to comment out some GICv2 support for this unfortunately, because IPIs are sent via the distributor in this version, whereas on GICv3 it's done on the CPU interface. I left a big comment on the affected code and in arm_boards; this code now panics.
- The `aarch64` part of `interrupt_controller` initializes these properly and implements `SystemInterruptControllerApi` & `LocalInterruptControllerApi` on top of them.

However, this is a draft PR because I'm not sure where to store these, since there have been work on cpu-local-storage recently and I'm not sure if it's ready or not.

### DIFF
--- a/kernel/gic/src/gic/cpu_interface_gicv2.rs
+++ b/kernel/gic/src/gic/cpu_interface_gicv2.rs
@@ -12,14 +12,13 @@ use super::InterruptNumber;
 use volatile::{Volatile, ReadOnly, WriteOnly};
 use zerocopy::FromBytes;
 
-/// Registers of a CPU interface of the GIC, pertaining
-/// to a specific CPU in the system.
+/// The GICv2 MMIO registers for interfacing with a specific CPU.
 ///
-/// Methods refer to that specific CPU by "current CPU".
+/// Methods herein apply to the "current" CPU only, i.e., the CPU
+/// on which the code that accesses these registers is currently running.
 ///
-/// Caution: Even though the physical address for this
-/// structure is the same for each CPU, the actual backing
-/// memory is per-CPU (different for each CPU).
+/// Note: the physical address for this structure is the same for all CPUs,
+/// but the actual backing memory refers to physically separate registers.
 #[derive(FromBytes)]
 #[repr(C)]
 pub struct CpuRegsP1 {            // base offset
--- a/kernel/gic/src/gic/cpu_interface_gicv2.rs
+++ b/kernel/gic/src/gic/cpu_interface_gicv2.rs
@@ -16,6 +16,10 @@ use zerocopy::FromBytes;
 /// to a specific CPU in the system.
 ///
 /// Methods refer to that specific CPU by "current CPU".
+///
+/// Caution: Even though the physical address for this
+/// structure is the same for each CPU, the actual backing
+/// memory is per-CPU (different for each CPU).
 #[derive(FromBytes)]
 #[repr(C)]
 pub struct CpuRegsP1 {            // base offset
--- a/kernel/gic/src/gic/cpu_interface_gicv2.rs
+++ b/kernel/gic/src/gic/cpu_interface_gicv2.rs
@@ -6,17 +6,21 @@
 //! - Acknowledging interrupt requests
 //! - Sending End-Of-Interrupts signals
 
-use super::GicRegisters;
 use super::Priority;
 use super::InterruptNumber;
 
-mod offset {
-    use crate::Offset32;
-    pub(crate) const CTLR: Offset32 = Offset32::from_byte_offset(0x00);
-    pub(crate) const PMR:  Offset32 = Offset32::from_byte_offset(0x04);
-    pub(crate) const IAR:  Offset32 = Offset32::from_byte_offset(0x0C);
-    pub(crate) const RPR:  Offset32 = Offset32::from_byte_offset(0x14);
-    pub(crate) const EOIR: Offset32 = Offset32::from_byte_offset(0x10);
+use volatile::{Volatile, ReadOnly, WriteOnly};
+use zerocopy::FromBytes;
+
+#[derive(FromBytes)]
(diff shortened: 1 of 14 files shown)


kernel/gic/src/gic/cpu_interface_gicv2.rs
```diff
@@ -25,45 +29,47 @@ mod offset {
...
-/// is lower or equal to this one, they're queued
-/// until this CPU or another one is ready to handle
-/// them
-pub fn set_minimum_priority(registers: &mut GicRegisters, priority: Priority) {
-    registers.write_volatile(offset::PMR, (u8::MAX - priority) as u32);
-}
+    /// Interrupts have a priority; if their priority
+    /// is lower or equal to this one, they're queued
+    /// until this CPU or another one is ready to handle
+    /// them
```
kevinaboos: these docs need some improvement. I like the description of interrupt priorities, but you don't actually explain what the function does. Devs would expect something like:

```rust
/// Returns the minimum interrupt priority for the current CPU.
///
/// Interrupt priority is ...
```

kernel/gic/src/gic/redist_interface.rs
```diff
@@ -9,20 +9,42 @@
...
+pub struct RedistRegsSgiPpi {            // base offset
+    _reserved0:   [u8;            0x80],
+
+    group:        [Volatile<u32>; 0x01], // 0x080
+    _reserved1:   [u32;           0x1f],
+
+    set_enable:   [Volatile<u32>; 0x01], // 0x100
+    _reserved2:   [u32;           0x1f],
+
+    clear_enable: [Volatile<u32>; 0x01], // 0x180
```
kevinaboos: for these 1-element `Volatile<u32>` fields, what's the point of doing it like that instead of just:
```rust
    group: Volatile<u32>,
```

kernel/gic/src/gic/mod.rs
```diff
@@ -349,72 +332,88 @@ impl ArmGic {
...
-    /// Will that interrupt be forwarded by the distributor?
+    /// Will that interrupt be received by this CPU?
     pub fn get_interrupt_state(&self, int: InterruptNumber) -> Enabled {
-        match (int, self) {
-            (0..=31, Self::V3(v3)) => {
-                let i = get_current_cpu_redist_index();
-                redist_interface::is_sgippi_enabled(&v3.redistributors[i].redist_sgippi, int)
-            },
-            (_, this) => dist_interface::is_spi_enabled(this.distributor(), int),
+        assert!(int < 32);
```
kevinaboos: all of these assertions are good, but we should make two improvements to each of them:
1. add a string message that is specific to each function, explaining why it failed.
2. Add docs to the function describing the conditions under which it will panic. You can do that either using a section header `/// # Panic` or as a description of the `int` argument, i.e., that is must be under 32.

kernel/interrupt_controller/src/aarch64.rs
```diff
@@ -111,126 +112,92 @@ impl SystemInterruptControllerApi for SystemInterruptController {
...
+    fn get() -> &'static Self {
+        // While we're waiting for cpu-local-storage, this loop will work as fine as an AtomicMap
 
-        int_ctlr.init_secondary_cpu_interface();
+        let cpu_id = current_cpu();
+        let index = BOARD_CONFIG.cpu_ids.iter().position(|mpidr| cpu_id == (*mpidr).into());
+        let index = index.expect("Invalid CpuId returned by current_cpu()");
+
+        let ctrls = LOCAL_INT_CTRL.get();
+        let ctrls = ctrls.expect("interrupt_controller wasn't initialized");
```
kevinaboos: this series of operations is confusing. You can clarify it by describing the order of indexing used in the `LOCAL_INT_CTRL` (on the docs for that static var), and how it is explicitly _not_ directly indexable by the CPU ID value itself.

kevinaboos (changes requested):
Kindly add docs to all of the top-level register structs describing what they represent and how they should be used. Even better would be to add docs to each of their fields too.

In general, please avoid abbreviations as much as you can. Acronyms/initialisms are okay when they're common, but too many abbreviations makes it hard for non-GIC-experts to read the code.

kernel/gic/src/gic/redist_interface.rs
NathanRoyer: Each (field + _reserved) pair corresponds to an array where only the first item is usable.
This is the best way I found to reflect that

NathanRoyer: Also the existing code for set-enable and other bit arrays requires this to be accessible as a slice

kernel/interrupt_controller/src/aarch64.rs
NathanRoyer: Ok I added this:
```rust
        // how this function works:
        //   a. get the current CpuId: this CpuId of the current CPU
        //   b. iterate on all valid CpuIds, find the index of the current CpuId.
        //      This is used as a current CPU index.
        //   c. get the global array of interrupt controllers
        //   d. index into this array based on the current CPU index
```

And this to the singleton:
```rust
/// Per-CPU local interrupt controller
///
/// To get the controller for a specific CPU:
///    a. Find the position of its CpuId in `BOARD_CONFIG.cpu_ids`
///    b. Index into array slice using that position
static LOCAL_INT_CTRL: Once<[LocalInterruptController; NUM_CPUS]> = Once::new();
```

kevinaboos (approved):

PR #1027, opened 2023-08-04
Make use of `create_identity_mapping` in `multicore_bringup`

NathanRoyer:
`multicore_bringup` contained workarounds in aarch64 code because the main function needed a way to identity map any page/frame. Now that Kevin implemented that, this removes the temporary workarounds and makes use of the new `memory::create_identity_mapping` function.

### DIFF
--- a/kernel/multicore_bringup/src/aarch64.rs
+++ b/kernel/multicore_bringup/src/aarch64.rs
@@ -1,4 +1,4 @@
-use memory::{MmiRef, VirtualAddress, PhysicalAddress, MappedPages, PteFlags};
+use memory::{create_identity_mapping, MmiRef, VirtualAddress, PhysicalAddress, MappedPages, PteFlags};
 use memory_aarch64::{read_mmu_config, asm_set_mmu_config_x2_x3};
 use kernel_config::memory::{PAGE_SIZE, KERNEL_STACK_SIZE_IN_PAGES};
 use psci::{cpu_on, error::Error::*};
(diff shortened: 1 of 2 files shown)


kernel/multicore_bringup/src/aarch64.rs
```diff
@@ -64,56 +84,8 @@ pub fn handle_ap_cores(
...
-            let src = unsafe { (ap_entry_point as *const [u8; PAGE_SIZE]).as_ref() }.unwrap();
-            dst.copy_from_slice(src);
-
-        } else {
-            // get the physical address of ap_entry_point
-            let entry_point_virt_addr = VirtualAddress::new_canonical(ap_entry_point as usize);
-            entry_point_phys_addr = page_table.translate(entry_point_virt_addr).unwrap();
-        }
+        // get physical address of generated entry point
+        entry_point_phys_addr = page_table.translate(virt_addr).unwrap();
```
kevinaboos: am i missing something, or is this the identity mapping that we created above? If so, we don't need to translate it, we already know the physical address -- it's the same as the virtual one.

kernel/multicore_bringup/src/aarch64.rs
```diff
@@ -50,9 +50,29 @@ pub fn handle_ap_cores(
...
+
+    // map this RWX
+    let flags = PteFlags::new().valid(true).writable(true).executable(true);
+    let mut ap_startup_mapped_pages = create_identity_mapping(1, flags)?;
+    let virt_addr = ap_startup_mapped_pages.start_address();
+
+    // copy the entry point code
+    // instructions must have an alignment of four bytes; a page's alignment will always be larger.
+    let dst = ap_startup_mapped_pages.as_slice_mut(0, PAGE_SIZE).unwrap();
+    let src = unsafe { (ap_entry_point as *const [u8; PAGE_SIZE]).as_ref() }.unwrap();
```
kevinaboos: hmm, i think i was able to do this safely on the x86 side, but I'm not 100% sure what differs here in the aarch64 world. Is it possible to map this address and then access it safely via the `MappedPages::as_slice()` API instead of just unsafely trusting that we can dereference it?

kernel/multicore_bringup/src/aarch64.rs
```diff
@@ -50,9 +50,29 @@ pub fn handle_ap_cores(
...
 
+    // -------
+    //
+    // This code identity-maps one page of writable+executable memory and
+    // copies the machine code of `ap_entry_point` to it. The fact that the
+    // mapping is identity-mapped allows the code to enable the MMU safely,
+    // as PC will be valid before and after the switch.
+
+    // map this RWX
+    let flags = PteFlags::new().valid(true).writable(true).executable(true);
```
kevinaboos: This mapping obviously needs to be writable initially for us to copy stuff to it. But does it need to be writable after that initial copy, or can we remap it as just R+X? 

Generally we want to map pages using the _minimum_ required permissions, and if those pages just contain executable text content, then we definitely don't want to map them as writable.

kevinaboos (changes requested):
Just a few questions/suggestions.

kernel/multicore_bringup/src/aarch64.rs
NathanRoyer: I figured this would be cleaner than a cast but I can comment this and write a cast instead

kernel/multicore_bringup/src/aarch64.rs
NathanRoyer: Sure, I'll use the mod_mgmt lookup

kernel/multicore_bringup/src/aarch64.rs
NathanRoyer: > can we remap it as just R+X?

yes will do

kernel/multicore_bringup/src/aarch64.rs
kevinaboos: Ah yikes, I didn't realize it was that tedious to obtain the backing `MappedPages` object for `ap_entry_point`. There might be a faster way but this is okay for now.

PR #1028, opened 2023-08-14
Add dynamic CPU-local variables

tsoutsman:
~I accidentally performed a drive-by reformatting in `mod_mgmt/src/lib.rs` which added \~500 lines to the diff, but it only removed trailing whitespace so hopefully that's fine, although I can extract the changes if you'd like me to.~

Also changed `LocalDataImage` to use a `Vec` internally to avoid wrapping the box in an option type. Happy to change it back.


kernel/ap_start/src/lib.rs
```diff
@@ -108,7 +108,7 @@ pub fn kstart_ap(
 
     // Now that the Local APIC has been initialized for this CPU, we can initialize the
     // per-CPU storage, tasking, and create the idle task for this CPU.
-    per_cpu::init(cpu_id).unwrap();
+    cls_allocator::reload_current_core();
```
kevinaboos: nit: we're moving away from "core" to "cpu" lingo, so use cpu here

cfg/partial_linking_combine_sections.ld
```diff
@@ -34,6 +34,11 @@ SECTIONS {
 		*(.gcc_except_table .gcc_except_table.*)
 	}
 
+	.cls :
```
kevinaboos, line 37: Q: did we ever experiment with or determine whether it's possible to separate .cls into a .bss and .data section? I assume this unified section ends up with a minor amount of wasted space in the object files that's just filled with zeroes (?).

I'm not sure if there's a way to address that easily, just curious if that was ever explored.

kernel/cls/cls_macros/src/lib.rs
```diff
@@ -264,3 +273,237 @@ pub fn cpu_local(args: TokenStream, input: TokenStream) -> TokenStream {
...
+                // a page boundary.
+                //
+                // Hence, CLS symbols have an offset that is incorrect by
+                // `tls_size.next_multiple_of(0x1000)`, or 0 if there is not TLS section. So, when
+                // loading CLS symbols on AArch64, `mod_mgmt` simply sets `__THESEUS_TLS_SIZE` to 0.
+
+                let tls_start_to_cls_start: u64 = if tls_size == 0 {
+                    0
+                } else {
+                    // TODO: Use `next_multiple_of` when stabilised.
```
kevinaboos: we already use it under the `int_roundings` feature in other parts of Theseus.

If you can't use it due to stability requirements of proc macro crates, np. You can still do this branchlessly with standard rounding up math, if desired.

kernel/cls_allocator/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "cls_allocator"
```
kevinaboos, line 2: nit: missing description & author

kernel/crate_metadata_serde/src/lib.rs
```diff
@@ -107,6 +110,7 @@ pub enum SectionType {
     /// A `.tbss` section is a read-only section that holds all-zero data for a thread-local storage (TLS) area.
     /// This is is effectively an empty placeholder: the all-zero data section doesn't actually exist in memory.
     TlsBss,
+    Cls,
```
kevinaboos, line 121: add doc comment

kernel/local_storage_initializer/src/lib.rs
```diff
@@ -1,22 +1,23 @@
-//! Logic for generating Thread-Local Storage (TLS) data image for TLS areas.
+//! Logic for generating thread-local storage (TLS) and CPU-local storage (CLS) images.
 //!
 //! The two key types are:
-//! 1. [`TlsInitializer`]: a "factory" that maintains a list of loaded TLS sections
-//!    in order to correctly generate new TLS data images.
-//! 2. [`TlsDataImage`]: a generated TLS data image that can be used as the TLS area
-//!    for a single task.
+//! 1. [`LocalStorageInitialiser`]: a "factory" that maintains a list of loaded sections
```
kevinaboos: US vs. AUS spelling mismatch, lol

kernel/mod_mgmt/src/lib.rs
```diff
@@ -2173,6 +2282,35 @@ impl CrateNamespace {
...
                                 };
+
+                                // See `cls_macros` for more details.
+                                if source_sec_name == "__THESEUS_CLS_SIZE" {
+                                    let relocation_entry = RelocationEntry::from_elf_relocation(rela_entry);
+                                    write_relocation(
+                                        relocation_entry,
+                                        target_sec_slice,
+                                        target_sec.mapped_pages_offset,
+                                        VirtualAddress::new(usize::MAX).unwrap(),
```
kevinaboos: for these, you can use `new_canonical` or the `MAX_VIRTUAL_ADDRESS` const from `kernel_config::memory`

kernel/mod_mgmt/src/parse_nano_core.rs
```diff
@@ -712,8 +751,10 @@ struct MainSectionInfo {
     data_shndx:      Shndx,
     bss_shndx:       Shndx,
     tls_data_info:   Option<(Shndx, VirtualAddress)>,
+    cls_info:       Option<(Shndx, VirtualAddress)>,
```
kevinaboos: tiny spacing fix 
```suggestion
    cls_info:        Option<(Shndx, VirtualAddress)>,
```

kernel/mod_mgmt/src/parse_nano_core.rs
```diff
@@ -850,14 +891,36 @@ fn add_new_section(
...
+            sec_name,
+            Arc::clone(rodata_pages),
+            // CLS sections are lumped into the ".rodata" MappedPages with the read-only data sections.
+            rodata_pages_locked.offset_of_address(cls_sec_data_vaddr).ok_or("nano_core CLS .cls section wasn't covered by the .rodata mapped pages!")?,
+            VirtualAddress::new(cls_offset).ok_or("new TLS .cls section had invalid virtual address (CLS offset)")?,
+            sec_size,
+            global,
+            new_crate_weak_ref.clone(),
+        );
+        // Add this new TLS section to this namespace's TLS area image.
```
kevinaboos: nit:  copy-paste issues with outdated references to TLS instead of CLS

kernel/nano_core/linker_higher_half-aarch64.ld
```diff
@@ -116,3 +121,7 @@ SECTIONS {
         *(.data.rel.ro.local*) *(.data.rel.ro .data.rel.ro.*)
     }
 }
+
+/* This definition must be placed at the end of the file so that the .cls,
+section is placed in the correct order. */
+__THESEUS_TLS_SIZE = SIZEOF(.tdata) + SIZEOF(.tbss);
```
kevinaboos, line 127: hmm, am i missing something, or is there no __THESEUS_CLS_SIZE const defined on aarch64? how does that pass linking?

kernel/preemption/src/lib.rs
```diff
@@ -117,7 +118,9 @@ impl Drop for PreemptionGuard {
         assert!(
             self.cpu_id == cpu_id,
             "PreemptionGuard::drop(): BUG: CPU IDs did not match! \
-            This indicates an unexpected task migration across CPUs."
+            Task migrated from CPU {} to CPU {}.",
```
kevinaboos: nit: directly state it's not expected

```suggestion
            Task unexpectedly migrated from CPU {} to CPU {}.",
```

tools/elf_cls/src/main.rs
```diff
@@ -0,0 +1,282 @@
...
+///
+/// This must be kept in sync with `mod_mgmt`.
+const CLS_SECTION_FLAG: u64 = 0x100000;
+
+const _: () = assert!(CLS_SECTION_FLAG & SHF_MASKOS as u64 == CLS_SECTION_FLAG);
+
+/// The flag identifying CLS symbols.
+///
+/// This must be kept in sync with `mod_mgmt`.
+const CLS_SYMBOL_TYPE: u8 = 0xa;
```
kevinaboos: can you define these two consts in `crate_metadata_serde` (which is intended for this purpose) such that this `elf_cls` tool can directly depend on that?

Then the `crate_metadata` (and thus also `mod_mgmt`) can re-export those symbols from `crate_metadata_serde` to have a single unified system-wide definition. This is similar to what we did with `SectionType` when you implemented nano_core serialization.

tools/elf_cls/Cargo.lock
kevinaboos, line 1: can you add a brief bullet point description of this `elf_cls` tool to `tools/README` (i think it's listed in alphabetical order)

tools/elf_cls/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "elf_cls"
```
kevinaboos, line 2: nit: missing author, description

Makefile
```diff
@@ -295,6 +295,10 @@ else
 $(error Error: unsupported option "debug=$(debug)". Options are 'full', 'none', or 'base')
 endif
 
+## Sixth, parse CPU local sections.
+	@echo -e "Parsing CPU local sections"
```
kevinaboos: nit: are we parsing them, or modifying/fixing up the ELF sections? Use whatever terms you think are most accurate.

Makefile
```diff
@@ -349,6 +353,8 @@ endif
 ## This builds the nano_core binary itself, which is the fully-linked code that first runs right after the bootloader
 $(nano_core_binary): cargo $(nano_core_static_lib) $(linker_script)
 	$(CROSS)ld -n -T $(linker_script) -o $(nano_core_binary) $(compiled_nano_core_asm) $(nano_core_static_lib)
+## Parse CLS sections.
```
kevinaboos: same comment as above.

kernel/local_storage_initializer/src/lib.rs
```diff
@@ -329,64 +393,132 @@ impl TlsInitializer {
...
+    /// Inherits the data from another data image.
+    ///
+    /// # Panics
     ///
-    /// * On x86_64, this writes to the `FsBase` MSR.
-    /// * On ARMv8, this writes to `TPIDR_EL0`.
-    pub fn set_as_current_tls_base(&self) {
+    /// Panics if the other image has a longer length, or (on x64) if the other
+    /// image has a differently sized static area.
+    pub fn inherit(&mut self, other: &Self) {
```
kevinaboos, line 451: perhaps i missed something, but I don't see where this is used or why we need it. What's the intended use of this?

kevinaboos (changes requested):

kernel/nano_core/linker_higher_half-aarch64.ld
kevinaboos: ah, nvm, i see it's gated by an arch-specific cfg.

In that case, kindly add a comment under this (and within the extern block that accesses the TLS size in `cls_offset_expr`) explaining that we don't need the CLS size on aarch64 and why.

kernel/local_storage_initializer/src/lib.rs
tsoutsman: we use it in `reload_current_core`.

kernel/cls/cls_macros/src/lib.rs
tsoutsman: It would require the implementor crate to add `#![feature(int_roundings)]` to the crate root for no real gain.

Rust does manage to make it branchless but yea probably best to make it explicit.

kernel/nano_core/linker_higher_half-aarch64.ld
tsoutsman: TBH I thought https://github.com/tsoutsman/Theseus/blob/6a195f54a2bbed2c1d105ea22aeee2d8854bf379/kernel/cls/cls_macros/src/lib.rs#L431-L456 sufficiently explains why we don't need TLS size.

cfg/partial_linking_combine_sections.ld
tsoutsman: I don't think it's possible with `link_section`. I did play around with ELFs a bit and I'm pretty certain that with a custom link section Rust will always fill out the bytes (e.g. the `preemption` CLS section could hypothetically be a .bss section).

kernel/mod_mgmt/src/lib.rs
tsoutsman: I'd say the problem is more with the `write_relocation` function accepting a `VirtualAddress` rather than a `usize`. The problem with reexporting `kernel_config` through `cls` is that `#cls_offset_expr` can't have a dependency on `cls` because then `preemption` won't compile (`preemption` cannot have a dependency on `cls`).

I've changed the code so that we only ever call
```rust
VirtualAddress::new(usize::MAX).unwrap();
```
on `x64`, even if encountering a `__THESEUS_CLS_SIZE` relocation on AArch64. I think that's the most correct solution.

cfg/partial_linking_combine_sections.ld
kevinaboos: yea i didn't think so, we'd need some (nonexistent) attribute like `link_section_prefix`/`suffix`.
Not a big deal either way.

kernel/mod_mgmt/src/lib.rs
kevinaboos: I agree with that and have considered storing a raw `usize` in the virtual address field of the `LoadedSection` type, considering how TLS (and now CLS) works. I never got around to doing it since most of the sections use a regular virtual address, but yeah, we should probably do that. That change will propagate to the `write_relocation` function too.

For now, go ahead and keep it as is, but can you open an issue with a reference to this comment chain so we don't forget about it? I can implement that later.

kernel/nano_core/linker_higher_half-aarch64.ld
kevinaboos: It does explain it, but it's still worth adding a comment such that when someone is reading or modifying the linker scripts across both architectures, we can prevent their likely confusion/surprise by the inconsistency there. Plus, without a comment, there's no way to know that they should dig deeply into the code of that cls_macros crate.

tsoutsman: Yep I added a comment to the linker script, but didn't change anything in `cls_macros`.

kernel/local_storage_initializer/src/lib.rs
kevinaboos: I see, thanks. This seems a little bit overly complex; I'm not sure why we need the extra complexity as compared to TLS (aside from, of course, needing to swap the new CLS data region into the list of all CLS data regions.

But no real need to change it, if it works it's fine.

tsoutsman: Well because we are appending new data onto the active CLS image. The old part of the CLS image has already been used, so its data will be in some different state that we need to copy over into the new image.

Not entirely sure if TLS does this/how it does this.

kernel/cls/cls_macros/src/lib.rs
```diff
@@ -400,15 +400,8 @@ fn cls_offset_expr(name: &Ident) -> proc_macro2::TokenStream {
...
-                            // TODO: Use `next_multiple_of` when stabilised.
-                            let remainder = cls_size % 0x1000;
-                            if remainder == 0 {
-                                cls_size
-                            } else {
-                                cls_size + 0x1000 - remainder
-                            }
-                        };
+                        // TODO: Use `next_multiple_of(0x1000)` when stabilised.
+                        let cls_start_to_tls_start = (cls_size + 0xfff) & !0xfff;
```
kevinaboos: ah, should've said this the first time, but we should stay away from magic numbers. Use `PAGE_SIZE - 1` instead to make it clear that we're aligning up to the next 4k page boundary due to the linker script section alignment.

Also

kernel/cls_allocator/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+/// Generates a new data image for the current CPU and sets the CLS register
+/// accordingly.
+pub fn reload_current_cpu() {
+    let current_cpu = cpu::current_cpu();
+
+    let mut data = CLS_INITIALIZER.lock().get_data();
+
+    let mut sections = CLS_SECTIONS.lock();
```
kevinaboos: in reading this function, i'm getting mildly confused by the terminology used and how it differs from TLS. 

This `CLS_SECTIONS` isn't a set of sections, right? It's a set of data images, according to your type def of it, so i think it's name should be changed to reflect that.

My understanding of the terminology is as follows (and tbh i'm likely not 100% consistent with this, but we should be moving forward):
* a TLS/CLS data image is the _starting_ value that gets populated by the content of `.tdata`/`.cls` sections. The term image is fitting, as we consider it read-only from the perspective of executing code... though it will be changed when dynamically loading a new section.
* a TLS/CLS data region or area is an instance of a data image that was allocated for and given to a specific task or CPU. 
* a TLS/CLS section is actually a symbol, but yea, that one is clear.

kernel/cls/cls_macros/src/lib.rs
```diff
@@ -148,19 +144,27 @@ pub fn cpu_local(args: TokenStream, input: TokenStream) -> TokenStream {
...
     let ref_expr = quote! {
         {
+            let offset = #offset_expr;
             #[cfg(target_arch = "x86_64")]
             let mut ptr = {
                 use cls::__private::x86_64::registers::segmentation::{GS, Segment64};
-                GS::read_base().as_u64()
+                let gs = GS::read_base().as_u64();
+
+                // If we were statically linked on x64, `offset` will be negative.
```
kevinaboos: minor clarification
```suggestion
                // If this CLS section was statically linked, its `offset` will be negative on x86_64 only.
```

kernel/cls/cls_macros/src/lib.rs
```diff
@@ -264,3 +273,221 @@ pub fn cpu_local(args: TokenStream, input: TokenStream) -> TokenStream {
     }
     .into()
 }
+
+fn cls_offset_expr(name: &Ident) -> proc_macro2::TokenStream {
+    quote! {
+        {
+            // There are two ways in which a crate can be linked:
```
kevinaboos: this documentation is really awesome, I think it should be promoted to a top crate-level doc so it gets rendered by rustdoc, which also allows us to link to it.

The same goes for the comment blocks starting at line 324 and 431.

Then, you can reference the crate-level docs with inline code comments here.

kevinaboos (changes requested):
just a few more minor suggestions.

kernel/local_storage_initializer/src/lib.rs
kevinaboos: Right, so I think TLS is similar but simpler overall. The (old) TLS initializer created and maintained a single global image, which it then hands out cloned instances of to each new task.

But yes, in this case, we do need to preserve the modified data on a per-CPU basis, which is an additional requirement beyond TLS. TLS doesn't need that due to the fact that each new task will only be created _after_ its crate (and its dependencies) have been linked in, which includes all of its TLS sections being loaded. There's no case (yet) where a task would suddenly need to expand it TLS data area in order to access sections that weren't loaded at the time when it initially began executing.

I think my original comment came from my old assumption/mental model about how CLS worked, from the old days of the `PerCpuData` struct that was stored in a dedicated 1-page `MappedPages` for each CPU -- I was expecting to see a single (or two) pages per CPU that stored its CLS data, and would only permit growing (not removing sections). But that approach isn't unlimited since it's restricted by the remaining space available in the page(s).

kernel/cls_allocator/src/lib.rs
tsoutsman: > This CLS_SECTIONS isn't a set of sections, right?

Yea that's a mistake on my part.

> a TLS/CLS data image is the starting value that gets populated by the content of .tdata/.cls sections. The term image is fitting, as we consider it read-only from the perspective of executing code... though it will be changed when dynamically loading a new section.

I don't think that's true. When we set an image as the current base (for both TLS/CLS) it can be modified by the process/cpu implicitly through TLS/CLS accesses. The `Tls/ClsInitializer` is what stores the starting value.

> a TLS/CLS data region or area is an instance of a data image that was allocated for and given to a specific task or CPU.

No, that's a data image.

kevinaboos: hmm, did i mix them up? That's how I recall reading about TLS images. 

Perhaps I created that set of terminology and it's not canonical? My above description matches how I use that language in Theseus's TLS-related crates. I mean, overall it's just a semantic choice to differentiate between "data image" and "data area/region", but we just need to be consistent. 

If you have a better suggestion for clear terminology here, I'm totally open to it. In my mind, an "image" is a sort of snapshot of something, so I wouldn't expect it to be modified in general usage, but rather just serve as a starting point. Kinda COW-style sematics.

kevinaboos: Looks like the best name for the image is "TLS initial[izer] image".  I don't see a canonical name for the data area instance owned by a specific task or CPU.

> I don't think that's true. When we set an image as the current base (for both TLS/CLS) it can be modified by the process/cpu implicitly through TLS/CLS accesses. The `Tls/ClsInitializer` is what stores the starting value.

Sorry, forgot to respond. What I said was correct -- the behavior currently on `theseus_main` is to call [`TlsInitializer::get_data()`](https://github.com/theseus-os/Theseus/blob/27429e57120be7b6725101fde4582dfd48b936d0/kernel/tls_initializer/src/lib.rs#L250C49-L251C1) to obtain a new instance of the TLS data image for a newly-spawned task, and the FSBASE register is set to point to that new instance, not the single instance cached in the `TlsInitializer`.


Looking at that linked code, I see the confusion. That `get_data()` function returns a `TlsDataImage`, which I refer to as a "TLS data area" -- it is precisely at that point (the call to `get_data()`) that I consider it to have become a TLS data area (or whatever we want to name it), as it is now specific to a task and can be modified by regular `%fs:[OFFSET]` access instructions.

kevinaboos: The [Fuchsia docs](https://fuchsia.dev/fuchsia-src/development/kernel/threads/tls) refer to each thread's instance as a "TLS Region", which itself is composed of multiple "TLS Block"s. 

We don't care about the distinction between various TLS Blocks since we only support the Local Exec TLS model. 

Let's go with `TLS Initial Image` and `TLS Region`.... later, in a future PR. We'll fix it up then for the sake of consistency.

kevinaboos: So, long story short, this indicates that `CLS_SECTIONS` should be called `CLS_REGIONS`.

kernel/cls/cls_macros/src/lib.rs
tsoutsman: > ah, should've said this the first time, but we should stay away from magic numbers. Use PAGE_SIZE - 1 instead to make it clear that we're aligning up to the next 4k page boundary due to the linker script section alignment.

👍 

> Also

Seems like you were going to write something else that got cut off.

kevinaboos: oops, just ignore it. thanks!

kevinaboos (approved):
tested and working!

PR #1031, opened 2023-08-28
Size-aware memory types

NIMogen:
Draft PR for huge pages without changes to the mapper.


kernel/memory_structs/src/lib.rs
```diff
@@ -24,6 +26,53 @@ use paste::paste;
...
+pub enum MemChunkSize {
+    Normal4K,
+    Huge2M,
+    Huge1G,
+}
+
+/// Trait used to group the size marker structs that paging-related types are parameterized with.
+pub trait PageSize: Ord + PartialOrd + Clone {
+    const SIZE: MemChunkSize;
+    const CONTAINED_4K: usize;
```
kevinaboos: nit: i find this name confusing, it'd probably be clearer to go with something like `SIZE_IN_4K_PAGES` or `NUM_4K_PAGES`

kernel/memory_structs/src/lib.rs
```diff
@@ -24,6 +26,53 @@ use paste::paste;
 use derive_more::*;
 use range_inclusive::{RangeInclusive, RangeInclusiveIterator};
 
+/// Numeric constants associated with different page types.
+pub const PAGE_4KB_SIZE: usize = 1 << 12;
+pub const PAGE_2MB_SIZE: usize = (1 << 12) * 512;
+pub const PAGE_1GB_SIZE: usize = (1 << 12) * 512 * 512;
```
kevinaboos: Doc comments only apply to the single item immediately following them, so you need to do this:
```suggestion
/// The size in bytes of a normal 4KiB page: 4096 bytes.
pub const PAGE_4KB_SIZE: usize = 1 << 12;
/// The size in bytes of a huge 2MiB page: 2097152 bytes.
pub const PAGE_2MB_SIZE: usize = (1 << 12) * 512;
/// The size in bytes of a normal 1GiB page: 1073741824 bytes.
pub const PAGE_1GB_SIZE: usize = (1 << 12) * 512 * 512;
```

Can you also use the definitions from the `kernel_config::memory` module:
* There is a const called `PAGE_SIZE` which is the same as your `PAGE_4KB_SIZE`
  * we can and will rename that `PAGE_SIZE` definition later, just in a separate PR
* There is a const called `ENTRIES_PER_PAGE_TABLE`, which has the value of 512.

Take a look at that module and see if any other consts better represent the semantics behind how you calculate the page size consts here.

kernel/memory_structs/src/lib.rs
```diff
@@ -24,6 +26,53 @@ use paste::paste;
...
+
+/// Enum used to indicate the size of a page or frame.
+#[derive(Debug)]
+pub enum MemChunkSize {
+    Normal4K,
+    Huge2M,
+    Huge1G,
+}
+
+/// Trait used to group the size marker structs that paging-related types are parameterized with.
```
kevinaboos: Your current docs describe how it's used, not really what it means. Suggestion below:
```suggestion
/// Trait used to represent the size of a page or frame, i.e., for normal or huge pages.
///
/// This is used to parameterize `Page`- and `Frame`-related types with a page size,
/// in order to define normal and huge pages in a generic manner.
```

kernel/memory_structs/src/lib.rs
```diff
@@ -24,6 +26,53 @@ use paste::paste;
...
+    const SIZE_IN_BYTES: usize = PAGE_SIZE;
+}
+
+/// Marker struct used to indicate a page size of 2MiB.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct Page2MiB;
+impl PageSize for Page2MiB {
+    const SIZE: MemChunkSize = MemChunkSize::Huge2M;
+    const CONTAINED_4K: usize = 512;
+    const SIZE_IN_BYTES: usize = PAGE_SIZE * 512;
```
kevinaboos: these consts (and the ones below in `Page1GiB`) are redundant with the const defintiions up top. You can either re-use the ones defined up top here, or just use these consts as the only consts and remove the ones up top.

kernel/memory_structs/src/lib.rs
```diff
@@ -266,87 +315,145 @@ macro_rules! implement_page_frame {
 
             #[doc = "A `" $TypeName "` is a chunk of **" $desc "** memory aligned to a [`PAGE_SIZE`] boundary."]
             #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-            pub struct $TypeName {
+            pub struct $TypeName<P: PageSize = Page4KiB> {
                 number: usize,
+                pub size: PhantomData::<P>,
```
kevinaboos: no need for this to be public.

kernel/memory_structs/src/lib.rs
```diff
@@ -266,87 +315,145 @@ macro_rules! implement_page_frame {
...
             impl $TypeName {
-                #[doc = "Returns the `" $address "` at the start of this `" $TypeName "`."]
-                pub const fn start_address(&self) -> $address {
-                    $address::new_canonical(self.number * PAGE_SIZE)
+                #[doc = "Returns the `" $TypeName "` containing the given `" $address "`."]
+                pub const fn containing_address(addr: $address) -> $TypeName {
+                    $TypeName {
+                        number: addr.value() / PAGE_SIZE,
+                        size: PhantomData::<Page4KiB>,
+                    }
```
kevinaboos: you can still make this generic, and combine it with both `containing_address_2mb` and `containing_address_1gb` into one single function.

Just put in the `impl<P: PageSize> $TypeName<P>` block down below, and then use the `P::SIZE` and other constants for the math to calculate the `number` correctly.

Also, with that, you shouldn't need to specify anything in the PhantomData type; I _think_ you can just do:
```rust
    size: PhantomData,
```
in the constructor
```rust

Comment (kevinaboos) [reply kernel/memory_structs/src/lib.rs]: In this way, there wouldn't be separate/dedicated functions for the 2mb and 1gb versions. The caller of this function would just specify what kind of page they want with type annotations, which is canonically Rusty:
```rust
let huge_2mb_page = Page::<Page2MiB>::containing_address(...);
let huge_1gb_page = Page::<Page1GiB>::containing_address(...);
```

kernel/memory_structs/src/lib.rs
```diff
@@ -266,87 +315,145 @@ macro_rules! implement_page_frame {
...
                 }
-
+            }
+            impl<P> $TypeName<P> where P: PageSize + 'static {
```
kevinaboos: does the compiler require these `'static` lifetime annotations? If so, that's fine, but I see some places without them. Let's make it consistent across this entire file.

kernel/memory_structs/src/lib.rs
```diff
@@ -266,87 +315,145 @@ macro_rules! implement_page_frame {
...
-                        number: self.number.saturating_sub(rhs),
+                        number: self.number.saturating_sub(rhs * P::CONTAINED_4K),
+                        size: self.size,
                     }
                 }
             }
             impl SubAssign<usize> for $TypeName {
                 fn sub_assign(&mut self, rhs: usize) {
                     *self = $TypeName {
                         number: self.number.saturating_sub(rhs),
```
kevinaboos: this isn't correct any more.

kernel/memory_structs/src/lib.rs
```diff
@@ -266,87 +315,145 @@ macro_rules! implement_page_frame {
...
+                        size: self.size,
                     }
                 }
             }
-            impl AddAssign<usize> for $TypeName {
+            impl<P: PageSize> AddAssign<usize> for $TypeName<P> {
                 fn add_assign(&mut self, rhs: usize) {
                     *self = $TypeName {
-                        number: core::cmp::min(MAX_PAGE_NUMBER, self.number.saturating_add(rhs)),
+                        number: core::cmp::min(MAX_PAGE_NUMBER, self.number.saturating_add(rhs * P::CONTAINED_4K)),
```
kevinaboos: Not sure why I didn't do this originally, but in `AddAssign` and `SubAssign` we should be able to re-use the logic from `Add` and `Sub` respectively.

For `Add`, something like:
```rust
    *self = (*self).add(rhs);
```

kernel/memory_structs/src/lib.rs
kevinaboos: same comment here as above

kernel/memory_structs/src/lib.rs
```diff
@@ -266,87 +315,145 @@ macro_rules! implement_page_frame {
...
-                    Step::steps_between(&start.number, &end.number)
+                fn steps_between(start: &$TypeName<P>, end: &$TypeName<P>) -> Option<usize> {
+                    match Step::steps_between(&start.number, &end.number) {
+                        None => {
+                            None
+                        }
+                        Some(n) => {
+                            Some(n / P::CONTAINED_4K)
+                        }
+                    }
```
kevinaboos:
```suggestion
                    Step::steps_between(&start.number, &end.number).map(|n| n / P::CONTAINED_4K)
```

kernel/memory_structs/src/lib.rs
```diff
@@ -266,87 +315,145 @@ macro_rules! implement_page_frame {
...
-                fn forward_checked(start: $TypeName, count: usize) -> Option<$TypeName> {
-                    Step::forward_checked(start.number, count).map(|n| $TypeName { number: n })
+                fn forward_checked(start: $TypeName<P>, count: usize) -> Option<$TypeName<P>> {
+                    Step::forward_checked(start.number, count +  P::CONTAINED_4K).map(|n| $TypeName { number: n, size: PhantomData })
                 }
                 #[inline]
-                fn backward_checked(start: $TypeName, count: usize) -> Option<$TypeName> {
-                    Step::backward_checked(start.number, count).map(|n| $TypeName { number: n })
+                fn backward_checked(start: $TypeName<P>, count: usize) -> Option<$TypeName<P>> {
+                    Step::backward_checked(start.number, count +  P::CONTAINED_4K).map(|n| $TypeName { number: n, size: PhantomData })
```
kevinaboos: shouldn't these be `count * P::CONTAINED_4K` rather than plus?

kernel/memory_structs/src/lib.rs
```diff
@@ -266,87 +315,145 @@ macro_rules! implement_page_frame {
...
+                }
+            }
+            impl TryFrom<&$TypeName> for $TypeName<Page2MiB> {
```
kevinaboos: `TryFrom` is a "consuming" function, meaning that it should be implemented for `Page`, not `&Page`.

kernel/memory_structs/src/lib.rs
```diff
@@ -392,17 +499,12 @@ macro_rules! implement_page_frame_range {
                         
             #[doc = "A range of [`" $chunk "`]s that are contiguous in " $desc " memory."]
             #[derive(Clone, PartialEq, Eq)]
-            pub struct $TypeName(RangeInclusive<$chunk>);
+            pub struct $TypeName<P: PageSize = Page4KiB>(RangeInclusive<$chunk::<P>>);
 
             impl $TypeName {
-                #[doc = "Creates a new range of [`" $chunk "`]s that spans from `start` to `end`, both inclusive bounds."]
```
kevinaboos, line 398: Any reason why you reordered the functions in this impl block? Can you restore the ordering so the diff is more readable?

kernel/memory_structs/src/lib.rs
```diff
@@ -392,17 +499,12 @@ macro_rules! implement_page_frame_range {
                         
             #[doc = "A range of [`" $chunk "`]s that are contiguous in " $desc " memory."]
             #[derive(Clone, PartialEq, Eq)]
-            pub struct $TypeName(RangeInclusive<$chunk>);
+            pub struct $TypeName<P: PageSize = Page4KiB>(RangeInclusive<$chunk::<P>>);
 
             impl $TypeName {
```
kevinaboos: the reason these implementations are difficult is because you didn't parameterize this impl block with the `P` parameter. You need that here in order to create `PageRange`s of huge pages. Right now you only allow for creating ranges of 4K normal pages.

Once you do this, you'll see how the `empty()` function can be properly implemented (the starting page will have its `number` be `1 * P::CONTAINED_4K` instead of `11`, which doesn't make sense). Same goes for all the other functions.

kernel/memory_structs/src/lib.rs
```diff
@@ -473,45 +607,50 @@ macro_rules! implement_page_frame_range {
...
+            }
+            impl $TypeName<Page1GiB> {
+                #[doc = "Returns the number of [`" $chunk "`]s covered by this iterator.\n\n \
+                Use this instead of [`Iterator::count()`] method. \
+                This is instant, because it doesn't need to iterate over each entry, unlike normal iterators."]
+                pub const fn [<size_in_ $chunk:lower s>](&self) -> usize {
+                   // add 1 because it's an inclusive range; divide by 262144 because it's a 2MiB page
+                   (self.0.end().number + 1).saturating_sub(self.0.start().number) / (512 * 512)
+                }
+            }
```
kevinaboos: all of these functions can be made generic by using the `P` parameter. You don't need special functions for any specific page size.

kernel/memory_structs/src/lib.rs
```diff
@@ -527,6 +666,20 @@ macro_rules! implement_page_frame_range {
...
+                    self.0.iter()
+                }
+            }
+            impl IntoIterator for $TypeName<Page1GiB> {
+                type Item = $chunk<Page1GiB>;
+                type IntoIter = RangeInclusiveIterator<$chunk<Page1GiB>>;
+                fn into_iter(self) -> Self::IntoIter {
+                    self.0.iter()
+                }
+            }
```
kevinaboos: why does this have to be done specifically for these two page sizes? you can just do it generically like with all other functions.

If you're not sure of the syntax, you can do this:
```rust
impl<P: PageSize> IntoIterator for $TypeName<P> {
    ...
}
```

kevinaboos (changes requested):
For the `TryFrom` conversions, you have some of the cases covered, but not all. Specifically, if you think of it as a 3x3 table where you're converting a type on the left to the type on the top:
|                | Page4K | Page2M | Page 1G |
| -------- | -------- | -------- | --------  |
| Page4K | auto.      | fallible.   | fallible.   |
| Page2M | infallible | auto      | fallible.   |
| Page1G  | infallible | infallible | auto.     |

Notes:
* "auto" means it's covered by Rust's built-in blanket implementation, where `From` is always implemented for converting a type to itself.
* "fallible" means that the conversion may fail due to improper page `number` alignment, so you must implement that with `TryFrom`.
* "infallible" means that the conversion is guaranteed to succeed, so you can implement it with `From` instead of `TryFrom`.

----------
Aside from that, my main comment is to use generics _everywhere_. You're using them in about half of the places they could be used, which is good, but you shouldn't need to implement methods for specific `Page<Page2MiB>`/`Page<Page1GiB>` types hardly at all, except for the aforementioned conversion methods.

kernel/memory_structs/src/lib.rs
```diff
@@ -294,59 +331,163 @@ macro_rules! implement_page_frame {
...
+                        size: PhantomData
+                    }
+                }
+            }
+            impl From<$TypeName<Page2MiB>> for $TypeName {
+                fn from(p: $TypeName<Page2MiB>) -> Self { 
+                    Self {                             
+                        number: p.number,
+                        size: PhantomData
+                    }
```
kevinaboos: can these 3 be combined into one impl? or does Rust's trait rules prevent that

kernel/memory_structs/src/lib.rs
```diff
@@ -392,17 +533,12 @@ macro_rules! implement_page_frame_range {
...
             impl $TypeName {
-                #[doc = "Creates a new range of [`" $chunk "`]s that spans from `start` to `end`, both inclusive bounds."]
-                pub const fn new(start: $chunk, end: $chunk) -> $TypeName {
-                    $TypeName(RangeInclusive::new(start, end))
-                }
-
                 #[doc = "Creates a `" $TypeName "` that will always yield `None` when iterated."]
                 pub const fn empty() -> $TypeName {
-                    $TypeName::new($chunk { number: 1 }, $chunk { number: 0 })
+                    $TypeName::new($chunk { number: 11, size: PhantomData }, $chunk { number: 0, size: PhantomData })
```
kevinaboos: why `11` instead of `1`? what's the difference?

kernel/memory_structs/src/lib.rs
```diff
@@ -419,6 +555,12 @@ macro_rules! implement_page_frame_range {
                         $TypeName::new(start, end)
                     }
                 }
+            }
+            impl<P: PageSize + 'static + Copy> $TypeName<P> {
```
kevinaboos: is the `Copy` bound required here?

kernel/memory_structs/src/lib.rs
```diff
@@ -531,20 +679,46 @@ macro_rules! implement_page_frame_range {
...
-                fn from(cr: [<Copyable $TypeName>]) -> Self {
+            impl<P: PageSize + Copy + 'static> From<[<Copyable $TypeName>]<P>> for $TypeName<P> {
+                fn from(cr: [<Copyable $TypeName>]<P>) -> Self {
                     Self::new(cr.start, cr.end)
                 }
             }
+            impl TryFrom<$TypeName> for $TypeName<Page2MiB> {
+                type Error = &'static str;
+                fn try_from(p: $TypeName) -> Result<Self, &'static str> {
+                    if (p.[<size_in_ $chunk:lower s>]() - 1) % 512 == 0 {
```
kevinaboos: why subtract `1` here? I don't think that's correct.

In fact, I don't think you even need this conditional at all. As long as the `try_from` calls for the `start` and `end` page succeed, that alone is sufficient to ensure the page range can be validly converted into a 2MiB range.

kernel/memory_structs/src/lib.rs
```diff
@@ -531,20 +679,46 @@ macro_rules! implement_page_frame_range {
...
+                        ));
+                    } else {
+                        return Err("Could not convert 4KiB page range into 2MiB page range.");
+                    }
+                }
+            }
+            impl TryFrom<$TypeName> for $TypeName<Page1GiB> {
+                type Error = &'static str;
+                fn try_from(p: $TypeName) -> Result<Self, &'static str> {
+                    if (p.[<size_in_ $chunk:lower s>]() - 1) % (512 * 512) == 0 {
```
kevinaboos: same question here as above

kevinaboos (changes requested):
a few more comments

kernel/memory_structs/src/lib.rs
NIMogen: That wasn't intentional; I'll remove it. Must have been mistakenly added when adding `size: PhantomData` to everything.

kernel/memory_structs/src/lib.rs
NIMogen: The `Copy` bound is required, yes. I added the `'static` bound everywhere for consistencies sake, even where the compiler didn't request it. I only added the `Copy` bound where needed to avoid clutter, but I can add it everywhere for consistency just like `'static,` if that's preferred.

kevinaboos: Ok, but `Copy` _shouldn't_  be required. I think it's only necessary because of the way you implemented `AddAssign` and `SubAssign`, which is why I must not have implemented it that way in the past.

Where else (in what other methods) would `Copy` be specifically needed? In general we do not treat a range type as being `Copy`, since Rust doesn't (even though they easily could be).

NIMogen: Using the original implementation of  `AddAssign` and `SubAssign` _does_ make it so that the `Copy` bound isn't needed for those traits.  
But removing the bound here results in an error message arising from a move from a shared reference.  `move occurs because value has type Page<P>, which does not implement the Copy trait`, for `*self.start()` and `*self.end()` in `overlap()`.

kevinaboos: Ah ok, right. In that case, no worries. 

I misinterpreted the bound as applying to the range type, but it applies to the `Page` type. So feel free to ignore this.

Only thing to change is that Rust typically places trait bounds before lifetime bounds, i.e., `P: PageSize + Copy + 'static`. Also I still don't see why `'static` is necessary, but we can look at that later.

kernel/memory_structs/src/lib.rs
```diff
@@ -294,59 +331,134 @@ macro_rules! implement_page_frame {
...
                     }
                 }
             }
-            impl AddAssign<usize> for $TypeName {
+            impl<P: PageSize + Copy + 'static> AddAssign<usize> for $TypeName<P> {
                 fn add_assign(&mut self, rhs: usize) {
-                    *self = $TypeName {
-                        number: core::cmp::min(MAX_PAGE_NUMBER, self.number.saturating_add(rhs)),
-                    };
+                    *self = (*self).add(rhs * P::NUM_4K_PAGES)
```
kevinaboos: this isn't correct. We need to _either_:
* add `rhs * P::NUM_4K_PAGES` to the inner `number`
  * Doesn't require `Copy`
* add `rhs` to the outer `Page`, which will call your above `Add` implementation that handles the `NUM_4K_PAGES` multiplier properly.
  * Requires `Copy`

Right now you're doing both, which means you end up multiplying it twice accidentally, i.e., doing `number += rhs * P::NUM_4K_PAGES * P::NUM_4K_PAGES`

I assume this is just an accident since you have it right in `SubAssign` below

kernel/memory_structs/src/lib.rs
```diff
@@ -419,6 +526,12 @@ macro_rules! implement_page_frame_range {
                         $TypeName::new(start, end)
                     }
                 }
+            }
+            impl<P: PageSize + Copy + 'static> $TypeName<P> {
```
kevinaboos: most of the functions in this impl block do not require the `Copy` bound. Generally we want impl blocks to specify the _minimum_ trait bounds required for each function to work. 

You should only include the functions that require `Copy`ing the Page/Frame type in this impl block; all others should be moved to a separate impl block that doesn't have the `Copy` bound.

kernel/memory_structs/src/lib.rs
```diff
@@ -430,17 +543,17 @@ macro_rules! implement_page_frame_range {
                     This is instant, because it doesn't need to iterate over each entry, unlike normal iterators."]
                 pub const fn [<size_in_ $chunk:lower s>](&self) -> usize {
                     // add 1 because it's an inclusive range
-                    (self.0.end().number + 1).saturating_sub(self.0.start().number)
+                    (self.0.end().number + (1 * P::NUM_4K_PAGES)).saturating_sub(self.0.start().number) / P::NUM_4K_PAGES
```
kevinaboos: to me, this function should return the size in terms of the number `P`-sized pages, not size in the number of 4k pages.  For example, a range of two 2MiB huge pages should return `2` here, not `2*512`. This is consistent with how we expect iteration to work too.

That means you also need to change `size_in_bytes` to reflect this (use `P`)

kevinaboos (changes requested):
Left a few more comments with a couple tiny mistakes found. 

For these cases, your tests ought to catch such mistakes. If you don't have test cases to cover these, please add them and then ensure they all pass.

kernel/memory_structs/src/lib.rs
NIMogen: That's what `size_in_chunks` currently does. I have tests for this as well.

But yeah, I had forgotten to account for this in `size_in_bytes`. I'll make that change and add tests for it.

kernel/memory_structs/src/lib.rs
NIMogen: My bad. If we want to reduce the trait bounds for impl blocks to a minimum, I can adopt the strategy that doesn't require `Copy` for both.

kernel/memory_structs/src/lib.rs
kevinaboos: ah, yes, i misread it.  My mistake.

kernel/memory_structs/src/lib.rs
kevinaboos: no worries, it's up to you. The `Copy` bound isn't really a problem here since we only implement the `PageSize` trait for our own 3 marker types, which are all `Copy` anyway.

kevinaboos (approved):
Thanks very much Noah! Looking forward to seeing this in #1016!

PR #1035, opened 2023-09-04
Add `Scheduler` trait

tsoutsman:
Adds the `Scheduler` trait to `task` which can be used to dynamically switch between schedulers, akin to `SELECT_NEXT_TASK_FUNC` but with more functionality. Merges the `runqueue_*` and `scheduler_*` crates as a byproduct because now the state must be stored with policy.

The PR makes run queues (i.e. schedulers) more opaque by removing the ability to directly access a run queue. I think this is the way we want to go because it prevents bugs such as #1000. Realistically, users shouldn't have complete access over the run queue.

Also I'm not sure about the API for `task::scheduler`. I was thinking instead of the function variants (e.g. `add_task` `add_task_to`, `add_task_to_current`) to instead have something like
```rust
// Or some better name
pub enum SchedulerSpecifier {
    Any,
    Specific(CpuId),
    Current,
}

pub fn add_task(task: TaskRef, scheduler: SchedulerSpecifier) {
    // ...
}
```
but I'm also fine with the current API.

~Depends on #1036.~


tsoutsman:
Forgot to add the `single_simd_task_optimisation` config blocks back. Will add that tomorrow.

kernel/scheduler_epoch/src/lib.rs
```diff
@@ -12,170 +12,213 @@
...
-            // apic_id);
-            return None;
+    fn assign_tokens(&mut self) {
+        // We begin with total priorities = 1 to avoid division by zero
+        let mut total_priorities: usize = 1;
+
+        // This loop calculates the total priorities of the runqueue
+        for (_i, t) in self.queue.iter().enumerate() {
+            // we skip the idle task, it contains zero tokens as it is picked last
+            if t.is_an_idle_task {
```
kevinaboos: we no longer keep idle tasks on the runqueues as of #982, right? so this can be removed

kernel/scheduler_epoch/src/lib.rs
```diff
@@ -12,170 +12,213 @@
...
-                Some(x) => x.saturating_sub(1),
-                None => 0,
+
+        // We keep each epoch for 100 tokens by default
+        // However since this granularity could miss low priority tasks when
+        // many concurrent tasks are running, we increase the epoch in such cases
+        let epoch: usize = core::cmp::max(total_priorities, 100);
+
+        // We iterate through each task in runqueue
+        // We dont use iterator as items are modified in the process
```
kevinaboos: this comment confuses me

kernel/scheduler_epoch/src/lib.rs
```diff
@@ -12,170 +12,213 @@
...
+        for (_i, t) in self.queue.iter_mut().enumerate() {
+            // we give zero tokens to the idle tasks
+            if t.is_an_idle_task {
+                continue;
+            }
+
+            // we give zero tokens to none runnable tasks
```
kevinaboos:
```suggestion
            // we give zero tokens to non-runnable tasks
```

kernel/scheduler_priority/src/lib.rs
```diff
@@ -4,43 +4,157 @@
...
+    fn drain(&mut self) -> alloc::boxed::Box<dyn Iterator<Item = TaskRef> + '_> {
+        Box::new(self.queue.drain().map(|priority_task| priority_task.task))
+    }
+
+    fn dump(&self) -> Vec<TaskRef> {
+        self.queue
+            .clone()
+            .into_iter()
+            .map(|priority_task| priority_task.task)
+            .collect()
```
kevinaboos, line 88: for this and the other impls of `dump`, i'd vote to change the signature to return an iterator over `TaskRef`s as `Item`s rather than forcibly `collect()` it into a `Vec`.  No need to eagerly `collect` if the caller is just going to iterate anyway.

Could also re-name it more accurately to something like `task_iter`, or a similar name of your choosing.

kernel/scheduler_priority/src/lib.rs
```diff
@@ -4,43 +4,157 @@
...
+impl task::scheduler::PriorityScheduler for Scheduler {
+    fn set_priority(&mut self, task: &TaskRef, priority: u8) -> bool {
+        let previous_len = self.queue.len();
+        self.queue.retain(|t| t.task != *task);
+
+        if previous_len != self.queue.len() {
+            // We should have at most removed one task from the run queue.
+            debug_assert_eq!(self.queue.len() + 1, previous_len);
+            self.queue.push(PriorityTaskRef {
+                // TODO: Don't take reference?
```
kevinaboos: lazy is better, i suppose. Also since this is a trait, i'd vote to keep the argument as `&TaskRef` since other impls may not need to clone the taskref.

kernel/task/src/scheduler.rs
```diff
@@ -0,0 +1,296 @@
...
+    // If preemption was not previously enabled (before we disabled it above),
+    // then we shouldn't perform a task switch here.
+    if !preemption_guard.preemption_was_enabled() {
+        // trace!("Note: preemption was disabled on CPU {}, skipping scheduler.", cpu::current_cpu());
+        return false;
+    }
+
+    let cpu_id = preemption_guard.cpu_id();
+
+    let next_task = SCHEDULER.update(|scheduler| scheduler.as_ref().unwrap().lock().next());
```
kevinaboos: can use `update_guarded` here, right?

kevinaboos (changes requested):

kernel/scheduler_epoch/src/lib.rs
tsoutsman: tbh it confuses me as well. i just copied the code over from `runqueue_epoch` because I wanted to minimise the changes. I'll remove it.

kernel/scheduler_priority/src/lib.rs
tsoutsman: We can't return an associated type because then the trait wouldn't be object safe.

> Could also re-name it more accurately to something like task_iter, or a similar name of your choosing.

As I've stated above it can't return an iterator so I've renamed it to `tasks` because it returns a vec of the tasks in the run queue.

kernel/scheduler_priority/src/lib.rs
tsoutsman: Again this was verbatim copied over from `runqueue_priority`, but I agree.

kernel/scheduler_priority/src/lib.rs
kevinaboos: Right, I didn't mean to suggest using an associated type like my old scheduler refactor design, but rather an `impl Iterator<...>`. But I just realized that I had only recently heard about [RPITIT](https://github.com/rust-lang/rust/issues/91611) being supported in soon-to-come future versions of Rust, so I was mixing up the fact that RPITIT support is coming soon with the fact that it doesn't actually exist yet in the version of Rust we're pinned to.

kernel/scheduler_epoch/src/lib.rs
kevinaboos: ah right, thanks for fixing it in any case.

kernel/scheduler_priority/src/lib.rs
kevinaboos: Actually it appears that it is already supported on nightly -- but we can save that for another PR.

kevinaboos (approved):

PR #1036, opened 2023-09-06
Save `rflags` during context switch

tsoutsman:
There are no guarantees about the value of the interrupt flag when context switching. If the context switch is voluntary, i.e. a thread called `schedule`, interrupts will most likely be enabled, whereas if a thread is preempted, interrupts will be disabled. But this means that if a preempted thread A switches to a thread B that voluntarily yielded, thread B will return from the call to `schedule` with interrupts disabled.

The AArch64 code also needs to be modified but I'll leave that to @NathanRoyer.

May also want to mark `task_switch` as `inline(never)`?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3471,7 +3471,6 @@ dependencies = [
  "fault_crate_swap",
  "fault_log",
  "fs_node",
- "irq_safety",
  "lazy_static",
  "log",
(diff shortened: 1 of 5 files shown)


kevinaboos (approved):
Thanks very much to both of you for this changeset!

PR #1038, opened 2023-09-11
Fix section order on x86_64 when using older versions of `ld`

tsoutsman:
Older versions of `ld` place the `.cls`, `.tdata`, and `.tbss` sections first because they are used in linker script variable definitions. This creates problems because `memory_x86_64` assumes a specific order of the sections. The fix is the same as for AArch64 --- move the definitions to the end of the file.


kevinaboos (approved):
thanks a lot!

PR #1039, opened 2023-09-12
aarch64: support FIQs and use them for TLB shootdown IPIs

NathanRoyer:
Sometimes when we tried spawning an interactive console in theseus on aarch64, by sending keystrokes through the UART port, theseus hanged.

We then found out that a complicated scenario was taking place:
- the console task was spawned and scheduled but the sub-tasks it spawned in turn were not running
- in fact two of the four cores were stopped, so they were ghosting the tasks randomly assigned to them...
- they were stopped because they tried to broadcast an interrupt to other cores, which is an NMI on x86 but I had implemented it as a regular interrupt on aarch64, early in the port process, without leaving a clear TODO/warning... big mistake: the two missing cores were silently clashing on this
- I reimplemented this using fast interrupts, which highlighted some other bugs but ultimately fixed the issue

This brings the following changes:
- The interrupt controller API only contains cross-platform features, other features are not part of the trait. This made it easier to support FIQs on aarch64, but it could become a separate trait: Aarch64LocalInterruptController.
- Aarch64-specific methods of the LocalInterruptController allow configuring/handling fast interrupts (FIQs).
- The interrupt controller now allows enabling/disabling SPIs too (this was missing).
- The GIC driver was updated with Group 0 (used by FIQs) interrupt support.
- A bug in memory_aarch64 was fixed (a shift operation was missing).
- A silent bug recently introduced in context_switch_regular was fixed (incorrect structure size assumption, fixed by padding the structure, which makes the assembly and memory layout easier to understand).
- The `interrupts` crate was modified to handle FIQs.
- The TLB Shootdown crate now uses FIQs on aarch64 as an NMI alternative.
- nano_core, captain and ap_start were modified to set FIQ masks correctly.

This depend on [a PR in irq_safety](https://github.com/theseus-os/irq_safety/pull/10), so it doesn't build at this moment.

### DIFF
--- a/kernel/context_switch_regular/src/aarch64.rs
+++ b/kernel/context_switch_regular/src/aarch64.rs
@@ -22,6 +22,7 @@ pub struct ContextRegular {
 
     // only NZCV & DAIF bits are saved and restored
     pstate: usize,
+    padding: usize,
 }
 
 impl ContextRegular {
@@ -42,6 +43,7 @@ impl ContextRegular {
 
             // interrupts are initially unmasked/enabled
             pstate: 0,
+            padding: 0,
         }
     }
 
--- a/kernel/context_switch_regular/src/aarch64.rs
+++ b/kernel/context_switch_regular/src/aarch64.rs
@@ -22,7 +22,6 @@ pub struct ContextRegular {
 
     // only NZCV & DAIF bits are saved and restored
     pstate: usize,
-    padding: usize,
 }
 
 impl ContextRegular {
@@ -43,7 +42,6 @@ impl ContextRegular {
 
             // interrupts are initially unmasked/enabled
             pstate: 0,
-            padding: 0,
         }
     }
 
@@ -87,21 +85,21 @@ macro_rules! save_registers_regular {
         // Save all general purpose registers into the previous task.
         r#"
             // Make room on the stack for the registers.
-            sub sp,  sp,  #8 * 2 * 6
+            sub sp,  sp,  #8 * 13
 
             // Push registers on the stack, two at a time.
-            stp x19, x20, [sp, #8 * 2 * 0]
-            stp x21, x22, [sp, #8 * 2 * 1]
-            stp x23, x24, [sp, #8 * 2 * 2]
-            stp x25, x26, [sp, #8 * 2 * 3]
-            stp x27, x28, [sp, #8 * 2 * 4]
-            stp x29, x30, [sp, #8 * 2 * 5]
+            stp x19, x20, [sp, #8 * 0]
+            stp x21, x22, [sp, #8 * 2]
+            stp x23, x24, [sp, #8 * 4]
+            stp x25, x26, [sp, #8 * 6]
+            stp x27, x28, [sp, #8 * 8]
+            stp x29, x30, [sp, #8 * 10]
 
             // Push an OR of DAIF and NZCV flags of PSTATE
             mrs x29, DAIF
             mrs x30, NZCV
             orr x29, x29, x30
-            str x29, [sp, #8 * 2 * 6]
+            str x29, [sp, #8 * 12]
         "#
     );
 }
@@ -135,20 +133,20 @@ macro_rules! restore_registers_regular {
         r#"
             // Pop DAIF and NZCV flags of PSTATE
             // These MSRs discard irrelevant bits; no AND is required.
-            ldr x29, [sp, #8 * 2 * 6]
+            ldr x29, [sp, #8 * 12]
             msr DAIF, x29
             msr NZCV, x29
 
             // Pop registers from the stack, two at a time.
-            ldp x29, x30, [sp, #8 * 2 * 5]
-            ldp x27, x28, [sp, #8 * 2 * 4]
-            ldp x25, x26, [sp, #8 * 2 * 3]
-            ldp x23, x24, [sp, #8 * 2 * 2]
-            ldp x21, x22, [sp, #8 * 2 * 1]
-            ldp x19, x20, [sp, #8 * 2 * 0]
+            ldp x29, x30, [sp, #8 * 10]
+            ldp x27, x28, [sp, #8 * 8]
+            ldp x25, x26, [sp, #8 * 6]
+            ldp x23, x24, [sp, #8 * 4]
+            ldp x21, x22, [sp, #8 * 2]
+            ldp x19, x20, [sp, #8 * 0]
 
             // Move the stack pointer back up.
-            add sp,  sp,  #8 * 2 * 6
+            add sp,  sp,  #8 * 13
         "#
     );
 }
--- a/kernel/gic/src/gic/cpu_interface_gicv2.rs
+++ b/kernel/gic/src/gic/cpu_interface_gicv2.rs
@@ -8,6 +8,7 @@
 
 use super::Priority;
 use super::InterruptNumber;
+use super::SPURIOUS_INTERRUPT_NUM;
 
 use volatile::{Volatile, ReadOnly, WriteOnly};
 use zerocopy::FromBytes;
@@ -97,7 +98,7 @@ impl CpuRegsP1 {
         let priority = self.running_prio.read() as u8;
 
         match int_num {
-            1023 => None,
+            SPURIOUS_INTERRUPT_NUM => None,
             _ => Some((int_num, priority))
         }
     }
--- a/kernel/gic/src/gic/cpu_interface_gicv2.rs
+++ b/kernel/gic/src/gic/cpu_interface_gicv2.rs
@@ -42,7 +42,7 @@ pub struct CpuRegsP1 {            // base offset
 }
 
 // enable group 0
-// const CTLR_ENGRP0: u32 = 0b01;
+const CTLR_ENGRP0: u32 = 0b01;
 
 // enable group 1
 const CTLR_ENGRP1: u32 = 0b10;
@@ -51,6 +51,7 @@ impl CpuRegsP1 {
     /// Enables routing of group 1 interrupts for the current CPU.
     pub fn init(&mut self) {
         let mut reg = self.ctlr.read();
+        reg |= CTLR_ENGRP0;
         reg |= CTLR_ENGRP1;
         self.ctlr.write(reg);
     }
@@ -87,12 +88,17 @@ impl CpuRegsP1 {
     ///
     /// This tells the GIC that the requested interrupt is being
     /// handled by this CPU.
-    pub fn acknowledge_interrupt(&mut self) -> (InterruptNumber, Priority) {
+    ///
+    /// Returns None if a spurious interrupt is detected.
+    pub fn acknowledge_interrupt(&mut self) -> Option<(InterruptNumber, Priority)> {
         // Reading the interrupt number has the side effect
         // of acknowledging the interrupt.
         let int_num = self.acknowledge.read() as InterruptNumber;
         let priority = self.running_prio.read() as u8;
 
-        (int_num, priority)
+        match int_num {
+            1023 => None,
+            _ => Some((int_num, priority))
+        }
     }
 }
--- a/kernel/gic/src/gic/cpu_interface_gicv3.rs
+++ b/kernel/gic/src/gic/cpu_interface_gicv3.rs
@@ -12,6 +12,7 @@ use super::IpiTargetCpu;
 use super::Priority;
 use super::InterruptNumber;
 use super::InterruptGroup;
+use super::SPURIOUS_INTERRUPT_NUM;
 
 const SGIR_TARGET_ALL_OTHER_PE: u64 = 1 << 40;
 const IGRPEN_ENABLED: u64 = 1;
@@ -95,9 +96,9 @@ pub fn acknowledge_interrupt(group: InterruptGroup) -> Option<(InterruptNumber, 
 
     let int_num = int_num & 0xffffff;
     let priority = priority & 0xff;
-    match int_num {
-        1023 => None,
-        _ => Some((int_num as InterruptNumber, priority as u8))
+    match int_num as InterruptNumber {
+        SPURIOUS_INTERRUPT_NUM => None,
+        n => Some((n, priority as u8)),
     }
 }
 
--- a/kernel/gic/src/gic/cpu_interface_gicv3.rs
+++ b/kernel/gic/src/gic/cpu_interface_gicv3.rs
@@ -11,6 +11,7 @@ use core::arch::asm;
 use super::IpiTargetCpu;
 use super::Priority;
 use super::InterruptNumber;
+use super::InterruptGroup;
 
 const SGIR_TARGET_ALL_OTHER_PE: u64 = 1 << 40;
 const IGRPEN_ENABLED: u64 = 1;
@@ -28,7 +29,7 @@ pub fn init() {
 
     // Enable Group 0
     // bit 0 = group 0 enable
-    // unsafe { asm!("msr ICC_IGRPEN0_EL1, {}", in(reg) IGRPEN_ENABLED) };
+    unsafe { asm!("msr ICC_IGRPEN0_EL1, {}", in(reg) IGRPEN_ENABLED) };
 
     // Enable Groupe 1 (non-secure)
     // bit 0 = group 1 (non-secure) enable
@@ -61,9 +62,13 @@ pub fn set_minimum_priority(priority: Priority) {
 /// the current CPU.
 ///
 /// This implies that the CPU is ready to process interrupts again.
-pub fn end_of_interrupt(int: InterruptNumber) {
+pub fn end_of_interrupt(int: InterruptNumber, group: InterruptGroup) {
     let reg_value = int as u64;
-    unsafe { asm!("msr ICC_EOIR1_EL1, {}", in(reg) reg_value) };
+    
+    match group {
+        InterruptGroup::Group0 => unsafe { asm!("msr ICC_EOIR0_EL1, {}", in(reg) reg_value) },
+        InterruptGroup::Group1 => unsafe { asm!("msr ICC_EOIR1_EL1, {}", in(reg) reg_value) },
+    }
 }
 
 /// Acknowledge the currently serviced interrupt and fetches its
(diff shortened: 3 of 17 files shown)


kernel/context_switch_regular/src/aarch64.rs
```diff
@@ -22,6 +22,7 @@ pub struct ContextRegular {
 
     // only NZCV & DAIF bits are saved and restored
     pstate: usize,
+    padding: usize,
```
kevinaboos: why do we need this?

kernel/gic/src/gic/cpu_interface_gicv2.rs
```diff
@@ -87,12 +88,17 @@ impl CpuRegsP1 {
...
+    /// Returns None if a spurious interrupt is detected.
+    pub fn acknowledge_interrupt(&mut self) -> Option<(InterruptNumber, Priority)> {
         // Reading the interrupt number has the side effect
         // of acknowledging the interrupt.
         let int_num = self.acknowledge.read() as InterruptNumber;
         let priority = self.running_prio.read() as u8;
 
-        (int_num, priority)
+        match int_num {
+            1023 => None,
```
kevinaboos: avoid magic numbers like this, use a `const` instead.

kernel/gic/src/gic/cpu_interface_gicv3.rs
```diff
@@ -61,34 +62,47 @@ pub fn set_minimum_priority(priority: Priority) {
...
     unsafe {
-        asm!("mrs {}, ICC_IAR1_EL1", out(reg) int_num);
         asm!("mrs {}, ICC_RPR_EL1", out(reg) priority);
     }
 
     let int_num = int_num & 0xffffff;
     let priority = priority & 0xff;
-    (int_num as InterruptNumber, priority as u8)
+    match int_num {
+        1023 => None,
```
kevinaboos: const instead of magic number

kernel/gic/src/gic/mod.rs
```diff
@@ -19,6 +19,14 @@ use dist_interface::{DistRegsP1, DistRegsP6};
 use cpu_interface_gicv2::CpuRegsP1;
 use redist_interface::{RedistRegsP1, RedistRegsSgiPpi};
 
+#[repr(u32)]
+pub enum InterruptGroup {
+    // FIQs
+    Group0 = 0,
+    // IRQs
+    Group1 = 1,
```
kevinaboos: here (and in lots of other places), you should use `///` triple slash rustdoc comments so the docs are actually generated correctly. `//` double slash comments are only for internal comments about general behavior of code blocks.

kernel/gic/src/gic/mod.rs
```diff
@@ -241,18 +249,18 @@ impl ArmGicDistributor {
...
-        self.distributor().is_spi_enabled(int)
+        self.distributor().get_spi_state(int)
     }
 
     /// Enables or disables the forwarding of the given interrupt
     /// by the distributor.
     ///
     /// Panics if `int` is not in the SPI range (>= 32).
-    pub fn set_spi_state(&mut self, int: InterruptNumber, enabled: Enabled) {
+    pub fn set_spi_state(&mut self, int: InterruptNumber, state: Option<InterruptGroup>) {
```
kevinaboos, line 273: previously this function and `get_spi_state()` was self-explanatory, but now the `state` parameter is not really clear.

Is it something like:
* `None`: not enabled at all
* `Some(Group0)`: enabled for Group 0
* `Some(Group1)`: enabled for Group 1

Is is possible for group 0 and group 1 to be enabled? I think the issue is that it's not clear what a group is -- you should document that as well.

If so, you could consider adding a `Neither` variant to the `InterruptGroup` enum, which would make this clearer, or if `InterruptGroup` is used in other ways, you can create another enum called `InterruptState`/`InterruptGroupState` or something that has those 3 variants.

kernel/interrupt_controller/src/aarch64.rs
```diff
@@ -69,7 +72,10 @@ pub struct SystemInterruptController(IrqSafeMutex<ArmGicDistributor>);
 /// Struct representing per-cpu-core interrupt controller chips.
 ///
 /// On aarch64 w/ GIC, this corresponds to a Redistributor & CPU interface.
-pub struct LocalInterruptController(IrqSafeMutex<ArmGicCpuComponents>);
+pub struct LocalInterruptController(UnsafeCell<IrqSafeMutex<ArmGicCpuComponents>>);
```
kevinaboos, line 75: Why use `UnsafeCell`; surely that isn't the best way to implement this? Cells operate primarily on Copy-like types, at least safely. I think it'd make more sense here to use a `force_unlock()` function, which is what most code does when it needs to get around Rust's language-level knowledge of lock safety.

Also, as an aside, we no longer want to use `irq_safety` directly for new code; instead we should use the `sync_irq` crate which is a bit more robust. But we can address that later.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -187,21 +184,18 @@ pub fn init_timer(timer_tick_handler: InterruptHandler) -> Result<(), &'static s
     Ok(())
 }
 
-/// This function registers an interrupt handler for an inter-processor interrupt
+/// This function registers an interrupt handler for the TLB Shootdown IPI
 /// and handles interrupt controller configuration for that interrupt.
-pub fn setup_ipi_handler(handler: InterruptHandler, local_num: InterruptNumber) -> Result<(), &'static str> {
```
kevinaboos, line 192: don't we still want this functionality? for ex, if i wanted to create another IPI used for something like inter-CPU scheduling or waking up a sleeping CPU

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -48,7 +48,7 @@ const MAX_IRQ_NUM: usize = 256;
 // it's an array of function pointers which are meant to handle IRQs.
 // Synchronous Exceptions (including syscalls) are not IRQs on aarch64;
 // this crate doesn't expose any way to handle them at the moment.
-static IRQ_HANDLERS: IrqSafeRwLock<[InterruptHandler; MAX_IRQ_NUM]> = IrqSafeRwLock::new([default_irq_handler; MAX_IRQ_NUM]);
+static IRQ_HANDLERS: IrqSafeRwLock<[Option<InterruptHandler>; MAX_IRQ_NUM]> = IrqSafeRwLock::new([None; MAX_IRQ_NUM]);
```
kevinaboos, line 51: I think we still want to have default interrupt handlers for all IRQs, right? What's the benefit of this change?

kevinaboos: Ok, i see where you print the unhandled IRQs. This change is still a bit worse though, since `Option<InterruptHandler>` uses more memory than just `InterruptHandler`... but it may be faster? hmm.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -276,27 +267,27 @@ pub fn register_interrupt(int_num: InterruptNumber, func: InterruptHandler) -> R
...
+        Err(handler)
+    } else {
+        handlers[index] = None;
+        Ok(())
     }
 }
 
-/// Broadcast an Inter-Processor Interrupt to all other
-/// cores in the system
-pub fn send_ipi_to_all_other_cpus(irq_num: InterruptNumber) {
```
kevinaboos, line 297: same comment here -- we still want IPI ability in addition to TLB shootdowns. The TLB shootdown is just one kind of IPI that we can send.

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -418,17 +409,49 @@ extern "C" fn current_elx_synchronous(e: &mut ExceptionContext) {
 extern "C" fn current_elx_irq(exc: &mut ExceptionContext) {
     let (irq_num, _priority) = {
         let int_ctrl = LocalInterruptController::get();
-        int_ctrl.acknowledge_interrupt()
+        match int_ctrl.acknowledge_interrupt() {
+            Some(irq_prio_tuple) => irq_prio_tuple,
+            None /* spurious interrupt */ => return,
```
kevinaboos, line 452: are you receiving spurious interrupts now? There's an awful lot of added complexity to deal with that now, and to me it sounds like something is buggy if we're just now getting lots of them but we didn't before.

kevinaboos: Spurious interrupts are a special case on x86; one would hope aarch64 doesn't suffer from that same problem.

kernel/interrupts/src/aarch64/table.s
```diff
@@ -77,9 +77,10 @@ __exception_vector_start:
 .org 0x200
     CALL_WITH_CONTEXT current_elx_synchronous
 .org 0x280
+    msr daifclr, #1
```
kevinaboos, line 81: is this needed? on x86 interrupts are disabled automatically by the hardware when entering an interrupt handler. Does aarch64 behave differently?

And if this is required, how do interrupts get re-enabled? Does the hardware do that automatically? Seems weird that it would do it one way but not the other.

kernel/memory_aarch64/src/lib.rs
```diff
@@ -16,24 +16,52 @@ use pte_flags::PteFlags;
...
 use kernel_config::memory::KERNEL_OFFSET;
 
-#[cfg(any(target_arch = "aarch64", doc))]
+#[cfg(target_arch = "aarch64")]
 use core::arch::asm;
 
 const THESEUS_ASID: u16 = 0;
 
-#[cfg(any(target_arch = "aarch64", doc))]
+/* allow attribute is there for cross-platform doc building */
```
kevinaboos: i think you can use something like `#[cfg_attr(doc, unused_variables)]` for this purpose; i forget the exact syntax.

kernel/memory_aarch64/src/lib.rs
```diff
@@ -16,24 +16,52 @@ use pte_flags::PteFlags;
...
-    unsafe { asm!("tlbi vae1, {}", in(reg) vaddr.value()) };
+    unsafe {
+        // unsure here: where should the original address ASID go?
+        // it's zero in theseus so it's not important for us
+
+        // about the 48 bit shift:
+        // If the implementation supports 16 bits of ASID, then the
+        // upper 8 bits of the ASID must be written to 0 by software
+        // when the context being invalidated only uses 8 bits
+        let value = ((THESEUS_ASID as usize) << 48) | (vaddr.value() >> 12);
```
kevinaboos: this isn't unsafe, can be moved out of the `unsafe` block

kernel/memory_aarch64/src/lib.rs
```diff
@@ -16,24 +16,52 @@ use pte_flags::PteFlags;
...
-///         using the supplied address
+/// "va" => all translations at execution level using the supplied address
 /// "e1" => execution level
+#[allow(unused_variables)]
 pub fn tlb_flush_virt_addr(vaddr: VirtualAddress) {
     #[cfg(target_arch = "aarch64")]
-    unsafe { asm!("tlbi vae1, {}", in(reg) vaddr.value()) };
+    unsafe {
+        // unsure here: where should the original address ASID go?
+        // it's zero in theseus so it's not important for us
```
kevinaboos: no clue, but your comment is fine.

kernel/tlb_shootdown/src/lib.rs
```diff
@@ -40,6 +40,7 @@ pub fn init() {
 pub fn handle_tlb_shootdown_ipi() -> bool {
     let pages_to_invalidate = TLB_SHOOTDOWN_IPI_PAGES.read().clone();
     if let Some(pages) = pages_to_invalidate {
+        // THIS LOG USAGE CAN CAUSE A DEADLOCK ON AARCH64
```
kevinaboos: it can cause a deadlock on x86 too, actually, since it's also executing in an NMI context.

```suggestion
        // Note: logging in a NMI (x86_64) or FIQ (aarch64) context can cause deadlock,
        // so this should only be used sparingly to help debug problems with TLB shootdowns.
```

kevinaboos (changes requested):
Dealing with FIQs is kind of annoying. I'm wondering if we should just create a separate/dedicated struct or series of functions for handling FIQs that doesn't go through the same LocalInterruptController API that regular IRQs do.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: > are you receiving spurious interrupts now? There's an awful lot of added complexity to deal with that now, and to me it sounds like something is buggy if we're just now getting lots of them but we didn't before.

They appear when FIQs happen during handling of IRQs. Once the FIQ is handled, we try to ACK the IRQ in its handler, but the GIC reports a spurious interrupt. If the interrupt is still pending upon return, a new IRQ will be sent to the core.

kernel/interrupts/src/aarch64/table.s
NathanRoyer: > is this needed? on x86 interrupts are disabled automatically by the hardware when entering an interrupt handler. Does aarch64 behave differently?
> 
> And if this is required, how do interrupts get re-enabled? Does the hardware do that automatically? Seems weird that it would do it one way but not the other.

I think you interpreted this instruction as "disable some interrupts", whereas it enables fast interrupts. This is required for FIQs (NMIs) to not be missed during IRQ handling. We don't need to do this at the FIQ entry point because there can only be one TLB shootdown sequence at a time, so one FIQ at a time per core.

kernel/interrupt_controller/src/aarch64.rs
NathanRoyer: > Why use `UnsafeCell`; surely that isn't the best way to implement this? Cells operate primarily on Copy-like types, at least safely. I think it'd make more sense here to use a `force_unlock()` function, which is what most code does when it needs to get around Rust's language-level knowledge of lock safety.

I initially wanted to add a `force_unlock` method to Mutex but thought that this modification would be more dangerous and difficult than using `UnsafeCell` (which I think I did properly).

> Also, as an aside, we no longer want to use `irq_safety` directly for new code; instead we should use the `sync_irq` crate which is a bit more robust. But we can address that later.

Ok no problem, I'll change this in another PR :thumbsup:

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: You're right. I'll re-add this.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer:
```rust
pub type InterruptHandler = extern "C" fn(&InterruptStackFrame) -> EoiBehaviour;
```

Are function pointers assumed as NonNull?
if they are, `Option<InterruptHandler>` won't use more memory, so there's no downgrade.
I actually don't know if they are. If they aren't, we can change the `InterruptHandler` to a `NonNull<Something>`.

kernel/context_switch_regular/src/aarch64.rs
NathanRoyer: I fixed this more properly in [this commit](https://github.com/theseus-os/Theseus/pull/1039/commits/8c6ce13fee7afff69f09b499ab1cd897dbd026e5).

kernel/gic/src/gic/mod.rs
NathanRoyer: An interrupt can either be disabled, enabled in group 0, or enabled in group 1, but not enabled in both groups.
When it is disabled, it's also technically in a group, but the group it's in doesn't have any impact: it's simply disabled.

Improved the documentation in [this commit](https://github.com/theseus-os/Theseus/pull/1039/commits/271f023ca7f99e6b05900b6cb43b44812c432857).

NathanRoyer:
> Dealing with FIQs is kind of annoying. I'm wondering if we should just create a separate/dedicated struct or series of functions for handling FIQs that doesn't go through the same LocalInterruptController API that regular IRQs do.

If there was only GICv3 we could, but as GICv2 uses MMIO for FIQ management I think we should keep this "Use the normal structures but bypass their sync primitives" design: it's pretty simple to understand, unsafe code is very small, and it retains GIC compatibility via the GICv2/GICv3 abstraction.

kernel/context_switch_regular/src/aarch64.rs
kevinaboos: checks out, thanks.

kernel/gic/src/gic/mod.rs
kevinaboos: Thanks for clarifying; I had _assumed_ it worked like that but it wasn't 100% clear initially.

kernel/interrupt_controller/src/aarch64.rs
kevinaboos: yeah, it's a tough call, as I'm not necessarily sure that `force_unlock` is actually better. I think you may be right to just completely circumvent the mutex entirely rather than attempt to modify its locked state at all (e.g., by force unlocking it). So i think we'll stick with your approach for now.

kernel/interrupts/src/aarch64/mod.rs
kevinaboos: i have no idea; it would make sense that they're non-null but i don't know if that optimization actually exists and is used in the compiler. You can check the size of a fn ptr vs the size of Option<fn ptr> to verify that.

kernel/interrupts/src/aarch64/mod.rs
kevinaboos: ok i see, thanks for clarifying. Let's document that somewhere publicly visible such that it's clear why we actually need to bother with handling the possibility of spurious IRQs.

kernel/interrupts/src/aarch64/table.s
kevinaboos: lol, yep, i did certainly misinterpret it. Thanks for clarifying. Again, I'd recommend putting what you just wrote directly in a comment here so that information is preserved for future readers. It'd also be nice to repeat it somewhere in a top-level/Rust-level docs so it's not just buried in an asm file that nobody will ever open 😆 

Feel free to mark this as resolved once you do that.

kevinaboos:
> > Dealing with FIQs is kind of annoying. I'm wondering if we should just create a separate/dedicated struct or series of functions for handling FIQs that doesn't go through the same LocalInterruptController API that regular IRQs do.
> 
> If there was only GICv3 we could, but as GICv2 uses MMIO for FIQ management I think we should keep this "Use the normal structures but bypass their sync primitives" design: it's pretty simple to understand, unsafe code is very small, and it retains GIC compatibility via the GICv2/GICv3 abstraction.

Fair enough, I agree. 

Left a few new comments, should be easy to address pretty quickly.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: Added a sentence about this in the doc of `current_elx_irq`.

kernel/interrupts/src/aarch64/mod.rs
NathanRoyer: Looks like the optimization is working: [Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b85d7763574fe18b4960ac773d5ac7dc)

kevinaboos: Sounds good to me, I prefer this since it more explicitly expresses our intent. Wish we could do that on x86 too.... but we don't control the format of the IDT.

kevinaboos (approved):
I've accepted this PR, thanks! A couple of follow-up changes I'd like to see:

* `InterruptDestination` seems redundant with `IpiTargetCpu`...?
* We can use type wrappers to ensure that the `acknowledge_fast_interrupt` and `end_of_fast_interrupt` functions are only invokable by `current_elx_fiq`. By that I mean create a new type `ExceptionContextFiq` that is a simple wrapper around `ExceptionContext`, and then require a reference to that type to be passed into both of the above fast interrupt functions. Then, we can make those functions safe on the outside, and just use unsafe on the inside. 
    ```rust
    #[transparent]
    #[non_exhaustive]
    pub struct ExceptionContextFiq(ExceptionContext);
    ```
    This design is better imo since there's no real way to ensure that the fast interrupt functions are actually called safely.

* It's a little bit odd to use the type `*const InterruptHandler`, since `InterruptHandler` is already a function pointer. We should be able to remove the leading `*const` part and just deal with the `InterruptHandler` directly. But do let me know if that is actually requried somewhere else.

PR #1040, opened 2023-09-12
Move the PL011 driver into the main workspace

NathanRoyer:
This moves the PL011 driver into the main workspace.


kevinaboos (approved):
Looks good, thanks!

I wonder if there's a way to redesign the serial port interface such that we don't need an inner `Option<PL011>` inside the aarch64-specific serial port implementation. I find that design choice to be less than ideal.

PR #1041, opened 2023-09-13
Add automated QEMU-based testing to CI passes (x86_64 only for now)

tsoutsman:
- Added `qemu_test` application that automatically runs all tests and returns a specific exit code from QEMU based on output of tests.
- Added a `test` Make target
- Added a CI workflow that runs the `test` target
- Changed `test_mlx5`, `test_ixgbe`, and `test_block_io` to return an exit code of 0 if the associated devices aren't connected. An alternative would be to add them to the skipped tests in `qemu_test`
- Changed `test_identity_mapping`, `test_aligned_page_allocation`, and `test_filerw` to fail rather than print if they encounter an error.
- Renamed `tls_test` to `test_tls` so it is detected by `qemu_test`.
- Changed `test_channel` and `test_tls` to run all tests rather than specifying specific tests in the arguments.
- Renamed `test_serial_echo` to `serial_echo` because it isn't really a test with defined success and failure conditions.
- Changed `test_task_cancel` to always return 0 as task cancellation is not yet implemented.

I've also temporarily added `test_channel` to the list of skipped tests, because there seems to be a bug that causes deadlock. I'll need to look into the bug, but it's outside of the scope of this PR.


Makefile
```diff
@@ -1079,3 +1079,16 @@ endif
 	@sudo cp -vf $(iso) /var/lib/tftpboot/theseus/
 	@sudo systemctl restart isc-dhcp-server 
 	@sudo systemctl restart tftpd-hpa
+
+test: export override QEMU_FLAGS += -device isa-debug-exit,iobase=0xf4,iosize=0x04
+test: export override QEMU_FLAGS += -nographic
+test: export override FEATURES +=--features theseus_tests
+test: export override FEATURES +=--features qemu_test
+test: export override RUSTFLAGS +=--cfg qemu_test
```
kevinaboos: i find it a bit confusing to have a rustc-level cfg _and_ a cargo-level feature both named `qemu_test`. Given that we can use cargo features to specify dependencies, let's use features instead of cfg values here.

I think you can add it to the top-level `theseus_features` crate, and then have that enable both the `qemu_test` crate as well as a similar feature in `first_application`. You may need to name it something else, e.g., `test_qemu`, to avoid conflicts with the crate named `qemu_test`.

applications/test_channel/src/lib.rs
kevinaboos, line 1: if `test_channel` isn't working, let's undo these changes, and potentially consider moving it to old_crates or just exclude it from the workspace entirely.

kernel/first_application/src/lib.rs
```diff
@@ -28,7 +28,8 @@ use path::Path;
 
 /// See the crate-level docs and this crate's `Cargo.toml` for more.
 const FIRST_APPLICATION_CRATE_NAME: &str = {
-    #[cfg(target_arch = "x86_64")] { "shell-" }
+    #[cfg(all(target_arch = "x86_64", qemu_test))] { "qemu_test-" }
+    #[cfg(all(target_arch = "x86_64", not(qemu_test)))] { "shell-" }
```
kevinaboos: as in my other comment, i'd vote to use cargo features here, which would enable you to conditionally depend on your new `qemu_test` crate in `first_application/Cargo.toml`.

I'm also trying to gradually migrate the repo to using cargo features exclusively instead of rustc cfg (where possible), mostly for that reason of visibility at the manifest level.

kevinaboos (changes requested):
Generally looks great, thanks so much! I left a few small comments.

One question -- how long does this take? I noticed a 10-min timeout in the action, hence my curiosity. Our existing CI passes already take over 5 min or so for basic PRs, so I'm wondering if there's a way to speed this up, e.g., by caching build results across separate action passes, and/or disabling actions when a PR's changeset doesn't touch files that require them to be re-run.

applications/test_channel/src/lib.rs
tsoutsman: `test_channel` isn't working because there is a bug in `async_channel`, not because the changes broke it. If you look in the [logs of the failed CI](https://pipelinesghubeus25.actions.githubusercontent.com/FETIoBJQ5HzlrexsDZU67Xr10mDwYSlx6msA7ijXwIiXeebNId/_apis/pipelines/1/runs/4639/signedlogcontent/2?urlExpires=2023-09-13T23:36:49.9592746Z&urlSigningMethod=HMACV1&urlSignature=4szDIxWd6wwE55n/KdFu8OGDV8gpW1Pq++SFiS0zRNs%3D), the receiver task never returns for some reason. I've been able to sporadically recreate it on my machine.

The changes to `test_channel` merely make it run all the tests (i.e. if we were to pass in all the flags).

tsoutsman:
> One question -- how long does this take?

The most recent commit took 7 min 27 sec.

EDIT: the most most recent commit took 5 min 26 sec, without any significant changes so 🤷 .

> caching build results across separate action passes

It's [possible](https://github.com/marketplace/actions/rust-cache), but I'm not sure how significant the speed up will be.

I don't think the time it takes to run CI is particularly important. CI speed is only relevant if a cleanup commit was pushed right before the PR is merged but Github has an [auto-merge feature](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request).

Makefile
tsoutsman: The only reason I didn't do that is because I don't think the feature could be considered "additive", but yea it's probably better than yet another `cfg`.

tsoutsman:
Added the feature shenanigans in this [commit](https://github.com/theseus-os/Theseus/pull/1041/commits/473d49bf399951fbc1952b36e6baccfa9660089e). Not quite sure if that's the best way to do it.

applications/test_channel/src/lib.rs
kevinaboos: i see, that's fine.

kevinaboos: and yes, i think the issue with async_channel is that we had a contributor add a partial amount of recovery functionality to it, but never finished the job.

kevinaboos:
> CI speed is only relevant if a cleanup commit was pushed right before the PR is merged but Github has an [auto-merge feature](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request).

Well that's my primary use case, I do that for almost every PR so it would be nice to not make CI passes super slow. But nbd, i can wait 5-7 mins.

> Added the feature shenanigans in this [commit](https://github.com/theseus-os/Theseus/pull/1041/commits/473d49bf399951fbc1952b36e6baccfa9660089e). Not quite sure if that's the best way to do it.

Ah i see, bitten yet again by the need for features to be additive. Yes, looking at that commit I do agree it's a bit weird, but we should be able to get around that by simply enabling `first_application/qemu_test` directly from the `theseus_features` crate. That way we wouldn't have to pass it through the `captain` for no reason.
Unfortunately, we won't be able to deactivate the `shell` dependency that way, as we can't use `cfg(not(feature = "..."))` in `Cargo.toml` manifests afaik, but you can do that in the Rust code itself. That would result in similar behavior to the `cfg(qemu_test)` approach, but with the added benefit of not including `qemu_test` in normal builds. You'll also have to remove the `compile_error` check, but basically the feature can be kept pseudo-additive by making `qemu_test` a preferential feature, in that it overrides the usage of `shell` in `first_application` whenever it is active.

tsoutsman:
Should be good.

kevinaboos (approved):

PR #1042, opened 2023-09-16
Change scheduler to only store runnable tasks on run queue

tsoutsman:
Note that this inherently adds task migration as when a task is unblocked it can be added to any core.

Depends on #1035, #1044, #1045.


kevinaboos:
all PR dependencies have been merged in. 👍

kernel/spawn/src/lib.rs
```diff
@@ -391,8 +391,8 @@ impl<F, A, R> TaskBuilder<F, A, R>
 
         let exposed = ExposedTask { task: new_task };
         exposed.inner().lock().pinned_cpu = self.pin_on_cpu;
-        let ExposedTask { task: mut new_task } = exposed;    
-
+        let ExposedTask { task: mut new_task } = exposed;
+
```
kevinaboos, line 395: first line seems good, second line seems wrong (extra whitespace for no reason?)

kernel/spawn/src/lib.rs
```diff
@@ -443,12 +443,8 @@ impl<F, A, R> TaskBuilder<F, A, R>
         fence(Ordering::Release);
         
         // Idle tasks are not stored on the run queue.
-        if !self.idle {
-            if let Some(cpu) = self.pin_on_cpu {
-                task::scheduler::add_task_to(cpu, task_ref.clone());
-            } else {
-                task::scheduler::add_task(task_ref.clone());
-            }
+        if !self.idle & !self.blocked {
```
kevinaboos: i think you meant `&&`

kernel/task/src/lib.rs
```diff
@@ -281,6 +284,55 @@ impl TaskRef {
...
     }
+
+    /// Blocks this `Task` by setting its runstate to [`RunState::Blocked`].
+    ///
+    /// Returns the previous runstate on success, and the current runstate on error.
+    /// This will only succeed if the task is runnable or already blocked.
+    pub fn block(&self) -> Result<RunState, RunState> {
+        use RunState::{Blocked, Runnable};
+
+        let run_state = self.0.task.runstate();
```
kevinaboos, line 295: Unless I'm missing something, I believe this (and the same logic in `unblock()`) simply obtains a copy of the task's current `RunState` and then modifies that copy; it doesn't actually modify the `RunState` of the task itself.

kernel/task/src/lib.rs
```diff
@@ -281,6 +284,55 @@ impl TaskRef {
...
+
+        if run_state.compare_exchange(Runnable, Blocked).is_ok() {
+            Ok(Runnable)
+        } else if run_state.compare_exchange(Blocked, Blocked).is_ok() {
+            // warn!("Blocked an already blocked task: {:?}", self);
+            Ok(Blocked)
```
kevinaboos, line 301: It looks like blocked tasks are not proactively removed from the runqueue they're on, but rather done lazily by each scheduler policy. 

This is fine, as it's much faster, but it  should be clearly documented somewhere (i.e., in the docs for the `Scheduler` trait's `next()` function) because it's a non-symmetric pair of operations. Otherwise, a scheduler policy implementor wouldn't know that it's okay to simply remove a task from the runqueue when it comes across one that is blocked.

kernel/task/src/lib.rs
kevinaboos: I think a good way to avoid this is to split up the functions:
* in `task_struct::Task::block()`, only the runstate is modified. Basically, this function as it was.
   * the functions in `task_struct::Task` may  need to be renamed.
* in `task::TaskRef::block()`, it calls `task_struct::Task::block()` and then modifies the runqueue accordingly.
* `unblock()` is similar.

kernel/task/src/scheduler.rs
```diff
@@ -103,25 +103,36 @@ where
...
+        let mut min_busyness = usize::MAX;
+        let mut least_busy_index = None;
+
+        for (i, (_, scheduler)) in locked.iter().enumerate() {
+            let busyness = scheduler.lock().busyness();
+            if busyness < min_busyness {
+                least_busy_index = Some(i);
+                min_busyness = busyness;
+            }
         }
```
kevinaboos, line 133: this function is incredibly slow. In general, I don't think we should ever really access `SCHEDULERS` except when initializing or changing scheduler policies, though we can think about that more later (and how to improve this, e.g., via a separate estimated utilization metric or something that avoids the need to iterate over all schedulers).

It's likely that using this function every time a task is unblocked is going to absolutely destroy performance... right? Also I imagine it will cause lots and lots of task migrations, which is generally something you want to keep to a minimum unless a given CPU core is at max utilization for a "long" time (which we currently don't track a metric for).

kernel/scheduler_epoch/src/lib.rs
```diff
@@ -28,92 +36,86 @@ const INITIAL_TOKENS: usize = 10;
...
-            // we assign tokens only to runnable tasks
-            if !t.is_runnable() {
-                continue;
-            }
-
-            total_priorities = total_priorities
-                .saturating_add(1)
-                .saturating_add(t.priority as usize);
+        while let Some(task) = self.out_of_tokens.pop() {
+            self.have_tokens.push_back(task);
```
kevinaboos, line 96: nit: Is there a way to do this in one batch rather than item-by-item? I guess not when we're using `VecDeque` for one list and `Vec` for the other...?

kernel/scheduler_epoch/src/lib.rs
```diff
@@ -28,92 +36,86 @@ const INITIAL_TOKENS: usize = 10;
...
+                    .expose_is_on_run_queue()
+                    .store(false, Ordering::Release);
+                // This check prevents an interleaving where `TaskRef::unblock` wouldn't add
+                // the task back onto the run queue. `TaskRef::unblock` sets the run state and
+                // then checks `is_on_run_queue` so we have to do the inverse.
+                if unlikely(task.task.is_runnable()) {
+                    if let Some(task) = self.add_epoch_task(task) {
+                        return Some(task);
+                    }
+                }
```
kevinaboos, line 70: So, IIUC, the crux of the issue surrounding this `is_on_run_queue()` atomic boolean is that we need to do two things in one "atomic" step:
1. change the task's runstate to `Runnable` 
2. add the task back to a runqueue

Since we cannot do that, you're using a separate atomic bool to indicate whether a task is on a runqueue. I understand this, but I don't love this design. It's complex without bringing any other benefits, and kind of a messy design choice, as it makes both current and future scheduler policies harder to write.

Ideally this shouldn't be a necessary step. Can we avoid this complexity via an alternative design that's a bit cleaner and doesn't burden the `Scheduler` policy implementor with additional concerns?

Side note: if we actually wanted to store some piece of state inside a task about whether that task is on a runqueue, we might as well store a reference to (or the ID of) the actual runqueue that it's currently on. This would at least make it faster to add the task back once it gets unblocked, which addresses my other comment as a plus.

kevinaboos (changes requested):
left a few questions. I'm a bit concerned the overhead associated with some changes in this PR is actually counterproductive for overall performance, but most of it is solvable.

PR #1043, opened 2023-09-16
Update `test_scheduler`

tsoutsman:
The new test is significantly more robust than the old one. As of right now, the test isn't particularly useful because we don't have task migration, but #1042 adds implicit task migration when unblocking a task. Hence, the test has a focus on blocking and unblocking tasks.


tsoutsman:
Closed because I accidentally used the wrong branch. Superseded by #1045.

PR #1044, opened 2023-09-16
fix: set `inner.pinned_cpu` for pinned tasks

tsoutsman:
When spawning a pinned task, `spawn` didn't previously set `inner.pinned_cpu`. This created problems in #1042 because the scheduler didn't know that tasks were pinned and freely migrated them across cores.


kevinaboos (approved):
Oops! good catch, thanks.

PR #1045, opened 2023-09-17
Update `test_scheduler`

tsoutsman:
The new test is significantly more robust than the old one. As of right now, the test isn't particularly useful because we don't have task migration, but https://github.com/theseus-os/Theseus/pull/1042 adds implicit task migration when unblocking a task. Hence, the test has a focus on blocking and unblocking tasks.


applications/test_scheduler/Cargo.toml
```diff
@@ -2,20 +2,18 @@
 name = "test_scheduler"
 version = "0.1.0"
 authors = ["Namitha Liyanage <namithaliyanage@gmail.com>"]
```
kevinaboos: this is an entirely new crate now, so you can update the authorship to only yourself.

kernel/task_struct/src/lib.rs
```diff
@@ -485,6 +485,18 @@ impl Task {
         }
     }
 
+    /// Same as [`block`], but doesn't print a warning if the task is already
+    /// blocked.
+    ///
+    /// This method can be useful if blocking potentially blocked tasks in a
+    /// loop (e.g. `test_scheduler`). Logging is very slow, and this function
+    /// can lead to a `100x` performance improvement.
+    pub fn block_no_log(&self) {
```
kevinaboos: for this and the related `unblock_no_log`, let's just remove the log statements entirely from the existing `block`/`unblock` functions. We have since determined that it is entirely valid to unblock an already runnable task, e.g., in the case of a deferred interrupt handler where the deferred task is already running when another interrupt occurs.

kevinaboos (changes requested):
#1035 has been merged in, thanks!

kevinaboos (approved):
thanks!

PR #1051, opened 2023-10-07
Redesign the interrupt controller crate; support x86_64

kevinaboos:
* The `interrupt_controller` crate still is not used on x86_64.
* Move initialization of x86-specific interrupt controllers (i.e., APIC)
  into the x86-specific module of `interrupt_controller`,
  which makes more sense than doing it from the `device_manager`.
  * As such, the `captain` can now unconditionally initialize
    the interrupt controller crate, which works on all architectures.

Two remaining TODOs:
1. Implement `enable_local_timer_interrupt()` for aarch64
   * This requires moving some aarch64-specific functionality for setting up the system's per-CPU timer out of the  `interrupts` crate and into another crate.
2. Implement `SystemInterruptController::get()` for x86_64.
   * This is a bit challenging because there can be _multiple_ system-wide interrupt controllers (I/O APICs) on x86_64, so we can't simply implement a basic getter function.
   * We may need to offer an iterator interface, or perhaps a getter that takes the current CPU or NUMA domain or some kind of other identifier into account.

In the future, we'll move each `LocalInterruptController` instance into CPU-local storage to allow for very fast access.

### DIFF
--- a/kernel/interrupt_controller/Cargo.toml
+++ b/kernel/interrupt_controller/Cargo.toml
@@ -17,7 +17,6 @@ sync_irq = { path = "../../libs/sync_irq" }
 
 [target.'cfg(target_arch = "aarch64")'.dependencies]
 arm_boards = { path = "../arm_boards" }
-memory = { path = "../memory" }
 gic = { path = "../gic" }
 
 [target.'cfg(target_arch = "x86_64")'.dependencies]
--- a/kernel/interrupt_controller/Cargo.toml
+++ b/kernel/interrupt_controller/Cargo.toml
@@ -1,21 +1,27 @@
 [package]
-authors = ["Nathan Royer <nathan.royer.pro@gmail.com>"]
-description = "Cross-platform abstraction over interrupt controllers"
-version = "0.1.0"
-edition = "2021"
 name = "interrupt_controller"
+version = "0.1.0"
+authors = [
+    "Nathan Royer <nathan.royer.pro@gmail.com>",
+    "Kevin Boos <kevinaboos@gmail.com>",
+]
+description = "Cross-platform abstraction over interrupt controllers"
+edition = "2021"
 
 [dependencies]
 log = "0.4.8"
 cpu = { path = "../cpu" }
+memory = { path = "../memory" }
+spin = "0.9.4"
+sync_irq = { path = "../../libs/sync_irq" }
 
 [target.'cfg(target_arch = "aarch64")'.dependencies]
-sync_irq = { path = "../../libs/sync_irq" }
 arm_boards = { path = "../arm_boards" }
 memory = { path = "../memory" }
 gic = { path = "../gic" }
-spin = "0.9.4"
 
 [target.'cfg(target_arch = "x86_64")'.dependencies]
+acpi = { path = "../acpi" }
 apic = { path = "../apic" }
 ioapic = { path = "../ioapic" }
+madt = { path = "../acpi/madt" }
--- a/kernel/interrupt_controller/src/x86_64.rs
+++ b/kernel/interrupt_controller/src/x86_64.rs
@@ -1,9 +1,10 @@
 use super::*;
 
-use {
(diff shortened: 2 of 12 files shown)


kevinaboos:
Marked as a draft because it's still a work-in-progress. However, everything still works.

kernel/interrupt_controller/Cargo.toml
```diff
@@ -1,21 +1,27 @@
...
 log = "0.4.8"
 cpu = { path = "../cpu" }
+memory = { path = "../memory" }
+spin = "0.9.4"
+sync_irq = { path = "../../libs/sync_irq" }
 
 [target.'cfg(target_arch = "aarch64")'.dependencies]
-sync_irq = { path = "../../libs/sync_irq" }
 arm_boards = { path = "../arm_boards" }
 memory = { path = "../memory" }
```
NathanRoyer: This aarch64-specific memory dep needs to be removed

kernel/interrupt_controller/src/x86_64.rs
```diff
@@ -14,35 +15,38 @@ pub struct LocalInterruptControllerId(pub u32);
...
+/// (bootstrap processor) and the system-wide IOAPIC(s).
+pub fn init(kernel_mmi: &memory::MmiRef) -> Result<(), &'static str> {
+    apic::init();
+
+    // Use the MADT ACPI table to initialize more interrupt controller details.
+    {
+        let acpi_tables = acpi::get_acpi_tables().lock();
+        let madt = Madt::get(&acpi_tables)
+            .ok_or("The required MADT ACPI table wasn't found (signature 'APIC')")?;
+        madt.bsp_init(&mut kernel_mmi.lock().page_table)?;
```
NathanRoyer, line 28: The API would be more consistent if this function obtained the page table from `memory` directly rather than getting it via call parameters. It would reduce the risk of deadlocks too? (here, not really)

kernel/serial_port/src/lib.rs
```diff
@@ -384,7 +384,10 @@ static INTERRUPT_ACTION_COM1_COM3: Once<Box<dyn Fn() + Send + Sync>> = Once::new
 static INTERRUPT_ACTION_COM2_COM4: Once<Box<dyn Fn() + Send + Sync>> = Once::new();
 
 
-// Cross-platform interrupt handler for COM1 and COM3 (IRQ 0x24 on x86_64).
+// Cross-platform interrupt handler for the primary serial port.
+//
+// * On x86_64, this is IRQ 0x24, used for COM1 and COM3 serial ports.
+// * On aarch64, we also use interrupt 0x24, used for the
```
NathanRoyer: On aarch64, we use interrupt 33 / 0x21 for serial port reception (we have no choice).
By the way, I think this comment is unfinised.

NathanRoyer (commented):
I added a few comments.
Your branch runs fine on my side :+1:

kernel/serial_port/src/lib.rs
kevinaboos: ah yes, thanks!

kernel/interrupt_controller/src/x86_64.rs
kevinaboos: hmm, i disagree actually --- in general, it's better state management (and clearer expression of dependencies) to pass a function everything it needs in order for it to execute. Obviously that's not always possible, but at least in other initialization code so far we do try to always pass in things like MMI.

kernel/interrupt_controller/Cargo.toml
kevinaboos: thanks!

kernel/interrupt_controller/src/x86_64.rs
kevinaboos: Also marginally more efficient, haha. Nothing significant, but it doesn't hurt to avoid the attempt to fallibly access global kernel mmi state.

Issue #1047, opened 2023-09-24
Follow-up improvements to aarch64 FIQ support


kevinaboos:
* `InterruptDestination` seems redundant with `IpiTargetCpu`...?
* We can use type wrappers to ensure that the `acknowledge_fast_interrupt` and `end_of_fast_interrupt` functions are only invokable by `current_elx_fiq`. By that I mean create a new type `ExceptionContextFiq` that is a simple wrapper around `ExceptionContext`, and then require a reference to that type to be passed into both of the above fast interrupt functions. Then, we can make those functions safe on the outside, and just use unsafe on the inside. 
    ```rust
    #[transparent]
    #[non_exhaustive]
    pub struct ExceptionContextFiq(ExceptionContext);
    ```
    This design is better imo since there's no real way to ensure that the fast interrupt functions are actually called safely.

* It's a little bit odd to use the type `*const InterruptHandler`, since `InterruptHandler` is already a function pointer. We should be able to remove the leading `*const` part and just deal with the `InterruptHandler` directly. But do let me know if that is actually requried somewhere else.

_Originally posted by @kevinaboos in https://github.com/theseus-os/Theseus/pull/1039#pullrequestreview-1641014977_

NathanRoyer:
Addressed in #1052.

PR #1052, opened 2023-10-09
Minor FIQ improvements (follow-up for issue #1047)

NathanRoyer:
#### `InterruptDestination` seems redundant with `IpiTargetCpu`...?

As discussed on discord, it makes sense to keep both types:
- `InterruptDestination`: the cross-platform public type that can be used everywhere in theseus. Has two variants: "All But Me" / "One Specific CPU core"
- `IpiTargetCpu` / `SpiDestination`: GIC-specific structures that must only be used in GIC-specific code. These have a `GICv2TargetList` variant.

-----

#### We can use type wrappers to ensure that the `acknowledge_fast_interrupt` and `end_of_fast_interrupt` functions are only invokable by `current_elx_fiq`.

Yes. I implemented a safe wrapper, but I'm not entirely sure I did what you had in mind: As `ExceptionContext` is defined in the `interrupts` crate, everything had to be done inside of it. `interrupt_controller` wasn't modified.

The relevant commit is the second one.

-----

#### It's a little bit odd to use the type `*const InterruptHandler`, since `InterruptHandler` is already a function pointer.

Yes. That's now fixed.

### DIFF
--- a/kernel/interrupts/src/aarch64/mod.rs
+++ b/kernel/interrupts/src/aarch64/mod.rs
@@ -170,7 +170,7 @@ pub fn init() -> Result<(), &'static str> {
 pub fn init_timer(timer_tick_handler: InterruptHandler) -> Result<(), &'static str> {
     // register/deregister the handler for the timer IRQ.
     if let Err(existing_handler) = register_interrupt(CPU_LOCAL_TIMER_IRQ, timer_tick_handler) {
-        if timer_tick_handler as *const InterruptHandler != existing_handler {
+        if timer_tick_handler as InterruptHandler != existing_handler {
             return Err("A different interrupt handler has already been setup for the timer IRQ number");
         }
     }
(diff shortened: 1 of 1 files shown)


kevinaboos:
> Yes. I implemented a safe wrapper, but I'm not entirely sure I did what you had in mind: As `ExceptionContext` is defined in the `interrupts` crate, everything had to be done inside of it. `interrupt_controller` wasn't modified.

Yeah, my idea was to have the actual `AArch64LocalInterruptControllerApi::acknowledge_fast_interrupt()` and `end_of_fast_interrupt()` functions to accept a reference to the `ExceptionContextFiq` type in order to ensure they are only callable from a fast interrupt context; that would allow us to mark them as externally safe.

I understand the dependency relationship between `interrupts` and `interrupt_controller` currently prevents that. What do you think is the best way to refactor/move this code around in order to achieve that?

On a related note, we could actually require the regular `acknowledge_interrupt` and `eoi` functions (even on x86) to accept a reference to `ExceptionContext` in order to ensure they are only invokable from an interrupt handler context.

NathanRoyer:
> I understand the dependency relationship between `interrupts` and `interrupt_controller` currently prevents that. What do you think is the best way to refactor/move this code around in order to achieve that?
> 
> On a related note, we could actually require the regular `acknowledge_interrupt` and `eoi` functions (even on x86) to accept a reference to `ExceptionContext` in order to ensure they are only invokable from an interrupt handler context.

I think it would be weird to define `ExceptionContext` in `interrupt_controller` (because it's an `interrupts` internal). On the other hand, we cannot swap the dependency link between `interrupt_controller` and `interrupts`, as the current dependency direction has multiple reasons to be (in multiple places, the int ctrl is manipulated during interrupt handling).

I must say I found the state before this PR consistent / satisfying -enough, or at least more so than what we're envisioning. The unsafe functions have clear explanations in their comments; I cannot think of a reason someone would explore `interrupt_controllers`, see these functions and misuse them.

kevinaboos (approved):

PR #1054, opened 2023-10-11
aarch64: move generic system timer to its own crate

kevinaboos:
Can now be used in the `LocalInterruptController` to manage the CPU-local timer interrupt on aarch64.

### DIFF
--- a/kernel/generic_timer_aarch64/src/lib.rs
+++ b/kernel/generic_timer_aarch64/src/lib.rs
@@ -75,11 +75,8 @@ pub fn set_next_timer_interrupt(ticks_to_elapse: u64) {
 pub fn enable_timer_interrupt(enable: bool) {
     // Unmask the interrupt (to enable it), and enable the timer.
     CNTP_CTL_EL0.write(
-          CNTP_CTL_EL0::IMASK.val(0)
-        + CNTP_CTL_EL0::ENABLE.val(match enable {
-            true => 1,
-            false => 0,
-        })
+          CNTP_CTL_EL0::IMASK.val(!enable as u64)
+        + CNTP_CTL_EL0::ENABLE.val(enable as u64)
     );
 
     if false {
--- a/kernel/generic_timer_aarch64/src/lib.rs
+++ b/kernel/generic_timer_aarch64/src/lib.rs
@@ -1,6 +1,9 @@
 //! Support for aarch64's generic timer and system counter.
+//!
+//! Docs: <https://developer.arm.com/documentation/102379/latest/>
 
 #![no_std]
+#![feature(negative_impls)]
 
 // This crate is only relevant on aarch64 systems,
 // but we use a cfg gate here to allow it to be included in x86 builds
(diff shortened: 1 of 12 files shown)


kernel/generic_timer_aarch64/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+    enable_timer_interrupt(true);
+}
+
+/// Enables/disables the generic system timer interrupt on the current CPU.
+///
+/// This writes the `CNTP_CTL_EL0` system register.
+pub fn enable_timer_interrupt(enable: bool) {
+    // Unmask the interrupt (to enable it), and enable the timer.
+    CNTP_CTL_EL0.write(
+          CNTP_CTL_EL0::IMASK.val(0)
```
kevinaboos: @NathanRoyer  question about this -- i copied this from your existing code, but it seems to unconditionally enable the timer _interrupt_ but correctly enable the timer itself conditionally.

I'm not sure what the difference is, but should this value be `!enable` here? Or does it not matter because the interrupt itself is auto-disabled when the timer itself is disabled? I'm not really sure what it means to disable the actual _timer_ ... does the timer stop ticking up? If so, perhaps we need two functions -- one to enable/disable the timer ticking, and another one to enable/disable the timer interrupt.

Generally we don't care about disabling the timer, only the interrupt. I can't imagine we'd ever want to disable the timer ticking, or even be allowed to by hardware.

kevinaboos:
I think this is complete now, but I may have forgotten something. Lmk if you notice anything flagrant.

I plan to make some more changes in the future, specifically regarding how the initialization flow occurs and the abstractions are layered atop each other:
* `interrupt_controller` should be inited by `interrupts`
* the generic timer interrupt on aarch64 should be accessed through the local interrupt controller API

kernel/generic_timer_aarch64/src/lib.rs
NathanRoyer: Your understanding is correct. We might prefer to keep the timer always enabled and mask the interrupt here whenever we want to pause timer interrupts.
I'm OK with both strategies, feel free to change this to your preference.

NathanRoyer (approved):
This looks alright :ok_hand:

kernel/generic_timer_aarch64/src/lib.rs
kevinaboos: Yeah, i'd say that it's pretty counterintuitive and unexpected to stop the actual timer ... I'm surprised the hardware even allows you to do that.

In this case, we should follow what other h/w timers do (e.g., LocalApic LVT timer on x86), which is only concerned with disabling the actual timer _interrupt_.  Do you know if it actually works correctly if we do that? Or do we actually have to disable the timer itself in order to properly disable the interrupt too?

NathanRoyer: > Do you know if it actually works correctly if we do that? Or do we actually have to disable the timer itself in order to properly disable the interrupt too?

I don't really know, I don't remember trying to mask the interrupt, but imo it should totally work as you're envisioning, with the timer always enabled and the mask being toggled.

kevinaboos: Based on some posts from the LKML, it seems that we actually do need to disable the timer itself in order to work around potential bugs in various VMMs/hypervisors. In any case, since we're only enabling/disabling it for a short interval, i doubt it makes much difference.

In the future we can separate the function to enable the timer from the function to enable the timer's interrupt, though I'm not sure if we'll ever actually need that ability.

PR #1056, opened 2023-10-16
Update README for USB

YtvwlD:
4231c5d447e4cb7f3047cc22e660e6c58136b3b9 changed the make target, but the README still had the old one.


kevinaboos (approved):
Thanks, good catch!

PR #1059, opened 2023-10-19
EHCI USB Controllers Support

NathanRoyer:
This PR brings early support for EHCI USB controllers. It currently supports directly connected High Speed devices (no hub / low-speed device / full speed device support), and allows:
- sending standard USB requests to them and
- getting interrupt transfers from them.

NB 1: USB Requests work on top of control transfers.

High Level Todos:
- bulk transfer support (async schedule)
- isochronous transfer support (periodic schedule)
- device disconnect support

NB 2: isochronous transfers are low-priority because they're a legacy concept and they were mainly use for audio, which we don't support in theseus.

Low Level Todos:
- for the first request (requesting the device descriptors), we don't know the max packet size of that device. For 1.x devices, it could be less than the size of the device descriptor; we currently assume max_packet_size incorrectly for the first request because I couldn't understand which max_packet_size to use initially.
- interrupt transfer support currently doesn't follow the spec regarding the polling interval we read from the device. That value should be used to populate the periodic schedule with a correct ratio, but right now we give all interrupt transfers equal importance.
- Various USB interrupts are handled with placeholders that simply log the event. That could result in log spam in events such as device disconnects.
- The MMIO region is identity-mapped for simplicity; without an identity mapping, a lot of virtual <-> physical address translations are going to be needed.
- The MMIO region must not cross a 4GiB boundary, however this is currently not ensured.
- in this crate, Controllers own the interface driver objects for their devices. The ownership story is kind of complicated because the interface drivers sometimes have to modify the controller objects. Currently, the interface driver objects are stored as a collection of `Option<Interface>` so that they can be removed from the controllers while they in turn modify the controllers themselves. It's pretty ugly.
- there's a bug where the driver misses PCI interrupts sometimes, causing device freezes. I spent a lot of time on it and couldn't find out why.


kevinaboos:
Once you merge in the changes from #1081, feel free to mark this as "ready for review" and then request a review from me.

NathanRoyer:
Ok this is ready for a first review.
I just added a command line toggle for USB emulation: `usb_hid=yes` (default=no).

PR #1061, opened 2023-10-21
Change `path` to mimic `std::path`

tsoutsman:
This PR changes `path` to export a `Path` and `PathBuf` struct, akin to `std::path`. Initially I made this change because it was necessary for Rust FFI for the `std` port, but it turns out the change also removed a lot of unnecessary allocations throughout the codebase.

* Change `Path::file_stem()` to return an `Option<&str>`
  * This causes `crate_name_from_path()` to return a Result

### DIFF
--- a/kernel/path/Cargo.toml
+++ b/kernel/path/Cargo.toml
@@ -1,8 +1,8 @@
 [package]
 name = "path"
 version = "0.1.0"
-authors = ["Andrew Pham <apham727@gmail.com>, Christine Wang <chrissywang54@gmail.com"]
-description = "contains functions for navigating the filesystem / getting pointers to specific directories via the Path struct"
+authors = ["Andrew Pham <apham727@gmail.com>, Christine Wang <chrissywang54@gmail.com", "Klim Tsoutsman <klim@tsoutsman.com>"]
+description = "File system path manipulation"
 edition = "2021"
 
 [dependencies]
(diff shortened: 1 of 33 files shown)


kernel/path/src/lib.rs
```diff
@@ -1,279 +1,628 @@
...
 }
 
-impl fmt::Display for Path {
-    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        write!(f, "{}", self.path)
+impl AsMut<Path> for str {
+    #[inline]
+    fn as_mut(&mut self) -> &mut Path {
+        // SAFETY: Path has the same type layout as str. This is the same
+        // implementation as std: https://doc.rust-lang.org/src/std/path.rs.html#2044-2048
```
kevinaboos: use a permalink to github instead, as this docs-based link could change.

kernel/path/Cargo.toml
```diff
@@ -3,26 +3,8 @@ name = "path"
 version = "0.1.0"
 authors = ["Andrew Pham <apham727@gmail.com>, Christine Wang <chrissywang54@gmail.com"]
```
kevinaboos: with this significant of a restructuring, you can add yourself as primary author.

kernel/path/src/lib.rs
```diff
@@ -1,279 +1,628 @@
...
+    /// Wraps a string slice as a path slice.
+    ///
+    /// This is a cost-free conversion.
+    #[inline]
+    pub fn new<S>(s: &S) -> &Self
+    where
+        S: AsRef<str> + ?Sized,
+    {
+        // SAFETY: Path has the same type layout as str. This is the same
+        // implementation as std: https://doc.rust-lang.org/src/std/path.rs.html#2040-2042
```
kevinaboos: same here re: permalink

kevinaboos (changes requested):
Thanks a lot! Looks good.

I left a few very minor comments. Code-wise everything is fine.

kevinaboos (approved):

PR #1065, opened 2023-10-28
Upgrade to latest Rust nightly v1.75

kevinaboos:
* Upgrade proc-macro2 versions used in `uefi-builder` tool to avoid this issue: https://github.com/rust-lang/rust/issues/113152

* Modify how the `.theseus_merged` special section is detected by the crate loading/linking code.


tsoutsman (approved):
Weird that the `ovmf-prebuilt` dependency was still in the lock file. Looks good other than clippy. Is there a reason we don't use `THESEUS_MERGED_SEC_SHNDX` any more?

tsoutsman:
From Discord:
> One new issue that I observed is that there are now GROUP sections (classified as COMDAT groups) for the rust_eh_personality sections (which are related to unwinding), but they only show up in aarch64 builds AFAICT. 
I'm not sure if this is something that we need to be aware of; in the past I did not pay any attention to personality sections because we don't use that in our unwinder.
In any case, I just added a quick workaround to allow crate loading to work by ignoring the section.

kevinaboos:
> Is there a reason we don't use `THESEUS_MERGED_SEC_SHNDX` any more?

the existence of a `GROUP` section type may come before it, so I took a less fragile approach of looking for the `.theseus_merged` section anywhere before the `.text*` sections, since that is definitely under the control of our linker script.

PR #1067, opened 2023-10-30
Fix `heap` allocator bug

tsoutsman:
The implementation of `GlobalAlloc` for `Heap` assumed that all memory under `INITIAL_HEAP_END_ADDR` was allocated using the initial allocator, but this isn't true.

`MultipleHeaps` allocates large objects using mapped pages leading to objects allocated in the lower half of memory. When deallocating these objects, `Heap` tried to deallocate them using the initial allocator rather than `MultipleHeaps`.


kevinaboos (approved):
Thanks for tracking this down!

PR #1069, opened 2023-11-06
Add support for `std`

tsoutsman:
The CI is too slow to even check out the rust submodule, let alone build the compiler. Probably best to create a separate branch rather than merging into `theseus_main`.

To upstream this we would have to:
- Implement the rest of `std` functionality
- Publish `theseus_ffi` and `theseus_shim` to crates.io


.ignore
```diff
@@ -0,0 +1 @@
+ports/rust
```
tsoutsman: So that editors don't try to load the entire rust repository.

shim/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+pub fn getpid() -> u32 {
+    unsafe { c::getpid() }
+}
+
+#[inline]
+pub unsafe fn register_dtor(t: *mut u8, dtor: unsafe extern "C" fn(*mut u8)) {
+    c::register_dtor(t, dtor)
+}
+
+// TODO: impl Send + Sync for Reader and Writer?
```
tsoutsman: Bump. It should be fine because `FatPointer` is created from an `Arc`.

shim/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+#[inline]
+pub fn stderr() -> Result<Writer> {
+    Ok(Writer {
+        inner: Result::from(unsafe { c::stderr() })?,
+    })
+}
+
+// TODO: Should we be taking a mutable or immutable reference?
```
tsoutsman: Bump.

kernel/libtheseus/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+        .try_into()
+        .expect("current task id too large")
+}
+
+#[export_name = "libtheseus::register_dtor"]
+pub unsafe extern "C" fn register_dtor(t: *mut u8, dtor: unsafe extern "C" fn(*mut u8)) {
+    unsafe { thread_local_macro::register_dtor(t, dtor) }
+}
+
+// TODO: Something better than transmutations?
```
tsoutsman, line 114: Bump.

Makefile
```diff
@@ -313,10 +315,33 @@ endif
...
+##	Remove the last line of std/Cargo.toml
+	@$(HEAD) -n -1 $(RUST_SOURCE)/library/std/Cargo.toml > $(ROOT_DIR)/temp-std-cargo.toml
+	@mv $(ROOT_DIR)/temp-std-cargo.toml $(RUST_SOURCE)/library/std/Cargo.toml
+## Restore previous std/Cargo.toml
+	mv $(ROOT_DIR)/std-cargo.toml $(RUST_SOURCE)/library/std/Cargo.toml
+
+##	Add .cargo/config.toml
+	@mkdir -p $(ROOT_DIR)/.cargo
+	@echo "[build]" > $(ROOT_DIR)/.cargo/config.toml
+	@echo "rustc = \"$(RUST_SOURCE)/build/$(HOST_TRIPLE)/stage1/bin/rustc\"" >> $(ROOT_DIR)/.cargo/config.toml
```
tsoutsman: This is terrible, but IDK if there's a better solution.

kernel/thread_local_macro/src/lib.rs
```diff
@@ -84,14 +84,19 @@ pub fn take_current_tls_destructors() -> Vec<TlsObjectDestructor> {
...
 ///   When the current task exits, this function will be invoked with `a`
 ///   as its only argument, at which point the `dtor` function should drop `a`.
 /// 
 /// Currently the only value of `dtor` that is used is a type-specific monomorphized
 /// version of the above [`fast::destroy_value()`] function.
-fn register_dtor(object_ptr: *mut u8, dtor: unsafe extern "C" fn(*mut u8)) {
+///
+/// # Safety
+///
+/// `object_ptr` must point to a valid object, and `dtor` must be the destructor for that object.
```
tsoutsman, line 98: Is this safety doc correct?

tsoutsman:
@kevinaboos should be ready for review

kevinaboos:
> @kevinaboos should be ready for review

Great! I'll take a look tonight or tomorrow late afternoon.

shim/src/lib.rs
kevinaboos: is it necessary? as in, where is it needed for `Reader`/`Writer` to be `Send`/`Sync`?

If it is required, then imo it should be an impl on `FatPointer` rather than any other types that use it. If that's sound, then perhaps we should change `FatPointer` to indicate that it's a shared non-raw pointer that always comes from an `Arc`, e.g., `FatSharedPtr`

shim/src/lib.rs
kevinaboos: All of the read/write traits take `&mut self`, so I think we ought to stick with that convention.

libs/theseus_ffi/src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+    Interrupted,
+    Unsupported,
+    UnexpectedEof,
+    OutOfMemory,
+    Other,
+}
+
+#[derive(Debug, Clone)]
+#[repr(C)]
+pub struct FatPointer {
```
kevinaboos, line 302: this can impl Send/Sync if and only if it actually comes from types that impl Send/Sync. Here I don't see that being the case yet, as it's always transmuted to/from something. 

Instead, it'd be cleaner and safer to use a From/Into impl on limited types that you know are safe to convert to/from fat pointers like this. Given a set of closed conversion methods like that, _then_ it might be okay to impl Send/Safe.

kernel/libtheseus/src/lib.rs
kevinaboos: yea, can't we use a typical conversion method for this? I know there are existing fat pointer types (pointers w/ metadata) in libcore, perhaps we can leverage that (or just DIY) to restrict the interface and types that can be used to convert to/from FatPointers

kevinaboos: Check out this section about strict provenance and the types included therein 
https://doc.rust-lang.org/beta/core/ptr/index.html#strict-provenance

kevinaboos: e.g., https://doc.rust-lang.org/beta/core/primitive.pointer.html#method.to_raw_parts

kernel/panic_entry/src/lib.rs
```diff
@@ -132,3 +132,16 @@ fn oom(_layout: core::alloc::Layout) -> ! {
     error!("\n(oom) Out of Heap Memory! requested allocation: {:?}", _layout);
     panic!("\n(oom) Out of Heap Memory! requested allocation: {:?}", _layout);
 }
+
+// TODO: Are these truly unreachable?
```
kevinaboos, line 136: these two function impls are fine for now, just replace this comment with one above each of these functions that explains how the build of std-reliant apps was erroring out without these present.

cfg/Config.mk
```diff
@@ -70,7 +70,7 @@ endif
 ## Also ensure that core memory functions (e.g., memcpy) are included in the build and not name-mangled.
 ## We keep these flags separate from the regular CARGOFLAGS for purposes of easily creating a sysroot directory.
 BUILD_STD_CARGOFLAGS += -Z unstable-options
-BUILD_STD_CARGOFLAGS += -Z build-std=core,alloc
+BUILD_STD_CARGOFLAGS += -Z build-std
```
kevinaboos: is this no longer needed? Or is this just a change that's left over from experimenting with building std?

If we only need this for "normal" builds of Theseus (without the new std lib) but we cannot use it for builds of the new std lib, then we can change it elsewhere, e.g., as an export override on one of the non-std make targets

Makefile
```diff
@@ -313,10 +315,33 @@ endif
...
 
 ## This target invokes the actual Rust build process via `cargo`.
-cargo:
+cargo: std
```
kevinaboos: bump: i think we decided that `std` should not be built by default (?). Same goes for the `clippy` and `doc` targets, i think  those ideally shouldn't depend on std.

Makefile
```diff
@@ -68,6 +69,7 @@ THESEUS_CARGO           := $(ROOT_DIR)/tools/theseus_cargo
 THESEUS_CARGO_BIN       := $(THESEUS_CARGO)/bin/theseus_cargo
 EXTRA_FILES             := $(ROOT_DIR)/extra_files
 LIMINE_DIR              := $(ROOT_DIR)/limine-prebuilt
+RUST_SOURCE             := $(ROOT_DIR)/ports/rust
```
kevinaboos: technically this is a port, but I think that we probably should move it into a top-level dir like `/rust` since it's so wildly different in nature than other port crates (which can typically just be built normally as regular crates)

kevinaboos: That also makes it easier to discover and document.

.ignore
kevinaboos: see other comment about moving this to `/rust`

.gitmodules
```diff
@@ -22,3 +22,6 @@
 	path = libs/core2
 	url = https://github.com/theseus-os/core2.git
 	shallow = true
+[submodule "ports/rust"]
+	path = ports/rust
+	url = https://github.com/theseus-os/rust.git
```
kevinaboos: let's not make this a submodule, but instead clone it (if it doesn't already exist) as part of a separate makefile target (on which the `std` target would depend).

.gitignore
```diff
@@ -39,3 +39,6 @@ github_pages/doc/
 
 # library lock files
 /libs/**/Cargo.lock
+
+# TODO: Remove once std is upstreamed
+.cargo/
```
kevinaboos, line 47: what is this, something that the build of `rust` generates?

kevinaboos (changes requested):
So i know that in our mtg, you said that if we used something like stabby or abi_stable, we'd have to infect the rest of the codebase with their specific Arc-like types. But I wonder how it would be any different than defining our own types in `theseus_ffi` -- couldn't we just have the shim and libtheseus layers convert to/from those existing ABI-stable FFI types, regardless of whether they came from our lib or another lib? Seems like we could save the effort (and potential soundness/safety sharp edges) of implementing and maintaining our own FFI types by just using their types for that specific set of crates only.

My other major concern is mostly just the transmutes used everywhere, hopefully we can use existing fat pointer types with explicit to/from conversions instead.

kernel/libtheseus/src/lib.rs
tsoutsman: Well the problem is that the types in `core` aren't FFI-safe.

> yea, can't we use a typical conversion method for this? 

I mean the function (at least for moving from a fat pointer to a trait object) is ultimately unsafe. There's nothing guaranteeing that the implementation of the function on the other side of the FFI boundary is really returning a fat pointer corresponding to e.g. `Writer`.

tsoutsman:
> couldn't we just have the shim and libtheseus layers convert to/from those existing ABI-stable FFI types, regardless of whether they came from our lib or another lib?

So not really. E.g. we call `app_io::stdout` and get an `alloc::sync::Arc<dyn Writer>`. We can't call `stabby::Arc::from_raw(alloc::sync::Arc::into_raw(writer))` because `stabby::Arc` and `sync::Arc` could have different layouts. And there's not really much else we could do. We can't get an owned reference to the `dyn Writer` and so we can't turn it into a `stabby` trait object and wrap it in a `stabby::Arc`.

And conceptually that makes sense. `writer` is layed out in memory as an `alloc::sync::Arc<dyn Writer>` and there is no way we can interpret it as a `stabby::Arc<vtable!(dyn Writer)>` because they have different layouts.

.gitignore
tsoutsman: Yes, we set the specific path to our custom `rustc` installation in `.cargo/config.toml`.

.gitmodules
tsoutsman: Yea, your review is of an older commit.

.ignore
tsoutsman: I still clone it to `ports/rust` rather than `rust` in the Makefile.

Makefile
tsoutsman: You're reviewing an older commit.

shim/src/lib.rs
tsoutsman: No, I don't think it's necessary.

PR #1071, opened 2023-11-08
PCI & interrupts modifications related to USB

NathanRoyer:
These are changes indirectly related to the USB PR. It doesn't build on x86 yet, because the changes were only made on aarch64.

The biggest change is that PCI legacy (INTx) interrupts are supported. Their handler wakes up a task that must have been bound to them, else it panics.

In USB, the woken up task simply checks the state of the USB controller, triggering various mechanisms when appropriate. When that task is done, it goes back to sleep using the blocker returned by [this function](https://www.theseus-os.com/Theseus/doc/waker/fn.new_waker.html).


kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -217,6 +217,23 @@ pub fn init_pl011_rx_interrupt() -> Result<(), &'static str> {
     int_ctrl.set_destination(PL011_RX_SPI, Some(current_cpu()), u8::MAX)
 }
 
+pub fn init_pci_interrupts(handler: InterruptHandler) -> Result<(), &'static str> {
```
kevinaboos: Need some docs about what this function does. I presume it registers the given `handler` to all legacy PCI interrupt vectors, but it's better to be clear.

Also, on that note, do we always want the same single interrupt handler to be registered for all PCI interrupts? It's probably better to allow the caller to specify individual handlers for each of the 4 legacy interrupts, e.g., 

```rust
pub fn init_pci_interrupts(
    pin_a_handler: InterruptHandler,
    pin_b_handler: InterruptHandler,
    pin_c_handler: InterruptHandler,
    pin_d_handler: InterruptHandler,
```

kernel/interrupts/src/aarch64/mod.rs
```diff
@@ -217,6 +217,23 @@ pub fn init_pl011_rx_interrupt() -> Result<(), &'static str> {
...
+    let int_ctrl = SystemInterruptController::get()
+        .ok_or("SystemInterruptController was not yet initialized")?;
+    let dst = Some(cpu::bootstrap_cpu().unwrap());
+
+    for int_num in BOARD_CONFIG.pci_intx {
+        if let Err(existing_handler) = register_interrupt(int_num, handler) {
+            if handler as InterruptHandler != existing_handler {
+                return Err("A different interrupt handler has already been setup for that IPI");
```
kevinaboos: outdated error msg

kernel/pci/src/lib.rs
```diff
@@ -207,6 +216,37 @@ pub fn pci_device_iter() -> Result<impl Iterator<Item = &'static PciDevice>, &'s
     Ok(get_pci_buses()?.iter().flat_map(|b| b.devices.iter()))
 }
 
+static INTX_DEVICES: RwLock<Vec<&'static PciDevice>> = RwLock::new(Vec::new());
+
+// Architecture-independent PCI interrupt handler
+// Aarch64-only because legacy interrupts aren't supported on x86
```
kevinaboos: just clarifying your comment, which made it sound like x86 itself doesn't support this 
```suggestion
// Currently aarch64-only, because legacy interrupts aren't supported on x86 yet.
```

kernel/pci/src/lib.rs
```diff
@@ -829,6 +879,36 @@ impl PciDevice {
...
+    pub fn pci_get_interrupt_status(&self, check_enabled: bool) -> bool {
+        const PCI_STATUS_INT: u16 = 1 << 3;
+
+        let interrupt_enabled = || (self.pci_read_16(PCI_COMMAND) & PCI_COMMAND_INT_DISABLED) == 0;
+        let pending_interrupt = || (self.pci_read_16(PCI_STATUS)  & PCI_STATUS_INT          ) != 0;
+
+        ((!check_enabled) || interrupt_enabled()) && pending_interrupt()
+    }
+
+    /// Sets a a task waker to be used when this device triggers an interrupt
```
kevinaboos: typo: repeated word "a"

kernel/pci/src/lib.rs
```diff
@@ -3,14 +3,17 @@
 //! Note: while pci currently uses port-io on x86 and mmio on aarch64,
 //! x86 may also support memory-based PCI configuration in the future;
 //! port-io is the legacy way to access the config space.
+//!
+//! For context on the various interrupt mechanisms (MSI/MSI-X/INTx):
+//! https://electronics.stackexchange.com/a/343218
```
kevinaboos: rustdoc wants you to surround raw links with `<` `>`
```suggestion
//! <https://electronics.stackexchange.com/a/343218>
```

Also, while this link is nice, it's not really anything official. Is there a more official source, such as a chapter/section of the PCI spec or an intel/arm technical reference manual?

kevinaboos (changes requested):
Looks good mostly, just some small requests.

One comment/question about RwLocks -- they are inefficient and much slower than Mutexes, so they're typically only used when absolutely necessary. It appears that the RwLock around `INTX_DEVICES` _might_ be necessary, but I can't tell for sure. If the point of using a RwLock there is to allow multiple interrupt handlers to access it simultaneously, then that is kind of defeated by using the interrupt-safe RwLock variant, which prevents other interrupts from firing while it is held.

What about the RwLock around the PciDevice's `Waker`? Are there any use cases where multiple tasks need to immutably access that waker concurrently?
If those are not needed, then you should switch to Mutex.

NathanRoyer:
> If those are not needed, then you should switch to Mutex.

You were right; I replaced these with Mutexes.

kernel/pci/src/lib.rs
```diff
@@ -506,17 +550,20 @@ impl PciLocation {
     }
 
     /// Sets the PCI device's command bit 10 to disable legacy interrupts
-    pub fn pci_set_interrupt_disable_bit(&self) {
+    pub fn pci_set_interrupt_disable_bit(&self, bit: bool) {
```
kevinaboos, line 553: I think we should rename this function to be more clear about what it actually does, and use the standard "enable_xyz" terminology. We can mention what it actually does (i.e., set the disable bit) in the function documentation, but as it stands now it's quite difficult to understand. 

I'll accept this now; kindly include that in the next PR about PCI/USB. 👍

kernel/pci/src/lib.rs
```diff
@@ -207,6 +216,37 @@ pub fn pci_device_iter() -> Result<impl Iterator<Item = &'static PciDevice>, &'s
...
+
+    for device in &*devices {
+        if device.pci_get_interrupt_status(true) {
+            device.pci_enable_interrupts(false);
+            log::info!("Device {} triggered an interrupt", device.location);
+
+            let reader = device.interrupt_waker.lock();
+            match &*reader {
+                Some(waker) => waker.wake_by_ref(),
+                None => log::error!("Device doesn't have an interrupt waker!"),
```
kevinaboos, line 235: i changed this to an error since i don't think it merits a panic, and also, panicking inside an interrupt handler should ideally be avoided.

kernel/pci/src/lib.rs
```diff
@@ -207,6 +216,37 @@ pub fn pci_device_iter() -> Result<impl Iterator<Item = &'static PciDevice>, &'s
...
+
+// Architecture-independent PCI interrupt handler
+// Currently aarch64-only, because legacy interrupts aren't supported on x86 yet.
+#[cfg(target_arch = "aarch64")]
+interrupt_handler!(pci_int_handler, None, _stack_frame, {
+    let devices = INTX_DEVICES.lock();
+
+    for device in &*devices {
+        if device.pci_get_interrupt_status(true) {
+            device.pci_enable_interrupts(false);
```
kevinaboos, line 229: this is also confusing, as upon first glance, the reader of this code would expect that you're just temporarily disabling PCI interrupts here while you handle the interrupt and then re-enabling them later. However, that doesn't happen, so this must mean something else.

We should rename and properly document the `pci_enable_interrupts()` function to be more clear about how and why it should be invoked (e.g., perhaps we just always want to disable this type of interrupt?)

kevinaboos (approved):
Accepting for now, but please see the 3 comments I left. Two of them can be addressed quickly with a follow-up PR or as part of another PR.

PR #1078, opened 2023-12-02
Rename `async_channel` to `sync_channel`

tsoutsman:
Renames the current `async_channel` to `sync_channel`.


kevinaboos (approved):
Thanks, lgtm!

PR #1079, opened 2023-12-04
Increase PS/2 timeouts for reset

YtvwlD:
The chosen arbitrary timeout value works in qemu, but causes a panic on boot on a Dell Optiplex 9020. I've tried a few values and with these, it boots with a Cherry and a HP keyboard and two Logitech mice.

-------------

I also had to change the `ScancodeSet` to `Set2` for the Cherry keyboard, but that's not in this PR. With these changes, Theseus *boots* on the Optiplex. I still can't input text or move the mouse.
Also, I think it would be nice if booting didn't require a mouse.

### DIFF
--- a/kernel/ps2/src/lib.rs
+++ b/kernel/ps2/src/lib.rs
@@ -309,7 +309,7 @@ impl<'c> PS2Mouse<'c> {
     /// Reset the mouse.
     pub fn reset(&self) -> Result<(), &'static str> {
         self.command_to_mouse(HostToMouseCommandOrData::MouseCommand(Reset))?;
-        const VERY_ARBITRARY_TIMEOUT_VALUE: u16 = u16::MAX;
+        const VERY_ARBITRARY_TIMEOUT_VALUE: u32 = u16::MAX as u32 * 3;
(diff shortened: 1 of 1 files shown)


kevinaboos (approved):
Thanks very much! Appreciate the report of experiences on real hardware.

Theseus "shouldn't" require a mouse to exist in order to boot, at least it didn't used to. We should also be able to select the correct scan code set by querying the PS2 controller, but I suppose were not actually doing that correctly. We'd happily accept that change too, if you're so inclined. 😊

hecatia-elegua:
@YtvwlD you could try to go through several `ScancodeSet` iterations, this notes at least one problem: https://github.com/theseus-os/Theseus/blob/ffb5e8b360fff655767b5ed13e6122db9a583857/kernel/keyboard/src/lib.rs#L49
iirc another problem was that rn we only have keycodes defined for set 1: https://github.com/theseus-os/Theseus/blob/ffb5e8b360fff655767b5ed13e6122db9a583857/libs/keycodes_ascii/src/lib.rs#L137

PR #1080, opened 2023-12-06
Handle paths with multiple components in `chdir`

tsoutsman:
Adds support for changing multiple levels of directories at once e.g. `cd task/2`.

Also since `components` normalizes the trailing slash, this fixes the bug which prevented changing into a directory followed by a trailing slash e.g. `cd namespaces/`.

This is indicative of the larger issue that each `FsNode` implementation needs its own path parsing algorithm e.g. `cd task && cd ..` doesn't work because `TaskFs` doesn't support `..`. I'll have a PR ready soon to fix that issue.


kevinaboos (approved):
Thanks!

PR #1081, opened 2023-12-06
Add `lspci` app, restructure PCI legacy interrupts (INTx) initialization

NathanRoyer:
(This PR depends on #1071)

This adds an `lspci` command:

image

It takes no argument and always lists all devices.


kernel/pci/src/lib.rs
```diff
@@ -731,6 +744,17 @@ impl PciDevice {
         mem_size
     }
 
+    /// Detectes MSI/MSI-X support on this device
+    ///
+    /// - Returns `(true, _)` if the device supports MSI.
+    /// - Returns `(_, true)` if the device supports MSI-X.
+    pub fn modern_interrupt_support(&self) -> (bool, bool) {
```
kevinaboos: this return type is prone to misuse, I'd return two enums to make it clearer: `(MsiSupport, MsiXSupport)` or something similar. Could also put it in a struct so you can name the fields.

kernel/pci/src/lib.rs
```diff
@@ -1,5 +1,20 @@
...
+//!   Four interrupt pins are shared among all devices.
+//!   This crate refers to that by "intx".
+//!
+//! * MSI (Message Signaled Interrupts) appeared with PCI express.
+//!   They allow devices to allocate up to 32 interrupt numbers.
+//!   This crate refers to that by "msi".
+//!
+//! * MSI-X messages appeared with PCIe 3.0.
+//!   They allow devices to allocate up to 2048 interrupt numbers.
+//!   This crate refers to that by "msix".
```
kevinaboos: formatting & clarifications
```suggestion
//! This crate deals with multiple types of interrupts:
//! * Legacy (INTx) interrupts are the oldest PCI interrupt representation.
//!   * Four interrupt pins are shared among all devices.
//!   * This crate refers to these legacy interrupts as "intx".
//!
//! * MSI (Message Signaled Interrupts) appeared with PCI express.
//!   * They allow devices to allocate up to 32 interrupt numbers.
//!   * This crate refers to these interrupts as "msi".
//!
//! * MSI-X messages appeared with PCIe 3.0.
//!   * They allow devices to allocate up to 2048 interrupt numbers.
//!   * This crate refers to these interrupts as "msix".
```

kernel/pci/src/lib.rs
```diff
@@ -218,18 +237,17 @@ pub fn pci_device_iter() -> Result<impl Iterator<Item = &'static PciDevice>, &'s
 
 static INTX_DEVICES: Mutex<Vec<&'static PciDevice>> = Mutex::new(Vec::new());
 
-// Architecture-independent PCI interrupt handler
-// Currently aarch64-only, because legacy interrupts aren't supported on x86 yet.
-#[cfg(target_arch = "aarch64")]
-interrupt_handler!(pci_int_handler, None, _stack_frame, {
+// Architecture-independent INTx handler
+// TODO: this zero value is incorrect
```
kevinaboos: ah, if this is incorrect because the IRQ vector differs for each handler, then you can't really use `HandlerDidNotSendEoi` here, right? You'd have to do it manually.

Is there another solution we could use, e.g., const generics here?

kevinaboos: Actually, we probably cannot use const generics on x86 since the IRQ numbers are discovered at runtime.

kernel/pci/src/lib.rs
```diff
@@ -881,25 +905,45 @@ impl PciDevice {
...
+                _ => panic!("Cannot use INTx on device {:?}: No INTx info", self),
+            };
+
+            if let Err(existing_handler) = register_interrupt(int_num, pci_intx_handler) {
+                if existing_handler != (pci_intx_handler as usize) {
+                    panic!("Couldn't set PCI INTx handler for device: {:?}", self);
+                }
+            }
+
+            // on aarch64, this is done earlier (see the public init function)
```
kevinaboos: this is quite a confusing workflow that spans multiple functions. Can't we unify it to be the same across both x86 and aarch64?

kevinaboos (changes requested):
The `lspci` app itself is fine. I left some comments about other PCI changes, the most significant of which is how the INTx interrupt handlers are initialized (and how it differs across architectures).

kernel/pci/src/lib.rs
NathanRoyer: Fixed it by allocating INTx handling slots as new interrupt numbers are discovered

kernel/pci/src/lib.rs
NathanRoyer: I don't think we should, as the x86 solution is rather hacky whereas the aarch64 solution is canonical.
I tried to clarify and document the code more.

kevinaboos: what's "hacky" about the x86 solution? You mean the way we "discover" (or assume) the IRQ numbers on x86 is hacky?

My comment here was just about the way that x86 and aarch64 initialization is split across multiple functions in a way that makes it really hard to follow, which means it is hard to maintain. From a design perspective, it'd be nice if a single function or a set of functions handled both aarch64 and x86 initialization using the same pattern/structure of code.

kernel/pci/src/lib.rs
```diff
@@ -217,33 +236,74 @@ pub fn pci_device_iter() -> Result<impl Iterator<Item = &'static PciDevice>, &'s
...
-                Some(waker) => waker.wake_by_ref(),
-                None => log::error!("Device doesn't have an interrupt waker!"),
+static INTX_NUMBERS: Mutex<[Option<InterruptNumber>; 4]> = Mutex::new([None; 4]);
+
+// Architecture-independent INTx handlers
+macro_rules! intx_handler {
+    ($name:ident, $num:literal) => {
+        interrupt_handler!($name, {
+            let intx_numbers = INTX_NUMBERS.lock();
+            intx_numbers[$num].expect("uninitialized x86 PCI INTx handler")
```
kevinaboos, line 246: This is a bit funky and kinda hard to read; I think this is a good place to just include this logic in the actual body of the interrupt handler and issue an EOI directly in the body instead of returning `HandlerDidNotSendEoi`. 

Another thing we could do is to include the interrupt number alongside the `waker` instance or alongside the `INTX_DEVICES` static, in order to "ensure" those are always available.

kevinaboos (approved):
Thanks, I've accepted it and left a comment for a minor future improvement. Let me know your thoughts on that.

PR #1082, opened 2023-12-12
Allow booting without keyboard or mouse

YtvwlD:
If a PS/2 controller exists, but no keyboard or mouse is plugged in, resetting those devices fails. Handle this gracefully by logging an error and continuing without them.

I really don't know much about PS/2 but this allows booting on a real PC without keyboard or mouse (and it correctly detects which of those is missing).


kevinaboos (approved):
Thanks, much appreciated!

PR #1086, opened 2023-12-23
Don't hardcode the path for bash

YtvwlD:
Some systems (eg. NixOS) don't have it in /bin.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 ### This makefile is the top-level build script that builds all the crates in subdirectories 
 ### and combines them into the final OS .iso image.
 ### It also provides convenient targets for running and debugging Theseus and using GDB on your host computer.
-SHELL := /bin/bash
+SHELL := /usr/bin/env bash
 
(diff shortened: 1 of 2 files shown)


kevinaboos:
Ah, good point! Do all systems have `/usr/bin/env` though? I thought `/bin` was the most common.

YtvwlD:
Most of them apparently? I failed to find an authoritative source ([this](https://unix.stackexchange.com/a/77586/70882) is the best I could come up with), but most modern systems seem to have it either in `/usr/bin` or in both places. Notable exceptions [seem to be](https://www.in-ulm.de/~mascheck/various/shebang/#env) "SCO OpenServer (…) and Cray Unicos", but I've never heard of them before.

kevinaboos:
Cool, I also looked into it and it seems that using env is okay, especially in this scenario because bash wouldn't suffer from different versioning issues like python and other interpreters might.

kevinaboos (approved):
Thanks!

PR #1089, opened 2024-01-04
Remove unnecessary `clone()` functions in `hull`

jklott:
There is a `TODO` specifically pointing out some unneeded `clone()` functions being used which this PR addresses. A total of 5 clones can be removed safely.


tsoutsman (approved):
Thank you!

Either between when that code was written and now the compiler gained the ability to reason about return statements in loop bodies, or I underestimated the compiler when writing the code. Judging by the last clearly unnecessary clone, it was probably the latter.

PR #1102, opened 2024-08-14
Add size option (-s) in the `ls` command

nnh12:
This pull request is to add the `ls -s` option in the Theseus operating environment. The `ls -s` is a standard linux command. It shows the sizes of files and directories in blocks, which is useful for getting a quick overview of file sizes. Below are some sample test cases:

```
/extra_files: ls -s
    --         wasm
    --         test_files
    1159    README.md
/extra_files: ls test_files -s
    --         zork_dat
    --         text
/extra_files/test_files/text: ls -s
    422     the_rise_of_skywalker.txt
    492     the_phanton_menace.txt
    477     the_last_jedi.txt
    499     the_force_awakens.txt
    486     the_empire_strikes_back.txt
    482     revenge_of_the_sith.txt
    462     return_of_the_jedi.txt
    490     attack_of_the_clones.txt
    503     a_new_hope.txt
```

Default `ls` command still works:

```
/extra_files/test_files: ls 
zork_dat
text
```

### DIFF
--- a/applications/ls/src/lib.rs
+++ b/applications/ls/src/lib.rs
@@ -19,6 +19,7 @@ use path::Path;
 pub fn main(args: Vec<String>) -> isize {
     let mut opts = Options::new();
     opts.optflag("h", "help", "print this help menu");
+    opts.optflag("s", "size", "print the size of each file in directory");
 
     let matches = match opts.parse(args) {
         Ok(m) => m,
@@ -34,13 +35,21 @@ pub fn main(args: Vec<String>) -> isize {
         return 0;
     }
 
+    let size_option = matches.opt_present("s");
+
     let Ok(curr_wd) = task::with_current_task(|t| t.get_env().lock().working_dir.clone()) else {
         println!("failed to get current task");
         return -1;
     };
+
+    if size_option && matches.free.is_empty() {
+        print_children(&curr_wd, true);
+        return 0;
+    }
+
     // print children of working directory if no child is specified
     if matches.free.is_empty() {
-        print_children(&curr_wd);
+        print_children(&curr_wd, false);
         return 0;
     }
 
@@ -49,7 +58,12 @@ pub fn main(args: Vec<String>) -> isize {
     // Navigate to the path specified by first argument
     match path.get(&curr_wd) {
         Some(FileOrDir::Dir(dir)) => {
-            print_children(&dir);
+            if size_option {
+                print_children(&dir, true);
+            }
+            else {
+                print_children(&dir, false);
+            }
             0
         }
         Some(FileOrDir::File(file)) => {
@@ -63,12 +77,18 @@ pub fn main(args: Vec<String>) -> isize {
     }
 }
 
-fn print_children(dir: &DirRef) {
+fn print_children(dir: &DirRef, print_size: bool) {
     let mut child_string = String::new();
     let mut child_list = dir.lock().list(); 
     child_list.reverse();
     for child in child_list.iter() {
-        writeln!(child_string, "{child}").expect("Failed to write child_string");
+        let child_path = dir.lock().get(child).expect("Failed to get child path");
+        if print_size {
+            let size = FileOrDir::get_file_size(&child_path);
+            writeln!(child_string, "   {}    {}", size, child).expect("Failed to write child_string");
+        } else {
+            writeln!(child_string, "{}", child).expect("Failed to write child_string");
+        }
     }
     println!("{}", child_string);
 }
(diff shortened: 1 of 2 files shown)


applications/ls/src/lib.rs
```diff
@@ -34,13 +35,21 @@ pub fn main(args: Vec<String>) -> isize {
...
+
+    if size_option && matches.free.is_empty() {
+        print_children(&curr_wd, true);
+        return 0;
+    }
+
     // print children of working directory if no child is specified
     if matches.free.is_empty() {
-        print_children(&curr_wd);
+        print_children(&curr_wd, false);
```
kevinaboos: Suggestion: reduce the above code duplication with just a single call:
```rust
    print_children(&curr_wd, size_option);
```

applications/ls/src/lib.rs
```diff
@@ -49,7 +58,12 @@ pub fn main(args: Vec<String>) -> isize {
     // Navigate to the path specified by first argument
     match path.get(&curr_wd) {
         Some(FileOrDir::Dir(dir)) => {
-            print_children(&dir);
+            if size_option {
+                print_children(&dir, true);
```
kevinaboos: Same comment here, no need to duplicate code, just pass in `size_option` directly as an argument.

kernel/fs_node/src/lib.rs
```diff
@@ -24,6 +24,13 @@ use alloc::sync::{Arc, Weak};
 use memory::MappedPages;
 use io::{ByteReader, ByteWriter, KnownLength};
 
+/// A value to represent the space on the disk (stored in meta-information) for directores
```
kevinaboos: What is this? I don't understand, a directory doesn't necessarily occupy 4096 bytes. For a directory, you should either calculate the full size of all of its children, or for an easier option, just use something like `--` to show that a size is N/A.

kernel/fs_node/src/lib.rs
```diff
@@ -190,4 +197,15 @@ impl FileOrDir {
             FileOrDir::Dir(_) => true,
         }
     }
+
+    /// Returns the size of the file in bytes, or directory disk space.
+    pub fn get_file_size(file_or_dir: &FileOrDir) -> usize {
```
kevinaboos: This is redundant, see the existing `len()` method for files (part of the `KnownLength` trait). See my other comment about handling sizes for directories.

kevinaboos (changes requested):
Thanks for the PR! I left a few comments.

My top-level comment is that the default output of size should be in units of bytes, not in disk sectors. Kindly make that change and address my other comments, and then I can merge it in. Thanks!

applications/ls/src/lib.rs
nnh12: Got it thank you - I will add that change

kernel/fs_node/src/lib.rs
nnh12: Thanks for letting me know (when researching, I found that most linux systems obtains the size of a directory from its metadata, which usually just indicate the directory size as 4096)

However, I will add the -- indication for directory size as that would may be easier and less misleading

nnh12:
I have created the PR to the requested changes 
The file size is returned from the `ls -s` in bytes (since it calls the `len()` function from `KnownLength`
The directory size is returned as `--` as you suggested

applications/ls/src/lib.rs
```diff
@@ -34,13 +35,21 @@ pub fn main(args: Vec<String>) -> isize {
...
+
     let Ok(curr_wd) = task::with_current_task(|t| t.get_env().lock().working_dir.clone()) else {
         println!("failed to get current task");
         return -1;
     };
+
+    if size_option && matches.free.is_empty() {
+        print_children(&curr_wd, size_option);
+        return 0;
+    }
```
kevinaboos: this duplicated code block is unnecessary, you can remove it as it's redundant with the block below it

kevinaboos (changes requested):
thanks! i left only one small comment about code duplication that you missed.

For future reference, kindly don't force push commits to your PR branch, as it removes all of my review comments (notice that they're all gone now). This makes it hard to track whether all of the comments have been addressed.

nnh12:
Got it thank you! 
Just to confirm, so should I just make a new branch and open a new PR addressing the feedback?

applications/ls/src/lib.rs
nnh12: Got it I will remove that

kevinaboos:
> Got it thank you! Just to confirm, so should I just make a new branch and open a new branch addressing the feedback?

Nope, all you have to do is make another change, commit it to this same branch, and push it. Github will automatically show the new commit in this PR.

nnh12:
Got it thanks - I won't force push

nnh12:
I just removed the duplicated code block- thanks

kevinaboos (approved):
Thanks, looks great!

PR #1103, opened 2024-08-26
update makefile for freebsd

hoanga:
good evening,

the following update updates the makefile so the build runs on freebsd (using gmake)


kevinaboos (approved):
Thanks, much appreciated! 

Sorry for the long delay, been occupied with other work.

PR #1104, opened 2024-08-28
Update `cd ` command to use correct path traversal from the `path` crate to allow for parent directory

nnh12:
This pull request adds the option to navigate to the parent directory respective of the current working directory, without specifying the full path. The `cd ..` is a standard operation in linux. For example, if current directory is `/extra_files/test_files`, running `cd ..` moves directory to `/extra_files`.

Below are some test cases ran on Theseus:

```
/extra_files/test_files: cd ..
task[20] existed with code 0 (0x0)
/extra_files: 
```

```
/extra_files: cd ..
task [21] exited with code 0 (0x0)
/: 
```

```
/extra_files/test_files/text: cd ..
task[28] exited with code 0 (0x0)
/extra_files/test_files:
```

The default `cd` operation still works too:

```
/extra_files/test_files/: cd text
task[29] exited with code 0 (0x0)
/extra_files/test_files/text: 
```

### DIFF
--- a/applications/cd/src/lib.rs
+++ b/applications/cd/src/lib.rs
@@ -17,7 +17,6 @@ use getopts::Options;
 pub fn main(args: Vec<String>) -> isize {
     let mut opts = Options::new();
     opts.optflag("h", "help", "print this help menu");
-    opts.optflag("p", "parent", "use '..' to change to parent directory");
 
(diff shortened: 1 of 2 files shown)


kevinaboos (approved):
Thanks, looks great now!

PR #1105, opened 2024-08-28
Frame Allocator can create duplicate free regions during initialization

Ramla-I:
The current code catches the mistake and reports an error. The duplicate region is created due to a bug in the `check_and_add_free_region` function. After returning from a recursive call, the original area's end frame should be updated to not include the region that was just added to the free list.

I believe there is an unspoken assumption that the list of reserved regions will be in ascending order, but that is not always the case.

In the screenshots below, the original code adds the region 0x2098000..=0x1FFDE000 twice.

Screenshot_w_bug
Screenshot_w_fix


kevinaboos (approved):
ah, good catch! Thanks.

